PR #2, opened 2019-12-17
BCRS UI Framework

cameron-freshworks:
*Issue #:* /bcgov/entity#1895

This is an initial PR for the UI Framework. Can add or remove anything based off of team preference. Please let me know if i missed anything!

*Description of changes:*
* Implements directory structure & Typescript related directories.
* Router set up ( Auth Not Included ).
* Store structure set up ( Examples for State Included, Actions and Mutation Examples to come ).
* Utilities methods to initialize the env variables.
* Libraries and dependencies Set up.
* Typescript Store/State examples.
* Typescript Mixin example.
* Peters Concept for static values being pulled in as resources with example.
* Linting & Unit testing.

 
*Not Included:*
* Launch Darkly Set up.


bcrs-ui/.env.production
```diff
@@ -0,0 +1,2 @@
+VUE_APP_PATH = basePath
```
severinbeauvais: Perfect content for now.
Please also check in `.env`.

cameron-freshworks: Oh right, i made it. It must have been ignored!

bcrs-ui/babel.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  presets: [
+    '@vue/cli-plugin-babel/preset'
+  ]
+}
```
severinbeauvais: We need to find out why this is different than coops-ui version. Does this one work? Does the coops-ui one need to be updated?
I think that if this requires more than 1 hour of research then we should create a new ticket.

bcrs-ui/jest.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  preset: '@vue/cli-plugin-unit-jest/presets/typescript-and-babel'
+}
```
severinbeauvais: We need to find out why this is different than coops-ui version. Does this one work? Does the coops-ui one need to be updated?
I think that if this requires more than 1 hour of research then we should create a new ticket.

bcrs-ui/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "scripts": {
+    "serve": "vue-cli-service serve",
+    "build": "vue-cli-service build",
+    "lint": "vue-cli-service lint",
+    "test:unit": "vue-cli-service test:unit --testPathPattern"
+  },
+  "dependencies": {
+    "@mdi/font": "^4.5.95",
+    "axios": "^0.18.0",
+    "core-js": "^3.4.3",
```
severinbeauvais: I think you also need `"regenerator-runtime": "^0.13.3",`.
See comment in coops-ui, main.ts.

bcrs-ui/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "sbc-common-components": "2.0.22",
+    "vue": "^2.6.10",
+    "vue-class-component": "^7.0.2",
+    "vue-property-decorator": "^8.3.0",
+    "vue-router": "^3.1.3",
+    "vuetify": "^2.0.11",
+    "vuex": "^3.1.2",
+    "vuex-class": "^0.3.2"
```
severinbeauvais: What's this one for? Should we be using it in coops-ui in lieu of something else?

bcrs-ui/package.json
```diff
@@ -0,0 +1,43 @@
...
+  },
+  "devDependencies": {
```
severinbeauvais: I'd love to see this list ordered alphabetically :)
(easier to compare with other projects' lists)

bcrs-ui/public/config/configuration.json
```diff
@@ -0,0 +1,6 @@
+{
+  "API_URL": "https://legal-api-dev.pathfinder.gov.bc.ca/api/v1/businesses/",
+  "AUTH_API_URL": "https://auth-api-dev.pathfinder.gov.bc.ca/api/v1/",
+  "PAY_API_URL": "https://pay-api-dev.pathfinder.gov.bc.ca/api/v1/",
+  "ADDRESS_COMPLETE_KEY": "RH99-RN99-GB72-FW86"
+}
```
severinbeauvais: Please add latest "KEYCLOAK_CONFIG_URL" key and value, and also its related code (in config-helper.ts, main.ts and vue.config.js).

severinbeauvais:
Please use the bcgov favicon from coops-ui (instead of the default Vue favicon).

bcrs-ui/public/index.html
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width,initial-scale=1.0">
+    <title>hello-world</title>
+    <link rel="icon" href="dummy-not-used-do-not-remove">
```
severinbeauvais: FYI, this line needs to be here for IE11 to work but does nothing in other browsers.

bcrs-ui/public/index.html
```diff
@@ -0,0 +1,21 @@
...
+    <link rel="icon" href="<%= BASE_URL %>favicon.ico">
+    <link rel="preload" href="<%= BASE_URL %>fonts/materialdesignicons-webfont.927457ed.woff2" as="font" type="font/woff2" crossorigin>
```
severinbeauvais: Thanks for including the fonts.
Do a `npm run build` and check the dist folder to confirm this font name.

bcrs-ui/vue.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  configureWebpack: {
+    devtool: 'source-map'
+  },
+  transpileDependencies: [
+    'vue-plugin-helper-decorator',
+    'vuetify'
+  ]
+}
```
severinbeauvais: Do you need this line here as well, or does this work as is?
`publicPath: process.env.VUE_APP_PATH,`
(this matches the properties in .env)

bcrs-ui/tsconfig.json
```diff
@@ -0,0 +1,40 @@
...
+    "moduleResolution": "node",
+    "experimentalDecorators": true,
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true,
+    "sourceMap": true,
+    "baseUrl": ".",
+    "types": [
+      "webpack-env",
+      "jest"
+    ],
```
severinbeauvais: May need to include "vuetify" here as well.

bcrs-ui/src/views/StateExample.vue
```diff
@@ -0,0 +1,18 @@
...
+    <ul>
+      <li>{{ stateText }}</li>
+    </ul>
+  </div>
+</template>
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator'
+import { State } from 'vuex-class'
+
+@Component
```
severinbeauvais: Shouldn't this be `@Component({})` to avoid a lint error?

bcrs-ui/src/views/MixinExample.vue
```diff
@@ -0,0 +1,13 @@
...
+  <div class="mixinExample">
+    <h1>This is an about page that contains an example for a Mixin</h1>
+    <h2>Testing Mixins: {{ sendMsg('My First Message')}}</h2>
+  </div>
+</template>
+<script>
+import { Component, Mixins } from 'vue-property-decorator'
+import { CommonMixin } from '@/mixins'
+
+@Component
```
severinbeauvais: Shouldn't this be @Component({}) to avoid a lint error?

bcrs-ui/src/views/MixinExample.vue
```diff
@@ -0,0 +1,13 @@
...
+export default class MixinExample extends Mixins(CommonMixin) {}
```
severinbeauvais: Thanks for awesome example!

bcrs-ui/src/views/Home.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+  <div class="home">
+    <img alt="Vue logo" src="../assets/logo.png">
+    <HelloWorld msg="Welcome to Your Vue.js App"/>
+  </div>
+</template>
+
+<script>
+import { Component, Vue } from 'vue-property-decorator'
+import HelloWorld from '@/components/HelloWorld.vue'
```
severinbeauvais: Should we have a `@/components/index.ts`, as in other folders? Just a thought, not saying we should.

bcrs-ui/src/App.vue
```diff
@@ -0,0 +1,33 @@
...
+
+<style lang="scss">
+#app {
+  font-family: 'Avenir', Helvetica, Arial, sans-serif;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+  text-align: center;
+  color: #2c3e50;
+  margin-top: 60px;
+}
```
severinbeauvais: Some of this is obsolete/invalid for us (like font). See if you can use our existing classes and get rid of CSS here.

severinbeauvais:
Are we using `bcrs-ui/src/assets/logo.png`? If not please delete it.

bcrs-ui/src/assets/styles/base.scss
```diff
@@ -0,0 +1,321 @@
...
+    flex: 1 1 auto;
+    width: 100%;
+  }
+
+  select {
+    flex: 1 1 auto;
+    -webkit-appearance: menulist;
+  }
+}
+// Payment Table
```
severinbeauvais: I think we aren't using these 3 classes anywhere (not this project and not coops-ui). Please verify and delete this obsolete CSS.

bcrs-ui/src/main.ts
```diff
@@ -0,0 +1,37 @@
+// Vue Libraries
+import Vue from 'vue'
```
severinbeauvais: I'm pretty sure you need these first:
```
import 'core-js/stable' // to polyfill ECMAScript features
import 'regenerator-runtime/runtime' // to use transpiled generator functions
import '@mdi/font/css/materialdesignicons.min.css' // ensure you are using css-loader
```

bcrs-ui/src/main.ts
```diff
@@ -0,0 +1,37 @@
...
+import Vuetify from 'vuetify'
+import router from './router'
+import { store } from '@/store'
+
+// Styles
+import '@/assets/styles/base.scss'
```
severinbeauvais: Also add this:
`import '@mdi/font/css/materialdesignicons.min.css' // ensure you are using css-loader`

bcrs-ui/src/main.ts
```diff
@@ -0,0 +1,37 @@
...
+import '@/assets/styles/layout.scss'
+import '@/assets/styles/overrides.scss'
+
+// Base App
+import App from './App.vue'
+
+// Helpers
+import { fetchConfig } from '@/utils'
+
+Vue.config.productionTip = false
```
severinbeauvais: Add comment for this one.

bcrs-ui/src/main.ts
```diff
@@ -0,0 +1,37 @@
...
+Vue.use(Vuetify)
+const vuetify = new Vuetify()
```
severinbeauvais: Please use parameter same as in coops-ui.

bcrs-ui/src/main.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Fetch config from server, then load Vue
+ */
+fetchConfig()
+  .then(() => {
+    new Vue({
```
severinbeauvais: See latest code in coops-ui as I think we'll need all of it here also.

bcrs-ui/src/router/index.ts
```diff
@@ -0,0 +1,13 @@
...
+import VueRouter from 'vue-router'
+import { routes } from './routes'
+
+Vue.use(VueRouter)
+
+const router = new VueRouter({
+  mode: 'history',
+  base: process.env.BASE_URL,
+  routes
+})
```
severinbeauvais: I think you should add the "scrollBehaviour" attribute here (same as coops-ui). Thoughts?

bcrs-ui/src/router/routes.ts
```diff
@@ -0,0 +1,25 @@
...
+  {
+    path: '/StateExample',
+    name: 'stateExample',
+    component: StateExample
+  },
+  {
+    // default/fallback route
+    path: '*',
+    redirect: '/'
+  }]
```
severinbeauvais: These should be on separate lines or may cause lint issue.

bcrs-ui/src/types.ts
```diff
@@ -0,0 +1,7 @@
+export interface State {
+  todos: ToDo[]
+}
+
+export interface ToDo {
+  text: string
+}
```
severinbeauvais: Nice boilerplate! Do you think this should be in a separate folder?

bcrs-ui/public/index.html
```diff
@@ -0,0 +1,21 @@
...
+    <link rel="stylesheet" href="<%= BASE_URL %>css/addresscomplete-2.00.min.css">
+    <script type="text/javascript" src="<%= BASE_URL %>js/addresscomplete-2.00.min.js" defer></script>
```
severinbeauvais: Please check in these 2 files above. (I think we'll probably want to use AC.)

severinbeauvais:
> Do we want to use models and services that we discussed in one of the retros? Models can be used to define the request/response structure and services can have one Service file for each API and that will list all the operations of that specific API that we are using in the UI.
> 
> Also do we have multiple unrelated functionalities that uses different APIs? I can't think of any at the moment. If we come across that it would be a good idea to organize the store into modules

Great idea! Do you think this should be part of the "template" for future BCR apps or should we just move that way for this app at the moment?

bcrs-ui/babel.config.js
cameron-freshworks: Good question!  I will look into it.

cameron-freshworks: At first glance, the vue/cli-plugin-babel/preset is a wrapper for a few things like Babel 7 + babel-loader + @vue/babel-preset-app. I think this just saves the time of adding these manually as options in the babel configuration. It does indeed work, but i don't know enough about babel and webpack to say for sure that the Coops-ui should be updated.

bcrs-ui/jest.config.js
cameron-freshworks: As mentioned above, i think this is just a bundle of sorts of the packages mentioned in the name. When you install Jest in the start up of VUE CLI, this is the bundle installed. I believe it saves having to set the `transform` & `fileextensions` options manually.  We can definitely dig into this deeper though, as i don't know what the full suite of options and presets are compared to what was used in the Coops-ui.

bcrs-ui/package.json
cameron-freshworks: k sounds good!

bcrs-ui/package.json
cameron-freshworks: This module provides a shortcut to binding store to our classes. Like the `@Props` decorator used by `Vue-class-component` it allows us to use `@State` , '@Actions` and so on. Proves a consistent looking class component :)  As far as updated Coops-ui, i'm not so sure. That would involve rewriting a lot of the store of as well. If you feel it's something we would be given time to do, i'm certain it could be done, but would be a fairly large task affecting many components ( everywhere store is used ).

bcrs-ui/package.json
cameron-freshworks: Yes sir, great idea, you got it!

bcrs-ui/public/config/configuration.json
cameron-freshworks: oh nice, some new stuff! I didn't notice the changes to master, i will get this work included. Thanks for the heads up :)

cameron-freshworks: I guess this leads to another question, since this work directly correlates with Authentication, are we thinking to pull that in as well?

bcrs-ui/public/index.html
cameron-freshworks: Interesting! Thanks for the info.

bcrs-ui/public/index.html
cameron-freshworks: K, good idea, will do!

bcrs-ui/babel.config.js
severinbeauvais: I had a quick look at `@vue/babel-preset-app/index.js` and it appears to set a bunch of default app config -- including some that appear different from what we do in coops-ui.

I think we should "resolve this conversation" but keep an eye on what gets built (ie, npm run build) and how it's loading and running in different browsers, etc. If things seem incorrect then we can revisit this config.

bcrs-ui/jest.config.js
severinbeauvais: I had a quick look at this one too, and you're right: it presets some options here. However, coops-ui has some different options.

As above, I think we should "resolve this conversation" but keep an eye on what gets built (ie, npm run build) and how it's loading and running in different browsers, etc. If things seem incorrect then we can revisit this config.

bcrs-ui/package.json
severinbeauvais: I've created #2179 to possibly implement this change in coops-ui.

bcrs-ui/public/config/configuration.json
severinbeauvais: Yes, this authorization code is needed in this project.

I also this this code should be included in the UI template. For the foreseeable future, this is the authorization code that we will use in UI projects. And the UI template can easily be updated later as needed.

cameron-freshworks: ok perfect, we will get it all in there, thanks!

bcrs-ui/vue.config.js
cameron-freshworks: The existing code does work as is.  I debated included this until we knew what our path would be, but on second thought we should include this code even though its a placeholder of sorts so it is not forgotten later. Thank you!

bcrs-ui/src/views/StateExample.vue
cameron-freshworks: Currently there is no lint error without in this framework or in coops-ui. Should there be?

bcrs-ui/src/views/MixinExample.vue
cameron-freshworks: As above, should this be a linting error? Currently not

bcrs-ui/src/views/Home.vue
cameron-freshworks: Hmm, i thought about that too. For the sake of examples, i think we should! Agreed 👍

bcrs-ui/src/App.vue
cameron-freshworks: It's definitely unrelated and obsolete to the project as a whole. It's just some styling for the VUE Splash page which will be removed. In that context, would you still see some value in removing this and restyling the splash page?

cameron-freshworks:
> Are we using `bcrs-ui/src/assets/logo.png`? If not please delete it.

Just for the splash page. Same as above i suppose, would like me to just remove the splash page from the framework?

bcrs-ui/src/assets/styles/base.scss
cameron-freshworks: okie doke, will do!

bcrs-ui/src/views/StateExample.vue
severinbeauvais: Don't change this for now.

bcrs-ui/src/views/MixinExample.vue
severinbeauvais: Don't change this for now.

bcrs-ui/src/App.vue
severinbeauvais: I guess it doesn't really matter either way. Leave the code as-is for now.

bcrs-ui/src/main.ts
cameron-freshworks: Added in!

bcrs-ui/src/main.ts
cameron-freshworks: Done!

bcrs-ui/src/main.ts
cameron-freshworks: Done, and added dependency 👍

bcrs-ui/src/main.ts
cameron-freshworks: Will do, thanks!

bcrs-ui/src/main.ts
cameron-freshworks: Will do, i will include the auth alongside your new refresh logic 👍

bcrs-ui/src/router/index.ts
cameron-freshworks: Yes i do believe that is a good idea. Interesting that it isn't the default behaviour, maybe i will do a little research.

severinbeauvais:
> > Are we using `bcrs-ui/src/assets/logo.png`? If not please delete it.
> 
> Just for the splash page. Same as above i suppose, would like me to just remove the splash page from the framework?

I didn't originally notice it in Home.vue.

The initial app should have a landing page -- Home -- so just leave it all as-is. And promise me we'll delete the logo when we rewrite Home :)

bcrs-ui/src/router/routes.ts
cameron-freshworks: Not an error in the frameworks linting or coops-ui, but will still separate for consistency 👍

bcrs-ui/src/types.ts
cameron-freshworks: 100%, this was me fiddling around and i left it in. My apologies, oops. I can add the examples to the interfaces :)

bcrs-ui/public/index.html
cameron-freshworks: Sounds good to me, out of curiosity was AddressComplete part of the app spec or was it something we chose. We're there other options out there? ( NOT suggesting we change, just curious! )

cameron-freshworks:
> > > Are we using `bcrs-ui/src/assets/logo.png`? If not please delete it.
> > 
> > 
> > Just for the splash page. Same as above i suppose, would like me to just remove the splash page from the framework?
> 
> I didn't originally notice it in Home.vue.
> 
> The initial app should have a landing page -- Home -- so just leave it all as-is. And promise me we'll delete the logo when we rewrite Home :)

I promise! ( fingers crossed behind my back )

cameron-freshworks:
> It would be good to add the request response interceptor utility. e.g axios-auth.ts in coops-ui

Good idea, definitely goes hand in hand with the Auth. So the plan is now to Add in Auth, Token refresh and the interceptor. Thanks!

cameron-freshworks:
> Do we want to use models and services that we discussed in one of the retros? Models can be used to define the request/response structure and services can have one Service file for each API and that will list all the operations of that specific API that we are using in the UI.
> 
> Also do we have multiple unrelated functionalities that uses different APIs? I can't think of any at the moment. If we come across that it would be a good idea to organize the store into modules

Great idea, i feel like now is a good time to incorporate with an example. As long as this framework isn't blocking any other work from being done, lets do it!

bcrs-ui/public/index.html
severinbeauvais: Dunno. In fact we're not using it to its potential (where it provides a key indicating the entered address is valid). I'll let the business drive any changes here...

bcrs-ui/src/resources/externalResourceExample.ts
```diff
@@ -0,0 +1,12 @@
+export const ExternalResource = [
+  {
+    id: 1,
+    displayName: 'Cameron',
+    message: 'Hola, Coma Estas?'
```
severinbeauvais: Lol, should be "cómo estás"

bcrs-ui/src/store/state/resource-model.ts
```diff
@@ -0,0 +1,14 @@
+import { ResourceExampleIF } from '@/interfaces'
+
+export const resourceModel: Array<ResourceExampleIF> = [
```
severinbeauvais: I don't understand how this array works. Comments please?

bcrs-ui/vue.config.js
severinbeauvais: Why do you need to prepend a \ for public path (where coops-ui doesn't)?

bcrs-ui/src/utils/axios-auth.ts
```diff
@@ -0,0 +1,18 @@
+import axios from 'axios'
```
severinbeauvais: Good decision to move this into utils/ !

bcrs-ui/src/main.ts
```diff
@@ -0,0 +1,75 @@
...
+import '@/assets/styles/base.scss'
+import '@/assets/styles/layout.scss'
+import '@/assets/styles/overrides.scss'
+import '@mdi/font/css/materialdesignicons.min.css' // ensure you are using css-loader
+
+// Base App
+import App from './App.vue'
+
+// Helpers
+import { fetchConfig, haveKcTokens } from '@/utils'
```
severinbeauvais: Good decision to import these in this way!

bcrs-ui/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "hello-world",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "serve": "vue-cli-service serve",
+    "build": "vue-cli-service build",
+    "lint": "vue-cli-service lint",
+    "test:unit": "vue-cli-service test:unit --testPathPattern"
```
severinbeauvais: I think this needs to have `--testPathPattern=.`.

cameron-freshworks: Hmm.. it actually works as is. It could be just for me locally but i don't think we need to add the `=.` to grab the matching patterns. I actually found by adding that it was pulling more tests into the match outside of the specified test in the coops-ui. Maybe you could try this out and let me know what you think. I know in coops-ui it would pull more tests into the scope than just the matching pattern.

bcrs-ui/src/main.ts
cameron-freshworks: Cool, thanks! Glad you like it. Getting more bang for our buck from that utils folder!

bcrs-ui/src/utils/axios-auth.ts
cameron-freshworks: awesome, glad you like, thanks!

bcrs-ui/src/store/state/resource-model.ts
cameron-freshworks: So in this instance, it's an Array of `ResourceExamples`. Applying the interface to the type of objects we want to see in that array so the resourceModel must conform to the interface, ie an array of our specified objects.     This was done purely as an example for the resource files, just to demonstrate setting a resource file whether it be JSON or an Array or an object and so on. I just randomly settled on using an Array 😄  Would you like a comment in the code as well?

bcrs-ui/src/resources/externalResourceExample.ts
cameron-freshworks: haha, your very right! This has been fixed 😆

bcrs-ui/vue.config.js
cameron-freshworks: That is a good question, i'm not sure if it was because i'm running it at localhost or because something is happening with the pathing behind the scenes in webpack, but in this instance it was required to set the correct path and find the public config file.

bcrs-ui/src/App.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+  <div id="app">
```
severinbeauvais: This `div` needs to be a `v-app` instead. Dunno why vue-cli scaffolded it this way!?
Ref: https://vuetifyjs.com/en/components/application

cameron-freshworks: Weird, let's make it right!

bcrs-ui/src/store/state/resource-model.ts
severinbeauvais: Yes, please add a short code comment. I'm wondering how to use this resource and when I would see message 1 vs message 2?

bcrs-ui/public/config/configuration.json
severinbeauvais: You need one last thing: the `devServer` config in vue.config.js so that, when running locally, the subject keycloak.json file is retrieved from the deployed auth project (so we don't need a local copy of it).

bcrs-ui/src/store/state/resource-model.ts
cameron-freshworks: Okie doke, no problem at all, thanks!

bcrs-ui/vue.config.js
severinbeauvais: OK for now.

bcrs-ui/public/index.html
severinbeauvais: Still need to either remove these 2 AddressComplete lines or check in the 2 files.

bcrs-ui/src/resources/externalResourceExample.ts
severinbeauvais: De nada

bcrs-ui/public/config/configuration.json
cameron-freshworks: K great idea. Will do!

bcrs-ui/public/index.html
cameron-freshworks: Will remove until we need 👍

bcrs-ui/src/utils/config-helper.ts
```diff
@@ -0,0 +1,72 @@
...
+      console.log('Set Pay API URL to: ' + payApiUrl)
+
+      const keycloakConfigUrl = response.data['KEYCLOAK_CONFIG_URL']
+      sessionStorage.setItem('KEYCLOAK_CONFIG_URL', keycloakConfigUrl)
+      console.info('Set KeyCloak config URL to: ' + keycloakConfigUrl)
+
+      // Extend the window with type
+      let myWindow = window as any
+
+      myWindow.addressCompleteKey = response.data['ADDRESS_COMPLETE_KEY']
```
severinbeauvais: Since you removed the AddressComplete JS and CSS files, could you please remove this also and the entry in configuration.json?

bcrs-ui/package.json
severinbeauvais: You're right, it works like that. I tested this in coops-ui but for me it matched the same tests as before so I dunno.

The reason that parameter is in there is to make it easier for devs (me) to specify a pattern to run just a single test. So I think either we remove it completely (and then I'll have to remember the parameter name when I want to run a single test) or we provide a generic pattern (`.`). Up to you.

bcrs-ui/src/utils/config-helper.ts
cameron-freshworks: yes, done!

bcrs-ui/package.json
cameron-freshworks: I'm impartial :) Added it back in!

PR #3, opened 2020-01-03
#1885 UI-Header & Vuetify Config Tweaks

cameron-freshworks:
*Issue #:* /bcgov/entity#1885

*Description of changes:*

* Restructured the UI file directory to root level.
* Implemented SBC Common components Header & Footer.
* Implemented an entity info header to display dynamic data from Name Request Response. **Currently Static**
* Added unit testing for entity info header.
* Updated configurations to fix Vuetify.


bcrs-ui/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "hello-world",
+  "name": "vue-frontEnd-Framework",
```
severinbeauvais: This should actually be this project's name (though it should be generic for the template project).

bcrs-ui/package.json
```diff
@@ -11,36 +11,36 @@
   "dependencies": {
     "@mdi/font": "^4.5.95",
     "axios": "^0.18.0",
-    "core-js": "^3.4.3",
+    "core-js": "^3.4.4",
     "css-loader": "^0.28.11",
     "keycloak-js": "^6.0.0",
```
severinbeauvais: Don't need 'keycloak.js' (as it's now imported by sbc-common-components).

bcrs-ui/src/plugins/vuetify.js
```diff
@@ -0,0 +1,7 @@
+import Vue from 'vue'
```
severinbeauvais: This is odd: I see another file with the same name and content below !?!?

bcrs-ui/package.json
cameron-freshworks: Good point! I will rename this, and i'm really glad you mentioned the template as i will have to PR these same updates to the Entities repo where we are keeping a snapshot. Moving forward after this i think it makes sense to personalize to the app ( like you suggested! )

bcrs-ui/package.json
cameron-freshworks: Thank you, will remove!

bcrs-ui/src/plugins/vuetify.js
cameron-freshworks: Very odd! I wonder how the .js file was created. I removed it.

cameron-freshworks:
@severinbeauvais  Updated, ready for re-review 👍

src/components/common/Header.vue
```diff
@@ -0,0 +1,73 @@
+<template>
```
severinbeauvais: Didn't we think we should use the header from sbc-common-components?

cameron-freshworks: You are correct sir, I implemented the common components but forgot to remove this placeholder. Will remove!

cameron-freshworks:
> Looks great!

Thanks!

cameron-freshworks:
Two approvals, merging!

PR #4, opened 2020-01-08
#2116 Front End Authentication Implementation

cameron-freshworks:
*Issue #:* /bcgov/entity#2116

** DRAFT **  

* Looking for feedback with regards to the extrapolation of the JWT and Auth Methods out of the APP component and this new approach to retrieving data pre-mounting of the app.

* Still need to apply the getters.
 
*Description of changes:*

* Implements an Authorization class which handles the fetching, parsing and storing the keycloak roles as well as the authorization roles.


src/interfaces/state-interface.ts
```diff
@@ -1,4 +1,10 @@
-// Interface to define the state model example
+// State model example
 export interface StateModelIF {
   stateText: string | undefined
 }
+
+// Tombstone State model
```
severinbeauvais: Should be in its own file.

src/interfaces/utils-interface.ts
```diff
@@ -0,0 +1,10 @@
+// Interface to define the JWT
+export interface JwtIF {
+  roles: Array<string>
+}
+
+export interface AuthIF {
+  data: {
```
severinbeauvais: Curious if this IF allows a variable of this type to be null?

src/store/state/tombstone-model.ts
```diff
@@ -0,0 +1,6 @@
+import { TombStoneIF } from '@/interfaces'
+
+export const tombStoneModel: TombStoneIF = {
+  keycloakRoles: [],
```
severinbeauvais: Is the intent for these to use the interfaces you declared above?

src/store/mutations/mutations-model.ts
```diff
@@ -5,3 +5,11 @@ export const mutateName = (state: any, name: string) => {
 export const mutateResource = (state: any, resource: object) => {
   state.resourceModel = resource
 }
+
+export const mutateKeycloakRoles = (state: any, keyCloakRoles: Array<string>) => {
```
severinbeauvais: Is the intent for 'keyCloakRoles' and 'authRoles' below to use the interfaces you declared above?

src/main.ts
```diff
@@ -54,10 +54,17 @@ fetchConfig()
...
-    tokenServices.initUsingUrl(sessionStorage.getItem('KEYCLOAK_CONFIG_URL') || '')
+    await tokenServices.initUsingUrl(sessionStorage.getItem('KEYCLOAK_CONFIG_URL') || '')
       .then(() => tokenServices.scheduleRefreshTimer())
       .catch((err: string) => console.error(err))
 
+    // Store KeyCloak & Auth roles
+    console.info('Fetching JWT Roles and Authentication roles...')
+    const authService = new AuthenticationService()
+    await authService.getJwtRoles()
+    await authService.getAuthorizations()
```
severinbeauvais: For these, what's the error handling if we can't get the roles/authorizations or they are invalid?

src/constants/constants.ts
```diff
@@ -0,0 +1,3 @@
+// action constants
+export const SET_KEYCLOAK_ROLES = 'setKeyCloakRoles'
```
severinbeauvais: As these are used for store purposes only, can they be moved to the store folder?

src/interfaces/state-interface.ts
cameron-freshworks: Can do!

src/interfaces/utils-interface.ts
cameron-freshworks: I believe without making it optional or giving it `| null`, it would expect a value.

src/store/state/tombstone-model.ts
cameron-freshworks: Yes, you bet. This is just the initialization of the model. So when we bind this model to the component we can call all of the properties simply. See the StateExample component line 48.
No matter how many properties we have in this matter we can access them all in the component with the short cut binding from Vuex-class. `@State modelName: type`.    I'm sure they is many ways to do this, so i'm totally open to ideas!

src/interfaces/utils-interface.ts
severinbeauvais: So you'd have to initialize (empty) such a variable to = { data: { roles: [ ] } } ?

I'm still learning how to use interfaces -- I think they're like structs -- but I'm thinking it would be nice to be able to initial the variable as null.  It may be possible as-is. Thoughts?

src/store/mutations/mutations-model.ts
cameron-freshworks: Not in these instances. The mutators refer to specific values and therefor just need the type of that specific property, whereas the interfaces were more for defining the model as a whole.

src/main.ts
cameron-freshworks: Thats a great point, if we wanted to handle the errors in the same way as with Coops-ui ( Dialogs ) it will definitely need a little more work. As you mentioned below this could be solved by doing this after we mount Vue, i will explore a little further though. And yes, 100% we can always do better :)

src/constants/constants.ts
cameron-freshworks: Yea, that makes sense. Will make a store constants for dispatch references!

cameron-freshworks:
> Cool!
> 
> I think my concerns about error handling might go away if you moved the Authentication Service to App.vue. Thoughts?

I definitely agree! I will explore the options a little further, but that is 100% a good option.

src/store/state/tombstone-model.ts
severinbeauvais: I'm wondering why the fields are initialized as empty arrays instead of using the interface types? I may be confused... might make more sense when I pull these changes and run it for myself :)

src/store/mutations/mutations-model.ts
severinbeauvais: Ah, I see!

src/interfaces/utils-interface.ts
cameron-freshworks: Great question for sure! Interfaces are just describing what we, as developers, expect to see within the object. It certainly doesn't always paint the whole picture, as there can be potentially be MUCH more than what we describe in the interface. 

So in this instance specifically, we don't need to initialize the object with empty values, just define what we expect, because we are assigning this type to a response object from the API. At this point, the only thing we expect from this response is { data: { roles: Array<string> } }.  We could make data optional, but i believe the way we are using it, it shouldn't be, as the error handling will throw Error if the data property isn't present in the response.

cameron-freshworks: Did any of that explanation help :) ?

PR #5, opened 2020-01-09
#2116 Front End Authentication

cameron-freshworks:
*Issue #:* /bcgov/entity#2116

*Description of changes:*

* Implements an Authorization class which handles the fetching, parsing and storing the keycloak roles as well as the authorization roles.

* Implements Store Getters.

* Expands Store interfaces & Constants.

* Added Store getters example usage.

** NOTES ** 

* I feel like the dialogs and error handling can be done outside of this this ticket, as it is still undecided how to best handle that. 

Current state: If the user cannot retrieve the authentication, it will return them to the log in url. 

Possible Future state: We relocate the AuthService calls to `App.vue` and display a dialog to redirect user back to Auth.

* Local Development hack to allow log in is still implemented in main.ts file for ease of development. Is pegged for removal when we move to deployment and coordinate login credentials with Relationships.


cameron-freshworks:
> Looks good. Please create a new ticket to capture the "future" items from your notes above.

100% Will do, thanks!

cameron-freshworks:
> Looks good. Please create a new ticket to capture the "future" items from your notes above.

Ticket created for reference, let me know if/what i should add to it! https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/2332

PR #7, opened 2020-01-13
created app shell and misc stubbed views and components

severinbeauvais:
*Issue #:* /bcgov/entity#2335

*Description of changes:*
- created app shell in App.vue (with Fee Summary component)
- created stub Stepper component
- created stub Actions component
- created stub Define Company component
- created stub Review Confirm component
- updated routes accordingly
- added vue-affix package for Fee Summary component
- added vue2-filters package for Fee Summary component
- temporarily disabled some strict options due to Fee Summary component
- added EntityTypes and FilingCodes
- checked in package-lock.json (needed for GH Actions)


src/App.vue
```diff
@@ -18,17 +57,42 @@ import { Component, Vue } from 'vue-property-decorator'
...
   }
 })
-export default class App extends Vue {}
-</script>
+export default class App extends Vue {
+  private filingData: Array<FilingDataIF> = []
+  private totalFee: number = 0
+
+  private created (): void {
+    this.filingData.push({ filingTypeCode: 'OTINC', entityType: 'CP' })
```
cameron-freshworks: Should we start filingCodes and entityTypes ENUMs for these property values?

src/components/common/EntityInfo.vue
```diff
@@ -54,7 +54,6 @@ export default class EntityInfo extends Vue {
 }
 </script>
 
-<!-- eslint-disable max-len -->
```
cameron-freshworks: Thanks!

src/components/common/Stepper.vue
```diff
@@ -0,0 +1,40 @@
...
+
+  </v-container>
+</template>
+
+<script lang="ts">
+import { Component, Vue, Watch } from 'vue-property-decorator'
+
+@Component
+export default class Stepper extends Vue {
+  @Watch('$route', { immediate: true })
```
cameron-freshworks: What does the option `immediate: true` do?

src/interfaces/index.ts
```diff
@@ -7,3 +7,6 @@ export * from './store-interfaces/state-interface'
 export * from './store-interfaces/action-interface'
 export * from './store-interfaces/tombstone-interface'
 export * from './store-interfaces/getter-interface'
+
+// Others
+export * from './filingData-interface'
```
cameron-freshworks: I like it! As far as directory structure ( thinking ahead ), would this fall into any interface type category that we might add too?

src/views/DefineCompany.vue
```diff
@@ -0,0 +1,28 @@
...
+      </section>
+    </v-card>
+  </div>
+</template>
+
+<script lang="ts">
+// Libraries
+import { Component, Vue } from 'vue-property-decorator'
+
+@Component({})
```
cameron-freshworks: I know we were used to putting in the `({options})` but it isn't required. Is this how you would like to see the decorator done?

src/views/ReviewConfirm.vue
```diff
@@ -0,0 +1,28 @@
...
+      </section>
+    </v-card>
+  </div>
+</template>
+
+<script lang="ts">
+// Libraries
+import { Component, Vue } from 'vue-property-decorator'
+
+@Component({})
```
cameron-freshworks: Same as above, just curious if you would prefer this!

tsconfig.json
```diff
@@ -2,7 +2,13 @@
   "compilerOptions": {
     "target": "esnext",
     "module": "esnext",
-    "strict": true,
```
cameron-freshworks: Why the change here? Is there a specific property that you didn't want strict that was covered in the list?

src/App.vue
severinbeauvais: Sure, I can add those.

severinbeauvais: Done.

src/components/common/Stepper.vue
severinbeauvais: You get an initial event (before the value changes).

There's also an option `deep: boolean` that generates an event when an object's properties change.

src/interfaces/index.ts
severinbeauvais: I'm not sure yet, which is why I didn't create one :)

I expect this will either be clearer in the future or not matter :)

src/views/DefineCompany.vue
severinbeauvais: Done like this. At one point I got lint errors when I didn't specify the `{}`.

src/views/ReviewConfirm.vue
severinbeauvais: I don't think I have a preference.

tsconfig.json
severinbeauvais: As you know, `strict` covers all the other properties. I needed to disable a couple of properties specifically (due to SbcFeeSummary) but I also wanted to keep the other checks enabled.

src/App.vue
```diff
@@ -78,7 +81,7 @@ export default class App extends Vue {
   private totalFee: number = 0
 
   private created (): void {
-    this.filingData.push({ filingTypeCode: 'OTINC', entityType: 'CP' })
+    this.filingData.push({ filingTypeCode: FilingCodes.INCORPORATION_BC, entityType: EntityTypes.BCOMP })
```
cameron-freshworks: Right on! 👍

src/components/common/Stepper.vue
cameron-freshworks: interesting, thanks!

src/interfaces/index.ts
cameron-freshworks: Sounds good, true agile ;)

tsconfig.json
cameron-freshworks: If you feel this doesn't compromise our goal of being strict and won't carry over into other elements of development, i'm 100% fine with it. Just at first glance, those 3 options certainly do lessen the rules. Is this the only way?

cameron-freshworks: Discussed with Severin that this is a temporary change until we can fix some type issues with sbc common components. Resolved!

severinbeauvais: I have created ticket /bcgov/entity#2357 to get the SbcFeeSummary updated so that we can go back to `strict: true`.

PR #11, opened 2020-01-15
Certify UI

leksmall:
*Issue #:* /bcgov/entity#2134

*Description of changes:*
Added the certify component
Added dynamic resource loading
Refactored the resource interface and model

Todo - Add unit test


src/App.vue
```diff
@@ -52,20 +52,28 @@
 
 <script lang="ts">
 // Libraries
-import { Component, Vue } from 'vue-property-decorator'
+import { Component, Vue, Mixins } from 'vue-property-decorator'
+import { mapState, mapActions } from 'vuex'
```
cameron-freshworks: Not sure we need these if we are using the Vuex-class decorators to shortcut the bindings.

src/components/ReviewConfirm/Certify.vue
```diff
@@ -0,0 +1,165 @@
...
+        {{certifyStatementResource.certifyClause}}
+      </p>
+    </div>
+  </v-card>
+</template>
+
+<script lang="ts">
+import { Component, Vue, Prop, Emit } from 'vue-property-decorator'
+import { State } from 'vuex-class'
+import { CertifyStatementIF } from '../../interfaces'
```
cameron-freshworks: I believe we can reference the import `@/interfaces`

src/mixins/resource-lookup-mixin.ts
```diff
@@ -3,22 +3,23 @@ import { Component, Vue } from 'vue-property-decorator'
 import { State } from 'vuex-class'
 
 // Interfaces
-import { ResourceExampleIF } from '@/interfaces'
+import { ExternalResourceIF } from '@/interfaces/resource-interfaces/ExternalResourceIF'
```
cameron-freshworks: Should be able to just import from `@/interfaces` if the index is exporting * from the specific interface folder.

src/store/state/state-model.ts
```diff
@@ -1,5 +1,10 @@
 import { StateModelIF } from '@/interfaces'
 
 export const stateModel: StateModelIF = {
-  stateText: 'Base State Text'
+  stateText: 'Base State Text',
+  currentDate: null,
+  certifyState: {
+    certifyFormValid: false,
+    certifiedBy: null
+  }
```
cameron-freshworks: This state model was initially put in just as an example. With the setup we have going we can group specific and related properties into these models. That being said we can totally have a generic model for this type of data but would we want to rename it? Possible also removed the example data property `stateText` if we were to use this model for generic data.

src/store/state/tombstone-model.ts
```diff
@@ -2,5 +2,6 @@ import { TombStoneIF } from '@/interfaces'
 
 export const tombStoneModel: TombStoneIF = {
   keycloakRoles: [],
-  authRoles: []
+  authRoles: [],
+  entityType: null
```
cameron-freshworks: When our typescript settings are set back to full strict, in this case you will get a `strictnullcheck` error for initializing with null. To allow this, you might want to add `| null` in the interface to allow this. Also, might not be a bad idea, during development to enable all the ts strict settings in tsconfig.json file, just to avoid these when we do revert back.

src/views/ReviewConfirm.vue
```diff
@@ -6,18 +6,72 @@
...
+                <h2>2. Completing Party Statement</h2>
+              </header>
+              <Certify
+                :isCertified.sync="isCertified"
+                :certifiedBy.sync="certifiedBy"
+                @valid="certifyFormValid=$event"
+                :date="currentDate"
+                :certifyStatementResource="certifyStatementResource"
+
+              />
```
cameron-freshworks: Spacing is a little bit off in this section.

thorwolpert:
I removed the image build from the PR script as it has issues with the tagging.  all of the unit tests and linting passed.

src/App.vue
leksmall: You are right. We dont need this. Will remove those

src/components/ReviewConfirm/Certify.vue
leksmall: Sure. I think these were auto imported. Will change the relative path

src/store/state/state-model.ts
leksmall: My understanding was this will hold the application state and any common properties like date etc which will be used by the whole application. Initially I thought of reorganizing the store into modules with namespaces. But that will be a bit too much because we are dealing with one application even though it has different steps. My preference would be using this to put anything that represents the current state of the application. We can rename it. I will remove the sample property

src/store/state/tombstone-model.ts
leksmall: Sure

src/App.vue
cameron-freshworks: cool!

src/store/state/state-model.ts
cameron-freshworks: Sounds good, totally agree!

src/resources/certifyStatementResource.ts
```diff
@@ -0,0 +1,37 @@
...
+    'the Articles and the Incorporation Agreement both contain a signature line for ' +
+    'each person identified as an incorporator in the Incorporation Application with ' +
+    'the name of that person set out legibly under the signature line,',
+    'an original signature has been placed on each of those signature lines, and',
+    'I have no reason to believe that the signature placed on a signature line is not ' +
+    'the signature of the person whose name is set out under that signature line.'
+  ],
+  certifyClause: 'It is an offence to make a false or misleading statement in respect ' +
+  'of a material fact in a record submitted to the Corporate Registry for filing. ' +
+  'See section 427 of the Business Corporations Act.'
```
severinbeauvais: This is not the correct text/reference for Cooperatives. We need to remember that this will need to be updated when we implement Coops.

src/views/ReviewConfirm.vue
```diff
@@ -6,18 +6,70 @@
...
+      {
+        certifyFormValid: val,
+        certifiedBy: this.certifiedBy
+      }
+    )
+  }
+
+  @Watch('certifiedBy')
+  private onCertifiedByChange (val: boolean): void {
+    this.setCertifyState(
```
severinbeauvais: This code is overly complicated as it has too much going on.

If I understand correctly, when the user enters "certified by", the certify component emits an event that updates the local property certifiedBy, which is watched in order to set the global certify state.

I think this should be simplified! I think you could remove the "sync" from the props and then call a handler method for each prop update event. The handlers could call setCertifyState(). This would basically replace the watchers with event handlers but it would cut out one interaction (ie, call handler directly instead of watch) and I think it would be much easier to understand.

Or maybe there's another way to bind the props to the states?

What do you think?

src/resources/certifyStatementResource.ts
leksmall: This is a placeholder for coops. Infact the text for bcomp is also not right from what I heard yesterday from Linda. The text will change for both coops and bcomp once the BA team has finalized the requirements. This is what we have at present in the UX pin

src/App.vue
```diff
@@ -52,20 +52,27 @@
...
 // Interfaces
-import { FilingDataIF } from '@/interfaces'
+import { FilingDataIF, ActionBindingIF, CertifyStatementIF } from '@/interfaces'
+
+import { CertifyStatementResource } from '@/resources'
 
 // Enums
 import { EntityTypes, FilingCodes } from '@/enums'
 
+import { State, Action } from 'vuex-class'
```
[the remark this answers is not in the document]
leksmall: Done

src/interfaces/store-interfaces/state-interface.ts
```diff
@@ -1,4 +1,8 @@
+import { CertifyIF } from '../stepper-interfaces/certify-interface'
+
 // State model example
 export interface StateModelIF {
-  stateText: string | undefined
+  stateText: string | undefined,
+  currentDate: string,
+  certifyState: CertifyIF
```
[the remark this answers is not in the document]
leksmall: We need to discuss about how each step will update the state.If each step individually updates the state, we may not need anything additional in step 5 because it is a read only view of what is present in other steps. But this needs to be discussed and refactored if needed

tests/unit/Certify.spec.ts
```diff
@@ -0,0 +1,214 @@
+//
+// Copyright © 2019 Province of British Columbia
```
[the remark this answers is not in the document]
leksmall: Sure will do

src/views/ReviewConfirm.vue
leksmall: I dont think it is complicated and we have similar interaction patterns in coops. But in this scenario I agree that use of a watcher and thus one interaction can be avoided. Will make the change to invoke a handler.

src/interfaces/store-interfaces/state-interface.ts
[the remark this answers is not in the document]
severinbeauvais: OK.

As I understand it, step 5 will have several sections to prompt for additional data, so I guess those will need state and validation as well.

src/views/ReviewConfirm.vue
leksmall: The component is modified to use handler. As part of this I removed the prop isCertify which does not make any sense. The parent component need not pass that to certify even in the case of save and resume

src/interfaces/stepper-interfaces/certify-interface.ts
```diff
@@ -0,0 +1,4 @@
+export interface CertifyIF {
+    certifyFormValid: boolean;
+    certifiedBy: string;
```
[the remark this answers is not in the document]
leksmall: I dint notice any lint errors. In general I noticed many errors when strict mode is enabled .That can go as a separate PR in my opinion and need not block this one

src/mixins/resource-lookup-mixin.ts
```diff
@@ -3,22 +3,23 @@ import { Component, Vue } from 'vue-property-decorator'
...
+  resourceModel!: Array<ExternalResourceIF>
 
   /**
    * Method to return the specified message
    *
    * @param id ID a number indicating the id of the resource to look up.
    */
   getName (id: number): string {
-    const user: ResourceExampleIF | undefined = this.resourceModel && this.resourceModel.find(user => user.id === id)
+    const user: ExternalResourceIF | undefined = this.resourceModel && this.resourceModel.find(user => user.id === id)
```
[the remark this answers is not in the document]
leksmall: Not sure about this. I just changed the type here. But I have seen similar pattern in method return type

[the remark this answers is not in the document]
severinbeauvais: I just tested this (with strictNullChecks:true)... and the answer is yes, you need this. Carry on :)

src/interfaces/stepper-interfaces/certify-interface.ts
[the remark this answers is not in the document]
severinbeauvais: I just tested this (with strictNullChecks:true)... and the answer is yes, you should declare a property with `| null` if you ever need to assign it to null (as you do in state-model.ts)... therefore please update this declaration.

src/components/ReviewConfirm/Certify.vue
```diff
@@ -0,0 +1,162 @@
...
+  /**
+   * Computed value.
+   * @returns The trimmed "Certified By" string (may be '').
+   */
+  private get trimmedCertifiedBy (): string {
+    // remove repeated inline whitespace, and leading/trailing whitespace
+    return this.certifiedBy && this.certifiedBy.replace(/\s+/g, ' ').trim()
+  }
+
+   @Emit('certifiedByChange')
```
severinbeauvais: According to https://vuejs.org/v2/guide/components-custom-events.html#sync-Modifier, you should have been able to retain the previous code `@Emit('update:certifiedBy')` and then changed the parent to `v-on:update:certifiedBy="onCertifiedByChange($event)"` (or `@update:certifiedBy...`). I recommend you try this as it should result in cleaner code (ie, prop name = event name).

src/views/ReviewConfirm.vue
severinbeauvais: Thank you for the handler usage. It makes the code easier for me to understand and review :)

src/App.vue
```diff
@@ -91,10 +96,18 @@ export default class App extends Vue {
...
+  private entityType: string = 'CP'
+
+  @Action('setEntityType') setEntityType!: ActionBindingIF
+  @Action('setCurrentDate') setCurrentDate!: ActionBindingIF
+  @Action('setCertifyStatementResource') setCertifyStatementResource!: ActionBindingIF
 
   // Lifecycle event
   private created (): void {
     this.filingData.push({ filingTypeCode: FilingCodes.INCORPORATION_BC, entityType: EntityTypes.BCOMP })
+    this.setEntityType(this.entityType)
```
severinbeauvais: Will this be moved to step1?

If so, it might make sense to have a watcher for entityType and then set the CertifyStatementResource in the watcher.

Thoughts?

src/interfaces/stepper-interfaces/certify-interface.ts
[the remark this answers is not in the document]
leksmall: I updated this and fixed a bunch of other places where strict null check was giving issues.

src/App.vue
leksmall: That will move from App.vue but as part of another pr which will have the finalized text for bcomp

src/components/ReviewConfirm/Certify.vue
severinbeauvais: OK

PR #12, opened 2020-01-15
implemented Define Company page + lots of state stuff + unit tests!

severinbeauvais:
*Issue #:* /bcgov/entity#2359

*Description of changes:*
- more work on Actions component
- more work on dummy Stepper component
- added vue-test-utils-helpers package
- added unit tests for Actions, App, DefineCompany, ReviewConfirm and Stepper
- lots of global state management
- lots of getters for state (business logic)
- app now automatically sets step number
- more usage of state
- added NR Generator buttons to Define Company page
- added startup check to Review Confirm page
- renamed state-getters to tombstone-getters
- misc cleanup

Note that I did not fully flesh out the unit tests because the components I updated are still in flux (Stepper, Actions, DefineCompany).


src/components/common/Actions.vue
```diff
@@ -41,28 +65,77 @@
...
-  private get showReviewConfirmBtn (): boolean {
-    return (this.$route.meta.step < 2)
+  /**
+   * Method called when Save and Resume button is clicked.
+   * @returns A promise (ie, this is an async method).
+   */
+  private async onClickSaveResume (): Promise<void> {
+    this.setIsSavingResuming(true)
+    await this.sleep(1000)
+    this.setIsSavingResuming(false)
```
cameron-freshworks: This is neat! but why not timeout the second method call to `false`? Do we expect to be using the sleep() method in more places than this?

src/interfaces/store-interfaces/tombstone-interface.ts
```diff
@@ -1,5 +1,6 @@
 // Tombstone State model
 export interface TombStoneIF {
-  keycloakRoles: Array<string>,
+  keycloakRoles: Array<string>
   authRoles: Array<string>
+  entityType: string | null
```
cameron-freshworks: right on 👍  Good catch, i'm sure your already doing this but i know for myself i will have the strict settings on just while developing locally until we can get the fee summary playing nice!

src/main.ts
```diff
@@ -48,26 +52,26 @@ sessionStorage.setItem('USER_FULL_NAME', 'Firstname Lastname')
...
+    const tokenServices = new TokenServices()
+    await tokenServices.initUsingUrl(sessionStorage.getItem('KEYCLOAK_CONFIG_URL') || '')
+      .then(() => tokenServices.scheduleRefreshTimer())
+      .catch((err: string) => console.error(err))
+
+    // Store KeyCloak & Auth roles
+    console.info('Fetching JWT Roles and Authentication roles...')
+    const authService = new AuthenticationService()
+    await authService.getJwtRoles()
+    await authService.getAuthorizations()
```
cameron-freshworks: Is Thor ok with the Auth enabled at this point?

src/router/routes.ts
```diff
@@ -1,6 +1,6 @@
-import Home from '@/views/Home.vue'
-import MixinExample from '@/views/MixinExample.vue'
-import StateExample from '@/views/StateExample.vue'
+// import Home from '@/views/Home.vue'
+// import MixinExample from '@/views/MixinExample.vue'
+// import StateExample from '@/views/StateExample.vue'
 import DefineCompany from '@/views/DefineCompany.vue'
 import ReviewConfirm from '@/views/ReviewConfirm.vue'
```
cameron-freshworks: cool! Not sure how you feel about this, but in this UI it appears we will have a `View` for every step. So more than we did with COOPS-UI. Would it make sense for us to include an Index file for the views to clean up our Imports as the Views grow in number?

src/router/routes.ts
```diff
@@ -43,12 +43,16 @@ export const routes = [
...
+  // TODO: handle steps differently because they
+  //       differ depending on company type (BC, CP)
+  //
   {
     path: '/review-confirm',
     name: 'review-confirm',
     component: ReviewConfirm,
     meta: {
-      step: 2,
+      step: 5,
```
cameron-freshworks: i like the use of the meta data here!

src/components/common/Actions.vue
severinbeauvais: This was just a quick hack to simulate something happening.

Yes, the second method could have been called after a timeout, but I think this is easier to understand (and will probably be the same code when we're done: simply replace the sleep by the actual work).

src/store/state/state-model.ts
```diff
@@ -1,5 +1,9 @@
 import { StateModelIF } from '@/interfaces'
 
 export const stateModel: StateModelIF = {
-  stateText: 'Base State Text'
+  stateText: 'Base State Text',
+  currentStep: 1,
+  isSaving: false,
+  isSavingResuming: false,
+  isFilingPaying: false
```
cameron-freshworks: I mentioned something very similar in Lekshmi's PR. This state model was initially just an example model. Would it make sense to group some of the properties in their own model? Also we could use this as a generic model, if you feel the data is just that. In that case, we should also remove the example property `stateText` 👍

src/main.ts
severinbeauvais: I'll undo this. I had it enabled so I could test the pay API and fee summary (which work btw).

(Done and pushed).

src/router/routes.ts
severinbeauvais: Maybe but I'm not exactly sure where to go with this, because we'll have different views depending on the selected business type. So for now, my vote is don't do unknown work. What do you think?

src/router/routes.ts
severinbeauvais: Thanks but again, we'll have different steps depending on business type, and how we manage those will affect the routing and knowing what step we're on. My vote is leave it for now. What do you think?

src/components/common/Actions.vue
cameron-freshworks: sounds great! 👍

src/main.ts
cameron-freshworks: awesome :) Was just curious, thank you!

src/router/routes.ts
cameron-freshworks: Totally, adding an index is 2 min job down the road, if we feel like we need one. Sounds good!

src/store/state/state-model.ts
severinbeauvais: I think it doesn't make sense to create other models yet, with just this data.

But soon, we will have more data (for each step) (and also page validation state). At that point, yes, it makes sense to group that data into its own model as needed.

src/router/routes.ts
cameron-freshworks: I agree to leave it for now as well. Down the road we might want to use the entityType to define what `steps` are involved in the stepper. Possible defining each step in an enum, and depending on what entity type we can build an array of steps from said enum. Which the Stepper can then use to navigate. How this will tie into the routing, not 100% sure lol. Just a random thought!

severinbeauvais: Separate enums might work, or an array of steps, I dunno. It might be neat to tie in to Peter's resource stuff too. I'll defer this to later!

src/store/state/state-model.ts
cameron-freshworks: Sounds good!

PR #13, opened 2020-01-16
Changes to consider the certify form state for enabling/disabling fil…

leksmall:
*Issue #:* /bcgov/entity#2129

*Description of changes:*
Changes to consider the certify form state for enabling/disabling file and pay button
Changes to the revised certify clause for BComp
Added more tests to actions
Changes to set the certify resource based on changes in entity type

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -96,7 +96,6 @@ export default class App extends Mixins(DateMixin) {
 
   private filingData: Array<FilingDataIF> = []
   private totalFee: number = 0
-  private entityType: string = 'CP'
 
   @Action('setEntityType') setEntityType!: ActionBindingIF
   @Action('setCurrentDate') setCurrentDate!: ActionBindingIF
@@ -105,9 +104,7 @@ export default class App extends Mixins(DateMixin) {
   // Lifecycle event
   private created (): void {
     this.filingData.push({ filingTypeCode: FilingCodes.INCORPORATION_BC, entityType: EntityTypes.BCOMP })
-    this.setEntityType(this.entityType)
     this.setCurrentDate(this.dateToUsableString(new Date()))
-    this.setCertifyStatementResource(CertifyStatementResource.find(x => x.entityType === this.entityType))
   }
 
   /**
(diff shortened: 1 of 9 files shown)


src/App.vue
```diff
@@ -140,5 +137,11 @@ export default class App extends Mixins(DateMixin) {
   private onRouteChanged (): void {
     this.setCurrentStep(this.$route.meta.step)
   }
+
+  @Watch('tombStoneModel.entityType')
+  private onEntityTypeChanged (val:string | null) : void{
+    this.setCertifyStatementResource(val ? CertifyStatementResource
+      .find(x => x.entityType === val) : null)
+  }
```
cameron-freshworks: This is interesting! I can see this watcher ( or something similar) being expanded in the future to cover more than just the certify statement resource but also dictate what type of steps an entity has or anything else entity specific that could be handle by a resource file. Very cool.

src/App.vue
```diff
@@ -96,7 +96,6 @@ export default class App extends Mixins(DateMixin) {
 
   private filingData: Array<FilingDataIF> = []
   private totalFee: number = 0
-  private entityType: string = 'CP'
 
   @Action('setEntityType') setEntityType!: ActionBindingIF
```
cameron-freshworks: I know you didn't work on this in this PR but are we using this Binding in this component?

leksmall: Good catch! Will remove it. Thanks

src/store/getters/state-getters.ts
```diff
@@ -32,7 +32,8 @@ export const isEnableFilePayBtn = (state: any): boolean => {
   const step3Valid = true // FOR TESTING ONLY
   const step4Valid = true // FOR TESTING ONLY
   const step5Valid = true // FOR TESTING ONLY
-  return (step1Valid && step2Valid && step3Valid && step4Valid && step5Valid)
+  return (step1Valid && step2Valid && step3Valid && step4Valid && step5Valid &&
+    state.stateModel.certifyState.certifyFormValid)
```
severinbeauvais: I agree to leave this in as "for testing" for now. Ultimately "certifyFormValid" state should roll up to "step5Valid" state.

PR #18, opened 2020-01-21
Business Contact UI + Store structure for stepper + Error Handling example

leksmall:
*Issue #:* /bcgov/entity#2344

*Description of changes:*
Added Business Contact UI
Refactored the store structure
Added validation error handling
Updated existing tests to be in sync with the changes


src/views/DefineCompany.vue
```diff
@@ -1,9 +1,9 @@
 <template>
   <div>
+    <section>
     <header>
```
severinbeauvais: Please indent code in section. Also below (lines 33+).

src/views/DefineCompany.vue
```diff
@@ -28,27 +28,56 @@
...
       </div>
     </v-card>
+    </section>
+    <p></p>
+    <section>
+      <header>
+        <h2>3. Business Contact Information</h2>
+      </header>
+      <BusinessContactInfo
+      :initialValue = businessContact
```
severinbeauvais: Please indent element additional lines.

src/views/ReviewConfirm.vue
```diff
@@ -76,8 +83,15 @@ export default class ReviewConfirm extends Mixins(ResourceLookupMixin) {
 </script>
 
 <style lang="scss" scoped>
-.step-container {
```
severinbeauvais: I think "step-container" class is now obsolete. If so, please remove it from layout.scss also.

src/interfaces/stepper-interfaces/DefineCompany/define-company-interface.ts
```diff
@@ -0,0 +1,6 @@
+import { BusinessContactIF } from './business-contact-interface'
+
+export interface DefineCompanyIF {
+    defineCompanyStepValid: boolean
```
severinbeauvais: I suggest renaming this since we already know we're in the Define Company IF, and also it would be nice if all step IFs had the same "valid" property name.

src/views/DefineCompany.vue
```diff
@@ -28,27 +28,56 @@
...
   // Global getters
   @Getter isEntityType!: GetterIF
 
   // Global actions
   @Action setEntityType!: ActionBindingIF
+  @Action setBusinessContact!: ActionBindingIF
+  @Action setDefineCompanyStepValidity!: ActionBindingIF
+
+  // Local Properties
+  private businessContactFormValid:boolean = false
```
severinbeauvais: This doesn't seem to be needed.

src/components/ReviewConfirm/Summary.vue
```diff
@@ -0,0 +1,31 @@
...
+  </div>
+</template>
+
+<script lang="ts">
+// Libraries
+import { Component, Vue } from 'vue-property-decorator'
+import { Getter, Action, State } from 'vuex-class'
+
+// Interfaces
+import { BusinessContactIF } from '@/interfaces'
```
severinbeauvais: Looks like this isn't needed here.

src/components/ReviewConfirm/Summary.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <div class="summary-container">
+    <SummaryDefineCompany/>
+  </div>
+</template>
+
+<script lang="ts">
+// Libraries
+import { Component, Vue } from 'vue-property-decorator'
+import { Getter, Action, State } from 'vuex-class'
```
severinbeauvais: Looks like this isn't needed here.

src/components/DefineCompany/BusinessContactInfo.vue
```diff
@@ -0,0 +1,154 @@
+<template>
+  <div>
+    <div v-if="!isEditing">
+      <div>Email Address: {{contact.email}}</div>
```
severinbeauvais: Do you alternate values in case we are in summary mode and some property hasn't been entered?

For example, `{{ contact.email || '-' }}` ? (Don't implement this. I'm just asking if something is needed.)

src/components/DefineCompany/BusinessContactInfo.vue
```diff
@@ -0,0 +1,154 @@
...
+      <div>Phone: {{contact.phone}}   ext:{{contact.phoneExtension}}</div>
```
severinbeauvais: Check with Scott but I think "ext" should not be displayed if phoneExtension isn't entered.

src/components/Summary/SummaryDefineCompany.vue
```diff
@@ -0,0 +1,58 @@
...
+          <v-icon color="#1976d2">mdi-information-outline</v-icon>
+          This step is not complete.
+          <router-link :to="{path:'/define-company', query: {showErrors: true}}">
+             Return to this step to complete it.
+          </router-link>
+         </span>
+      </div>
+      <div class="business-contact-container">
+        <BusinessContactInfo
+        :initialValue = businessContact
```
severinbeauvais: Indent.

src/components/DefineCompany/BusinessContactInfo.vue
```diff
@@ -103,7 +103,7 @@ export default class BusinessContactInfo extends Vue {
   ]
 
   private extensionRules = [
-    (v: any) => !v || (v.length >= 0 || v.length <= 3) || 'Extension is invalid'
+    (v: any) => !v || (v.length >= 0 || v.length <= 5) || 'Extension is invalid'
```
severinbeauvais: Should this be `length >= 1 && length <= 5`?

src/components/DefineCompany/BusinessContactInfo.vue
```diff
@@ -0,0 +1,154 @@
...
+  // Props
+ @Prop({
+   default: ():BusinessContactIF => ({
+     email: '',
+     confirmEmail: '',
+     phone: '',
+     phoneExtension: ''
+   })
+ })
+  private initialValue: BusinessContactIF
```
cameron-freshworks: Just curious, do we need to re-initialize these values. From what i can see they are defined in the store as strings ( not null or undefined ) and we then pass down that from DefineCompany into the component. 

Was it throwing errors when trying to just pass the prop as `@Prop() private initialValue: BusinessContactIF` ? If there was a type error, we could also tell typescript we can be sure it isn't null or undefined by added a `!` at the end of the property name.

src/components/DefineCompany/BusinessContactInfo.vue
```diff
@@ -0,0 +1,154 @@
...
+  private showErrors: boolean
+
+  // Properties
+  private contact: BusinessContactIF = this.initialValue
+
+  private formValid : boolean = false
+
+  // Rules
+  private emailRules = [
+    (v: any) => !!v || 'Email address is required',
```
cameron-freshworks: I'm not 100% sure but i believe we know these values will be strings.  So possibly `(v: string)` might be more suitable?

src/components/Summary/SummaryDefineCompany.vue
```diff
@@ -0,0 +1,58 @@
...
+
+@Component({
+  components: {
+    BusinessContactInfo
+  }
+})
+export default class SummaryDefineCompany extends Vue {
+  // State
+  @State(state => state.stateModel.defineCompanyStep.defineCompanyStepValid)
+  readonly defineCompanyStepValid!: boolean
```
cameron-freshworks: This is pretty neat. Is this a way of deconstructing the model at the same time as binding it?

src/views/DefineCompany.vue
```diff
@@ -28,27 +28,56 @@
         </div>
       </div>
     </v-card>
+    </section>
+    <p></p>
```
cameron-freshworks: Empty <p> tag here intentional?

src/views/DefineCompany.vue
```diff
@@ -28,27 +28,56 @@
...
+
+@Component({
+  components: {
+    BusinessContactInfo
+  }
+})
 export default class DefineCompany extends Vue {
+  // State
+  @State(state => state.stateModel.defineCompanyStep.businessContact)
+  readonly businessContact!: BusinessContactIF
```
cameron-freshworks: Once again, i think this is pretty cool! Is this to make accessing the sub properties of the model easier?

src/views/DefineCompany.vue
severinbeauvais: Agreed. Please use margin or padding instead. (Could even use class="mt-4" in next section tag.)

severinbeauvais:
> Overall looks great! Just had a few questions and comments.
> 
> Also i would recommend changing all the typescript settings to strict just to verify all the work before we merge 👍 ( IMO we should just leave them on strict and accept the type errors the common components until thats fixed, to help prevent any more sneaking in ). Thoughts on that?

I like that idea, but IIRC the errors due to sbc-common-components were reported in OUR components. I wonder if we could get around that or something.

src/interfaces/stepper-interfaces/DefineCompany/define-company-interface.ts
leksmall: Thats a great suggestion. Will rename it so that all steps has the property 'valid'

src/components/DefineCompany/BusinessContactInfo.vue
leksmall: I will check that

src/components/DefineCompany/BusinessContactInfo.vue
leksmall: Yes. will change the type to string.

src/components/Summary/SummaryDefineCompany.vue
leksmall: Yes :)

src/views/DefineCompany.vue
leksmall: No it was not intentional. This was supposed to be cleaned up. Will add the style

src/components/DefineCompany/BusinessContactInfo.vue
leksmall: I tested that. default value is not needed. Removed that from the prop

src/components/DefineCompany/BusinessContactInfo.vue
leksmall: Not sure as this works as expected with the current rule

src/components/DefineCompany/BusinessContactInfo.vue
cameron-freshworks: Right on, thanks!

src/components/DefineCompany/BusinessContactInfo.vue
cameron-freshworks: 👍

src/views/DefineCompany.vue
cameron-freshworks: Sounds good!

src/components/DefineCompany/BusinessContactInfo.vue
severinbeauvais: As we discussed, this works currently due to the input mask that allows 0-5 digits. Therefore this 'rules' is not needed.

PR #19, opened 2020-01-23
Business Contact Info unit test + Adding id to the ui elements

leksmall:
*Issue #:* /bcgov/entity#2344

*Description of changes:*
Added unit tests for Business Contact Info
Added id to the ui elements

### DIFF
--- a/src/components/DefineCompany/BusinessContactInfo.vue
+++ b/src/components/DefineCompany/BusinessContactInfo.vue
@@ -1,16 +1,23 @@
 <template>
   <div>
     <div v-if="!isEditing">
-      <div>Email Address: {{ !!contact.email ? contact.email : 'Not entered'}}</div>
-      <div>Phone: {{ !!contact.phone ? contact.phone : 'Not entered' }}
+      <div id="lbl-email">Email Address: {{ !!contact.email ? contact.email : 'Not entered'}}</div>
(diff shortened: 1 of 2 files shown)


src/components/DefineCompany/BusinessContactInfo.vue
```diff
@@ -1,16 +1,23 @@
 <template>
   <div>
     <div v-if="!isEditing">
-      <div>Email Address: {{ !!contact.email ? contact.email : 'Not entered'}}</div>
-      <div>Phone: {{ !!contact.phone ? contact.phone : 'Not entered' }}
+      <div id="lbl-email">Email Address: {{ !!contact.email ? contact.email : 'Not entered'}}</div>
```
cameron-freshworks: Just curious, was there a standard set for what we should be using as id names or did you come up with this? Really just asking for my own knowledge :)

leksmall: No No :) No standards at all.. i just added it like that

leksmall:
> Code looks good! Though it seems there was an issue with one of the unit tests.

The test runs fine in local. Looks like a timing issue/ value not getting set while running as part of this build. Looking into that

severinbeauvais:
I have confirmed that the unit tests are running successfully in the GHA build.

PR #21, opened 2020-01-28
#1896 Save & File Incorporations 

cameron-freshworks:
*Issue #:* /bcgov/entity#1896

*Description of changes:*
* Implements an api mixin
* Implements a mixin to build our filings
* Updated Interfaces
* Updated Store structure

* Notes: unit testing will follow when the API is deployed and testable.


src/interfaces/address-interfaces/address-interface.ts
```diff
@@ -0,0 +1,22 @@
...
+// Interface to define the joint base address response
+export interface BaseAddressObjIF {
+  deliveryAddress: AddressIF;
+  mailingAddress: AddressIF;
+}
+
+// Interface to define the Bcorps address response
+export interface BcorpAddressIf {
+  registeredOffice: BaseAddressObjIF;
+  recordsOffice: BaseAddressObjIF;
```
leksmall: No action item for you . But I will be changing the type name to  something like IncorporationAddressIF (because it is not really BcorpAddressIf alone) and make recordsOffice as an optional property.

cameron-freshworks: Totally, makes perfect sense to me! Now that i look at it, i didn't even implement this Interface and i just described the registered or records office as a baseAddressIF.

src/mixins/api-mixin.ts
```diff
@@ -0,0 +1,129 @@
...
+      }
+    }
+  }
+
+  /**
+   * Method to make a simple axios Post request.
+   * @param isDraft Boolean indicating if this filing is a draft.
+   * @param data The object body of the request.
+   */
+  axiosPost (isDraft: boolean, data: object): Promise<any> {
```
leksmall: It would be better if the method name is createIncorporationFiling / createFiling or something in that line

src/mixins/api-mixin.ts
```diff
@@ -0,0 +1,129 @@
...
+    })
+  }
+
+  /**
+   * Method to make a simple axios Put request.
+   * @param isDraft Boolean indicating if this filing is a draft
+   * @param data The object body of the request.
+   * @param filingId Optional filing id if this resuming an existing draft
+   */
+  axiosPut (isDraft: boolean, data: object, filingId: number): Promise<any> {
```
leksmall: Method name should be something like updateFiling

src/mixins/api-mixin.ts
```diff
@@ -0,0 +1,129 @@
...
+        throw new Error('invalid API response')
+      }
+    })
+  }
+
+  /**
+   * Method to make a simple axios Get request.
+   * @param filingId Optional filing id if this resuming an existing draft
+   */
+  axiosGet (filingId: number): Promise<any> {
```
leksmall: It would be better to call this method getFiling

src/mixins/api-mixin.ts
```diff
@@ -0,0 +1,129 @@
...
+import { StateModelIF, ActionBindingIF, GetterIF } from '@/interfaces'
+
+// Utils
+import { FilingTemplateMixin } from '@/mixins'
+
+/**
+ * Mixin that provides the integration with the legal api.
+ */
+@Component
+export default class ApiMixin extends Mixins(FilingTemplateMixin) {
```
leksmall: I am ok with having this as a mixin opposed to the service pattern. In that case the name needs to be more specific I think. We will be calling Name Requests API, Auth API etc.

src/mixins/api-mixin.ts
cameron-freshworks: make sense to me, no problem at all!

src/mixins/api-mixin.ts
cameron-freshworks: same as above, will do!

src/mixins/api-mixin.ts
cameron-freshworks: 👍 👍

src/mixins/api-mixin.ts
cameron-freshworks: This is outdated now, the filingtemplate is its own mixin and is now built when the save filing is called and passes it to the api mixin

cameron-freshworks:
> Looks great! Just some minor comments which can very well be addressed in another PR if they need to be addressed.

Thanks! I'll tackle them asap 👍

src/interfaces/store-interfaces/state-interface.ts
```diff
@@ -1,14 +1,13 @@
...
 export interface StateModelIF {
-  stateText: string | undefined
-  currentDate: string | null
-  certifyState: CertifyIF
-  currentStep: number
-  isSaving: boolean
-  isSavingResuming: boolean
-  isFilingPaying: boolean
-  defineCompanyStep: DefineCompanyIF
+  tombstone: TombStoneIF,
```
severinbeauvais: Commas? Semicolons? Nothing? I just had this conversation with the Relationships Team and they prefer no punctuation. Should we be consistent with that?

src/mixins/legal-api-mixin.ts
```diff
@@ -0,0 +1,118 @@
...
+   * @param isDraft Boolean indicating if this filing is a draft
+   * @param filingId Optional filing id if this resuming an existing draft
+   */
+  async saveFiling (filing: IncorporationFilingIF, isDraft: boolean): Promise<any> {
+    try {
+      let filingId = this.getFilingId
+
+      // If have a filing id, update an existing filing
+      if (filingId && filingId > 0) {
+        this.updateFiling(isDraft, filing, filingId)
```
severinbeauvais: Don't we need to 'await' updateFiling() (and also createFiling() below)?

src/mixins/legal-api-mixin.ts
```diff
@@ -0,0 +1,118 @@
...
+      }
+    }
+  }
+
+  /**
+   * Method to make a simple axios Post request.
+   * @param isDraft Boolean indicating if this filing is a draft.
+   * @param data The object body of the request.
+   */
+  createFiling (data: object): Promise<any> {
```
severinbeauvais: If this does not need to be called externally, I recommend setting its visibility to "private". Also for updateFiling() and getFiling() below.

src/interfaces/store-interfaces/state-interface.ts
severinbeauvais: I thought you would merge the tombstone properties with the state properties, but I guess this is OK too.

src/interfaces/stepper-interfaces/DefineCompany/define-company-interface.ts
```diff
@@ -1,6 +1,7 @@
-import { BusinessContactIF } from './business-contact-interface'
+import { BusinessContactIF, NameRequestIF } from '@/interfaces'
 
 export interface DefineCompanyIF {
-    valid: boolean
+    valid: boolean;
     businessContact : BusinessContactIF;
+    nameRequest: NameRequestIF;
```
severinbeauvais: I consider NR data as a separate block of data, separate from whatever the Define Company step does, so I'd prefer to see this at a "top level" instead of nested here. Thoughts?

src/components/common/Actions.vue
```diff
@@ -90,13 +93,15 @@ export default class Actions extends Vue {
   @Action setIsSavingResuming!: ActionBindingIF
   @Action setIsFilingPaying!: ActionBindingIF
 
+  // Local Properties
+  private authUrl = sessionStorage.getItem('AUTH_URL') || ''
```
severinbeauvais: Please verify that this property gets the proper value, since getting AUTH_URL and putting in session storage is asynchronous with program startup. I'm concerned that the assignment above takes place before AUTH_URL session storage is set.

src/components/common/Actions.vue
```diff
@@ -115,8 +121,10 @@ export default class Actions extends Vue {
    */
   private async onClickSaveResume (): Promise<void> {
     this.setIsSavingResuming(true)
-    await this.sleep(1000)
+    const filing = await this.buildFiling()
+    await this.saveFiling(filing, true)
     this.setIsSavingResuming(false)
+    window.location.assign(this.authUrl)
```
severinbeauvais: FYI, this will eventually be a redirect to a dashboard, but we don't know the URL for that yet so this is an OK assumption for now.
(no change required)

src/components/common/Actions.vue
```diff
@@ -90,13 +93,15 @@ export default class Actions extends Vue {
...
+  private authUrl = sessionStorage.getItem('AUTH_URL') || ''
+
   /**
    * Method called when Cancel button is clicked.
    */
   private onCancel (): void {
-    const authUrl = sessionStorage.getItem('AUTH_URL') || ''
     // assume Auth URL is always reachable
-    window.location.assign(authUrl)
+    window.location.assign(this.authUrl)
```
severinbeauvais: FYI, I'm not sure about this one. I'd say we redirect to a dashboard, but does it matter if we have saved a draft -- different dashboard??? Dunno atm.
(no change required)

src/components/common/Actions.vue
```diff
@@ -105,7 +110,8 @@ export default class Actions extends Vue {
    */
   private async onClickSave (): Promise<void> {
     this.setIsSaving(true)
-    await this.sleep(1000)
```
severinbeauvais: Is sleep() method obsolete now? Can it be deleted now?

src/interfaces/store-interfaces/state-interface.ts
cameron-freshworks: ok, no problem at all. I'm definitely fine with that 👍

src/mixins/legal-api-mixin.ts
severinbeauvais: Also for getFiling() ?

cameron-freshworks: Definitely, especially when we decide we want to do something with the response.

src/mixins/legal-api-mixin.ts
cameron-freshworks: Will do, good idea, thanks!

src/interfaces/store-interfaces/state-interface.ts
cameron-freshworks: ohh i see, i can put them at root level of the state as well. I wasn't sure if it was still as a subset of sorts.

src/interfaces/stepper-interfaces/DefineCompany/define-company-interface.ts
cameron-freshworks: Good point for sure, it certainly feels a little more universal beyond step one as the data won't be determined by step one but just display during it. I'm good with that change!

src/components/common/Actions.vue
cameron-freshworks: Yes, done!

src/components/common/Actions.vue
cameron-freshworks: Definitely, future state will be elsewhere.

PR #22, opened 2020-01-30
Save& File - Unit testing & completeFiling Functionality

cameron-freshworks:
*Issue #:* /bcgov/entity#1896

*Description of changes:*

This code is an update to the frontend regarding some new information about the endpoints.
Essentially the need for a `draft` parameter is no longer required, also the complete filing will be a 2 step request to ensure there is a `filingId` even if the user goes through the flow in one shot without saving. 

* Implemented Unit testing
* updated competeFilings functionality

### DIFF
--- a/src/components/common/Actions.vue
+++ b/src/components/common/Actions.vue
@@ -114,7 +114,6 @@ export default class Actions extends Mixins(FilingTemplateMixin, LegalApiMixin) 
       const filing = await this.buildFiling()
       await this.saveFiling(filing, true)
     } catch (e) {
-      alert(e)
       // TODO:  Trigger some error dialog. Will catch any errors from the Api calls
     }
     this.setIsSaving(false)
@@ -130,7 +129,6 @@ export default class Actions extends Mixins(FilingTemplateMixin, LegalApiMixin) 
       const filing = await this.buildFiling()
       await this.saveFiling(filing, true)
     } catch (e) {
-      alert(e)
       // TODO:  Trigger some error dialog. Will catch any errors from the Api calls
     }
     this.setIsSavingResuming(false)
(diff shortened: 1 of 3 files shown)


src/components/common/Actions.vue
```diff
@@ -122,7 +122,7 @@ export default class Actions extends Mixins(FilingTemplateMixin, LegalApiMixin)
   private async onClickSaveResume (): Promise<void> {
     this.setIsSavingResuming(true)
     const filing = await this.buildFiling()
-    await this.saveFiling(filing, true)
+    await this.saveFiling(filing)
```
severinbeauvais: I think you should wrap this statement in a try-catch block, and unwrap the methods in legal-api-mixin. Then here you can display some error dialog. (Error dialogs are future so you could just alert() here for now.)
Basically, I think Actions.vue should display the error, not the mixin.

src/components/common/Actions.vue
```diff
@@ -122,7 +122,7 @@ export default class Actions extends Mixins(FilingTemplateMixin, LegalApiMixin)
...
     this.setIsSavingResuming(false)
     window.location.assign(this.authUrl)
```
severinbeauvais: Don't redirect if the save failed :)

src/components/common/Actions.vue
```diff
@@ -134,7 +134,7 @@ export default class Actions extends Mixins(FilingTemplateMixin, LegalApiMixin)
   private async onClickFilePay (): Promise<void> {
     this.setIsFilingPaying(true)
     const filing = await this.buildFiling()
-    await this.saveFiling(filing, false)
+    await this.completeFiling(filing)
```
severinbeauvais: If we do error handling in this component, do we need to splits up completeFiling into 2 separate calls -- the first to save and the second to file?

src/components/common/Actions.vue
cameron-freshworks: Makes sense, much easier to handle in the component itself than the mixin, will do!

src/components/common/Actions.vue
cameron-freshworks: kk, will do.

src/components/common/Actions.vue
cameron-freshworks: I'll look into it!

src/components/common/Actions.vue
severinbeauvais: Assumption is that we'd `return` inside the catch block.

PR #23, opened 2020-01-30
UI component for addresses

leksmall:
*Issue #:* /bcgov/entity#2469

*Description of changes:*
Create Address UI
Address Summary
Modified the DefineCompany Summary Layout
Modified the Business Contact Info UI
Test fixes


src/utils/config-helper.ts
```diff
@@ -51,6 +51,10 @@ export function fetchConfig (): Promise<any> {
       const keycloakConfigUrl = response.data['KEYCLOAK_CONFIG_URL']
       sessionStorage.setItem('KEYCLOAK_CONFIG_URL', keycloakConfigUrl)
       console.info('Set Keycloak config URL to: ' + keycloakConfigUrl)
+
+      const addressCompleteKey = response.data['ADDRESS_COMPLETE_KEY']
+      window['addressCompleteKey'] = addressCompleteKey
```
cameron-freshworks: most likely be a type error here. Maybe try `const someArbitraryWindowName = window as any` before assigning the completeKey value?

src/views/DefineCompany.vue
```diff
@@ -29,54 +29,89 @@
         </div>
       </v-card>
     </section>
+    <section class="mt-4" v-if="isEntityType">
+      <header>
+        <h2 v-if="entityFilter(EntityTypes.BCOMP)">2. Registered and Records Office Addresses</h2>
+        <h2 v-else-if="entityFilter(EntityTypes.COOP)">2. Registered Office Addresses</h2>
```
cameron-freshworks: As far as i know ( assumption 😆 ), most incorporations will have both addresses, Coops seem to be the exception.
Would it make sense to cut this down 1 line with 1 filter?  something like ...
`<h2>2. Registered <span v-if="entityFilter(EntityTypes.BCOMP)">and Records</span> Office Addresses</h2>`

src/views/DefineCompany.vue
```diff
@@ -29,54 +29,89 @@
...
+        <p v-if="entityFilter(EntityTypes.BCOMP)">
+          Enter the business' Registered Office and Records Office Mailing and Delivery Addresses.
+        </p>
+        <p v-else-if="entityFilter(EntityTypes.COOP)">
+          Enter the business' Registered Office Mailing and Delivery Addresses.
+        </p>
```
cameron-freshworks: Same idea as above, maybe just utilizing one filter? Perhaps something like..
`<p>Enter the business' Registered Office <span v-if="entityFilter(EntityTypes.BCOMP)" >and Records Office</span> Mailing and Delivery Addresses.</p>`

src/components/DefineCompany/OfficeAddresses.vue
```diff
@@ -0,0 +1,518 @@
...
+import { isEmpty } from 'lodash'
+
+// Schemas
+import { officeAddressSchema } from '@/schemas'
+
+// Components
+import BaseAddress from 'sbc-common-components/src/components/BaseAddress.vue'
+
+// Interfaces
+import { BaseAddressObjIF, IncorporationAddressIf, AddressIF } from '@/interfaces'
```
cameron-freshworks: Is the BaseAddressIF used here?

src/components/DefineCompany/OfficeAddresses.vue
```diff
@@ -0,0 +1,518 @@
...
+
+// Constants
+import { ADDRESSCHANGED } from '@/constants'
```
cameron-freshworks: Not sure if this Constant is being used?

public/config/configuration.json
```diff
@@ -4,5 +4,6 @@
   "SIGNIN_URL": "https://dev.bcregistry.ca/cooperatives/auth/signin/bcsc/",
   "AUTH_API_URL": "https://auth-api-dev.pathfinder.gov.bc.ca/api/v1/",
   "PAY_API_URL": "https://pay-api-dev.pathfinder.gov.bc.ca/api/v1/",
-  "KEYCLOAK_CONFIG_URL": "/local-keycloak-config-url/keycloak.json"
+  "KEYCLOAK_CONFIG_URL": "/local-keycloak-config-url/keycloak.json",
+  "ADDRESS_COMPLETE_KEY": "RH99-RN99-GB72-FW86"
```
severinbeauvais: Need to update release ticket (for Create Business v0.2.0) with a note to add this new key to the Config Map (either in OpenShift or in create script in k8s folder) when deploying to Dev/Test/Prod.

src/components/Summary/SummaryDefineCompany.vue
```diff
@@ -9,35 +13,68 @@
...
       <BusinessContactInfo :initialValue="businessContact" :isEditing="false" />
     </div>
   </v-card>
 </template>
 
 <script lang="ts">
 // Libraries
-import { Component, Vue } from 'vue-property-decorator'
+import { Component, Vue, Mixins } from 'vue-property-decorator'
 import { Getter, Action, State } from 'vuex-class'
```
cameron-freshworks: Just have a few unused imports, Vue, Getter & Action

public/index.html
```diff
@@ -7,6 +7,9 @@
     <title>Incorporation Application</title>
     <link rel="icon" href="<%= BASE_URL %>favicon.png">
     <link rel="preload" href="<%= BASE_URL %>fonts/materialdesignicons-webfont.927457ed.woff2" as="font" type="font/woff2" crossorigin>
+    <link rel="stylesheet" href="<%= BASE_URL %>css/addresscomplete-2.00.min.css">
+    <script type="text/javascript" src="<%= BASE_URL %>js/addresscomplete-2.00.min.js" defer></script>
+    <script type="text/javascript" src="https://unpkg.com/date-time-format-timezone@latest/build/browserified/date-time-format-timezone-complete-min.js" defer></script>
```
severinbeauvais: Please remove date js not needed in this project.

src/components/DefineCompany/OfficeAddresses.vue
```diff
@@ -0,0 +1,518 @@
+<template>
+  <div id="office-addresses">
+    <div v-if="!isEditing">
+      <v-layout row>
+        <v-flex md4><label><strong>Registered Office</strong></label></v-flex>
+        <v-flex md4>
+          <label><strong>Mailing Address</strong></label>
+          <mailing-address
+            :address="mailingAddress"
+            :editing="isEditing"
```
severinbeauvais: I think this prop defaults to false, and this is the "read only" markup anyway, so I suggest either removing this prop or explicitly setting it to false (so it's obvious that this value doesn't change here).

And ditto for other base address components below.

src/components/DefineCompany/OfficeAddresses.vue
```diff
@@ -0,0 +1,518 @@
...
+            :editing="isEditing"/>
+          <div v-else>
+            <span v-if="isEmptyAddress(recDeliveryAddress)">Not entered</span>
+            <span v-else-if="inheritRegisteredAddress">Same as Registered Office</span>
+            <span v-else>Same as Mailing Address</span>
+          </div>
+        </v-flex>
+      </v-layout>
+    </div>
+    <v-card flat v-else>
```
severinbeauvais: Please add a blank line and a comment to delineate the read-only section above with this editing section below. (easier to read code)

src/components/DefineCompany/OfficeAddresses.vue
```diff
@@ -0,0 +1,518 @@
...
+
+        <!-- Registered Mailing Address -->
+        <li class="address-list-container">
+          <div class="meta-container">
+            <label>Mailing Address</label>
+            <div class="meta-container__inner">
+              <div class="address-wrapper">
+                <mailing-address
+                  :address="mailingAddress"
+                  :editing="isEditing"
```
severinbeauvais: Is editing prop always true? Consider using 'true' instead of variable to be explicit that this value doesn't change here.
Knowing that this value doesn't change here makes it easy to understand this code (imho).

And ditto for other base address components below.

src/components/DefineCompany/OfficeAddresses.vue
```diff
@@ -0,0 +1,518 @@
...
+    if (this.inheritRecMailingAddress) {
+      this.recDeliveryAddress = { ...this.recMailingAddress }
+      this.emitAddresses()
+    }
+  }
+
+  private updateAddress (addressChangedEntity:string, baseAddress: AddressIF, newAddress: AddressIF): void {
+    Object.assign(baseAddress, newAddress)
+    switch (addressChangedEntity) {
+      case 'mailingAddress':
```
severinbeauvais: Please use constant. Also in calls to this method. Also in cases below.

Also in updateAddressValid() below.

src/utils/config-helper.ts
severinbeauvais: Or you could ignore the lint error on the next line... something like:
`// eslint-disable-line no-console`
or
`// eslint-disable-next-line no-console`
(substitute the lint error accordingly)

src/store/state/state-model.ts
```diff
@@ -1,4 +1,4 @@
-import { StateModelIF } from '@/interfaces'
+import { StateModelIF, AddressIF } from '@/interfaces'
```
severinbeauvais: AddressIF not needed here?

src/schemas/office-address-schema.ts
```diff
@@ -0,0 +1,33 @@
...
+    maxLength: maxLength(50)
+  },
+  addressCity: {
+    required,
+    maxLength: maxLength(40)
+  },
+  addressCountry: {
+    required,
+    // FUTURE: create new validation function isCountry('CA')
+    isCanada: (val:string) => Boolean(val === 'CA')
```
severinbeauvais: Please follow up on whether this rule (and the region rule below) are needed here. If you're not sure, just comment it out for now and add a TODO comment :)

src/views/DefineCompany.vue
severinbeauvais: I prefer cleaner code (2 simple lines), but I'm flexible.

leksmall: Thats a great idea. Dint think that way. Will change the code

src/views/DefineCompany.vue
leksmall: Sure. Will change it

src/utils/config-helper.ts
leksmall: (<any>window).addressCompleteKey = addressCompleteKey seems to be working

cameron-freshworks: oh nice, i like that even more!

src/views/DefineCompany.vue
cameron-freshworks: Totally, all does the same thing, i'm fine with anything as well! 100% your call.

src/schemas/office-address-schema.ts
cameron-freshworks: I believe Sienna confirmed this today when i inquired about the validators on the api, that the BC and Canada rules will apply to our addresses in Incorporation filings similar to Change of address filings.So I believe the backend will have a validator on the api checking both, as far as i know.

severinbeauvais: The FE needs validation to warn the user up-front (user-friendly / prevent errors) and the BE needs validation to protect itself from bad data.

I suspect we will need eventually need 2 address schemas -- one for business address(es) and one for director address(es) -- same as coops-ui.

Here, just handle the current case (office).

public/config/configuration.json
leksmall: Done. The ticket is https://github.com/bcgov/entity/issues/2538

src/schemas/office-address-schema.ts
leksmall: Sienna confirmed that the address rules are the same as in coops. So no change here

PR #26, opened 2020-02-04
improved prototype stepper

severinbeauvais:
*Issue #:* /bcgov/entity#1892

*Description of changes:*
- updated prototype stepper to use objects
- updated prototype stepper with correct icons
- updated prototype stepper with 'valid' checkmark
- set filing data according to entity type
- changed some v-if to v-show
- updated version to 0.2.0
- ignored template param files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,6 +88,9 @@ venv.bak/
 .env.local
 .env.*.local
 
+# Template parameter files
+k8s/*.param
+
 # Spyder project settings
 .spyderproject
 .spyproject
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bcrs-business-create-ui",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "private": true,
   "scripts": {
     "serve": "vue-cli-service serve",
--- a/src/components/common/Stepper.vue
+++ b/src/components/common/Stepper.vue
@@ -1,53 +1,13 @@
 <template>
-  <v-container id="step-buttons-container" class="list-item justify-space-between">
-    <v-btn id="step-1-btn" large to="/define-company" :disabled="isBusySaving">
-      <span>1. Define Your<br>Company</span>
-    </v-btn>
-
-    <!-- Benefit Company steps -->
-    <template v-if="isTypeBcomp">
-      <v-btn id="step-2-btn" large disabled>
-        <span>2. Add People<br>and Roles</span>
-      </v-btn>
-
-      <v-btn id="step-3-btn" large disabled>
-        <span>3. Create Share<br>Structure</span>
-      </v-btn>
-
-      <v-btn id="step-4-btn" large disabled>
-        <span>4. Incorporation<br>Agreement</span>
-      </v-btn>
-
-      <v-btn id="step-5-btn" large to="/review-confirm" :disabled="isBusySaving">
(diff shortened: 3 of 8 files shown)


.gitignore
```diff
@@ -88,6 +88,9 @@ venv.bak/
 .env.local
 .env.*.local
 
+# Template parameter files
+k8s/*.param
```
cameron-freshworks: What is this?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bcrs-business-create-ui",
-  "version": "0.1.0",
+  "version": "0.2.0",
```
cameron-freshworks: Our little app is growing up so fast.. 😢

.gitignore
severinbeauvais: This is the file of config parameters to build the Device Config (in OpenShift) from a script. So it doesn't get checked in :)

package.json
severinbeauvais: Another sprint, another release!

src/components/common/Stepper.vue
```diff
@@ -69,6 +29,147 @@ export default class Stepper extends Vue {
   @Getter isBusySaving!: GetterIF
   @Getter isTypeBcomp!: GetterIF
   @Getter isTypeCoop!: GetterIF
+
+  // State
+  @State(state => state.stateModel.defineCompanyStep.valid)
+  readonly step1Valid!: boolean;
+
+  // FUTURE - map these to states as above
+  readonly step2Valid!: boolean;
```
cameron-freshworks: Definitely like the mapping! It might get trickier once a step needs to map to more than one view. ie step4valid will be different based on entityType.

src/components/common/Stepper.vue
```diff
@@ -69,6 +29,147 @@ export default class Stepper extends Vue {
...
+      icon: 'mdi-file-document-box-check-outline',
+      text: 'Review\nand Confirm',
+      to: '/review-confirm',
+      disabled: false,
+      valid: this.step5Valid
+    }]
+  }
+
+  // Step configuration for fallback
+  private get StepsElse (): Array<any> {
```
cameron-freshworks: Well done, nice to have a default!

.gitignore
cameron-freshworks: ok right on! Maybe tomorrow you can tell me a bit more, i'm not sure exactly what means :)

src/components/common/Stepper.vue
```diff
@@ -69,6 +29,147 @@ export default class Stepper extends Vue {
...
+      to: '/incorp-agreement',
+      disabled: true,
+      valid: this.step4Valid
+    },
+    {
+      id: 'step-5-btn',
+      icon: 'mdi-file-document-box-check-outline',
+      text: 'Review\nand Confirm',
+      to: '/review-confirm',
+      disabled: false,
```
severinbeauvais: Oops, this should be `this.isBusySaving` same as step 1.

Same below.

src/components/common/Stepper.vue
severinbeauvais: Yes, but we can deal with that when we know more.

src/components/common/Stepper.vue
severinbeauvais: Fixed.

PR #28, opened 2020-02-06
updated Create DC file (requires updated Param file - not checked in)

severinbeauvais:
*Issue #:* /bcgov/entity#2391

*Description of changes:*
- this is the Create DC file, updated with 2 new config map params
- this file requires a param file that contains the necessary config map parameters, but the param file is not checked in so I will email it separately

### DIFF
--- a/k8s/templates/bcrs-entities-create-dc.json
+++ b/k8s/templates/bcrs-entities-create-dc.json
@@ -111,7 +111,7 @@
             "kind": "ConfigMap",
             "apiVersion": "v1",
             "data": {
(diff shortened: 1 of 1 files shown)


thorwolpert:
I noticed the URL name pattern varies, do we want to choose one?

severinbeauvais:
> I noticed the URL name pattern varies, do we want to choose one?

We could but it would be a breaking change. This hasn't bugged me so far. What do you want to do?

severinbeauvais:
Using a standard URL name pattern can be done in a future commit.

PR #29, opened 2020-02-11
#2582 Handling Saving and Fetching Draft Filings

cameron-freshworks:
*Issue #:* /bcgov/entity#2582

*Description of changes:*
* Updates the saving / filing incorporations filings.
* Implements a means to fetch and parse draft filings into the ui.
* Updates unit testing.

* Notes: Currently there is no dashboard to launch a draft from. This is mocked by setting a nrNumber in the App.vue ( line 100 ). If you save a filing to the specified number you will retrieve and parse the draft based off what nrNumber is set. 

* Future state: We will inject an nrNumber or other equivalent identifier into the front end from wherever we launch incorporation filings from.

### DIFF
--- a/public/config/configuration.json
+++ b/public/config/configuration.json
@@ -1,5 +1,5 @@
 {
-  "API_URL": "https://legal-api-dev.pathfinder.gov.bc.ca/api/v1/businesses",
+  "API_URL": "https://legal-api-dev.pathfinder.gov.bc.ca/api/v1/",
   "AUTH_URL": "https://dev.bcregistry.ca/cooperatives/auth/",
   "SIGNIN_URL": "https://dev.bcregistry.ca/cooperatives/auth/signin/bcsc/",
   "AUTH_API_URL": "https://auth-api-dev.pathfinder.gov.bc.ca/api/v1/",
--- a/public/config/configuration.json
+++ b/public/config/configuration.json
@@ -1,5 +1,5 @@
 {
-  "API_URL": "https://legal-api-dev.pathfinder.gov.bc.ca/api/v1/businesses/",
+  "API_URL": "https://legal-api-dev.pathfinder.gov.bc.ca/api/v1/businesses",
   "AUTH_URL": "https://dev.bcregistry.ca/cooperatives/auth/",
   "SIGNIN_URL": "https://dev.bcregistry.ca/cooperatives/auth/signin/bcsc/",
   "AUTH_API_URL": "https://auth-api-dev.pathfinder.gov.bc.ca/api/v1/",
--- a/src/App.vue
+++ b/src/App.vue
@@ -26,7 +26,7 @@
 
             <stepper class="mt-10" />
 
-            <router-view />
+            <router-view :key="isDraft"/>
           </v-col>
 
           <v-col cols="12" lg="3" style="position: relative">
@@ -61,7 +61,7 @@ import SbcFeeSummary from 'sbc-common-components/src/components/SbcFeeSummary.vu
 import { EntityInfo, Stepper, Actions } from '@/components/common'
 
 // Mixins
-import { DateMixin } from '@/mixins'
+import { DateMixin, FilingTemplateMixin, LegalApiMixin } from '@/mixins'
 
 // Interfaces
 import { FilingDataIF, ActionBindingIF, StateModelIF } from '@/interfaces'
@@ -81,7 +81,7 @@ import { EntityTypes, FilingCodes } from '@/enums'
     Actions
   }
 })
-export default class App extends Mixins(DateMixin) {
+export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApiMixin) {
   // Global state
   @State stateModel!: StateModelIF
 
@@ -90,11 +90,27 @@ export default class App extends Mixins(DateMixin) {
   @Action setCurrentDate!: ActionBindingIF
   @Action setCertifyStatementResource!: ActionBindingIF
 
+  // Local Properties
   private filingData: Array<FilingDataIF> = []
+  private isDraft: boolean = false
 
-  // Lifecycle event
-  private created (): void {
+  private async created (): Promise<any> {
+    // Mock the nrNumber and Data
+    await this.setNameRequestState({ nrNumber: 'NR7654446', entityType: 'BC', filingId: null })
+
     this.setCurrentDate(this.dateToUsableString(new Date()))
+    try {
+      // Retrieve draft filing if it exists for the nrNumber specified
+      const draft = await this.fetchDraft()
+
+      // Parse the draft data into the ui if it exists
+      draft && await this.parseDraft(draft)
+
+      // Inform the router view we are resuming a draft and to update ui
+      this.isDraft = true
+    } catch (e) {
+      // TODO: Catch a flag from the api, if there is an error to be handled.
+    }
   }
 
   /**
--- a/src/App.vue
+++ b/src/App.vue
@@ -89,6 +89,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
   @Action setCurrentStep!: ActionBindingIF
   @Action setCurrentDate!: ActionBindingIF
   @Action setCertifyStatementResource!: ActionBindingIF
+  @Action setNameRequestState!: ActionBindingIF
 
   // Local Properties
   private filingData: Array<FilingDataIF> = []
@@ -97,7 +98,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
 
   private async created (): Promise<any> {
     // Mock the nrNumber and Data
-    await this.setNameRequestState({ nrNumber: 'NR7654458', entityType: 'BC', filingId: null })
+    this.setNameRequestState({ nrNumber: 'NR7654459', entityType: 'BC', filingId: null })
     this.setCurrentDate(this.dateToUsableString(new Date()))
 
     try {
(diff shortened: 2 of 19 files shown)


src/App.vue
```diff
@@ -90,11 +90,28 @@ export default class App extends Mixins(DateMixin) {
...
+  // Local Properties
   private filingData: Array<FilingDataIF> = []
+  private draftFiling: IncorporationFilingIF
+  private isDraft: boolean = false
 
-  // Lifecycle event
-  private created (): void {
+  private async created (): Promise<any> {
+    // Mock the nrNumber and Data
+    await this.setNameRequestState({ nrNumber: 'NR7654458', entityType: 'BC', filingId: null })
```
severinbeauvais: Looks like this is missing:
```
@Action setNameRequestState!: ActionBindingIF
```

Also, why is `await` needed here since it doesn't look like this is an async method?

Otherwise, this is fine here for now.

public/config/configuration.json
```diff
@@ -1,5 +1,5 @@
 {
-  "API_URL": "https://legal-api-dev.pathfinder.gov.bc.ca/api/v1/businesses/",
+  "API_URL": "https://legal-api-dev.pathfinder.gov.bc.ca/api/v1/businesses",
```
severinbeauvais: I disagree with this. All our other URLs end with a slash and the danger with this one being different is that code may use this key, assume it has a slash and therefore not add one, and cause an error. I think it's better to be consistent within this project (and across projects).

src/App.vue
```diff
@@ -26,7 +26,7 @@
 
             <stepper class="mt-10" />
 
-            <router-view />
+            <router-view :key="isDraft"/>
```
severinbeauvais: Is this to cause the router-view page to redisplay?

Interesting... This could be used to re-route to the current page (which is currently not possible in Vue -- or Angular and probably others). (For example, we might re-route to the dashboard when the user clicks the header logo, forcing the page to refresh all data. Just a thought.)

src/App.vue
severinbeauvais: OK I'm confused. Is setNameRequestState a store action or a mixin function? It seems to be used inconsistently.

severinbeauvais:
By the way, I noticed some errors and warnings in the unit test output... I think some tests are erroring out instead of "failing" the test. However, instead of blocking THIS task and PR, I raised a new task in the current sprint: /bcgov/entity#2777

src/mixins/legal-api-mixin.ts
```diff
@@ -117,7 +112,7 @@ export default class LegalApiMixin extends Vue {
 
     if (filingId) {
       // Assign the url business identifier
-      let url = `${this.getBusinessIdentifier}/filings/${filingId}`
+      let url = `/${this.getBusinessIdentifier}/filings/${filingId}`
```
severinbeauvais: I just realized (after chatting with Katie) that there are separate endpoints to post/put/get a business and then to post the completed filing. Can we make this more clear in this code via extra comments or something?

What would help is renaming createFiling/updateFiling to createBusiness/updateBusiness :)

public/config/configuration.json
severinbeauvais: Please update where AUTH_URL is set in Actions.spec.ts -- it should end with a / just like the real configuration.

severinbeauvais: Suggested solution for API_URL issue: change the URL to "https://legal-api-dev.pathfinder.gov.bc.ca/api/v1/" and then prepend "businesses" or "businesses/" when making the API calls.

src/App.vue
cameron-freshworks: You are 100% right, this is confusing. I just threw this in to temporarily simulate a user returning to an incorporation filing in progress, but i left out the Action Binding ( by mistake ). However, the action is still bound to the LegalApiMixin which this class extends which allows ( my mistake ) it to still function.    Either way, good catch. I will add the binding in here for now. Will be removed when the Name Request portion of work is complete. Removing the await as well!

public/config/configuration.json
cameron-freshworks: I like this suggestion, thanks!

src/App.vue
cameron-freshworks: Exactly!  I did a little research on how people would 'forceUpdates()' of the UI. Though it was widely suggested you shouldn't want/have to do that. 

What was recommended was using a key on the components you would want to re-render. When the key changes the components flagged will update.

 I found that none of the traditional component lifecycle hooks would time properly because the components would mount before the data was retrieved ( if there was any draft data ).  I'm sure we could still expand and improve on this approach, and i'm 100% open to any suggestions!

src/App.vue
```diff
@@ -90,11 +90,28 @@ export default class App extends Mixins(DateMixin) {
...
+    // Mock the nrNumber and Data
+    await this.setNameRequestState({ nrNumber: 'NR7654458', entityType: 'BC', filingId: null })
     this.setCurrentDate(this.dateToUsableString(new Date()))
+
+    try {
+      // Retrieve draft filing if it exists for the nrNumber specified
+      this.draftFiling = await this.fetchDraft()
+
+      // Parse the draft data into the store if it exists
+      this.draftFiling && await this.parseDraft(this.draftFiling)
```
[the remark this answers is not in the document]
cameron-freshworks: My mistake! I think when i was trying to troubleshoot the data being fetch, stored and displayed i went a little overboard with my use of `await`. Thanks for the heads up, will change!

src/mixins/filing-template-mixin.ts
```diff
@@ -16,45 +19,64 @@ export default class FilingTemplateMixin extends Vue {
...
+      // Set nameRequest data
+      this.setNameRequestState({
+        nrNumber: draftFiling.incorporationApplication.nameRequest.nrNumber,
+        entityType: draftFiling.incorporationApplication.nameRequest.legalType,
+        filingId: draftFiling.header.filingId
+      })
+
+      // Set Office Addresses
+      this.setRegisteredOffice(draftFiling.incorporationApplication.offices.registeredOffice)
+      this.setRecordsOffice(draftFiling.incorporationApplication.offices.recordsOffice)
```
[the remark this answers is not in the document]
cameron-freshworks: Great point, i think i prefer saving the object as whole. I'm not 100% why i created two new setters to handle them separately. But for now, i'll revert to setting `officeAddresses` as a whole. Thanks for mentioning this!

src/mixins/legal-api-mixin.ts
cameron-freshworks: Yes, this certainly is a little confusing. We can definitely discuss this in person! 

I will update the method names to help with clarity, thanks for the info!

The jist of it ( As far as i know )..

We POST to businesses using our NR identifier = We receive a filingId in return ( therefore a creating a filing in a sense ). This is essentially a draft filing at this point, as it isn't complete until we do the PUT to the actual filings endpoint with the filingId.

Whenever we go to File and pay ( `completeFiling()` ) it will always save the filing first to `/business/identifier` first. 

So i could be wrong here, but i believe after the initial POST we just need to PUT.

cameron-freshworks:
> By the way, I noticed some errors and warnings in the unit test output... I think some tests are erroring out instead of "failing" the test. However, instead of blocking THIS task and PR, I raised a new task in the current sprint: /[bcgov/entity#2777](https://github.com/bcgov/entity/issues/2777)

Ahh, i will look into this!

src/mixins/filing-template-mixin.ts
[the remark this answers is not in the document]
severinbeauvais: I too prefer using the object as a whole.

src/mixins/legal-api-mixin.ts
severinbeauvais: Thanks for the explanation. I get it now.

When we continue work on this, we (you or I) should update the comments to better describe what's going on. But for now let's get this code in.

src/mixins/legal-api-mixin.ts
```diff
@@ -94,15 +87,17 @@ export default class LegalApiMixin extends Vue {
...
   /**
-   * Method to make a simple axios Get request.
-   * @param filingId Optional filing id if this resuming an existing draft.
+   * Method to make a simple axios Get request on the draft filings.
    */
-  private getFiling (filingId: number): Promise<any> {
+  private getDraftFiling (): Promise<any> {
     // Assign the url business identifier
-    let url = `${this.getBusinessIdentifier}/filings/${filingId}`
+    let url = `businesses/${this.getBusinessIdentifier}/tasks`
```
katiemcgoff: Are we expecting a "tasks" endpoint for businesses? I hadn't thought so.

katiemcgoff: What I mean is, isn't this supposed to be the same as getting a draft filing (businesses/<identifier>/filings/<filing id>)?

cameron-freshworks: Hmm, from my understanding when a client returns to the Incorporation flow, all we have is an identifier ( NRnumber for example and possibly the type ). So without a filingId, we needed to get the draft information and filingId ( from tasks as above ).

Also doing a GET on the `businesses/identifier/filings` will only return paid and or completed filings. (Not drafts, as far as i know )

We currently get all the draft information back from tasks as it's the only task filing that exists for that identifier, which is also retrieving it in one api request opposed to two.

I could be wrong here but was just working within what currently existed! Open to any suggestions 100% :)

cameron-freshworks:
After speaking with @katiemcgoff, I have approvals to merge 👍

cameron-freshworks:
> Looks great. Very clean. Awesome job!

Thanks!

PR #34, opened 2020-03-04
added line between steps in prototype stepper

severinbeauvais:
*Issue #:* /bcgov/entity#1892

### DIFF
--- a/src/components/common/Stepper.vue
+++ b/src/components/common/Stepper.vue
@@ -1,12 +1,17 @@
 <template>
-  <v-container id="step-buttons-container" class="list-item justify-space-around py-8">
+  <v-container id="step-buttons-container">
     <template v-for="(step, index) in steps">
-      <div class="text-center align-self-start" v-bind:key="index">
-        <v-btn :id=step.id outlined fab color="primary" :to=step.to :disabled=step.disabled>
-          <v-icon>{{step.icon}}</v-icon>
-        </v-btn>
-        <v-icon size="30" color="green darken-1" v-show=step.valid>mdi-check-circle</v-icon>
-        <div class="pre-line">{{step.text}}</div>
+      <div class="step" v-bind:key="index">
+        <div class="step__indicator">
+          <div class="step__line"></div>
+          <v-btn class="step__icon" :id=step.id outlined fab color="primary" :to=step.to :disabled=step.disabled>
+            <v-icon>{{step.icon}}</v-icon>
+          </v-btn>
+          <v-icon class="step__icon2" size="30" color="green darken-1" v-show=step.valid>
+            mdi-check-circle
+          </v-icon>
+        </div>
+        <div class="step__label pre-line">{{step.text}}</div>
       </div>
     </template>
   </v-container>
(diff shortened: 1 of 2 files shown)


src/components/common/Stepper.vue
```diff
@@ -1,12 +1,17 @@
...
-  <v-container id="step-buttons-container" class="list-item justify-space-around py-8">
+  <v-container id="step-buttons-container">
     <template v-for="(step, index) in steps">
-      <div class="text-center align-self-start" v-bind:key="index">
-        <v-btn :id=step.id outlined fab color="primary" :to=step.to :disabled=step.disabled>
-          <v-icon>{{step.icon}}</v-icon>
-        </v-btn>
-        <v-icon size="30" color="green darken-1" v-show=step.valid>mdi-check-circle</v-icon>
-        <div class="pre-line">{{step.text}}</div>
+      <div class="step" v-bind:key="index">
```
cameron-freshworks: Not sure you 'have' too put v-bind before the `:key` but thats up to you!

src/components/common/Stepper.vue
```diff
@@ -177,14 +182,58 @@ export default class Stepper extends Vue {
...
+  position: relative;
+  width: 100%;
+  text-align: center;
+}
+
+.step__line {
+  position: absolute;
+  top: 50%;
+  left: 0;
+  height: 2px;
```
cameron-freshworks: Just curious why we are using px here and rem in other spots?

src/components/common/Stepper.vue
```diff
@@ -177,14 +182,58 @@ export default class Stepper extends Vue {
...
   position: relative;
   margin-top: -32px;
   margin-left: -16px;
   background: $BCgovInputBG;
   border-radius: 50%;
+  z-index: 3;
+}
+
+.step__label {
+  margin-top: 10px;
```
cameron-freshworks: Same q as above!

src/components/common/Stepper.vue
severinbeauvais: You're right, not required. But hey it's a "prototype".

src/components/common/Stepper.vue
severinbeauvais: This is the height of the connector line. It never needs to be scaled, which is the usual reason to use "root ems". 1 px seems too thin to me and 3 px seems too thick.

src/components/common/Stepper.vue
severinbeauvais: I guess I didn't have a good reason to use rems. This just seemed like a good value. It's also easier to read than 0.625rem.
To keep the number of font sizes to a minimum I use a small set of rem values for fonts. And I use a small set of values for margins and padding. But here I'm just tweaking the space between elements so px is more specific (and there's no need to scale it).

src/components/common/Stepper.vue
severinbeauvais: Ok, I didn't see this earlier and remove it...

cameron-freshworks: right on!

src/components/common/Stepper.vue
cameron-freshworks: Sounds good, was just curious!

PR #38, opened 2020-03-09
updated typescript and other packages

severinbeauvais:
*Issue #:* /bcgov/entity#2517

*Description of changes:*
- updated project version to 0.3.0
- updated typescript to 3.8.3
- updated other packages to latest (except where incompatible)
- added missing peer dependencies
- fixed some lint warnings
- fixed broken MDI


tsconfig.json
```diff
@@ -42,6 +42,7 @@
     "tests/**/*.tsx"
   ],
   "exclude": [
-    "node_modules"
+    "node_modules",
+    "dist"
```
cameron-freshworks: Interesting, is this to exclude the dependencies that are being imported/compiled in a bundle with our src files? Does this also allow us to set all the TS settings back to strict?  Either way, well done, never considered this!

severinbeauvais: I think it's a best practice rather than being a solution to the strictness problem. But I will test it!

severinbeauvais: The answer is no, it does not allow us to use `strict: true`. However, after looking into it, I found that you can do this:
```
"strict": true,
"noImplicitAny": false,
"strictNullChecks": false,
```
In other words, you can start with full strictness and then override just the checks that fail. I believe this is cleaner than before so I'm check this in :)

PR #47, opened 2020-03-12
Incorporations draft refactor

cameron-freshworks:
*Issue #:* /bcgov/entity#2958

*Description of changes:*

* Updated the filing endpoint URLS to adapt to the recent back end changes.

* Added Draft flags to filing requests where required. 

* Removed the :key from router view and now correctly allowing VUE to handle refreshing the components.

* Implemented 'vuex-persistedstate' to handle storing State in between login redirects as well as refreshing the page. One request to fetch a draft filing is made when the app loads.

* Minor unit test fixes.

* Added auth required to the root level route '/' for the redirect to pick up before redirecting to/from '/define-company'

### DIFF
--- a/src/mixins/legal-api-mixin.ts
+++ b/src/mixins/legal-api-mixin.ts
@@ -32,10 +32,10 @@ export default class LegalApiMixin extends Vue {
     let filingId = this.getFilingId
     // If have a filing id, update an existing filing
     if (filingId && filingId > 0) {
-      await this.updateFiling(filing)
+      await this.updateFiling(filing, isDraft)
     } else {
       // Set the filingId to store
-      await this.createFiling(filing)
+      await this.createFiling(filing, isDraft)
     }
 
     // Complete a filing if not draft
@@ -52,16 +52,21 @@ export default class LegalApiMixin extends Vue {
       return await this.getDraftFiling()
     } catch (e) {
       // TODO: Throw a flag to the ui from here, if we want to trigger error handling in ui
-      throw new Error('Invalid API Response')
     }
   }
 
   /**
    * Method to make a simple axios Post request.
    * @param data The object body of the request.
+   * @param isDraft Boolean indicating whether to complete filing.
    */
-  private createFiling (data: object): Promise<any> {
-    return axios.post('businesses', data).then(res => {
+  private createFiling (data: object, isDraft: boolean): Promise<any> {
+    let url = `businesses`
+    if (isDraft) {
+      url += '?draft=true'
+    }
+
+    return axios.post(url, data).then(res => {
       // Assign a filing Id from the response to the state
       if (res && res.data && res.data.filing && res.data.filing.header && res.data.filing.header.filingId) {
         this.setFilingId(res.data.filing.header.filingId)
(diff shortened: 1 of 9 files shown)


src/mixins/legal-api-mixin.ts
```diff
@@ -52,16 +52,21 @@ export default class LegalApiMixin extends Vue {
       return await this.getDraftFiling()
     } catch (e) {
       // TODO: Throw a flag to the ui from here, if we want to trigger error handling in ui
-      throw new Error('Invalid API Response')
```
severinbeauvais: This line of code would have caught the error thrown by the functions above and re-thrown it, so I agree there's not much benefit here.

What does happen here when an error was thrown -- is it just ignored now?

Another thought is that the throw calls below should be more specific about what call failed (POST, PUT, GET) and/or why they think it's an error; currently they all display the same error :>

src/router/routes.ts
```diff
@@ -9,7 +6,9 @@ import SignoutView from '@/views/auth/Signout.vue'
 export const routes = [
   {
     path: '/',
-    redirect: '/define-company'
+    meta: {
+      requiresAuth: true
+    }
```
severinbeauvais: Good change but this raises an issue elsewhere.

Normally, the auth process (signin page) would redirect back to the starting route (ie, /), but this route doesn't do anything. So I propose that / continue to redirect to define-company, and that the router code / signin page redirect back to the starting route; this can be done by passing the starting route as a param / prop to the signin component.

tests/unit/App.spec.ts
```diff
@@ -25,7 +25,7 @@ describe('App component', () => {
     const localVue = createLocalVue()
     localVue.use(VueRouter)
     const router = mockRouter.mock()
-    wrapper = shallowMount(App, { localVue, store, router, vuetify, stubs: { Affix: true } })
+    wrapper = shallowMount(App, { sync: false, localVue, store, router, vuetify, stubs: { Affix: true } })
```
severinbeauvais: Separately from this PR (ie, possibly create a new ticket if this looks significant), please investigate whether the latest version of the test framework / test tools are now async by default. I read somewhere that the tests were initial synchronous, but that breaks some tests (which is why sync: false is needed) and they were changing to async. Maybe we can just go to the next version of their packages now?

src/mixins/legal-api-mixin.ts
cameron-freshworks: I'll look into this, i'd like to do something consistent throughout. I'm curious if we have designs/tickets to implement error handling in UI ( dialogs or notifications ). It might be a good thing to do sooner than later!

src/router/routes.ts
cameron-freshworks: Okie doke, i will try this out!

tests/unit/App.spec.ts
cameron-freshworks: Very interesting, i will look into this.

Thanks for bringing this up, it appears the Jest devs will be ( if not already have ) moving away from the sync mode as it causes bugs in the tests and doesn't allow simulate how the code runs naturally. So in this case i will update these tests. If we move up dependencies now it may involve a wider refactor ( though i'm not sure we have used sync mode too often ). Either way, ill found out, great topic! 👍

src/mixins/legal-api-mixin.ts
severinbeauvais: /bcgov/entity#2332

cameron-freshworks: 😆  Exactly! Don't recall making that..

src/router/routes.ts
severinbeauvais: Can you debug this and figure out what's happening?

Use cases: user should be able to:
- browse to http://our_app/, login if needed, and return to / (which logically should redirect to /define-company
- browse to http://our_app/review-confirm, login if needed, and return to /review-confirm

cameron-freshworks:
3 approvals, will merge into master to help with other pending PR's. Will address comments in a follow up PR where required 👍

src/router/routes.ts
cameron-freshworks: You bet, looking at this now!

severinbeauvais: Cam mentioned this is working atm.

tests/unit/App.spec.ts
cameron-freshworks: So after researching, can confirm that the usage of the `sync` mode in Jest is not recommended. Will be refactoring these unit tests in the current PR, https://github.com/bcgov/bcrs-business-create-ui/pull/49 . 👍

PR #49, opened 2020-03-13
Incoporations step 2 and step 5

lekshmimallika-aot:
*Issue #:* /bcgov/entity#2704, /bcgov/entity#2670, /bcgov/entity#2674, /bcgov/entity#2698, /bcgov/entity#2687

*Description of changes:* Implementation of Step 2 Add People and Roles + Step 5 summary of people and roles

Note : The test for Add/Edit person will be a separate PR to get the main features going. Tests for other components are in place.


src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -0,0 +1,280 @@
+<template>
+  <div>
+    <p>
+      Add the people and corporations/firms who will have a role in your company. People can have multiple roles;
+      Corporation/firms can only be Incorporators.
```
severinbeauvais: Mockup shows "Corporation" as lower case in this description. Please confirm with Scott.

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -0,0 +1,280 @@
...
+        <span v-else><v-icon color='red' v-if="showErrors" class="incorp-invalid">mdi-close</v-icon></span>
+        <span class='chk-list-item-txt'>At least one Incorporator</span>
+      </li>
+      <li>
+        <v-icon color='blue' v-if="hasRole(Roles.DIRECTOR, 1, 'ATLEAST')" class="dir-valid">mdi-check</v-icon>
+        <span v-else><v-icon color='red' v-if="showErrors" class="dir-invalid">mdi-close</v-icon></span>
+        <span class='chk-list-item-txt'>At least one Director</span>
+      </li>
+    </ul>
+    <div class="btn-panel" v-if="orgPersonList.length === 0">
```
severinbeauvais: This div needs a little more whitespace above it... maybe 1 rem padding?

src/components/dialogs/ConfirmDialog.vue
```diff
@@ -0,0 +1,127 @@
...
+ *   <v-app>
+ *     ...
+ *     <confirm ref="confirm"></confirm>
+ *   </v-app>
+ * </template>
+ *
+ * mounted() {
+ *   this.$root.$confirm = this.$refs.confirm.open
+ * }
+ */
```
cameron-freshworks: Was this a quick reference for development?

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -0,0 +1,533 @@
...
+                    v-model="orgPerson.orgName"
+                    :rules="orgNameRules"
+                  />
+                </div>
+                <label class="sub-header">Roles</label>
+                <v-row>
+                  <v-col cols="4" v-if="isPerson">
+                    <v-checkbox v-model="isCompletingParty" label="Completing Party"
+                    :disabled="isRoleLocked(Roles.COMPLETING_PARTY)"
+                    v-bind:class="{'highlightedRole': isRoleLocked(Roles.COMPLETING_PARTY)}"
```
severinbeauvais: "highlightedRole" makes checkbox hard to read. If possible, should be white text on green background. See mockup.

Also, if possible, checkboxes should be "vertically centered"... check the vertical margins/padding to even this out.

severinbeauvais: Check the styles of the other checkboxes here. According to the mockup they should have a gray background if not checked or a green background if checked.

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -0,0 +1,533 @@
...
+                        :editing="true"
+                        :schema="personAddressSchema"
+                        :address="inProgressDeliveryAddress"
+                        @update:address="updateDeliveryAddress"
+                        @valid="updateDeliveryAddressValidity"/>
+                    </div>
+                  </div>
+                </div>
+
+                <div class="form__row form__btns">
```
severinbeauvais: Check if "large" buttons here match the mockup better. (They seem a little small atm.)

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -0,0 +1,533 @@
...
+                    :rules="lastNameRules"
+                  />
+                </div>
+                <div v-if="isOrg" class="org-name-container">
+                  <v-text-field
+                    filled
+                    class="item"
+                    label="Full Legal Corporation or Firm Name"
+                    id="firm-name"
+                    v-model="orgPerson.orgName"
```
severinbeauvais: Check if there's a maximum size for corp name.

src/interfaces/store-interfaces/state-interface.ts
```diff
@@ -1,4 +1,5 @@
 import { CertifyIF, DefineCompanyIF, NameRequestIF, TombStoneIF } from '@/interfaces'
+import { PeopleAndRoleIF } from '../stepper-interfaces/AddPeopleAndRole/add-people-role-interface'
```
cameron-freshworks: Not 100% sure but i believe this could be added to the above line. If exported correctly from its sub directory i think all the interfaces should be accessible from '@/interfaces'.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -0,0 +1,253 @@
...
+        </v-col>
+        <v-col>
+          <div v-if="person.roles.length>0">
+            <v-col v-for="(role, index) in person.roles" :key="index" class="col-roles">
+              <span>{{role}}</span>
+            </v-col>
+          </div>
+          <div v-else>
+             <v-icon color="rgb(232, 156, 48)" small>mdi-alert</v-icon>
+             <span class="warningText">Add Role</span>
```
severinbeauvais: The icon and text are not vertically aligned per mockup.

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -0,0 +1,533 @@
...
+          </div>
+        </li>
+      </ul>
+    </v-expand-transition>
+  </div>
+</template>
+
+<script lang="ts">
+// Libraries
+import { Component, Vue, Prop, Watch, Emit, Mixins } from 'vue-property-decorator'
```
cameron-freshworks: Vue & Watch appear to be un-used here.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -0,0 +1,253 @@
...
+          <base-address v-else class="peoples-roles-delivery-address" :address="person.address.deliveryAddress"/>
+        </v-col>
+        <v-col>
+          <div v-if="person.roles.length>0">
+            <v-col v-for="(role, index) in person.roles" :key="index" class="col-roles">
+              <span>{{role}}</span>
+            </v-col>
+          </div>
+          <div v-else>
+             <v-icon color="rgb(232, 156, 48)" small>mdi-alert</v-icon>
```
severinbeauvais: Check if there is named colour you can use.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -0,0 +1,253 @@
...
+
+  .col-roles {
+    padding: 0rem!important;
+  }
+}
+
+.warningText {
+  color: rgb(232, 156, 48);
+  margin-left: 0.1rem;
+  padding-top: 0.1rem;
```
severinbeauvais: For small offsets like this, you should use px values. I prefer to use rem values for "standard spacing" between sections or the like. YMMV.

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -0,0 +1,533 @@
...
+  @Emit('removePersonEvent')
+  private emitRemovePersonEvent (activeIndex: Number): void { }
+
+  @Emit('removeCompletingPartyRole')
+  private emitReassignCompletingPartyEvent () : void {}
+}
+</script>
+
+<style lang="scss" scoped>
+[class^="col"] {
```
cameron-freshworks: I don't actually understand this bit, what is going on here?

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -0,0 +1,280 @@
...
+        :isSummary="false"
+        @editPerson="editOrgPerson($event)"
+        @removePerson="onRemovePerson($event)"/>
+    </v-card>
+  </div>
+</template>
+
+<script lang="ts">
+// Libraries
+import { Component, Prop, Watch, Emit, Mixins } from 'vue-property-decorator'
```
cameron-freshworks: I think Prop, Watch and emit are unused here.

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -0,0 +1,280 @@
...
+import { Action, State } from 'vuex-class'
+
+// Interfaces
+import { OrgPersonIF, BaseAddressObjIF, BaseAddressType, FormType, ActionBindingIF } from '@/interfaces'
```
cameron-freshworks: BaseAddressObjIF i don't think is used here.

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -0,0 +1,280 @@
...
+
+// Enums
+import { EntityTypes, Roles, IncorporatorTypes } from '@/enums'
+
+// Components
+import OrgPerson from './OrgPerson.vue'
+import ListPeopleAndRoles from './ListPeopleAndRoles.vue'
+
+// Schemas
+import { addressSchema } from '@/schemas'
```
cameron-freshworks: Is the schema used in this component?

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -0,0 +1,533 @@
...
+  private setPersonRoles (): string [] {
+    let roles: string[] = []
+    if (this.isCompletingParty) {
+      roles.push(Roles.COMPLETING_PARTY)
+    }
+    if (this.isDirector) {
+      roles.push(Roles.DIRECTOR)
+    }
+    if (this.isIncorporator) {
+      roles.push(Roles.INCORPORATOR)
```
severinbeauvais: Per mockup, Incorporator should come before Director.

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -0,0 +1,533 @@
...
+      <ul class="list add-person">
+        <li class="add-person-container">
+          <div class="meta-container">
+            <label class="add-person-header" v-if="isPerson">
+              <span v-if="activeIndex===-1">Add Person</span>
+              <span v-else>Edit Person</span>
+            </label>
+            <label class="add-person-header" v-if="isOrg">
+              <span v-if="activeIndex===-1">Add Corporation or Firm</span>
+              <span v-else>Edit Corporation or Firm</span>
```
severinbeauvais: Per mockup, please add in handling for BCOMP vs COOP (ie, different labels).

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -0,0 +1,280 @@
...
+    <div class="btn-panel"  v-if="orgPersonList.length > 0">
+      <v-btn outlined color="primary" @click="addOrgPerson([], IncorporatorTypes.PERSON)"
+      :disabled="showOrgPersonForm" id="btn-add-person">
+        <v-icon>mdi-account-plus</v-icon>
+        <span>Add a Person</span>
+      </v-btn>
+      <v-btn outlined color="primary" :disabled="showOrgPersonForm" class="spacedButton"
+      @click="addOrgPerson([Roles.INCORPORATOR], IncorporatorTypes.CORPORATION)" id="btn-add-corp">
+        <v-icon>mdi-domain-plus</v-icon>
+        <span>Add a Corporation or Firm</span>
```
severinbeauvais: Per mockup, please add handling for BCOMP vs COOP (ie, different label).

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -0,0 +1,253 @@
...
+             <span class="warningText">Add Role</span>
+          </div>
+        </v-col>
+
+        <!-- Actions Column -->
+        <v-col sm="1" v-if="!isSummary">
+          <div class="actions">
+            <span class="edit-action">
+              <v-btn small text color="primary"
+                :id="'person-' + person.id + '-change-btn'"
```
cameron-freshworks: Do we need ID's on the buttons? Is this primarily for testing?

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -0,0 +1,253 @@
...
+import { Component, Prop, Mixins, Emit } from 'vue-property-decorator'
+
+// Components
+import BaseAddress from 'sbc-common-components/src/components/BaseAddress.vue'
+
+// Dialogs
+import { ConfirmRemoveDialog } from '@/components/dialogs'
+
+// Mixins
+import CommonMixin from '@/mixins/common-mixin'
```
cameron-freshworks: My mistake here, i should have wrote `import { CommonMixin } from '@/mixins'`

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -0,0 +1,253 @@
...
+})
+export default class ListPeopleAndRoles extends Mixins(CommonMixin) {
+  // Store Properties
+  @Prop({ default: '' })
+  private personList: Array<OrgPersonIF>
+
+  @Prop({ default: false })
+  private showErrorSummary: boolean
+
+  @Prop({ default: false })
```
cameron-freshworks: I know i wrote this, but a question for my own curiosity, should we be defining `default` values for booleans? Should we be defining defaults for everything? or nothing? When is a good time??

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -0,0 +1,253 @@
...
+  private isSummary: boolean
+
+  // Local Properties
+  readonly tableHeaders: Array<string> = ['Name', 'Mailing Address', 'Delivery Address', 'Roles']
```
cameron-freshworks: Can i do this a better way? I wanted to make the table somewhat dynamic for future expansion/changes, so took this approach to hardcoding all them into the template. Any other ideas here that are simple?

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -0,0 +1,533 @@
+<template>
+  <div>
+    <confirm-dialog ref="reassignCPDialog" attach="#addEditPersonContainer"/>
+    <v-expand-transition id="addEditPersonContainer">
```
cameron-freshworks: Not sure we have a standard, but how do you feel about line spaces in the template, separating sections/dialogs etc from one another?

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -0,0 +1,533 @@
...
+  }
+
+  private assignCompletingPartyRole (): void {
+    if (this.isCompletingParty && this.existingCompletingParty &&
+      this.orgPerson.id !== this.existingCompletingParty.id) {
+      this.confirmReassignPerson()
+    }
+  }
+
+  // Methods
```
cameron-freshworks: Do you think there is value in expanding the comments on these methods?

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -0,0 +1,533 @@
...
+  get isPerson (): boolean {
+    return this.orgPerson && this.orgPerson.type === IncorporatorTypes.PERSON
+  }
+
+  get isOrg (): boolean {
+    return this.orgPerson && this.orgPerson.type === IncorporatorTypes.CORPORATION
+  }
+
+  get incorporatorLabel () : string {
+    return this.entityFilter(EntityTypes.BCOMP) ? Roles.INCORPORATOR : 'Subscriber'
```
severinbeauvais: Should "Subscriber" be another enum? Seems weird here to use an enum for one value and string for the other value.

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -0,0 +1,533 @@
...
+     <p>Selecting "Completing Party" here will change the Completing Party.</p>`
+    return errorMessage
+  }
+
+  get isPerson (): boolean {
+    return this.orgPerson && this.orgPerson.type === IncorporatorTypes.PERSON
+  }
+
+  get isOrg (): boolean {
+    return this.orgPerson && this.orgPerson.type === IncorporatorTypes.CORPORATION
```
severinbeauvais: As we are now using TS 3.8, you might be able to use Nullish Coalescing here for cleaner code :)

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -0,0 +1,533 @@
...
+  }
+}
+
+.highlightedRole {
+    opacity: 0.5;
+    mix-blend-mode: normal;
+    border-radius: 2px;
+    border-color: rgb(140, 140, 140);
+    background-color: rgb(55, 164, 71);
+    color: rgb(255, 255, 255) s!important;
```
severinbeauvais: Typo.

src/store/getters/state-getters.ts
```diff
@@ -94,7 +94,7 @@ export const isShowFilePayBtn = (state: any): boolean => {
  */
 export const isEnableFilePayBtn = (state: any): boolean => {
   const step1Valid = state.stateModel.defineCompanyStep.valid
-  const step2Valid = true // FOR TESTING ONLY
+  const step2Valid = state.stateModel.addPeopleAndRoleStep.valid // FOR TESTING ONLY
```
severinbeauvais: You can remove the "TESTING" comment now.

src/schemas/office-address-schema.ts
```diff
@@ -31,3 +31,30 @@ export const officeAddressSchema = {
     maxLength: maxLength(80)
   }
 }
+
+export const addressSchema = {
```
severinbeauvais: What's the difference between this address schema and the one above? Consider giving it a more descriptive name and/or add a code comment explaining why we have 2 address schemas.

src/components/common/Stepper.vue
```diff
@@ -71,8 +72,8 @@ export default class Stepper extends Vue {
       id: 'step-2-btn',
       icon: 'mdi-account-multiple-plus',
       text: 'Add People\nand Roles',
-      to: '/people-roles',
-      disabled: true,
+      to: '/add-people-roles',
```
severinbeauvais: Please also update Actions.vue so that clicking on back / next buttons include this new page in the navigation flow.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -0,0 +1,253 @@
...
+  private formatName (filing: any): string {
+    return filing.orgName ? filing.orgName
+      : `${filing.firstName} ${filing.middleName || ''} ${filing.lastName}`
+  }
+
+  /**
+   * Identify and confirm the removal of a person/org through dialog.
+   * @param index The active index which is subject to removal.
+   */
+  confirmRemove (index: number): void {
```
cameron-freshworks: I should have made this private.

src/components/common/Stepper.vue
```diff
@@ -40,6 +40,7 @@ export default class Stepper extends Vue {
   readonly step1Valid!: boolean;
 
   // FUTURE - map these to states as above
+  @State(state => state.stateModel.addPeopleAndRoleStep.valid)
```
severinbeauvais: Please also fix the Step 5 Valid logic (which is currently not working).

src/components/common/Stepper.vue
severinbeauvais: > Please also update Actions.vue so that clicking on back / next buttons include this new page in the navigation flow.

Lekshmi says we can implement something clever for this once all the pages are implemented :)

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -0,0 +1,280 @@
+<template>
+  <div>
+    <p>
+      Add the people and corporations/firms who will have a role in your company. People can have multiple roles;
+      Corporation/firms can only be Incorporators.
+    </p>
+    Your application must include the following:
+    <ul>
+      <li>
+        <v-icon color='blue' v-if="hasRole(Roles.COMPLETING_PARTY, 1, 'EXACT')" class="cp-valid">mdi-check</v-icon>
```
severinbeauvais: I'm wondering out loud if we should have a red X here if the completing party has been removed (or all incorporators or all directors)? Please follow up with Scott.

I don't think we should have a red X initially, but only if the respective person is removed.

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
severinbeauvais: This button panel also needs more whitespace below it -- either bottom margin for this div or top margin for the divs below it.

Also, I think the Add Person / Add Corporation containers (v-cards) have too much padding. I think they should look like the list container. Try padding: 0.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -0,0 +1,253 @@
...
+            <span>
+              <v-menu offset-y>
+                <template v-slot:activator="{ on }">
+                  <v-btn text small color="primary" class="actions__more-actions__btn" v-on="on">
+                    <v-icon>mdi-menu-down</v-icon>
+                  </v-btn>
+                </template>
+                <v-list class="actions__more-actions">
+                  <v-list-item @click="confirmRemove(index)">
+                    <v-list-item-title><v-icon>mdi-delete</v-icon>Remove</v-list-item-title>
```
severinbeauvais: This item looks too big, has too much vertical padding, and should be better aligned vertically.

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -0,0 +1,533 @@
...
+                        :schema="personAddressSchema"
+                        :address="inProgressDeliveryAddress"
+                        @update:address="updateDeliveryAddress"
+                        @valid="updateDeliveryAddressValidity"/>
+                    </div>
+                  </div>
+                </div>
+
+                <div class="form__row form__btns">
+                  <v-btn color="error" :disabled="activeIndex===-1" @click="removePerson()">Remove</v-btn>
```
severinbeauvais: IMHO we should have the "are you sure" dialog here. Please confirm with Scott.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -0,0 +1,253 @@
...
+@Component({
+  components: {
+    BaseAddress,
+    ConfirmRemoveDialog
+  }
+})
+export default class ListPeopleAndRoles extends Mixins(CommonMixin) {
+  // Store Properties
+  @Prop({ default: '' })
+  private personList: Array<OrgPersonIF>
```
severinbeauvais: Default for this should probably be [] ?

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: I think it depends on the code. If props have no default and they are not declared in the parent then they are probably undefined, which can give you 3 states for a boolean (which may be handy).

Conversely, if you must have a default and the prop might not be declared when the component is created (ie, updated later) then you might want to provide defaults. A good example is an array that you may iterate over... I'm not sure what iterating over "null" does but [].length is 0 which is clean.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: This is pretty nice I think.

You could specify the array right in the template but I'm not sure that's cleaner.

We might decide that all constants (incl. readonly) should be all caps. (That was a C convention.) Thoughts?

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: Yes and yes.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: I read somewhere that "private" didn't really work until the latest version of TS.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -0,0 +1,253 @@
...
+</script>
+
+<style lang="scss" scoped>
+#people-roles {
+  margin-top: 1rem;
+}
+
+.people-roles-summary-header {
+  display: flex;
+  background-color: rgba(1, 51, 102, 0.15);
```
severinbeauvais: Is "$gray1" a suitable color variable that could be used here instead. (If yes, consider replacing it globally.)

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -0,0 +1,253 @@
...
+.people-roles-header {
+  padding: .5rem 1.25rem .5rem 1.25rem;
+  font-size: 12px;
+  margin-top: 1rem;
+}
+
+.people-roles-content {
+  margin-top: .5rem;
+  padding: .5rem 1.25rem .5rem 1.25rem;
+  border-top: 1px solid #d1d1d1;
```
severinbeauvais: Check for a named color (or color variable). This color is used in a couple of places. It would be better if it wasn't hard-coded.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -0,0 +1,253 @@
...
+      margin-left: 0.5rem;
+    }
+  }
+}
+
+.col {
+  padding: .25rem;
+
+  .col-roles {
+    padding: 0rem!important;
```
severinbeauvais: Missing space?

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -0,0 +1,280 @@
...
+    }
+  }
+
+  private get showErrors (): boolean {
+    return Boolean(this.$route.query.showErrors)
+  }
+
+  private get completingParty () : OrgPersonIF {
+    let cp: OrgPersonIF | null = null
+    const cpList = this.orgPersonList.filter(people => people.roles.includes(Roles.COMPLETING_PARTY))
```
severinbeauvais: Check if this could all be replaced with a "find" instead. Would be cleaner and a bit faster :)

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -0,0 +1,280 @@
...
+      return numOfCompletingParty === 1 && numOfIncorporator >= 1 && numOfDirector >= 1 && numOfPeopleWithNoRoles === 0
+    } else if (this.entityFilter(EntityTypes.COOP)) {
+      return numOfCompletingParty === 1 && numOfIncorporator >= 3 && numOfDirector >= 3 && numOfPeopleWithNoRoles === 0
+    }
+  }
+
+  private hasRole (roleName: string, count:number, mode:string) : boolean {
+    const orgPersonWithSpecifiedRole: OrgPersonIF[] =
+    this.orgPersonList.filter(people => people.roles.includes(roleName))
+    if (mode === 'EXACT') {
```
severinbeauvais: If possible, declare constants or enums for mode, so that caller and callee are always in sync.

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -0,0 +1,280 @@
...
+  }
+
+   // State
+  @State(state => state.stateModel.addPeopleAndRoleStep.orgPeople)
+  readonly orgPersonList: OrgPersonIF[]
+
+  @Action setOrgPersonList!: ActionBindingIF
+  @Action setAddPeopleAndRoleStepValidity!: ActionBindingIF
+
+  private newOrgPerson: OrgPersonIF = {
```
severinbeauvais: This is a good use case for a constructor. In TS in a previous project (albeit using a different framework) we defined "classes" that always had a constructor (for initial values) and a copy-constructor (to create a new object from an existing one).

I'm not sure if implementing a constructor is any cleaner for an interface... Maybe look into it quickly; if not then current code is OK.

severinbeauvais: Or you could use a factory method to return a new, initialized object.

src/components/ReviewConfirm/Summary.vue
```diff
@@ -1,22 +1,36 @@
 <template>
   <div class="summary-container">
-    <SummaryDefineCompany/>
+    <SummaryDefineCompany />
+    <ListPeopleAndRoles :personList="orgPersonList" :isSummary="true" :showErrorSummary="!stepValid"/>
```
severinbeauvais: I realize you are reusing this component from Step 2, which is different from the dedicated SummaryDefineCompany component... I just wanted to say that I think this is OK! (Ie, Define Company needed a new summary component but People and Roles does not.)

src/components/ReviewConfirm/Summary.vue
```diff
@@ -1,22 +1,36 @@
...
   }
 })
-export default class Summary extends Vue { }
+export default class Summary extends Vue {
+  // State
+  @State(state => state.stateModel.addPeopleAndRoleStep.orgPeople)
+  readonly orgPersonList: OrgPersonIF[]
+
+  @State(state => state.stateModel.addPeopleAndRoleStep.valid)
+  readonly stepValid: boolean
```
severinbeauvais: I recommend that you rename this, as you may need something similar for steps 3 and 4.

src/components/dialogs/ConfimRemoveDialog.vue
```diff
@@ -0,0 +1,40 @@
...
+      </v-card-actions>
+    </v-card>
+  </v-dialog>
+</template>
+
+<script lang="ts">
+import { Component, Vue, Prop, Emit } from 'vue-property-decorator'
+
+@Component
+export default class ConfirmRemoveDialog extends Vue {
```
severinbeauvais: Would it make more sense to use the generic Confirm Dialog here (as this is such a simple component)?

src/components/dialogs/ConfirmDialog.vue
severinbeauvais: Yes, this is doc.

I think this is the exact same component used in business-filings-ui.

src/views/AddPeopleAndRoles.vue
```diff
@@ -0,0 +1,80 @@
...
+  components: {
+    PeopleAndRoles
+  }
+})
+export default class AddPeopleAndRoles extends Mixins(EntityFilterMixin) {
+  // Global getters
+  @Getter isEntityType!: GetterIF
+
+  // Global actions
+  @Action setPeopleAndRoles!: ActionBindingIF
```
severinbeauvais: Please remove unused actions, imports, etc.

tests/unit/ListPeopleAndRoles.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+let vuetify = new Vuetify({})
+
+describe('List People And Roles', () => {
+  let wrapperFactory: any
+
+  const mockPersonList = [
+    {
+      'id': 0,
+      'firstName': 'Cameron',
+      'lastName': 'Bowler',
```
severinbeauvais: Forever immortalized, haha!

tests/unit/ListPeopleAndRoles.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+        router
+      })
+    expect(wrapper.vm.$route.name).toBeNull()
+
+    const errorLink = wrapper.find('#router-link')
+    errorLink.trigger('click')
+
+    expect(wrapper.vm.$route.name).toBe('add-people-roles')
+  })
+})
```
severinbeauvais: Should also have tests for roles (ie, ensure correct roles are displayed; ensure Add Role gets displayed).

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -0,0 +1,280 @@
...
+
+  private showOrgPersonForm: boolean = false
+  private activeIndex: number = -1
+  private addEditInProgress: boolean = false
+  private currentOrgPerson: OrgPersonIF | null = null
+  private nextId: number = -1
+
+  readonly EntityTypes: {} = EntityTypes
+  readonly Roles: {} = Roles
+  readonly IncorporatorTypes: {} = IncorporatorTypes
```
severinbeauvais: You don't need to declare types in a TS file (only in the old Vue class object).

tests/unit/PeopleAndRoles.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+  it('Shows check mark next to roles added', () => {
+    store.state.stateModel.addPeopleAndRoleStep.orgPeople = getPersonList(['Director', 'Incorporator'])
+    const wrapper = wrapperFactory()
+    expect(wrapper.find(checkIncorporator).exists()).toBeTruthy()
+    expect(wrapper.find(checkDirector).exists()).toBeTruthy()
+    expect(wrapper.find(checkCompletingParty).exists()).toBeFalsy()
+    wrapper.destroy()
+    resetStore()
+  })
+})
```
severinbeauvais: Should also have tests for Show Errors.

severinbeauvais: Should also have tests for role logic (see role methods) and the other methods.

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -0,0 +1,533 @@
...
+  ]
+
+  // Life cycle methods
+  private created (): void {
+    if (this.initialValue) {
+      this.orgPerson = { ...this.initialValue }
+      this.isDirector = this.orgPerson.roles.includes(Roles.DIRECTOR)
+      this.isIncorporator = this.orgPerson.roles.includes(Roles.INCORPORATOR)
+      this.isCompletingParty = this.orgPerson.roles.includes(Roles.COMPLETING_PARTY)
+      this.inProgressMailingAddress = { ...this.orgPerson.address.mailingAddress }
```
severinbeauvais: Should any of these flags be watchers/computed properties so that they are automatically updated?

severinbeauvais:
Could one of you run `npm i` and check in the latest package lock file please?

src/components/AddPeopleAndRoles/OrgPerson.vue
severinbeauvais: Classes that not col.

severinbeauvais: Classes that not col.

src/components/AddPeopleAndRoles/OrgPerson.vue
severinbeauvais: I'm in favour of whitespace between blocks of code (including markup).

src/components/AddPeopleAndRoles/OrgPerson.vue
severinbeauvais: I wasn't confused by most of these methods and that's my litmus test for whether something needs comments, but then again helpful comments are always welcome.

severinbeauvais: I think you guys did a great job naming methods here and adding comments wouldn't add huge value.

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: Will follow up with Scott! 👍

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: Fixed!

src/components/dialogs/ConfirmDialog.vue
cameron-freshworks: Oh interesting, thanks!

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
lekshmimallika-aot: I checked this with Scott and even showed him how it is working. He mentioned that the red X should be shown only when you return to the page by clicking the link

src/components/AddPeopleAndRoles/OrgPerson.vue
lekshmimallika-aot: This is specifically mentioned in the ux pin that no warning should be shown when the user is removed by clicking the remove button from the popup

src/components/AddPeopleAndRoles/OrgPerson.vue
lekshmimallika-aot: There is no further update to the properties in this component other than the address which is handled in a different way altogether. So I dont think we need watchers here

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: Looking into this. It's definitely a result of some inherited css from Vuetify, should be able to override them! Working on er

cameron-freshworks: Spacing has been fixed but currently WIP on overriding the text color.

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: 155 chars for Corp Name. Will apply!

cameron-freshworks: Applied the rule of 30 chars to the first, middle and last names 👍

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: ok great, good to hear! I guess thats the true test, if someone can jump into the code and understand with ease!

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: Sure, makes sense, added!

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: Do you mean, returning a default value using it to return a default value if null?

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: fixed

src/store/getters/state-getters.ts
cameron-freshworks: removed

src/schemas/office-address-schema.ts
cameron-freshworks: Only difference is the validation on the Country and Region. I'm assuming the same rules don't apply to the offices as it does the directors being within Canada and BC. I believe we can use one schema and apply these conditionally. Will try!

cameron-freshworks: Renamed the `addressSchema` to `personAddressSchema` where applicable to differentiate between the Offices and Persons validation.

cameron-freshworks: * Future, perhaps we can refactor to one schema.

src/components/common/Stepper.vue
cameron-freshworks: K, noted and will update when all the pages are implemented!

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: oh, good to know! :P

src/components/common/Stepper.vue
cameron-freshworks: implemented!

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Fixed!

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Done!

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Good question. The designs don't actually translate to the specific colours in the themes. Same goes for the blue hex colour given in the design. I feel like i'm just guessing at which colour in the theme is closest..
Suggestions?

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: fixed!

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Wanted padding for the column but didn't want that inherited ( double padded ) for the child column

src/components/AddPeopleAndRoles/OrgPerson.vue
severinbeauvais: This.
```
return (this.orgPerson?.type === IncorporatorTypes.CORPORATION)
```

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: Done and done 👍

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
severinbeauvais: OK.

src/components/AddPeopleAndRoles/OrgPerson.vue
severinbeauvais: Weird, because we have that dialog when we delete the user from the table.

Oh well.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: Check this page for named colours and try to find a visual match. (I don't think we need exact RGB values.)
https://vuetifyjs.com/en/styles/colors/

severinbeauvais: ... And we should ask the designers to pick colours from the theme :)

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: Don't we need a space between `rem` and `!important`?

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: Done

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: I believe this is something we could refactor in the near future! If possible, since this is just a recommendation, can we leave this comment un-resolved so i can return to it post-PR?

src/components/ReviewConfirm/Summary.vue
cameron-freshworks: I also thought on that a bit! Would have loved for everything to be a single component with a flag but i also agree this is 100% situational.

src/components/ReviewConfirm/Summary.vue
cameron-freshworks: Good point, thanks. Done!

src/components/dialogs/ConfimRemoveDialog.vue
cameron-freshworks: I think anything is possible! This might require some more time than we have at this present moment. Once again i would request we extend this refactor until we time to revisit this.

src/views/AddPeopleAndRoles.vue
cameron-freshworks: Good catch, thanks!

tests/unit/ListPeopleAndRoles.spec.ts
cameron-freshworks: hahah... put down the first name that came into mind!

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: Done!

src/components/AddPeopleAndRoles/OrgPerson.vue
severinbeauvais: Was this because of style typo? Fixed now?

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
severinbeauvais: Yep, don't fix today. This was more of a observation although if there's a future opportunity to improve this then let's consider it.

src/components/dialogs/ConfimRemoveDialog.vue
severinbeauvais: OK, don't worry about it.

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: 100% agree!

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: ohh. The linters say no, my sanity says yes. But, i just removed the `rem` anyways because thats not required with a falsy value.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Agreed!

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: Fixed now! 👍

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: fixed!

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: Fixed

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: fixed 👍

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: fixed

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: fixed

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: fixed!

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: I can't actually find what you mean in regards to these two comments.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Thanks for the clarification!

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: I'm definitely ok with it as is, i will add this to the list of future code improvements tho!

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: I'll add some!

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: ah, yes! Can do that 👍

src/components/AddPeopleAndRoles/OrgPerson.vue
severinbeauvais: The labels are supposed to be different depending on entity type.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: I don't think it's interpreted as falsy. I think 0px = 0em = 0rem = just 0. There's probably actual guidance on this somewhere but I'm going on experience here :)

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: fixed

tests/unit/ListPeopleAndRoles.spec.ts
cameron-freshworks: As per discussions, we will expand on the unit testing in the next pr! 👍

tests/unit/PeopleAndRoles.spec.ts
cameron-freshworks: As per discussions, we will expand on the unit testing in the next pr! 👍

cameron-freshworks:
> Could one of you run `npm i` and check in the latest package lock file please?

Absolutely 👍

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: Found the bit in the design, updated 👍

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Fixed!

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Interesting, i trust you! Removed the rem either way 👍

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: Found this, updated!

severinbeauvais:
(This was co-authored between Cameron and Lekshmi, and both reviewed each other's work, so after I reviewed then I merged it.)

PR #59, opened 2020-03-18
Framework for actions and stepper

lekshmimallika-aot:
*Issue #:* /bcgov/entity#3024

*Description of changes:*
Framework changes for stepper and actions


src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -194,7 +194,7 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, EntityFilter
 
 .people-roles-summary-header {
   display: flex;
-  background-color: $gray1;
+  background-color: rgba(1, 51, 102, 0.15);
```
cameron-freshworks: oops, this was my request but come to think of we should change this to colour to one in our base theme that is as close to this colour as possible.  We might have to change this in the `SummaryDefineCompany` line 98 as well.  If i had to guess.... the theme colour closest to this might be `$BCgovBlue8`.

src/components/common/Actions.vue
```diff
@@ -5,39 +5,35 @@
...
+        @click="onClickSaveResume()">
         <span>Save and Resume Later</span>
       </v-btn>
     </div>
 
     <div class="buttons-right">
       <v-fade-transition hide-on-leave>
         <v-btn id="back-btn" large outlined
-          to="/define-company"
+          v-bind:to="previousRoute"
```
cameron-freshworks: Not sure but i don't believe `v-bind` is required if your prefer the shorthand `:`. Up to you!

src/components/common/Actions.vue
```diff
@@ -5,39 +5,35 @@
...
+          :disabled="isBusySaving">
           <v-icon>mdi-chevron-left</v-icon>
           <span>Back</span>
         </v-btn>
       </v-fade-transition>
 
       <v-fade-transition hide-on-leave>
         <v-btn id="review-confirm-btn" large color="primary"
-          to="/review-confirm"
+          v-bind:to="nextRoute"
```
cameron-freshworks: Same as above, up to you!

src/components/common/Actions.vue
```diff
@@ -149,6 +144,43 @@ export default class Actions extends Mixins(FilingTemplateMixin, LegalApiMixin)
...
+    const nextStep = this.next()
+    if (nextStep) {
+      return nextStep.text.replace('\n', ' ')
+    }
+    return ''
+  }
+
+  private next () : any {
+    const currentStep: number| null = this.$router.currentRoute.meta
+      ? this.$router.currentRoute.meta.step : null
```
cameron-freshworks: I'm not 100% sure but this could potentially be refactoring using optional chaining? Worth a try :)  
`const currentStep: number| null = this.$router.currentRoute.meta?.step`

src/components/common/Actions.vue
```diff
@@ -149,6 +144,43 @@ export default class Actions extends Mixins(FilingTemplateMixin, LegalApiMixin)
...
+    if (currentStep && currentStep < this.stateModel.maxStep) {
```
cameron-freshworks: Maybe applies here as well. `currentStep? < this.stateModel.MaxStep`. 

No real action required here, just purely suggestion if you feel optional chaining is a feature we should implement more to do away with short-circuiting properties.

src/components/common/Actions.vue
```diff
@@ -149,6 +144,43 @@ export default class Actions extends Mixins(FilingTemplateMixin, LegalApiMixin)
...
+      return this.getSteps.find(step => step.step === currentStep + 1)
+    }
+    return null
+  }
+
+  private get previousRoute (): string | null {
+    const currentStep: number| null = this.$router.currentRoute.meta
+      ? this.$router.currentRoute.meta.step : null
```
cameron-freshworks: Potentially same as above if optional chaining works. Your call!

src/components/common/Stepper.vue
```diff
@@ -27,153 +27,7 @@ import { GetterIF } from '@/interfaces'
...
-      icon: 'mdi-dots-horizontal',
-      disabled: true
-    },
-    {
-      id: 'step-5-btn',
-      icon: 'mdi-dots-horizontal',
-      disabled: true
-    }]
-  }
+  @Getter getSteps!: GetterIF
```
cameron-freshworks: This is a big change!

src/router/routes.ts
```diff
@@ -59,16 +61,32 @@ export const routes = [
...
+  {
+    path: '/step-4',
+    name: 'Step-4',
+    component: Step4Container,
+    meta: {
+      step: -1,
+      label: 'Add People and Roles',
+      requiresAuth: true
+    }
+  },
```
cameron-freshworks: Curious about the routes, not a question we need to answer now but we have to consider that steps 3 & 4 will be different based off on Entity type. 

It appears you've done that by establishing the steps 3 & 4 framework to be a generic container. I guess the labels just threw me off at first :)

lekshmimallika-aot: Step 3 and step 4 container (view) will be the same. Which component is rendered inside that will depend on the entity type. That way we dont have to worry abt handling all this in routes or steps

src/components/common/Stepper.vue
lekshmimallika-aot: Yes :) IMHO actions and stepper should have access to the same steps. So it is better that it is in a common place. Also the stepper should just display and bind the properties according to the data passed to it

src/store/getters/state-getters.ts
```diff
@@ -115,3 +115,55 @@ export const isBusySaving = (state: any): boolean => {
...
+ */
+export const getSteps = (state: any, getters:any): Array<any> => {
+  const steps: Array<any> = [{
+    id: 'step-1-btn',
+    step: 1,
+    icon: 'mdi-domain',
+    text: 'Define Your\nCompany',
+    to: '/define-company',
+    disabled: getters.isBusySaving,
+    valid: state.stateModel.defineCompanyStep.valid
```
cameron-freshworks: Not sure if it is overkill, but you could* make a steps Interface and apply that type definition to the `const steps: Array<StepIF>`.  As well as the the return on the method `getSteps`.

src/components/common/Actions.vue
lekshmimallika-aot: I tried optional chaining first. It was showing some error. Will look into it again

src/store/getters/state-getters.ts
```diff
@@ -115,3 +115,55 @@ export const isBusySaving = (state: any): boolean => {
...
+    id: 'step-5-btn',
+    step: 3,
+    icon: getters.isTypeBcomp ? 'mdi-text-box-check-outline' : 'mdi-file-document-box-check-outline',
+    text: 'Review\nand Confirm',
+    to: '/review-confirm',
+    disabled: getters.isBusySaving,
+    valid: getters.isApplicationValid
+  }]
+  return steps
+}
```
cameron-freshworks: Interesting change!

src/components/common/Actions.vue
cameron-freshworks: It might be that it returns `undefined` not `null`? Not 100% tho, i'll have to pull the code and try as well!

src/components/common/Stepper.vue
cameron-freshworks: Absolutely, i think it's a great idea!

src/components/common/Actions.vue
lekshmimallika-aot: Done

src/components/common/Actions.vue
lekshmimallika-aot: Done

src/components/common/Actions.vue
lekshmimallika-aot: Done

src/components/common/Actions.vue
lekshmimallika-aot: Done

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
lekshmimallika-aot: I tried $BCgovBlue8. The text is not visible. So will unbranch the file from this PR. looks like more styling is required

cameron-freshworks: This is fine for now! I would leave it in so it at the very least matches the `summarydefinecompany`. We can define and update to a theme colour later 👍

src/store/getters/state-getters.ts
```diff
@@ -115,3 +115,55 @@ export const isBusySaving = (state: any): boolean => {
 export const isApplicationValid = (state: any): boolean => {
   return (state.stateModel.defineCompanyStep.valid && state.stateModel.addPeopleAndRoleStep.valid)
 }
+
+/**
+ * Steps to be displayed in the stepper
+ */
+export const getSteps = (state: any, getters:any): Array<any> => {
+  const steps: Array<any> = [{
```
severinbeauvais: There is a lot of overlap between the steps and the routes, and I thought this would the opportunity to combine them.

In particular, each route could specify (in meta) the "next" and "prev" steps, if any (else null).

What do you think about this, instead of having 2 structures with the same number of steps and step ids?

src/store/state/state-model.ts
```diff
@@ -16,6 +16,8 @@ export const stateModel: StateModelIF = {
     certifiedBy: ''
   },
   currentStep: 1,
+  minStep: 1,
+  maxStep: 3,
```
severinbeauvais: Isn't minStep always 1?

Couldn't maxStep be computed?

I'm thinking there has to be a clean way to implement this without additional state properties.

src/views/Step3Container.vue
```diff
@@ -0,0 +1,16 @@
+<!--Container for share structure or rules depending on the type of entity-->
```
severinbeauvais: I think it would be better to have separate views for each entity type.

src/store/getters/state-getters.ts
lekshmimallika-aot: Actually I dont see much overlap. The two common properties are 'step' and 'to'. 'to' can be removed from the steps as it can be derived from the routes based on the step number. But I am not in agreement with pusing many extra properties to meta of the route. Stepper needs additional properties like valid, enabled disabled etc and in my opinion it shouldnot go to the routes as routes is for the full application and it need not have any information related to the component validity or state.

Also previous and next steps are not needed as in this case you will go to the next higher step or next lower step only when you click the next or previous button. That is why it is a computed property now.

src/store/state/state-model.ts
lekshmimallika-aot: Sure I can take it from that state and make it a local or computed property

src/views/Step3Container.vue
lekshmimallika-aot: So most of the top level pages are just containers for holding the child components. Max it does is pass the state as prop to the child component. I think having a top level container will hold and limit the logic with in that view

src/store/getters/state-getters.ts
severinbeauvais: OK for now then.

src/views/Step3Container.vue
severinbeauvais: But what if one entity type has 5 steps and another entity type has 6 steps?

lekshmimallika-aot: Ya sure. That is one of the advantage of retaining path in the step as well. If this has to be separate views you the step 3 'to' property in the step array can be modified as  to:getters.isTypeBcomp ? '/create-share-structure':'/create-rules' and there can be two routes pointing to two different views both with step as 3 in meta.

If one entity has different number of steps the step array returned would be different altogether and the same logic as above can be applied.

src/store/state/state-model.ts
lekshmimallika-aot: Have made the changes to make this a computed property

src/views/Step3Container.vue
severinbeauvais: OK

PR #62, opened 2020-03-20
Step 3 framework

lekshmimallika-aot:
*Issue #:* /bcgov/entity#3027

*Description of changes:* 

Step 3 data structure and framework updates. 
 - From what I understand, share class and share series has the same set of attributes. Share class c 
   an have an array of share series as child elements based on an attribute value.

Step 3 is enabled now  so that everyone can check the flow. We can leave it as enabled or make it disabled till step 3 is fully ready.

### DIFF
--- a/src/interfaces/stepper-interfaces/CreateShareStructure/create-share-structure-interface.ts
+++ b/src/interfaces/stepper-interfaces/CreateShareStructure/create-share-structure-interface.ts
@@ -1,0 +1,18 @@
+export interface ShareStructureIF {
+    valid: boolean;
+    shareClasses : ShareClassIF[];
+}
+
+export interface ShareClassIF {
+    id: number| null;
+    displayOrder: number; // To support move up and down functionality
+    type: string; // Indicates whether class or series
+    name: string;
+    maxNumberOfShares: number | null;
+    parValue: number | null;
+    currency: string;
+    hasNoMaximumShares: boolean;
+    hasNoPar: boolean;
+    hasRightsOrRestrictions: boolean;
+    series: ShareClassIF[];
+}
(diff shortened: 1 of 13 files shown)


src/interfaces/stepper-interfaces/CreateShareStructure/create-share-structure-interface.ts
```diff
@@ -0,0 +1,18 @@
+export interface ShareStructureIF {
+    valid: boolean;
+    shareClasses : ShareClassIF[];
+}
+
+export interface ShareClassIF {
+    id: number| null;
+    displayOrder: number; // To support move up and down functionality
```
severinbeauvais: Or do we want to simply reorder the array of shares?

severinbeauvais: Then again, when we restore a draft, we can't guarantee the order of an array, so leave this in.

severinbeauvais:
> * From what I understand, share class and share series has the same set of attributes. Share class c
>   an have an array of share series as child elements based on an attribute value.

Will you update the data store for share series in this ticket?

> Step 3 is enabled now so that everyone can check the flow. We can leave it as enabled or make it disabled till step 3 is fully ready.

I think you can leave it enabled.

src/views/CreateShareStructure.vue
```diff
@@ -0,0 +1,18 @@
...
+      </header>
+    </section>
+</template>
+
+<script lang="ts">
+// Libraries
+import { Component, Vue } from 'vue-property-decorator'
+
+@Component({})
+export default class CreateShareStructure extends Vue {}
```
cameron-freshworks: one thing i've noticed from trying to resume drafts is we need a way to re-validate after reloading a draft. I'm not sure the best solution on this, but i've fiddle around with the validating methods in a `mounted` hook in the frameworks. Just something to think about, maybe you have a better idea :)

PR #64, opened 2020-03-23
#2762 Step2 Drafts/Filing + NR Integration Updates.

cameron-freshworks:
*Issue #:* /bcgov/entity#2762

*Description of changes:*

* Includes data from step 2 into our filing body.
* Includes parsing data for step 2 from draft filings.
* Adjusts some legal api methods to adapt to some back end changes to the endpoints.
* Updates the unit tests to include the updated data set.
* Adjusts some string manipulation with the Namex NR Number handling.
* Removes hardcoded NR data from App.
* Assigns the EntityType as BCOMP until a time when Namex provides a Type. 

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -65,7 +65,7 @@
 
 <script lang="ts">
 // Libraries
-import { Component, Watch, Mixins } from 'vue-property-decorator'
+import { Component, Vue, Watch, Mixins } from 'vue-property-decorator'
 import { Action, State } from 'vuex-class'
 import axios from '@/utils/axios-auth'
 
--- a/src/App.vue
+++ b/src/App.vue
@@ -146,11 +146,11 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     // Evaluate name request pre conditions
     const nameRequest = await this.evaluateNRPreconditions()
     if (nameRequest && nameRequest.nrNum && nameRequest.isConsumable) {
-      // TODO: Handling different NR Formats
+      // TODO: Handling different NR Formats & accept the EntityType dynamically once Namex provides it.
       // Will probably change once we proxy through legal API with a consistent format
       this.setNameRequestState(
         { nrNumber: nameRequest.nrNum,
-          entityType: nameRequest.requestTypeCd,
+          entityType: EntityTypes.BCOMP,
           filingId: null })
       this.setCurrentDate(this.dateToUsableString(new Date()))
       this.onEntityTypeChanged(this.stateModel.nameRequest.entityType)
--- a/src/App.vue
+++ b/src/App.vue
@@ -149,10 +149,11 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
       // TODO: Handling different NR Formats
       // Will probably change once we proxy through legal API with a consistent format
       this.setNameRequestState(
-        { nrNumber: nameRequest.nrNum.replace('NR ', ''),
-          entityType: nameRequest.requestTypeCd,
+        { nrNumber: nameRequest.nrNum,
+          entityType: EntityTypes.BCOMP,
           filingId: null })
       this.setCurrentDate(this.dateToUsableString(new Date()))
+      this.onEntityTypeChanged(this.stateModel.nameRequest.entityType)
 
       try {
         // Retrieve draft filing if it exists for the nrNumber specified
--- a/src/App.vue
+++ b/src/App.vue
@@ -150,7 +150,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
       // Will probably change once we proxy through legal API with a consistent format
       this.setNameRequestState(
         { nrNumber: nameRequest.nrNum,
-          entityType: EntityTypes.BCOMP,
+          entityType: nameRequest.requestTypeCd,
           filingId: null })
       this.setCurrentDate(this.dateToUsableString(new Date()))
       this.onEntityTypeChanged(this.stateModel.nameRequest.entityType)
@@ -193,7 +193,6 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
   private storeAuthRoles (response): void {
     // NB: roles array may contain 'view', 'edit' or nothing
     const authRoles = response && response.data && response.data.roles
-    console.log(response)
     if (authRoles && authRoles.length > 0) {
       this.setAuthRoles(authRoles)
     } else {
--- a/src/App.vue
+++ b/src/App.vue
@@ -150,10 +150,10 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
       // Will probably change once we proxy through legal API with a consistent format
       this.setNameRequestState(
         { nrNumber: nameRequest.nrNum,
+          // TODO: Implement this check when POST-MVP: entityType: nameRequest.requestTypeCd,
           entityType: EntityTypes.BCOMP,
           filingId: null })
       this.setCurrentDate(this.dateToUsableString(new Date()))
-      this.onEntityTypeChanged(this.stateModel.nameRequest.entityType)
 
       try {
         // Retrieve draft filing if it exists for the nrNumber specified
@@ -280,7 +280,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     }
   }
 
-  @Watch('state.nameRequest.entityType')
+  @Watch('stateModel.nameRequest.entityType')
   private onEntityTypeChanged (val: string | null) : void {
     switch (val) {
       case EntityTypes.BCOMP:
--- a/src/App.vue
+++ b/src/App.vue
@@ -193,6 +193,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
   private storeAuthRoles (response): void {
     // NB: roles array may contain 'view', 'edit' or nothing
     const authRoles = response && response.data && response.data.roles
+    console.log(response)
     if (authRoles && authRoles.length > 0) {
       this.setAuthRoles(authRoles)
     } else {
--- a/src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
@@ -154,7 +154,7 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, EntityFilter
    * @returns The appropriate Corporation or Person name.
    */
   private formatName (filing: any): string {
-    return filing.person.orgName ? filing.person.orgName
+    return filing?.person?.orgName ? filing?.person?.orgName
       : `${filing.person.firstName} ${filing.person.middleName || ''} ${filing.person.lastName}`
   }
 
--- a/src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
@@ -194,7 +194,7 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, EntityFilter
 
 .people-roles-summary-header {
   display: flex;
-  background-color: rgba(1, 51, 102, 0.15);
+  background-color: $BCgovBlue5O;
   padding: 1.25rem;
 }
 
--- a/src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
@@ -194,7 +194,7 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, EntityFilter
 
 .people-roles-summary-header {
   display: flex;
-  background-color: $gray1;
+  background-color: rgba(1, 51, 102, 0.15);
   padding: 1.25rem;
 }
 
--- a/src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
@@ -134,7 +134,7 @@ import { OrgPersonIF } from '@/interfaces'
 })
 export default class ListPeopleAndRoles extends Mixins(CommonMixin, EntityFilterMixin) {
   // Store Properties
-  @Prop()
+  @Prop({ default: () => [] })
   private personList: Array<OrgPersonIF>
 
   @Prop({ default: false })
--- a/src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
+++ b/src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
@@ -134,7 +134,7 @@ import { OrgPersonIF } from '@/interfaces'
 })
 export default class ListPeopleAndRoles extends Mixins(CommonMixin, EntityFilterMixin) {
   // Store Properties
-  @Prop({ default: [] })
+  @Prop()
   private personList: Array<OrgPersonIF>
 
   @Prop({ default: false })
@@ -154,8 +154,8 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, EntityFilter
    * @returns The appropriate Corporation or Person name.
    */
   private formatName (filing: any): string {
-    return filing.orgName ? filing.orgName
-      : `${filing.firstName} ${filing.middleName || ''} ${filing.lastName}`
+    return filing.person.orgName ? filing.person.orgName
+      : `${filing.person.firstName} ${filing.person.middleName || ''} ${filing.person.lastName}`
   }
 
   /**
--- a/src/components/AddPeopleAndRoles/OrgPerson.vue
+++ b/src/components/AddPeopleAndRoles/OrgPerson.vue
@@ -343,8 +343,8 @@ export default class OrgPerson extends Mixins(EntityFilterMixin, CommonMixin) {
     return personAddress
   }
 
-  private setPersonRoles (): string [] {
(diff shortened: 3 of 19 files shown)


src/App.vue
```diff
@@ -146,13 +146,14 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     // Evaluate name request pre conditions
     const nameRequest = await this.evaluateNRPreconditions()
     if (nameRequest && nameRequest.nrNum && nameRequest.isConsumable) {
-      // TODO: Handling different NR Formats
+      // TODO: Handling different NR Formats & accept the EntityType dynamically once Namex provides it.
```
severinbeauvais: I think Namex does provide it.

Is there already a check that this is a "BC" NR? (I know atm there's only Corporation NRs so the current check should be commented out and enabled in the future; but the check should be there.)

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -154,8 +154,8 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, EntityFilter
    * @returns The appropriate Corporation or Person name.
    */
   private formatName (filing: any): string {
-    return filing.orgName ? filing.orgName
-      : `${filing.firstName} ${filing.middleName || ''} ${filing.lastName}`
+    return filing.person.orgName ? filing.person.orgName
```
severinbeauvais: For safety, you should check that filing isn't null and filing.person isn't null.
You may be able to use Optional Chaining (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining).

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -194,7 +194,7 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, EntityFilter
 
 .people-roles-summary-header {
   display: flex;
-  background-color: $gray1;
+  background-color: rgba(1, 51, 102, 0.15);
```
severinbeauvais: I liked the old code better. Why did this change?

src/App.vue
```diff
@@ -146,13 +146,14 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
...
       // Will probably change once we proxy through legal API with a consistent format
       this.setNameRequestState(
-        { nrNumber: nameRequest.nrNum.replace('NR ', ''),
-          entityType: nameRequest.requestTypeCd,
+        { nrNumber: nameRequest.nrNum,
+          entityType: EntityTypes.BCOMP,
           filingId: null })
       this.setCurrentDate(this.dateToUsableString(new Date()))
+      this.onEntityTypeChanged(this.stateModel.nameRequest.entityType)
```
severinbeauvais: You shouldn't need this. Perhaps you need a deep watch on line 283?

src/interfaces/stepper-interfaces/AddPeopleAndRole/org-person-interface.ts
```diff
@@ -1,12 +1,14 @@
 import { BaseAddressObjIF } from '@/interfaces'
 
 export interface OrgPersonIF {
-    id: number| null
-    type: string;
+  person: {
+    id: number | null
+    partyType: string;
```
severinbeauvais: Should this be an enum type?

src/interfaces/stepper-interfaces/AddPeopleAndRole/org-person-interface.ts
```diff
@@ -1,12 +1,14 @@
...
     firstName: string;
     middleName?: string;
     lastName: string;
     orgName: string
-    roles: string[];
-    address: BaseAddressObjIF;
+  },
+  roles: string[];
```
severinbeauvais: Should this be an array of enum types?

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -110,12 +110,14 @@ export default class PeopleAndRoles extends Mixins(EntityFilterMixin) {
...
-    middleName: '',
-    orgName: '',
-    type: '',
+    person: {
+      id: null,
+      firstName: '',
+      lastName: '',
+      middleName: '',
+      orgName: '',
+      partyType: ''
```
severinbeauvais: If you change this property to an enum type, you may need to initialize it to null.

src/App.vue
cameron-freshworks: The only property i could see that is close is the `requestTypeCd` prop, which is sending back an unrecognizable type for our front end. 

I updated to leave the check in Commented out for MVP 👍

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: For sure, will update!

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -140,13 +142,18 @@ export default class PeopleAndRoles extends Mixins(EntityFilterMixin) {
   readonly Roles = Roles
   readonly IncorporatorTypes = IncorporatorTypes
 
+  mounted (): void {
+    this.setAddPeopleAndRoleStepValidity(this.hasValidRoles())
+  }
+
   // Methods
   private addOrgPerson (rolesToInitialize: string[], type: string): void {
```
severinbeauvais: Can "type" be a type (enum) instead of a string?

src/mixins/legal-api-mixin.ts
```diff
@@ -61,7 +61,7 @@ export default class LegalApiMixin extends Vue {
    * @param isDraft Boolean indicating whether to complete filing.
    */
   private createFiling (data: object, isDraft: boolean): Promise<any> {
-    let url = ''
+    let url = 'https://legal-api-dev.pathfinder.gov.bc.ca/api/v1/businesses'
```
severinbeauvais: This needs to be a property from configuration.json (locally) or config map (deployed). Same as in Business Filings.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Sorry, the theme colour was wrong. It is a header that is used in a few places but we hadn't established which theme colour it most closely resembles yet. I was updating this so it was the same colour used in the Step1 Review Header.

cameron-freshworks: I'll try to match to a theme colour and update this in step 1 and step 2 review headers 👍

src/mixins/namex-api-mixin.ts
```diff
@@ -43,7 +43,7 @@ export default class NameXApiMixin extends Vue {
   async queryNameRequest (nrNumber: string): Promise<any> {
     if (nrNumber) {
       // TODO FIX - Override until we can proxy through legal api
-      let url = `https://namex-dev.pathfinder.gov.bc.ca/api/v1/requests/NR ${nrNumber}`
+      let url = `https://namex-dev.pathfinder.gov.bc.ca/api/v1/requests/${nrNumber}`
```
severinbeauvais: Needs to be a property.

src/App.vue
cameron-freshworks: Interesting, i'll look into the deep watch.

src/mixins/legal-api-mixin.ts
cameron-freshworks: I did this so i could post to businesses as the URL thats set in the config map has the `/` on the end which is invalid to a request made directly to businesses.

tests/unit/Actions.spec.ts
```diff
@@ -113,7 +113,41 @@ describe('Actions Filing Functionality', () => {
...
+        },
+        parties: [
+          {
+            person: {
+              id: 0,
+              firstName: 'Cameron',
+              lastName: 'Bowler',
+              middleName: 'D',
+              orgName: '',
+              partyType: 'Person'
```
severinbeauvais: NB: in the unit tests, use a string like this instead of an enum type.
1. this is "data" that maybe the data provider didn't type, so no type here either
2. this makes the unit test more fragile, which is good for verifying that enum changes are tested

tests/unit/Actions.spec.ts
```diff
@@ -113,7 +113,41 @@ describe('Actions Filing Functionality', () => {
...
+              deliveryAddress: {
+                streetAddress: '122-12210 Boul De Pierrefonds',
+                streetAddressAdditional: '',
+                addressCity: 'Pierrefonds',
+                addressRegion: 'QC',
+                postalCode: 'H9A 2X6',
+                addressCountry: 'CA'
+              }
+            },
+            roles: [
```
severinbeauvais: As above, strings are good in the unit tests.

src/mixins/namex-api-mixin.ts
cameron-freshworks: Sorry, this was code already in. You'd like to see this in the config map?

ochiu: @severinbeauvais @cameron-freshworks this call is meant to be proxied through legal api when that endpoint is ready, do we need a property for it?

src/App.vue
cameron-freshworks: As it was, it wasn't updating.

src/interfaces/stepper-interfaces/AddPeopleAndRole/org-person-interface.ts
cameron-freshworks: Can be for sure!

src/interfaces/stepper-interfaces/AddPeopleAndRole/org-person-interface.ts
cameron-freshworks: Can be for sure!

tests/unit/Actions.spec.ts
cameron-freshworks: Sounds good

src/App.vue
cameron-freshworks: * Fixed, the reference to store was incorrect.

src/mixins/legal-api-mixin.ts
cameron-freshworks: My idea was to change the urls in the config map, but for congruency they were appended with the /. I'm open to any suggestions tho!

src/mixins/namex-api-mixin.ts
cameron-freshworks: Need clarification here :)

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: Or create a new colour that can be used elsewhere.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -154,8 +154,8 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, EntityFilter
    * @returns The appropriate Corporation or Person name.
    */
   private formatName (filing: any): string {
-    return filing.orgName ? filing.orgName
-      : `${filing.firstName} ${filing.middleName || ''} ${filing.lastName}`
+    return filing?.person?.orgName ? filing?.person?.orgName
```
severinbeauvais: FYI, this will return null if filing is null or person is null (or orgName is null). This is probably OK.

src/mixins/legal-api-mixin.ts
severinbeauvais: Well, we could remove "businesses" from the config key and add it every time we use it, either with no slash or with a slash and other params.
It would make the LEGAL_API consistent with the other URL definitions.
Eg,
```
"LEGAL_API_URL": "https://legal-api-dev.pathfinder.gov.bc.ca/api/v1/"
```

src/mixins/namex-api-mixin.ts
severinbeauvais: I forgot this will eventually be proxied and changed. Leave this for now.

severinbeauvais: Wow, I even missed Ody's comments. Sorry!

src/mixins/legal-api-mixin.ts
severinbeauvais: If you go this route, add a comment to the release ticket to update the config map in the different Openshift environments.

cameron-freshworks:
> I approve.
> Up to you whether to resolve enums here or in a future PR.
> Would be nice to resolve the API_URL key issue in this PR.

I think i'd like to resolve this here! So hold off the merge for now :) thanks!

src/mixins/legal-api-mixin.ts
cameron-freshworks: Ok sounds great! Will update this.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Great idea!

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: Good catch here. Applies to both arguments!

src/mixins/legal-api-mixin.ts
cameron-freshworks: This has been updated. Worth another look tho! If your ok with this, this PR can be merged 👍

src/mixins/legal-api-mixin.ts
```diff
@@ -30,16 +30,27 @@ export default class LegalApiMixin extends Vue {
...
-      await this.createFiling(filing, isDraft)
+      filingComplete = await this.createFiling(filing, isDraft)
     }
 
-    // Complete a filing if not draft
-    if (!isDraft) await this.completeFiling(filing)
+    if (!isDraft && filingComplete) {
+      const paymentToken = filingComplete.header.paymentToken
+      const baseUrl = sessionStorage.getItem('BASE_URL')
+      const returnURL = encodeURIComponent(baseUrl + 'review-confirm?nrNumber=' +
```
severinbeauvais: Return URL should be "draft incorporation UI", which is the Business Filings UI, which we haven't updated yet...
For now, I suggest returning to the businesses main page. (Anyway, returning to this UI is incorrect.)
... Unless you're thinking we come back here for now (while developing), as long as it doesn't break anything to file...?

src/mixins/legal-api-mixin.ts
```diff
@@ -61,7 +72,7 @@ export default class LegalApiMixin extends Vue {
    * @param isDraft Boolean indicating whether to complete filing.
    */
   private createFiling (data: object, isDraft: boolean): Promise<any> {
-    let url = 'https://legal-api-dev.pathfinder.gov.bc.ca/api/v1/businesses'
+    let url = 'businesses'
```
severinbeauvais: Great!

src/components/AddPeopleAndRoles/OrgPerson.vue
```diff
@@ -368,7 +368,7 @@ export default class OrgPerson extends Mixins(EntityFilterMixin, CommonMixin) {
     }
   }
 
-  private isRoleLocked (role: string) : boolean {
+  private isRoleLocked (role: Roles) : boolean {
```
severinbeauvais: Convention is that a singular would be called "Role" and a list or array of them would be called "Roles".
Small change, sorry, but what would you call an array of Roles? Roless ? (haha)

severinbeauvais: BTW, it's OK for enums to be plural, because they're passed in as the enum name but are always dereferenced in use, eg. `IncorporatorTypes.PERSON`.

src/mixins/legal-api-mixin.ts
cameron-freshworks: Bonehead move, we already discussed too. I will update to return to the businesses main page 👍 I don't think we anticipate coming back to the incorporations flow. If payment fails for example i think we still resume from the manage businesses page?

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: 😆 I agre. Change required here?

src/mixins/legal-api-mixin.ts
severinbeauvais: Payment retries and resuming will be from the "draft incorporations UI"... next sprint. For now, the businesses main page is good.

src/components/AddPeopleAndRoles/OrgPerson.vue
severinbeauvais: Nah, you can do it in a future PR.

By the way, I can go-go-go, but we should work at a comfortable pace. No one needs to work evenings or weekends atm. Don't let me push you :)

src/mixins/legal-api-mixin.ts
cameron-freshworks: Ah right! We  can update once we have that piece :) Thanks!

src/components/AddPeopleAndRoles/OrgPerson.vue
cameron-freshworks: Sounds good to me!  😆 Don't worry about that one bit, no pressure in the world will outdo the pressure i get from inside the house!

severinbeauvais:
Looks good to me.

PR #65, opened 2020-03-23
renamed config key + cleanup

severinbeauvais:
*Issue #:* /bcgov/entity#3083

*Description of changes:*
- renamed VUE_APP_AUTH_ROOT_API to AUTH_API_URL
- renamed API_URL to LEGAL_API_URL
- renamed KEYCLOAK_CONFIG_URL to KEYCLOAK_CONFIG_PATH
- deleted obsolete  SIGNIN_URL
- delete obsolete auth functions, classes, constants and interfaces
- added some missing types
- simplified Vuetify init


src/App.vue
```diff
@@ -280,8 +280,8 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     }
   }
 
-  @Watch('stateModel.nameRequest.entityType')
-  private onEntityTypeChanged (val: string | null) : void {
+  @Watch('state.nameRequest.entityType')
```
cameron-freshworks: I'm not sure this is correct. I had to change the reference to the `stateModel` which we had bound to `App`, as there is no `state` property bound here.

src/main.ts
```diff
@@ -4,7 +4,7 @@ import 'regenerator-runtime/runtime' // to use transpiled generator functions
 
 // Vue Libraries
 import Vue from 'vue'
-import vuetify from '@/plugins/vuetify'
+import Vuetify from 'vuetify/lib'
```
cameron-freshworks: interesting, what is the difference here?

src/App.vue
severinbeauvais: Oops, merge issue. Fixing now.

src/utils/auth-helpers.ts
```diff
@@ -1,119 +0,0 @@
...
-  private storeAuthRoles (response: AuthIF): void {
-    const authRoles = response && response.data && response.data.roles
-    if (authRoles && authRoles.length > 0) {
-      // set auth roles to store
-      store.dispatch(SET_AUTH_ROLES, authRoles)
-    } else {
-      throw new Error('Invalid auth roles')
-    }
-  }
-}
```
cameron-freshworks: Good clean up! I guess all of this is obsolete with the new kc integration?

src/App.vue
severinbeauvais: I found another place where my rebase restored some deleted whitespace :/

cameron-freshworks: excellent!

src/main.ts
severinbeauvais: We were calling a plugin that did very little so I moved its code here. I think it's cleaner :)

severinbeauvais: That you can't tell the difference proves the plugin did very little :)

src/App.vue
lekshmimallika-aot: Should be stateModel I think

src/utils/auth-helpers.ts
severinbeauvais: Yep!

src/App.vue
severinbeauvais: Yes, fixed.

PR #71, opened 2020-04-01
2903 - update NR to use legal api proxy endpoint

ochiu:
*Issue #:* /bcgov/entity#2903

*Description of changes:*
- Renamed namex-api-mixin to name-request-mixin
- Removed old temporary token code
- Added query name request call to legal-api-mixin
- Moved NR Authorizations helper into name-request-mixin
- Updated NR expiration status check to look at expirationDate rather than status


src/mixins/name-request-mixin.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    return { isConsumable: true, expired: false, approved: true }
+  }
+
+  /**
+   * Fetch Authorizations by NR number
+   * @param nr the name request number (e.g NR 1234567)
+   */
+  getNRAuthorizations (nrNumber: string): Promise<any> {
```
severinbeauvais, line 48: Might make sense to move this to Legal API mixin. Thoughts?

src/App.vue
```diff
@@ -220,7 +209,6 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
 
     return this.getNRAuthorizations(queryNrNumber).then(async data => {
       this.storeAuthRoles(data) // throws if no role
-      await this.intializeNameXToken() // TODO : temporary token retrieval code
       const nrResponse = await this.queryNameRequest(queryNrNumber)
 
       // NR not found
```
severinbeauvais, line 214: In Business Filings UI, I added some extra validity checks that you might also want to implement.
See https://github.com/bcgov/business-filings-ui/blob/cb421eed27d97bc63dc99596d42277d3be7e0726/src/App.vue#L422

src/mixins/name-request-mixin.ts
ochiu: I was pondering this as well, so good to have a second opinion. I moved it to the legal api mixin now.

src/App.vue
ochiu: Thanks! I added that function to the name request mixin and added a state to the NR invalid dialog as well as updated the tests.

PR #74, opened 2020-04-07
2465 - Display NR Data

ochiu:
*Issue #:* /bcgov/entity#2465

*Description of changes:*
- Add in NameRequestInfo component
- Updated EntityInfo component to use NR Data
- Added helper function to create name request state from NR payload
- Updated review section to use NR Data for the company name
- Added name request data structure interfaces (NR details, NR applicant)
- Added in vuex getters for NR details/applicant, approved name
- Removed BCOMP, COOP, Reset buttons in Name request section in define company and updated tests
- Added Name request info component tests

### DIFF
--- a/src/components/common/EntityInfo.vue
+++ b/src/components/common/EntityInfo.vue
@@ -48,7 +48,7 @@ import { Getter } from 'vuex-class'
 import { GetterIF } from '@/interfaces'
 
 @Component
-export default class EntityInfo extends Mixins() {
+export default class EntityInfo extends Vue {
   // Global getters
   @Getter isEntityType!: GetterIF
   @Getter isTypeBcomp!: GetterIF
@@ -57,7 +57,7 @@ export default class EntityInfo extends Mixins() {
   @Getter getApprovedName!: GetterIF
 
   /** The entity title  */
-  entityTitle (): string {
+  private entityTitle (): string {
     if (this.isTypeBcomp) {
       return 'Incorporate a BC Benefit Company'
     } else if (this.isTypeCoop) {
--- a/src/components/common/EntityInfo.vue
+++ b/src/components/common/EntityInfo.vue
@@ -15,16 +15,9 @@
 
         <!-- Header With NR Data -->
         <v-list-item-content id="nr-header" v-show="isEntityType">
-
-          <!-- Subtitle -->
-          <v-list-item-subtitle class="entity-subtitle">
-            <span v-if="isTypeBcomp">Incorporate a BC Benefit Company</span>
-            <span v-if="isTypeCoop">Incorporate a BC Cooperative Association</span>
-          </v-list-item-subtitle>
-
           <!-- Company Name -->
           <v-list-item-title class="header-title" id="entity-legal-name">
-            <span>XYZ Inc.</span>
+            <span>{{ getApprovedName }}</span>
           </v-list-item-title>
 
           <!-- TODO: display Designation? See mockup. -->
@@ -32,9 +25,9 @@
           <!-- Company Number -->
           <v-list-item-subtitle class="business-info">
             <dl>
-              <dt>Name Request No:</dt>
+              <dt> {{ entityTitle() }} Name Request No:</dt>
               <dd class="ml-2" id="entity-nr-number">
-                <span>NR1234567</span>
+                <span>{{ getBusinessIdentifier }}</span>
               </dd>
             </dl>
           </v-list-item-subtitle>
@@ -48,19 +41,33 @@
 
 <script lang="ts">
 // Libraries
-import { Component, Vue } from 'vue-property-decorator'
+import { Component, Mixins, Vue } from 'vue-property-decorator'
 import { Getter } from 'vuex-class'
 
 // Interfaces
 import { GetterIF } from '@/interfaces'
 
 @Component
-export default class EntityInfo extends Vue {
+export default class EntityInfo extends Mixins() {
   // Global getters
   @Getter isEntityType!: GetterIF
   @Getter isTypeBcomp!: GetterIF
   @Getter isTypeCoop!: GetterIF
+  @Getter getBusinessIdentifier!: GetterIF
+  @Getter getApprovedName!: GetterIF
+
+  /** The entity title  */
+  entityTitle (): string {
+    if (this.isTypeBcomp) {
+      return 'Incorporate a BC Benefit Company'
+    } else if (this.isTypeCoop) {
+      return 'Incorporate a BC Cooperative Association'
+    }
+
+    return ''
+  }
 }
+
 </script>
 
 <style lang="scss" scoped>
--- a/src/components/common/EntityInfo.vue
+++ b/src/components/common/EntityInfo.vue
@@ -41,7 +41,7 @@
 
 <script lang="ts">
 // Libraries
-import { Component, Mixins, Vue } from 'vue-property-decorator'
+import { Component, Vue } from 'vue-property-decorator'
 import { Getter } from 'vuex-class'
 
 // Interfaces
--- a/src/components/common/NameRequestInfo.vue
+++ b/src/components/common/NameRequestInfo.vue
@@ -8,12 +8,12 @@
       </v-col>
       <v-col>
         <ul>
-          <li class="name-request-title mb-4">
+          <li class="name-request-title">
             <strong>{{ getBusinessIdentifier }}</strong> - {{ getNameRequestDetails.approvedName}}
           </li>
-          <li>Entity Type: {{ entityTypeDescription() }}</li>
-          <li class="mb-4"> Request Type: {{ requestType() }}</li>
-          <li>Expiry Date: {{ formattedExpirationDate() }}</li>
+          <li class="mt-4">Entity Type: {{ entityTypeDescription() }}</li>
+          <li> Request Type: {{ requestType() }}</li>
+          <li class="mt-4">Expiry Date: {{ formattedExpirationDate() }}</li>
           <li>Status: {{ getNameRequestDetails.status }}</li>
           <li>Condition/Consent: {{ conditionConsent() }}</li>
         </ul>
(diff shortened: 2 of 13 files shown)


src/components/common/EntityInfo.vue
```diff
@@ -15,26 +15,19 @@
 
         <!-- Header With NR Data -->
         <v-list-item-content id="nr-header" v-show="isEntityType">
-
-          <!-- Subtitle -->
-          <v-list-item-subtitle class="entity-subtitle">
-            <span v-if="isTypeBcomp">Incorporate a BC Benefit Company</span>
-            <span v-if="isTypeCoop">Incorporate a BC Cooperative Association</span>
-          </v-list-item-subtitle>
```
severinbeauvais, line 23: Did you confirm with Scott that we no longer need this subtitle in the header?

The reason I think we might want to keep this is that the user doesn't have the company yet... they are still in the process of incorporating.

Anyway I'm just assuming things so please check with Scott. Thanks.

src/components/common/EntityInfo.vue
```diff
@@ -48,19 +41,33 @@
...
-import { Component, Vue } from 'vue-property-decorator'
+import { Component, Mixins, Vue } from 'vue-property-decorator'
 import { Getter } from 'vuex-class'
 
 // Interfaces
 import { GetterIF } from '@/interfaces'
 
 @Component
-export default class EntityInfo extends Vue {
+export default class EntityInfo extends Mixins() {
```
severinbeauvais, line 51: You don't need to do this if you're not importing any mixins.

src/components/common/NameRequestInfo.vue
```diff
@@ -0,0 +1,146 @@
...
+      return 'BC Benefit Company'
+    } else if (this.isTypeCoop) {
+      return 'BC Cooperative Association'
+    }
+
+    return ''
+  }
+
+  /** The request type */
+  requestType () : string {
```
severinbeauvais, line 78: FYI, these could be getters (but I'm not saying to change it), and they should be private.

src/components/common/EntityInfo.vue
ochiu: Ahh yes, I moved it to match the mock ups.

src/components/common/NameRequestInfo.vue
```diff
@@ -0,0 +1,146 @@
...
+          <strong>Name Request</strong>
+        </label>
+      </v-col>
+      <v-col>
+        <ul>
+          <li class="name-request-title mb-4">
+            <strong>{{ getBusinessIdentifier }}</strong> - {{ getNameRequestDetails.approvedName}}
+          </li>
+          <li>Entity Type: {{ entityTypeDescription() }}</li>
+          <li class="mb-4"> Request Type: {{ requestType() }}</li>
```
severinbeauvais, line 15: FWIW, this is more of a personal preference thing (so I think it's OK as long as you're consistent -- and you are here), but I prefer to add top margins/padding to the lower element (since it "knows" what it comes after) instead of adding bottom margins/padding to the upper element (since it can only "assume" what come after it. This makes more sense when there's a lot of conditionals.

src/components/common/NameRequestInfo.vue
```diff
@@ -0,0 +1,146 @@
...
+@Component
+export default class NameRequestInfo extends Mixins(DateMixin) {
+  // Global getters
+  @Getter isEntityType!: GetterIF;
+  @Getter isTypeBcomp!: GetterIF;
+  @Getter isTypeCoop!: GetterIF;
+  @Getter getBusinessIdentifier!: GetterIF;
+  @Getter getNameRequestDetails!: NameRequestDetailsIF;
+  @Getter getNameRequestApplicant!: NameRequestApplicantIF;
+  getName = getName
```
severinbeauvais, line 64: You may not need this, since you don't use it in the template. I think if you leave it out, the script can still find it in scope (from the mixin).

src/components/common/NameRequestInfo.vue
```diff
@@ -0,0 +1,146 @@
...
+    // Build address lines
+    let address = this.getNameRequestApplicant.addressLine1
+    if (this.getNameRequestApplicant.addressLine2) {
+      address = `${address}, ${this.getNameRequestApplicant.addressLine2}`
+    }
+    if (this.getNameRequestApplicant.addressLine3) {
+      address = `${address}, ${this.getNameRequestApplicant.addressLine3}`
+    }
+
+    return `${address}, ${city}, ${stateProvince}, ${postal}, ${country}`
```
severinbeauvais, line 125: Nice and straighforward code. I like it :)

src/components/common/NameRequestInfo.vue
ochiu: I thought so as well, but when I remove that. I get:
`114:14 Property 'getName' does not exist on type 'NameRequestInfo'.
    112 |     const postal = this.getNameRequestApplicant.postalCode
    113 |     const country = this.getNameRequestApplicant.countryTypeCode
  > 114 |       ? this.getName(this.getNameRequestApplicant.countryTypeCode) : ''`

src/store/getters/state-getters.ts
```diff
@@ -68,6 +69,27 @@ export const getBusinessIdentifier = (state: any): string => {
   return state.stateModel.nameRequest.nrNumber
 }
 
+/**
+ * Return an approved name request name if it exists
```
severinbeauvais, line 73: This is a small nit but please remove "if it exists", since the code returns something whether or not the state property exists :]

src/components/common/NameRequestInfo.vue
ochiu: nvm! I was using `this`. Will fix.

src/store/state/state-model.ts
```diff
@@ -9,6 +9,8 @@ export const stateModel: StateModelIF = {
   nameRequest: {
     nrNumber: '',
     entityType: '',
+    details: {},
+    applicant: {},
```
severinbeauvais, line 13: I'm not sure if it makes any difference here, but you could initialize the data as
`applicant: {} as NameRequestApplicantIF ,`

Thoughts?

src/mixins/name-request-mixin.ts
```diff
@@ -1,18 +1,59 @@
...
  */
 @Component
 export default class NameRequestMixin extends Vue {
+  /**
+   * Generate name request state for the store
+   * @param nr the name request response payload
+   * @param filingId the filing id
+   */
+  generateNameRequestState (nr: any, filingId: number): NameRequestIF {
+    const approvedName = nr.names.filter(name => name.state === NameRequestStates.APPROVED)[0].name
```
severinbeauvais, line 18: You could also use `find`, which returns a single entry (or undefined). Theoretically it's faster because it will return when it finds the first entry, instead of filtering the entire array.

src/components/common/EntityInfo.vue
ochiu: Fixed.

src/components/common/NameRequestInfo.vue
ochiu: Updated to private

src/components/common/NameRequestInfo.vue
ochiu: Makes a lot of sense thanks! Updated.

src/store/getters/state-getters.ts
ochiu: Updated

src/store/state/state-model.ts
ochiu: Done. Good to know the interfaces at a glance.

src/mixins/name-request-mixin.ts
ochiu: Updated! Thanks!

src/components/common/EntityInfo.vue
```diff
@@ -48,7 +41,7 @@
 
 <script lang="ts">
 // Libraries
-import { Component, Vue } from 'vue-property-decorator'
+import { Component, Mixins, Vue } from 'vue-property-decorator'
```
severinbeauvais, line 44: Unneeded import.

src/components/common/NameRequestInfo.vue
```diff
@@ -0,0 +1,145 @@
...
+      return 'BC Cooperative Association'
+    }
+
+    return ''
+  }
+
+  /** The request type */
+  private requestType () : string {
+    return 'New Business'
+  }
```
cameron-freshworks, line 79: Whats the purpose of this method?

src/components/common/EntityInfo.vue
severinbeauvais: Fixed.

src/components/common/NameRequestInfo.vue
ochiu: The request type for our UI doesn't seem to match the request type for the actual NR payload. I added this as a function as I thought we may need some processing logic after some feedback similar to how the consent value is derived. 

Maybe I am overthinking it though :)

severinbeauvais: I like it, as it provides the value of the key-value pair in the template (like all the other values).

src/components/common/EntityInfo.vue
ochiu: Fixed.

src/components/common/NameRequestInfo.vue
```diff
@@ -0,0 +1,145 @@
...
+    return this.toReadableDate(this.getNameRequestDetails.expirationDate)
+  }
+
+  /** Return consent received string by checking if conditional and if consent has been received */
+  private conditionConsent (): string {
+    if (this.getNameRequestDetails.status === NameRequestStates.CONDITIONAL) {
+      return this.getNameRequestDetails.consentFlag === 'R' ? 'Received' : 'Not Received'
+    }
+
+    return ''
```
severinbeauvais, line 92: Should this return "Not applicable", or hide this line, if no consent is not required? Ask Scott...

ochiu: Checking with Scott.

severinbeauvais: Great. Feel free to merge this PR after this conversation is resolved.

ochiu: Conclusion was to hide the consent field when it isn't relevant. 
- added logic in the template to only show consent field when relevant
- updated tests

PR #94, opened 2020-04-15
added refresh/close/cancel warning

severinbeauvais:
*Issue #:* /bcgov/entity#3338

*Description of changes:*
- added Unsaved Changes dialog
- added onbeforeunload event handler
- added goToDashboard event handler to check for data changes
- reset data flag after fetch
- ignore data changes while pages load
- fixed template errors when certifyStatementResource isn't loaded yet
- reset data flag after saves
- deleted obsolete External Resources
- updated unit tests
- misc cleanup


src/App.vue
```diff
@@ -142,16 +154,6 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
...
-  /** True if route is Signin. */
-  private get isSigninRoute (): boolean {
-    return Boolean(this.$route.name === 'signin')
-  }
-
-  /** True if route is Signout. */
-  private get isSignoutRoute (): boolean {
-    return Boolean(this.$route.name === 'signout')
-  }
-
```
severinbeauvais, line 154: These were only used in one place so I moved them inside that method.

src/App.vue
```diff
@@ -162,6 +164,52 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     return (process.env.JEST_WORKER_ID !== undefined)
   }
 
+  /** Called when component is created. */
+  private created (): void {
+    // before unloading this page, if there are changes then prompt user
+    window.onbeforeunload = (event) => {
```
severinbeauvais, line 170: This catches the "window closed" or "window reloaded" events.

src/App.vue
```diff
@@ -162,6 +164,52 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
...
+      if (this.haveChanges) {
+        event.preventDefault()
+        // NB: custom text is not supported in all browsers
+        event.returnValue = 'You have unsaved changes. Are you sure you want to leave?'
+      }
+    }
+  }
+
+  /** Called to redirect to dashboard. */
+  private goToDashboard (): void {
```
severinbeauvais, line 180: This method has to be here (instead of in Actions.vue) because the Confirm Dialog is on this page.

src/App.vue
```diff
@@ -307,26 +356,24 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
       default:
         this.filingData = []
     }
-
-    this.setCertifyStatementResource(
-      entityType ? CertifyStatementResource.find(x => x.entityType === entityType) : null
-    )
```
severinbeauvais, line 313: This didn't belong here so I moved it up.

src/components/dialogs/ConfirmDialog.vue
```diff
@@ -9,7 +9,7 @@
 
     <v-card>
       <v-card-title>{{ title }}</v-card-title>
-      <v-card-text class="black--text" v-show="!!message" v-html="message"></v-card-text>
+      <v-card-text class="black--text pre-wrap" v-show="!!message">{{ message }}</v-card-text>
```
severinbeauvais, line 12: This is all the same cleanup that I did previously in Filings UI.

src/interfaces/utils-interfaces/confirm-dialog-type.ts
```diff
@@ -1,5 +1,5 @@
 import Vue from 'vue'
 
-export interface DialogType extends Vue {
+export interface ConfirmDialogType extends Vue {
```
severinbeauvais, line 3: I renamed this in case we have other dialog types :)

src/interfaces/stepper-interfaces/ReviewConfirm/certify-interface.ts
```diff
@@ -1,4 +1,4 @@
 export interface CertifyIF {
-    certifyFormValid: boolean
+    valid: boolean
```
severinbeauvais, line 2: I renamed this to make it consistent with the other state objects.

src/store/mutations/mutations-model.ts
```diff
@@ -1,5 +1,17 @@
...
+  state.resourceModel.certifyStatementResource = certifyStatementResource
+}
+
+export const mutateKeycloakRoles = (state: any, keyCloakRoles: Array<string>) => {
+  state.stateModel.tombstone.keycloakRoles = keyCloakRoles
+}
+
+export const mutateAuthRoles = (state: any, authRoles: Array<string>) => {
+  state.stateModel.tombstone.authRoles = authRoles
+}
```
severinbeauvais, line 14: I moved these up from below to keep them together and apart from the rest (ie, tombstone vs other properties).

src/store/mutations/mutations-model.ts
```diff
@@ -17,32 +29,19 @@ export const mutateIsFilingPaying = (state: any, isFilingPaying: boolean) => {
...
-}
-
-export const mutateCertifyStatementResource = (state: any, certifyStatementResource: CertifyStatementIF) => {
-  state.resourceModel.certifyStatementResource = certifyStatementResource
+  if (!state.stateModel.ignoreChanges) mutateHaveChanges(state, true)
 }
 
 export const mutateCertifyState = (state: any, certifyState: CertifyIF) => {
   state.stateModel.certifyState = certifyState
+  if (!state.stateModel.ignoreChanges) mutateHaveChanges(state, true)
```
severinbeauvais, line 39: This is a special check to allow pages to load/update data without counting that as a data change.

If you add new state data, please copy this code.

src/views/AddPeopleAndRoles.vue
```diff
@@ -37,9 +37,19 @@ export default class AddPeopleAndRoles extends Mixins(EntityFilterMixin) {
...
   readonly EntityTypes = EntityTypes
+
+  /** Called when component is created. */
+  private created (): void {
+    // ignore data changes until page has loaded
+    this.setIgnoreChanges(true)
+    Vue.nextTick(() => {
+      this.setIgnoreChanges(false)
+    })
+  }
```
severinbeauvais, line 52: I found that some pages load and update the data in the period before they (and their children) are created and the next tick. (Note that 'mounted' is also called in there before the next tick.)

tests/unit/Actions.spec.ts
```diff
@@ -220,10 +220,9 @@ describe('Actions Filing Functionality', () => {
     expect(mockBuildFiling).toHaveBeenCalled()
     expect(mockBuildFiling).toHaveReturned()
 
-    // verify redirection
-    const baseUrl = 'myhost/cooperatives/NR 1234567'
-
-    expect(window.location.assign).toHaveBeenCalledWith(baseUrl)
+    // also verify event emission
+    const events = wrapper.emitted('goToDashboard')
+    expect(events.length).toBe(1)
```
severinbeauvais, line 225: "Actions" now emits this event instead of redirecting. ("App" does the actual redirection.)

tests/unit/App.spec.ts
```diff
@@ -159,17 +171,26 @@ describe('App component', () => {
...
     const router = mockRouter.mock()
     router.push({ name: 'define-company', query: { nrNumber: 'NR 1234567' } })
-    wrapper = shallowMount(App, { sync: false, localVue, store, router, vuetify, stubs: { Affix: true } })
+    wrapper = shallowMount(App, {
+      sync: false,
+      localVue,
+      store,
+      router,
+      vuetify,
+      stubs: { Affix: true, ConfirmDialog }
```
severinbeauvais, line 185: Note the "ConfirmDialog" component here. This is so that it is correctly mounted/rendered. All other components are stubbed because this is a shallow mount.

tests/unit/Certify.spec.ts
```diff
@@ -198,23 +198,36 @@ describe('Certify', () => {
...
   })
 
-  it('has update event for isCertified', () => {
+  it('has update event for isCertified = true', () => {
     const wrapper: Wrapper<Certify> = createComponent()
     const checkboxElement: Wrapper<Vue> = wrapper.find(isCertifiedSelector)
-    checkboxElement.setChecked()
+    // toggle checkbox to guarantee a change event
+    checkboxElement.setChecked(false)
+    checkboxElement.setChecked(true)
```
severinbeauvais, line 217: This is because a change event isn't emitted if the property is set to its existing value.

src/components/ReviewConfirm/Certify.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <v-card flat id="step-5-container">
-    <div class="certify-container">
+    <div class="certify-container" v-if="certifyStatementResource">
```
severinbeauvais, line 3: Without this, there was a template error if ReviewConfirm was the initial page, because this prop isn't initially defined, so you can't de-reference it on lines 24, 30, and 38 below.

src/components/common/Actions.vue
```diff
@@ -94,21 +94,11 @@ export default class Actions extends Mixins(FilingTemplateMixin, LegalApiMixin)
...
-
-  /** The URL to the Authentication site. */
-  private get authUrl (): string {
-    return sessionStorage.getItem('AUTH_URL')
-  }
-
-  /** The URL to the Business Dashboard. */
-  private get dashboardUrl (): string {
-    return sessionStorage.getItem('DASHBOARD_URL')
-  }
```
severinbeauvais, line 106: These were only used in one place so I moved them inside that method.

src/App.vue
```diff
@@ -48,16 +53,19 @@
               <aside>
                 <affix relative-element-selector=".col-lg-9" :offset="{ top: 86, bottom: 12 }">
                   <sbc-fee-summary
-                    v-bind:filingData="[...filingData]"
-                    v-bind:payURL="payApiUrl"
+                    :filingData="[...filingData]"
+                    :payURL="payApiUrl"
```
cameron-freshworks, line 57: cooool,  shorthand bindings are so 2020 😆

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -142,7 +142,7 @@ export default class PeopleAndRoles extends Mixins(EntityFilterMixin) {
   readonly Roles = Roles
   readonly IncorporatorTypes = IncorporatorTypes
 
-  mounted (): void {
+  private mounted (): void {
```
cameron-freshworks, line 145: Do we need to protect vue lifecycle hooks?

src/components/common/Actions.vue
```diff
@@ -176,32 +172,34 @@ export default class Actions extends Mixins(FilingTemplateMixin, LegalApiMixin)
...
     const paymentToken = filingComplete?.header?.paymentToken
     if (paymentToken) {
+      // redirect to pay and return to the dashboard
+      const authUrl = sessionStorage.getItem('AUTH_URL')
+      const dashboardUrl = sessionStorage.getItem('DASHBOARD_URL')
       const queryNrNumber = this.$route.query.nrNumber as string
-      const returnURL = encodeURIComponent(this.dashboardUrl + queryNrNumber)
-      const payURL = this.authUrl + 'makepayment/' + paymentToken + '/' + returnURL
+      const returnUrl = encodeURIComponent(dashboardUrl + queryNrNumber)
+      const payUrl = authUrl + 'makepayment/' + paymentToken + '/' + returnUrl
```
cameron-freshworks, line 180: Good clean up here!

src/interfaces/utils-interfaces/form-type.ts
```diff
@@ -1,6 +1,6 @@
 import Vue from 'vue'
 
 export interface FormType extends Vue {
-  reset(): void;
-  validate(): boolean;
+  reset(): void
+  validate(): boolean
```
cameron-freshworks, line 5: no action required here, but i feel we have some interfaces with semicolons and some without. Should we go without, for all?

src/store/mutations/mutations-model.ts
cameron-freshworks: Just for clarity, what was the driving force behind this change? I don't fully grasp the complexity of it, but it seems like a bit of duplication.

src/App.vue
```diff
@@ -197,13 +243,18 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
...
+
+        // Set the resources
+        this.setCertifyStatementResource(CertifyStatementResource.find(x => x.entityType === this.entityType))
       } catch (e) {
         // TODO: Catch a flag from the api, if there is an error to be handled.
       }
     }
     this.haveData = true
+    // wait for things to stabilize, then reset flag
+    Vue.nextTick(() => this.setHaveChanges(false))
```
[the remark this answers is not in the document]
severinbeauvais: Yes, some components (in Filings and sbc-common) use it to let Vue(tify) tasks finish before doing something with the new state.
I've also used the equivalent in Angular projects to wait for the system to finish its "background processing" before using the new state for something.

src/store/mutations/mutations-model.ts
severinbeauvais: The problem is that some pages "set" data when they mount, but the purpose of this feature is to track unsaved changed and that sabotages this.
I looked at changing some components to NOT set data when they mount but that got complicated and it would be fragile.
Instead, my solution is to temporarily ignore "haveChanges" while the pages load (ie, the period between created() and the next tick).
So, this line of code (and yes, it's duplicated in this file but it's centralized) is to only keep track of changes if "ignoreChanges" isn't set.
I thought this was a decent solution to update haveChanges centrally, but I'm open to better solutions :)

cameron-freshworks: Absolutely, i think it's a great idea! Totally fine with it and as with all things, we can always adjust things later if we feel the need 👍

src/App.vue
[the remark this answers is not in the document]
cameron-freshworks: The order of which these front end technologies process things is pretty interesting.  If Vue created it, why not take advantage of it right!

src/interfaces/utils-interfaces/form-type.ts
cameron-freshworks: Totally, i'm impartial. So we can agree going forward with this format!

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
cameron-freshworks: definitely agree.

PR #95, opened 2020-04-15
Share Structure UI - Step 3

lekshmimallika-aot:
*Issue #:* /bcgov/entity#3032, /bcgov/entity#3029

*Description of changes:*
Component to add/edit/remove share class/series
Component to display the share structure
Unit tests


src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -196,6 +196,11 @@ export default class ListPeopleAndRoles extends Mixins(CommonMixin, EntityFilter
   display: flex;
   background-color: $BCgovBlue5O;
   padding: 1.25rem;
+
+  .people-roles-title {
+    padding-left: .5rem;
```
severinbeauvais, line 201: Yes, this looks good, but I compared this label with the one for "Your Company" and that one has some top padding that doesn't look good. Could you please remove the top padding for class "define-company-class"?

src/components/CreateShareStructure/ListShareClass.vue
```diff
@@ -1,14 +1,368 @@
...
+      <!-- Summary Header -->
+      <div class="share-summary-header" >
+        <v-icon>mdi-file-tree</v-icon>
+        <label class="share-summary-header-title"><strong> Share Structure</strong></label>
+      </div>
+
+      <!-- Summary Warning -->
+      <div v-if="showErrorSummary" class="share-summary-invalid-message">
+        <span>
+          <v-icon color="#1976d2">mdi-information-outline</v-icon>
```
severinbeauvais, line 15: We should avoid using hex colors.
Vuetify includes color names that we can use instead. See https://vuetifyjs.com/en/styles/colors/

In this case, the name of that hex color is "blue darken-2". Please replace it here and elsewhere in the code.

src/components/CreateShareStructure/ListShareClass.vue
```diff
@@ -1,14 +1,368 @@
...
+        </span>
+      </div>
+    </div>
+
+    <v-data-table
+      :headers="headers"
+      :items="shareClasses"
+      :disable-pagination="true"
+      :disable-sort="true"
+      :hide-default-footer="true"
```
severinbeauvais, line 29: I think these props have a shortcut: you can just specify their name with no value, eg
```
<v-data-table
   disable-pagination
   disable-sort
   ...
```

src/components/CreateShareStructure/ListShareClass.vue
```diff
@@ -1,14 +1,368 @@
...
+          <td>{{row.item.parValue ? `$${row.item.parValue}` : 'No Par Value'}}</td>
+          <td>{{row.item.currency}}</td>
+          <td>{{row.item.hasRightsOrRestrictions ? 'Yes' : 'No'}}</td>
+
+          <!-- Share Class Edit Btn -->
+          <td v-if="!isSummary">
+            <div class="actions">
+              <span class="edit-action">
+                <v-btn small text color="primary"
+                 :id="'Class-' + row.index + '-change-btn'"
```
severinbeauvais, line 46: IDs should be in lower-case kebab-case, eg "abc-def".

src/mixins/currency-lookup-mixin.ts
```diff
@@ -0,0 +1,176 @@
+import { Component, Vue } from 'vue-property-decorator'
```
severinbeauvais, line 1: Did you consider using a npm package to provide these (ie, something we don't have to maintain)?
For example, https://www.npmjs.com/package/currency-codes ?

Why implement this way instead?

src/views/CreateShareStructure.vue
```diff
@@ -1,18 +1,346 @@
...
+      </p>
+      <p>
+        The sample below is for a typical share structure with just one class of shares. If there is more than one class
+        of shares, each class must be assigned an identifying name such as Class A, Class B, etc.
+      </p>
+      <div id="share-structure-sample">
+        <ListShareClass :shareClasses="sharesHelpSample"/>
+      </div>
+      <p><small>Sample Share Structure</small></p>
+      <h4>Important Information About the Company's Share Structure</h4>
```
severinbeauvais, line 32: Sorry for being picky, but this H4 follows an H2 (on line 4). This goes against accessibility guidelines (where accessibility tools try to lay out a page's structure using Header tags, but this confuses them). This should be a H3. If you need to style it smaller, so be it.

src/views/CreateShareStructure.vue
```diff
@@ -1,18 +1,346 @@
...
+      <p>
+        The staff at the Corporate Registry cannot provide advice on how to set up your company's share structure. If
+        you do not understand what an authorized share structure is or what its purpose is or believe you need a more
+        complex share structure, you should seek professional advice or purchase an incorporation guide for detailed
+        information and infrastructure on establishing an authorized share structure.
+      </p>
+      <p>Refer to <a href="">helpLinkGoesHere</a> to obtain more information on incorporating a company</p>
+      <u class="help-btn" @click="helpToggle = !helpToggle"><small>Hide Help</small></u>
```
severinbeauvais, line 40: ow are we tracking that this link needs to be updated? New ticket?

src/views/CreateShareStructure.vue
```diff
@@ -1,18 +1,346 @@
...
+        @resetEvent="resetData()"/>
+    </v-card>
+    <ListShareClass v-if="this.shareClasses.length > 0"
+    :shareClasses="shareClasses"
+    :componentDisabled="showShareStructureForm"
+    @editClass="initShareClassForEdit($event)"
+    @removeClass="removeShareClass($event)"
+    @addSeries="initNewShareSeries($event)"
+    @editSeries="editSeries"
+    @removeSeries="removeSeries"/>
```
severinbeauvais, line 80: To be consistent with existing code style, please format this as follows:
```
<ListShareClass ...
   ...
   ...
/>
```

src/views/CreateShareStructure.vue
```diff
@@ -1,18 +1,346 @@
...
+  }]
+
+  private showShareStructureForm: boolean = false
+  private addEditInProgress: boolean = false
+  private currentShareStructure: ShareClassIF | null = null
+
+  private activeIndex: number = -1
+  private parentIndex: number = -1
+  private nextId: number = -1
+  private helpToggle: boolean = false
```
severinbeauvais, line 158: Don't change this but FYI: some linters will complain about the type being redundant here because it's implied by the initial value. So I try to not include type when it's obvious: I mouse over the property and see if it's already typed, otherwise I specify it.

src/views/CreateShareStructure.vue
```diff
@@ -1,18 +1,346 @@
...
+
+  // Lifecycle methods
+  mounted (): void {
```
severinbeauvais, line 161: If you merge after https://github.com/bcgov/bcrs-business-create-ui/pull/94, please verify that you pick up the new created() method here.

src/views/CreateShareStructure.vue
```diff
@@ -1,18 +1,346 @@
...
+  private showShareStructureForm: boolean = false
+  private addEditInProgress: boolean = false
+  private currentShareStructure: ShareClassIF | null = null
+
+  private activeIndex: number = -1
+  private parentIndex: number = -1
+  private nextId: number = -1
+  private helpToggle: boolean = false
+
+  // Lifecycle methods
```
severinbeauvais, line 160: If you merge after #94, please verify that you pick up the new created() method here.

src/views/CreateShareStructure.vue
```diff
@@ -1,18 +1,346 @@
...
+    this.addEditInProgress = true
+    this.showShareStructureForm = true
+  }
+
+  private initNewShareSeries (shareClassIndex: number): void {
+    this.activeIndex = -1
+    this.parentIndex = shareClassIndex
+
+    let newList: ShareClassIF[] = [...this.shareClasses]
+    const parentShareClass = newList[shareClassIndex]
```
severinbeauvais, line 192: It would be a good safety check to verify that shareClassIndex is valid before using it to deref an array.

severinbeauvais: Ditto for index on line 180 above.

severinbeauvais: Ditto for indices in all methods that use indices.

src/views/CreateShareStructure.vue
```diff
@@ -1,18 +1,346 @@
...
+li {
+  padding-top:0.25rem
+}
+
+p{
+  padding-top: 0.5rem;
+}
+
+.btn-panel {
+  padding: 2rem 0 2rem 0;
```
severinbeauvais, line 309: FYI, a shorthand for this is `2rem 0`.
https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties

tests/unit/entityInfo.spec.ts
```diff
@@ -31,17 +31,13 @@ describe('Entity Info component', () => {
...
 
-  it('renders the Name Request header when the EntityType(BC) is present and not the initial header', () => {
+  it('renders the Name Request header when the EntityType(BC) is present and not the initial header', async () => {
     wrapper.vm.$store.state.stateModel.nameRequest.entityType = 'BC'
-    expect(wrapper.vm.$el.querySelector('#nr-header').textContent)
-      .toContain('Incorporate a BC Benefit Company')
-    expect(wrapper.vm.$el.querySelector('#no-nr-header').style.display).toBe('none')
-  })
 
-  it('renders the Name Request header when the EntityType(CP) is present and not the initial header', () => {
```
severinbeauvais, line 41: Why remove this?

tests/unit/ShareStructure.spec.ts
```diff
@@ -0,0 +1,393 @@
...
+
+/**
+ * Returns the last event for a given name, to be used for testing event propagation in response to component changes.
+ *
+ * @param wrapper the wrapper for the component that is being tested.
+ * @param name the name of the event that is to be returned.
+ *
+ * @returns the value of the last named event for the wrapper.
+ */
+function getLastEvent (wrapper: Wrapper<ShareStructure>, name: string): any {
```
severinbeauvais, line 46: Instead of this helper method, you can also do this:
`expect(wrapper.emitted('close').pop()).toEqual([false])`

tests/unit/ShareStructure.spec.ts
```diff
@@ -0,0 +1,393 @@
...
+  it('Shows error message if class name contains the word value', async () => {
+    const existingShareClass = createShareStructure(null, 1, 'Class', 'Class A', true, 100, true, 0.50, 'CAD', true)
+    const shareClass = createShareStructure(null, 1, 'Class', 'Class B', true, 100, true, 0.50, 'CAD', true)
+    const wrapper: Wrapper<ShareStructure> = createComponent(shareClass, -1, 1, null, [existingShareClass])
+    const inputElement: Wrapper<Vue> = wrapper.find(nameSelector)
+    inputElement.setValue('Class A value')
+    inputElement.trigger('change')
+    await wrapper.vm.$nextTick()
+    await flushPromises()
+    await wrapper.vm.$nextTick()
```
severinbeauvais, line 248: I (we) will have to remember that sometimes all these must be awaited for a change event to complete. I know that in other places I (we) just gave up and called the handler method directly, but if the above works then we can go back and fix those other tests.
Thanks for persevering and finding a solution here! 👍 👍

severinbeauvais: I see that this set of 3 awaits is used in a few places...
How about creating a helper method to replace them?

tests/unit/ShareStructure.spec.ts
```diff
@@ -0,0 +1,393 @@
...
+
+  it('Do not display series no max share option if class has max share', async () => {
+    const shareClass = createShareStructure(null, 1, 'Class', 'Class A', true, 100, true, 0.50, 'CAD', true)
+    const shareSeries = createShareStructure(1, 1, 'Series', 'Series A', true, 100, true, 0.50, 'CAD', true)
+    shareClass.series.push(shareSeries)
+    const wrapper: Wrapper<ShareStructure> = createComponent(shareSeries, 0, -1, 0, [shareClass])
+    expect(wrapper.find('#lbl-no-maximum').exists()).toBe(false)
+    wrapper.destroy()
+  })
+})
```
severinbeauvais, line 393: Need a test that changing currency (eg, to USD) works (and keeps working).

tests/unit/ListShareClass.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+  let wrapper
+
+  const shareClasses = [
+    {
+      id: 1,
+      name: 'Common Shares',
+      priority: 0,
+      maxNumberOfShares: 10000,
+      parValue: 1.00,
+      currency: 'Canadian Dollar ($CDN)',
```
severinbeauvais, line 27: Should this match what the currency list (mixin) provides?

tests/unit/ListShareClass.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+    expect(classListItem1.querySelector('.list-item__title').textContent).toContain('Common Shares')
+    expect(seriesListItem1.querySelector('.series-name').textContent).toContain('Share Series 1')
+  })
+
+  it('displays the correct data in the selected table rows for shareClass', () => {
+    const classListItem1 = wrapper.vm.$el.querySelectorAll('.v-data-table .class-row')[2]
+
+    expect(classListItem1.querySelectorAll('td')[0].textContent).toContain('Common Shares 2')
+    expect(classListItem1.querySelectorAll('td')[1].textContent).toContain(10000)
+    expect(classListItem1.querySelectorAll('td')[2].textContent).toContain('$1')
```
severinbeauvais, line 126: Might as well check for longest match here, eg, "$1.00".

package.json
```diff
@@ -26,6 +26,7 @@
     "vuex": "^3.1.3"
   },
   "devDependencies": {
+    "array.prototype.move": "0.0.4",
```
severinbeauvais, line 29: This is probably a prod dependency instead of a dev dependency!

src/components/CreateShareStructure/ListShareClass.vue
```diff
@@ -1,14 +1,368 @@
...
+      :disable-sort="true"
+      :hide-default-footer="true"
+    >
+      <template v-slot:item="row" class="share-data-table">
+
+        <!-- Share Class Rows-->
+        <tr :key="row.item.id" class="class-row" :class="{ 'class-row-has-series': row.item.series.length}">
+          <td class="list-item__title">{{row.item.name}}</td>
+          <td>{{row.item.maxNumberOfShares || 'No Maximum'}}</td>
+          <td>{{row.item.parValue ? `$${row.item.parValue}` : 'No Par Value'}}</td>
```
severinbeauvais, line 37: This should use localized currency formatting... What if the selected currency isn't dollars?
Also on line 106.
Or, we could limit currencies to CAD and USD only... but need to check with Scott/BAs first.

src/components/CreateShareStructure/ListShareClass.vue
```diff
@@ -1,14 +1,368 @@
...
+                    </v-list-item>
+                  </v-list>
+                </v-menu>
+              </span>
+            </div>
+          </td>
+        </tr>
+
+        <!-- Share Series rows -->
+        <tr v-for="(seriesItem, index) in row.item.series" :key="`Class:${row.index}-Series:${index}`"
```
severinbeauvais, line 100: Should keys follow lower-case kebab-case also?

src/components/CreateShareStructure/ListShareClass.vue
lekshmimallika-aot: I already talked to BAs and Scott about this. Currency symbol wont be displayed in the table. The symbol has to be taken off

src/components/CreateShareStructure/ListShareClass.vue
```diff
@@ -1,14 +1,368 @@
...
+import { Component, Prop, Vue, Emit } from 'vue-property-decorator'
+import 'array.prototype.move'
+
+// Components
+
+// Dialogs
+
+// Mixins
+
+// Interfaces
```
severinbeauvais, line 175: Remove the ones you're not using (line 169-175).

src/components/CreateShareStructure/ListShareClass.vue
lekshmimallika-aot: They will be updating the docs and wireframes

src/components/CreateShareStructure/ListShareClass.vue
```diff
@@ -1,14 +1,368 @@
...
+    { text: '', value: 'actions' }
+  ]
+
+  /**
+   * Adjust the priority of the list share class
+   * @param indexFrom The index of the class
+   * @param direction The direction of the move
+   * @param seriesIndex The index of the series
+   */
+  private moveIndex (indexFrom: number, direction: string, seriesIndex: number = -1): void {
```
severinbeauvais, line 211: This is super-cool. Kudos for a nice, clean implementation!!!

severinbeauvais: I didn't see a test for moving classes/series up or down. If there isn't one could you please add one? (Ideally the test would click the buttons instead of calling this method.)

src/mixins/currency-lookup-mixin.ts
lekshmimallika-aot: Yes. I also considered couple of other options. The order in which the currency needs to be displayed , the item text will still need a lot of data reordering and stuff. Also whether all  options to be shown in the dropdown is another open question. This will be looked into by business once this is deployed to dev. This question was also asked in the demo.

severinbeauvais: OK, perfect, this conversation is resolved!

src/views/CreateShareStructure.vue
severinbeauvais: Since I'm familiar with the recent changes, I fixed the merge conflicts.

src/views/CreateShareStructure.vue
```diff
@@ -158,11 +159,18 @@ export default class CreateShareStructure extends Vue {
...
+    Vue.nextTick(() => {
+      this.setIgnoreChanges(false)
+    })
+  }
+
   mounted (): void {
     this.setCreateShareStructureStepValidity(this.shareClasses.length > 0)
   }
 
-  // Methods
```
severinbeauvais: Oops I didn't mean to take this out.

src/views/CreateShareStructure.vue
lekshmimallika-aot: BA team is yet to provide the link. But there is no ticket to track it

src/components/CreateShareStructure/ListShareClass.vue
lekshmimallika-aot: done

src/components/CreateShareStructure/ListShareClass.vue
lekshmimallika-aot: Fixed

src/components/CreateShareStructure/ListShareClass.vue
lekshmimallika-aot: Fixed

src/views/CreateShareStructure.vue
lekshmimallika-aot: I tried h3. It looks very big and odd. h4 looks better here

src/views/CreateShareStructure.vue
lekshmimallika-aot: Fixed

src/views/CreateShareStructure.vue
lekshmimallika-aot: Fixed

tests/unit/ListShareClass.spec.ts
lekshmimallika-aot: Yes. it must be the code. Fixed

src/components/CreateShareStructure/ListShareClass.vue
lekshmimallika-aot: Done

src/views/CreateShareStructure.vue
severinbeauvais: Can you change this to a H3 and style it as an H4? (There may be a class name "h4".)

tests/unit/ListShareClass.spec.ts
lekshmimallika-aot: I changed data set to give a better representation of possible values

tests/unit/ShareStructure.spec.ts
lekshmimallika-aot: This method is not used actually. Removed it.

tests/unit/ShareStructure.spec.ts
lekshmimallika-aot: Moved it into separate method

src/components/CreateShareStructure/ListShareClass.vue
lekshmimallika-aot: @cameron-freshworks  added a test for this

src/views/CreateShareStructure.vue
```diff
@@ -1,18 +1,346 @@
...
+  private currentShareStructure: ShareClassIF | null = null
+
+  private activeIndex: number = -1
+  private parentIndex: number = -1
+  private nextId: number = -1
+  private helpToggle: boolean = false
+
+  // Lifecycle methods
+  mounted (): void {
+    this.setCreateShareStructureStepValidity(this.shareClasses.length > 0)
```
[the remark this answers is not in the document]
lekshmimallika-aot: @cameron-freshworks had mentioned this before. There is another PR coming up for packaging + resuming. Can we have this as part of that PR

tests/unit/entityInfo.spec.ts
lekshmimallika-aot: Not sure. was this part of the unit test fix @ochiu did this morning?

src/views/CreateShareStructure.vue
lekshmimallika-aot: Updated the link href with the value Sienna provided

package.json
lekshmimallika-aot: Looks like that to me. Updated

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
lekshmimallika-aot: Fixed

tests/unit/entityInfo.spec.ts
ochiu: @lekshmimallika-aot 
I updated the nameRequestInfo.spec.ts and the App.spec.ts as they were causing the checks to fail. Entity Info tests were running fine so I didn't touch them.

src/views/CreateShareStructure.vue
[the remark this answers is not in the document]
severinbeauvais: Yes, let's leave that for later. (I have an idea how we could do it -- but out of scope here.)

tests/unit/entityInfo.spec.ts
severinbeauvais: Then please restore that unit test. Thanks!

src/components/CreateShareStructure/ListShareClass.vue
severinbeauvais: Thanks @cameron-freshworks ! Looks good.

src/views/CreateShareStructure.vue
lekshmimallika-aot: Will leave it as such for now

tests/unit/entityInfo.spec.ts
lekshmimallika-aot: Restored

tests/unit/ShareStructure.spec.ts
lekshmimallika-aot: Added a test for this

src/views/CreateShareStructure.vue
lekshmimallika-aot: I was looking at the flow of events and data and I dont think an invalid index will be passed back as the parent index / active index unless something goes wrong in reordering. In that case the fix would be on that side. I think we can leave it like this.

PR #102, opened 2020-04-24
3342

ochiu:
*Issue #:* /bcgov/entity#3342

*Description of changes:*
- Removed router-view and added v-if and v-show for view components so they can be pre-mounted so validation for each step can be evaluated at start up
- Added enum for route names


src/App.vue
```diff
@@ -45,8 +45,18 @@
...
+              <sign-out v-if="isRouteName(RouteNames.SIGN_OUT)" />
+
+              <!-- Only render when data is ready, or validation can't be properly evaluated. -->
+              <template v-if="haveData">
+                <!-- Using v-show to pre-create/mount components so validation on stepper is shown -->
+                <define-company v-show="isRouteName(RouteNames.DEFINE_COMPANY)" />
+                <add-people-and-roles v-show="isRouteName(RouteNames.ADD_PEOPLE_AND_ROLES)" />
+                <create-share-structure v-show="isRouteName(RouteNames.CREATE_SHARE_STRUCTURE)" />
+                <review-confirm v-show="isRouteName(RouteNames.REVIEW_CONFIRM)" />
+              </template>
```
ochiu, line 59: - Sign in and sign out components are using v-if as they are not needed all the time. Using v-show instead cause state issues and occasionally a redirect loop when refreshing the screen
- a template is used to wrap the view components to eliminate any timing issues, if mounting occurs too soon and the data is not ready, the validation is not properly evaluated and properly reflected in the stepper

severinbeauvais: We will have to see if there's a way to specify the steps components programmatically so that we can just loop over them to support other corp types in the future. (I know this is possible in Angular.)

Pseudo-code:
```
for (step in steps)
{
  <programmatic-component v-show="is current component" ... />
}
```

Maybe this? https://medium.com/js-dojo/simply-switching-in-between-the-components-dynamically-vue-js-a436fe10008a

src/App.vue
```diff
@@ -85,13 +95,17 @@ import SbcHeader from 'sbc-common-components/src/components/SbcHeader.vue'
 import SbcFooter from 'sbc-common-components/src/components/SbcFooter.vue'
 import SbcFeeSummary from 'sbc-common-components/src/components/SbcFeeSummary.vue'
 import { EntityInfo, Stepper, Actions } from '@/components/common'
+import {
+  AddPeopleAndRoles, CreateShareStructure, DefineCompany, ReviewConfirm,
+  SignIn, SignOut
+} from '@/views'
```
severinbeauvais, line 101: Is it possible to `import * from ...` for future support?

src/App.vue
```diff
@@ -103,7 +117,13 @@ import { EntityTypes, FilingCodes, NameRequestStates } from '@/enums'
...
     NameRequestInvalidErrorDialog,
     AccountAuthorizationDialog,
-    ConfirmDialog
+    ConfirmDialog,
+    AddPeopleAndRoles,
+    CreateShareStructure,
+    DefineCompany,
+    ReviewConfirm,
+    SignIn,
+    SignOut
```
severinbeauvais, line 126: I have no idea how to specify "all components in views" here :/

If you have time left in this ticket (and/or the will to look it up), research this now, otherwise we'll cross this bridge later.

src/App.vue
lekshmimallika-aot: Some routes are applicable for some entity types only. Example Share structure is only for BCOMP and Rules is only for CP (Yet to be implemented).

severinbeauvais: This would be nice too (idea from my Angular days):
https://github.com/vuejs/vue/issues/8097

(Not mandatory at this time, and might not be usable with "for" loop.)

ochiu: I will take a look! Thanks

src/App.vue
ochiu: I will do a little research on this as I am curious if it can be done as well.

src/App.vue
ochiu: Updated.

src/App.vue
ochiu: Updated.

severinbeauvais: Of course! Brilliant!

src/App.vue
ochiu: Updated the components to be rendered based off of the step config.

My thoughts were to update `getSteps()` in the future to return the relevant steps config depending on entity type when it is needed and it will just automatically show what is relevant.

severinbeauvais: Woohoo! Looks great... love it!

PR #103, opened 2020-04-27
UI - Future Effective Selector 

cameron-freshworks:
*Issue #:* /bcgov/entity#3337

*Description of changes:*
* Implements a UI component to select a Filing Type, Date and Time.
* Includes Validations on the routes/steppers/file and pay.
* Updates the Fee Summary initializer to be reactive to the change of filing Date/Time.
* Updated Unit testing and added in IncorporationDateTime unit testing.

** Issues:
* Currently the backend is validating properties when Drafts. So once an effectiveDate is selected and saved into a draft, we are unable to change it back to an immediate filing.
* The presentation of the validations are currently pending design, tho the logic has been implemented.

* Can also update the unit testing for validations once that piece is applied ( designs pending )


src/store/getters/state-getters.ts
```diff
@@ -138,9 +138,8 @@ export const isEnableFilePayBtn = (state: any): boolean => {
   const step2Valid = state.stateModel.addPeopleAndRoleStep.valid
   const step3Valid = true // FOR TESTING ONLY
   const step4Valid = true // FOR TESTING ONLY
-  const step5Valid = true // FOR TESTING ONLY
-  return (step1Valid && step2Valid && step3Valid && step4Valid && step5Valid &&
-    state.stateModel.certifyState.valid)
+  const step5Valid = state.stateModel.certifyState.valid && state.stateModel.incorporationDateTime.valid
```
severinbeauvais, line 141: Yay, finally! :)

src/store/getters/state-getters.ts
```diff
@@ -154,7 +153,8 @@ export const isBusySaving = (state: any): boolean => {
  * Whether all the incorporation steps are valid.
  */
 export const isApplicationValid = (state: any): boolean => {
-  return (state.stateModel.defineCompanyStep.valid && state.stateModel.addPeopleAndRoleStep.valid)
+  return (state.stateModel.defineCompanyStep.valid && state.stateModel.addPeopleAndRoleStep.valid &&
+    state.stateModel.createShareStructureStep.valid && state.stateModel.incorporationDateTime.valid)
```
severinbeauvais, line 157: Should this also look at certifyState?

src/store/mutations/mutations-model.ts
```diff
@@ -1,5 +1,7 @@
-import { CertifyStatementIF, CertifyIF, IncorporationAddressIf, NameRequestIF,
-  BusinessContactIF, OrgPersonIF, ShareClassIF } from '@/interfaces'
+import {
+  CertifyStatementIF, CertifyIF, IncorporationAddressIf, NameRequestIF,
+  BusinessContactIF, OrgPersonIF, ShareClassIF, DateTimeIF
+} from '@/interfaces'
```
severinbeauvais, line 4: If we are, in fact, importing all the IFs, can we use the syntax to import all?

src/interfaces/filing-interfaces/filingData-interface.ts
```diff
@@ -3,4 +3,5 @@ export interface FilingDataIF {
   filingTypeCode: string;
   entityType: string;
   waiveFees?: boolean;
+  futureEffective?: boolean
```
severinbeauvais, line 6: I like the code style where we name booleans "isSomething". Thoughts?

src/store/mutations/mutations-model.ts
```diff
@@ -34,6 +36,26 @@ export const mutateCurrentDate = (state: any, currentDate: string) => {
...
+  state.stateModel.incorporationDateTime.isFutureEffective = isFutureEffective
+  if (!state.stateModel.ignoreChanges) mutateHaveChanges(state, true)
+}
+
+export const mutateFutureEffectiveDate = (state: any, futureEffectiveDate: string) => {
+  state.stateModel.incorporationDateTime.futureEffectiveDate = futureEffectiveDate
+  if (!state.stateModel.ignoreChanges) mutateHaveChanges(state, true)
+}
+
+export const mutateIncorporationDateTimeValid = (state: any, incorporationDateTimeValid: boolean) => {
```
severinbeauvais, line 54: mutate**Is**IncorporationDateTimeValid  ?

src/store/actions/actions-model.ts
```diff
@@ -28,6 +28,22 @@ export const setCurrentDate: ActionIF = ({ commit }, currentDate): void => {
...
+
+export const setIsFutureEffective: ActionIF = ({ commit }, isFutureEffective): void => {
+  commit('mutateIsFutureEffective', isFutureEffective)
+}
+
+export const setFutureEffectiveDate: ActionIF = ({ commit }, futureEffectiveDate): void => {
+  commit('mutateFutureEffectiveDate', futureEffectiveDate)
+}
+
+export const setIncorporationDateTimeValid: ActionIF = ({ commit }, incorporationDateTimeValid): void => {
```
severinbeauvais, line 43: set**Is**IncorporationDateTimeValid ?

src/interfaces/filing-interfaces/filing-interface.ts
```diff
@@ -8,6 +8,7 @@ export interface IncorporationFilingIF {
       certifiedBy: string
       email: string
       date: string
+      effectiveDate: string
```
severinbeauvais, line 11: Would it make sense to name this consistent with the same property in incorporationDateTime?

src/mixins/filing-template-mixin.ts
```diff
@@ -87,6 +97,11 @@ export default class FilingTemplateMixin extends Vue {
         valid: false,
         certifiedBy: draftFiling.header.certifiedBy
       })
+
+      // Set Future Effective Time
+      this.setFutureEffectiveDate(draftFiling.header.effectiveDate)
+      this.setIsImmediate(!draftFiling.header.effectiveDate)
+      this.setIsFutureEffective(!!draftFiling.header.effectiveDate)
```
severinbeauvais, line 104: Why do we need 2 properties, since they are (always?) opposite of each other?

tests/unit/IncorporationDateTime.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+Vue.use(Vuetify)
+Vue.use(Vuelidate)
+const vuetify = new Vuetify({})
+
+// Store
+const store = getVuexStore()
+document.body.setAttribute('data-app', 'true')
+
+describe('List Shares and Series', () => {
```
severinbeauvais, line 21: Should describe THIS component :)

src/mixins/filing-template-mixin.ts
severinbeauvais: I see you use this for the radio buttons, but the fact that both radio buttons are initialized as "not set" should be abstracted in the component. At the app level, we only need to know if immediate or future, right?

src/store/mutations/mutations-model.ts
cameron-freshworks: Yea, good idea!

src/store/getters/state-getters.ts
cameron-freshworks: I'll look into this!

src/interfaces/filing-interfaces/filingData-interface.ts
cameron-freshworks: totally, gap on my part! Will update

src/store/mutations/mutations-model.ts
cameron-freshworks: Agree!

src/store/actions/actions-model.ts
cameron-freshworks: Agree!

src/components/ReviewConfirm/IncorporationDateTime.vue
```diff
@@ -0,0 +1,368 @@
...
+    if (this.incorporationDateTime.futureEffectiveDate) {
+      const effectiveDate = this.incorporationDateTime.futureEffectiveDate
+      const dateToParse = new Date(effectiveDate)
+      const startDate = new Date()
+
+      const timeDiff = dateToParse.getTime() - startDate.getTime()
+      const timeDiffInMinutes = Math.floor(timeDiff / 1000 / 60)
+
+      // Time set must be more than 2 minutes and less than 10 days
+      return timeDiffInMinutes > 2 && timeDiffInMinutes < 14400
```
severinbeauvais, line 229: Spec says "at least 2 min" and "at most 10 days", so these should be >= and <=.

https://preview.uxpin.com/6bd383a032304acc6ef186307cf186993fb02a39#/pages/127750603/simulate/no-panels

src/interfaces/filing-interfaces/filing-interface.ts
cameron-freshworks: Hmm, yes could do. Tho this is to make it consistent with the property the backend is expecting. I'll update the incorporationDateTime prop to be in line with this i think.

src/mixins/filing-template-mixin.ts
cameron-freshworks: Yes for sure, let me look into this a bit deeper. If we can eliminate the isImmediate prop, will do so! Good idea, will get back to this asap.

tests/unit/IncorporationDateTime.spec.ts
cameron-freshworks: Oops! 😆

src/components/ReviewConfirm/IncorporationDateTime.vue
cameron-freshworks: ahh, good catch. Will update, thanks!

tests/unit/IncorporationDateTime.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    const radioInput = wrapper.findAll('input[type="radio"]')
+    const radioIsImmediate = radioInput.at(0)
+    const radioIsFutureEffective = radioInput.at(1)
+
+    // Verify radios are false
+    expect(radioIsImmediate.attributes('aria-checked')).toBe('false')
+    expect(radioIsFutureEffective.attributes('aria-checked')).toBe('false')
+  })
+
+  it('confirms the selector fields are disabled if future effective is NOT selected', () => {
```
severinbeauvais, line 62: I'd like to see the Immediate radio button clicked in this test (opposite of test below).

cameron-freshworks: Good call, can do. Will update this

tests/unit/IncorporationDateTime.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    const radioInput = wrapper.findAll('input[type="radio"]')
+    const radioIsImmediate = radioInput.at(0)
+    await radioIsImmediate.trigger('click')
+
+    // Verify the Valid emit event is true
+    expect(wrapper.emitted().valid).toEqual([[true]])
+  })
+
+  it('emits an invalid state when the Future Effective is selected and no date is selected', async () => {
```
severinbeauvais, line 118: Add test for checking valid Future Effective state.

tests/unit/Actions.spec.ts
```diff
@@ -241,6 +245,7 @@ describe('Actions Filing Functionality', () => {
     store.state.stateModel.certifyState.certifiedBy = 'somePerson'
     store.state.stateModel.defineCompanyStep.businessContact = { email: 'someEmail', phone: '123-456-7890' }
     store.state.stateModel.currentDate = '2020/01/29'
+    store.state.stateModel.incorporationDateTime.futureEffectiveDate = new Date('2020-05-05T00:30:00.000Z')
```
severinbeauvais, line 248: To keep the tests running for a long time, use a date far in the future, say 2099 :)

src/mixins/filing-template-mixin.ts
```diff
@@ -29,18 +29,28 @@ export default class FilingTemplateMixin extends Vue {
...
+  @Action setIsFutureEffective!: ActionBindingIF
 
   /**
    * Method to construct a filing body when making an api request
    */
   buildFiling (): IncorporationFilingIF {
+    // Format DateTime for Filing
+    const effectiveDate = this.stateModel.incorporationDateTime.futureEffectiveDate
+    const formattedDateTime = effectiveDate &&
+      (effectiveDate.toISOString()).replace('Z', '+00:00')
```
severinbeauvais, line 43: Need to verify that saved and restored date/time is OK with respect to local time. For example, if I specify "effective at 12:00PM (noon) today", does it get restored the same way? And is it correct in the Lear db? I ask this because I don't see conversions to/from localtime.

cameron-freshworks: The date is restored in the `deconstructDateTime` method where it converts it to local time. It is always saved into store and filed as UTC. This does line up with the backend ( afaik ) and the display of the time after being parsed from a draft doesn't change its actual value in store. But this is definitely worth more research, i could certainly be missing something.

tests/unit/Actions.spec.ts
cameron-freshworks: This is a great idea. I didn't even consider this thanks :) 

So your telling me tests with an expiry date is not a good practice 😆

tests/unit/IncorporationDateTime.spec.ts
cameron-freshworks: for sure!

tests/unit/Actions.spec.ts
ochiu: 👍  for today+5!

src/components/ReviewConfirm/IncorporationDateTime.vue
```diff
@@ -0,0 +1,368 @@
...
+      // Format the selected date string and create Date
+      const dateString = this.dateText.replace(/-/g, ', ')
+      const dateToValidate = new Date(dateString)
+
+      // Create references & Apply time period
+      let hours = +this.selectHour
+      const minutes = +this.selectMinute
+
+      if (this.selectPeriod === 'AM' && +this.selectHour === 12) {
+        dateToValidate.setDate(dateToValidate.getDate() - 1)
```
lekshmimallika-aot, line 165: What is the purpose of this?

cameron-freshworks: Since we set each Date, Hours, minutes and AmPm separately to construct a DateTime, it was tricky setting to 12:00AM on the selected Date. What it was doing when i chose to set the Hours & Minutes of a selected day wasn't taking into consideration the AM/PM. So when i user selected 12 hours AND AM... that essentially equates to 00:00 time. 
So to get that time on the selected day, i had to set it to the day prior and Add the 12hrs to reset the clock essentially.  This is mainly a result of the hour selector only being 12hr selector opposed to a 24hr and using the 'AmPm' field.
It was tricky having them all in isolation but working together :P. 
Does this make sense? Could this be done better?

lekshmimallika-aot: I think I  got it. Is it converting to previous day 24:00? Is there any reason for not converting to current day 00:00? Just curious

src/store/getters/state-getters.ts
cameron-freshworks: Good question, this directly correlates to the stepper appearance of the green Checkmark. Do we want that checkmark to appear when all previous steps are valid or only once all of Step 5 is complete as well and the file and pay is ready? 

This means the user will need enter their name and give a certify Check, which will then result in the stepper appearance of the green Checkmark. 

Can totally update this tho, thoughts on this?

src/store/mutations/mutations-model.ts
cameron-freshworks: We could do this `import * as Interfaces from '@/interfaces' but would then have to prefix every interface we use as `Interfaces.SomeInterface`. 

Prefer longer import or a little more code?

src/interfaces/filing-interfaces/filingData-interface.ts
cameron-freshworks: shoot, sorry. This property is to tie into the Fee Summary `futureEffective` option property. I knew it had to be named this for a reason :P 
So the update would need to be on that common component for us to change this, i think.

src/store/getters/state-getters.ts
severinbeauvais: I think the step 5 page forms fields should affect the step 5 green checkmark, whereas the overall application state should affect the File + Pay button.
However, Scott may have a different experience in mind... ask/confirm with him.

PS - Before Ody's validation change, step 5 page would get recreated each time you landed on it, and the Certify would get cleared -- which I didn't like (IMO). I assume that now, once I have checked Certify, it stays checked during this session (and only gets cleared when I resume)?

src/interfaces/filing-interfaces/filing-interface.ts
cameron-freshworks: K, updated. All instances of `futureEffectiveDate` are now `effectiveDate` to be congruent with the backend properties 👍

src/store/mutations/mutations-model.ts
severinbeauvais: I prefer the longer import.

src/interfaces/filing-interfaces/filingData-interface.ts
severinbeauvais: Oh right. Never mind then.

src/store/getters/state-getters.ts
cameron-freshworks: K, will follow up with Scott on this interaction!

I'm not 100% sure. I will rebase in Ody's work and verify this 👍

src/store/mutations/mutations-model.ts
cameron-freshworks: cool, me as well :)

src/components/ReviewConfirm/IncorporationDateTime.vue
```diff
@@ -0,0 +1,368 @@
...
+<script lang="ts">
+// Libraries
+import { Component, Mixins, Watch, Emit, Prop } from 'vue-property-decorator'
+import { Action } from 'vuex-class'
+
+// Mixins
+import { DateMixin } from '@/mixins'
+
+// Constants
+import { ISIMMEDIATE, ISFUTUREEFFECTIVE } from '@/constants'
```
severinbeauvais, line 97: If these aren't used anywhere else then you could simply declare them as `readonly` properties in this module.

src/views/ReviewConfirm.vue
```diff
@@ -6,6 +6,16 @@
       </header>
       <Summary />
     </section>
+    <section class="mt-10">
+      <header>
+        <h2>Incorporation Date Time</h2>
```
severinbeauvais, line 11: Header is incorrect and you're missing the instructions, per https://preview.uxpin.com/6bd383a032304acc6ef186307cf186993fb02a39#/pages/127750603/simulate/no-panels .

severinbeauvais: Or https://preview.uxpin.com/d7266b8b6fb8bd80acc0ff38ff5c946434492f5e#/pages/122406228/simulate/no-panels .

cameron-freshworks:
> > Setting the effective date back to immediate does not clear the effective date field in db. This can be a problem because in this case the filing will work as a future effective filing and wont be applied immediately. Not sure whether there is a ticket for this already. If not please create one
> 
> If this is a FE fix then it can be done in this ticket. If it's a BE fix then yes, create a new ticket.

Sorry, i did specify this in the PR. The back end is currently running validations on the drafts so it is unable to expunge the effectiveDate in the backend due to those validators blocking it. 

Thor did mention this was on his radar and currently in progress ( afaik ). Not sure the exact ticket # however.

src/views/ReviewConfirm.vue
cameron-freshworks: Definitely, thanks for the reminder. Scott mentioned to add this in!

src/components/ReviewConfirm/IncorporationDateTime.vue
cameron-freshworks: For sure! Not used anywhere else at his time

lekshmimallika-aot:
> > > Setting the effective date back to immediate does not clear the effective date field in db. This can be a problem because in this case the filing will work as a future effective filing and wont be applied immediately. Not sure whether there is a ticket for this already. If not please create one
> > 
> > 
> > If this is a FE fix then it can be done in this ticket. If it's a BE fix then yes, create a new ticket.
> 
> Sorry, i did specify this in the PR. The back end is currently running validations on the drafts so it is unable to expunge the effectiveDate in the backend due to those validators blocking it.
> 
> Thor did mention this was on his radar and currently in progress ( afaik ). Not sure the exact ticket # however.

Yes I saw that comment. But looking at the BE code I think it will apply effective date if a valid value is present in json. The effective date is passed as an empty string now while resetting to immediate. Can we try by passing null?

src/mixins/filing-template-mixin.ts
cameron-freshworks: Removed the `isImmediate` property outside its local use to the component 👍

cameron-freshworks:
> > > > Setting the effective date back to immediate does not clear the effective date field in db. This can be a problem because in this case the filing will work as a future effective filing and wont be applied immediately. Not sure whether there is a ticket for this already. If not please create one
> > > 
> > > 
> > > If this is a FE fix then it can be done in this ticket. If it's a BE fix then yes, create a new ticket.
> > 
> > 
> > Sorry, i did specify this in the PR. The back end is currently running validations on the drafts so it is unable to expunge the effectiveDate in the backend due to those validators blocking it.
> > Thor did mention this was on his radar and currently in progress ( afaik ). Not sure the exact ticket # however.
> 
> Yes I saw that comment. But looking at the BE code I think it will apply effective date if a valid value is present in json. The effective date is passed as an empty string now while resetting to immediate. Can we try by passing null?

Definitely can, though the validators will say `NONE` type is not allowed in the same manner that `''` is not a valid DateTime.

src/components/ReviewConfirm/IncorporationDateTime.vue
cameron-freshworks: Are you ok if i leave them as constants? I'd prefer to have less properties on the component if i could. Referencing them also means more code as `this.property` opposed to `constant`. :)

severinbeauvais: It's 6 of one, 1/2 dozen of the other...
As constants, you're polluting the app namespace instead of just your component, but yes, you can just use `constant`.
As properties, you have to use `this.constant` (which I don't mind personally), but then they're localized to just your property.
Up to you (author's prerogative)!

tests/unit/IncorporationDateTime.spec.ts
cameron-freshworks: added!

tests/unit/Actions.spec.ts
cameron-freshworks: done!

src/components/ReviewConfirm/IncorporationDateTime.vue
cameron-freshworks: Great question, it IS converting to prior day at 24:00. You're right for sure, i could have set it explicitly, but i think i just structured like this so i only have to `Date.setHours()` once inside the function. 
i was using the `hours` selected to set the hour and there is no `0` in the selectors, so essentially selecting `12`th hour  in conjunction with `am` equates to 00 on the same day selected ( or in this case the code is 24:00 the previous day ). 

phew.. long winded, sorry 😆

src/store/getters/state-getters.ts
cameron-freshworks: rebased and ody's work is excellent, really helps! Things are looking good. 
The certify is cleared once resuming, correct 👍 
Also, enquired with Scott about when we want to present the validity on the stepper, will update when i find out!

src/mixins/filing-template-mixin.ts
cameron-freshworks: Can confirm it is restored in the same manner. Essentially local time is just used for the presentation, whereas everything being stored is UTC.

src/components/ReviewConfirm/IncorporationDateTime.vue
cameron-freshworks: Fair enough, thanks for the help on this. Food for thought!

src/views/ReviewConfirm.vue
cameron-freshworks: Added!

severinbeauvais: H2 text is still incorrect 😬

tests/unit/IncorporationDateTime.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+    const radioIsFutureEffective = radioInput.at(1)
+    await radioIsFutureEffective.trigger('click')
+
+    const validEvent = getLastEvent(wrapper, 'valid')
+
+    // Verify the Valid emit event is false at this point
+    expect(validEvent).toEqual(true)
+  })
+
+  it('emits a invalid state when the Future Effecting is selected and DateTime is invalid', async () => {
```
severinbeauvais, line 174: "Future Effective"

cameron-freshworks: 🤦

severinbeauvais:
If you file at 5pm (or, say, right now), what's the day number... Today or tomorrow?
________________________________

cameron-freshworks:
> If you file at 5pm (or, say, right now), what's the day number... Today or tomorrow?

Well, if you choose immediately it would be Null. 

If you choose future effective and chose the minimum 2 minutes in the future ( edge case, pretty expensive 2 minutes lol ) and file it, the time saved to the back end is the same at GMT time ( +7 hrs right now due to daylight savings ), this is just how the back end handles date times from what i know. The day in the effectiveDate would be 7 hrs from now, 2020/04/28:03:23:00+00:00 ... something along those lines. Since the job is running on that time format, it should still file in 2 mins from now because ( once again, afaik ) the job and or back end is running on GMT.

src/components/ReviewConfirm/IncorporationDateTime.vue
```diff
@@ -20,7 +20,7 @@
         </v-radio-group>
         <div class="date-time-selectors">
           <v-menu
-            :close-on-content-click="false"
+            :close-on-content-click="true"
```
severinbeauvais, line 23: FYI, shortcut, you can just use:
```
close-on-content-click
...
```

cameron-freshworks: I like it, will update!

src/components/ReviewConfirm/IncorporationDateTime.vue
```diff
@@ -67,7 +66,6 @@
                 :disabled="!isFutureEffective"
                 :rules="minuteRules"
                 filled
-                dense
```
ochiu: Thanks for updating, looks good!

ochiu:
Just curious about the presentation of the validation error messages. For example, if you remove an input on minutes you get `A second between 1 and 59 is required.`, but it is very squished. Is this okay?

cameron-freshworks:
> Just curious about the presentation of the validation error messages. For example, if you remove an input on minutes you get `A second between 1 and 59 is required.`, but it is very squished. Is this okay?

Yea, i've noticed that was well! Designs for the presentations of validations are currently pending, as well as the verbiage. This was just a first stab it, most likely to change in the near future.

lekshmimallika-aot:
> > > > > Setting the effective date back to immediate does not clear the effective date field in db. This can be a problem because in this case the filing will work as a future effective filing and wont be applied immediately. Not sure whether there is a ticket for this already. If not please create one
> > > > 
> > > > 
> > > > If this is a FE fix then it can be done in this ticket. If it's a BE fix then yes, create a new ticket.
> > > 
> > > 
> > > Sorry, i did specify this in the PR. The back end is currently running validations on the drafts so it is unable to expunge the effectiveDate in the backend due to those validators blocking it.
> > > Thor did mention this was on his radar and currently in progress ( afaik ). Not sure the exact ticket # however.
> > 
> > 
> > Yes I saw that comment. But looking at the BE code I think it will apply effective date if a valid value is present in json. The effective date is passed as an empty string now while resetting to immediate. Can we try by passing null?
> 
> Definitely can, though the validators will say `NONE` type is not allowed in the same manner that `''` is not a valid DateTime.

The backend had to change slightly to support this. Validators was not the issue. This is already in PR. FE does not need any change as long as it passes null value for immediate effective filing

PR #104, opened 2020-04-28
2556 - update email validation

ochiu:
*Issue #:* /bcgov/entity#2556

*Description of changes:*
- update email validation to match auth profile page

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,7 +11,8 @@ module.exports = {
   rules: {
     'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
     'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
-    'max-len': ['warn', { code: 120 }]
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -11,7 +11,8 @@ module.exports = {
   rules: {
     'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
     'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
-    'max-len': ['warn', { code: 120 }]
+    'max-len': ['warn', { code: 120, 
+                          ignoreRegExpLiterals: true }]
```
ochiu, line 15: I added this for long regular expressions. Thoughts?

severinbeauvais: Yuck... but do we have a choice?

ochiu: Not really! I believe we can do string concatenation and instantiate RegExp(), but that is worse in my opinion!

PR #106, opened 2020-04-30
UI fixes

lekshmimallika-aot:
*Issue #:* /bcgov/entity#3541,/bcgov/entity#3544

*Description of changes:*

Changes to pass the effective date string only for future effective filing
Change to pass approved name as legal name in the NR section of the filing request
Updated the tests
Lint fix

### DIFF
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -71,7 +71,7 @@ export default class FilingTemplateMixin extends Vue {
     }
     // Pass the effective date only for a future effective filing.
     if (formattedDateTime) {
-      filing['filing']['header']['effectiveDate'] = formattedDateTime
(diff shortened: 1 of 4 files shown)


src/mixins/filing-template-mixin.ts
```diff
@@ -67,6 +69,11 @@ export default class FilingTemplateMixin extends Vue {
         }
       }
     }
+    // Pass the effective date only for a future effective filing.
+    if (formattedDateTime) {
+      filing['filing']['header']['effectiveDate'] = formattedDateTime
```
severinbeauvais, line 74: Would be better to use dot notation. I'm surprised the linter didn't complain about this.

(The named property notation you are using ignored typings/interfaces.)

lekshmimallika-aot: Oops.. sorry.. Corrected this

PR #107, opened 2020-05-01
#3525 UI - Future Effective Validations

cameron-freshworks:
*Issue #:* /bcgov/entity#3525

*Description of changes:*
* Implemented UI indicators for the various invalid states of Dates and Times.
* Refactored some Date functionality into the DateMixin.
* Applied DateTime Validations at File&Pay.
* Applied DateTime Validations when returning from a draft.
* Fixed unit tests affected by refactors.
* Added new unit testing.

### DIFF
--- a/src/components/ReviewConfirm/IncorporationDateTime.vue
+++ b/src/components/ReviewConfirm/IncorporationDateTime.vue
@@ -79,12 +79,15 @@
             </v-col>
             <v-col cols="12" sm="6" md="2" class="label-col">
               <span class="time-zone-label" :class="{ 'disabled': !isFutureEffective }">Pacific Time</span>
-            </v-col>
+            </v-col>{{isUnderTime}}
           </v-row>
           <v-row>
-            <v-col class="validation-alert" v-if="isFutureEffective && dateText && !isValidDateTime()">
+            <v-col
+              class="validation-alert"
+              v-show="isFutureEffective && dateText && !isValidDateTime(incorporationDateTime.effectiveDate)"
+            >
               <p class="validation-alert-msg">
-                <span v-if="!isOverMaxDateTime">The time must be at least {{minTime()}} for the selected date</span>
+                <span v-if="isUnderTime">The time must be at least {{minTime()}} for the selected date</span>
                 <span v-else>The time can't be greater than {{maxTime()}} for the selected date</span>
               </p>
             </v-col>
@@ -147,7 +150,8 @@ export default class IncorporationDateTime extends Mixins(DateMixin) {
     const expectedDateFormat = /^(19|20)\d\d[-.](0[1-9]|1[012])[-.](0[1-9]|[12][0-9]|3[01])$/
 
     return [
-      v => this.isFutureEffective && ((expectedDateFormat.test(v) && this.isValidDateTime(true)) ||
+      v => this.isFutureEffective && ((expectedDateFormat.test(v) &&
+        this.isValidDateTime(this.incorporationDateTime.effectiveDate, true)) ||
         `Date must be between ${minDate} and ${maxDate}`)
     ]
   }
@@ -245,31 +249,7 @@ export default class IncorporationDateTime extends Mixins(DateMixin) {
     }
   }
 
-  /** Validate the DateTime is within the allowed range */
-  private isValidDateTime (ignoreTime: boolean = false): boolean {
-    if (this.incorporationDateTime.effectiveDate) {
-      const effectiveDate = this.incorporationDateTime.effectiveDate
-      const startDate = new Date()
-
-      // Ignore the time difference if the user has input a Date before time.
-      if (ignoreTime && effectiveDate.getDate() === startDate.getDate()) {
-        return true
-      }
-
-      // Calculate time diff
-      const timeDiff = effectiveDate.getTime() - startDate.getTime()
-      const timeDiffInMinutes = Math.floor(timeDiff / 1000 / 60)
-
-      // Flag for identifying over/under
-      this.isOverMaxDateTime = timeDiffInMinutes > 14400
-
-      // Time set must be more than 2 minutes and less than 10 days
-      return timeDiffInMinutes >= 2 && timeDiffInMinutes <= 14400
-    }
-    return false
-  }
-
-  /** The maximum date that can be entered. */
+  /** The minimum date that can be entered. */
   private get minDate (): string {
     const minDate = new Date()
 
@@ -297,6 +277,15 @@ export default class IncorporationDateTime extends Mixins(DateMixin) {
       .toLocaleTimeString('en-US', { hour: '2-digit', minute: '2-digit', hour12: true })
   }
 
+  private get isUnderTime (): boolean {
+    if (this.incorporationDateTime.effectiveDate) {
+      const startDate = new Date()
+      const effectiveDate = this.incorporationDateTime.effectiveDate
+      const diff = (effectiveDate.getTime() - startDate.getTime()) / 1000
+      return diff <= 0
+    }
+  }
+
   /**
    * Set date text to Date chosen in date picker.
    * @param val The date selected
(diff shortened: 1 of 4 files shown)


src/components/ReviewConfirm/IncorporationDateTime.vue
```diff
@@ -204,65 +234,66 @@ export default class IncorporationDateTime extends Mixins(DateMixin) {
...
+
+  /** The minimum time that can be entered. */
+  private get minTime (): string {
+    return new Date(new Date().getTime() + 180000)
+      .toLocaleTimeString('en-US', { hour: '2-digit', minute: '2-digit', hour12: true })
+  }
+
+  /** The maximum time that can be entered. */
+  private get maxTime (): string {
+    const maxDate = new Date(new Date().setDate(new Date().getDate() + 10))
```
lekshmimallika-aot, line 279: Are these many new Date() needed?

src/components/ReviewConfirm/IncorporationDateTime.vue
```diff
@@ -73,15 +73,26 @@
...
+                :disabled="!isFutureEffective || !dateText"
                 filled
               ></v-select>
             </v-col>
             <v-col cols="12" sm="6" md="2" class="label-col">
               <span class="time-zone-label" :class="{ 'disabled': !isFutureEffective }">Pacific Time</span>
             </v-col>
           </v-row>
-        </div>
+          <v-row>
```
severinbeauvais, line 84: I don't think this is the right way to do validations and errors.
Have you looked at Vuelidate usage in https://github.com/bcgov/sbc-common-components/blob/master/vue/sbc-common-components/src/components/BaseAddress.vue ?
There's also built-in Vuetify validation (using rules -- which can be dynamic, so you can have different messages).

src/components/ReviewConfirm/IncorporationDateTime.vue
severinbeauvais: Out of interest, set a break point (or add a console.log) to this method to see how often it's called. If it's called a lot, I'd be tempted to optimize it.

cameron-freshworks: Each Date.prototype method used here returns a number ( one in days and one in milliseconds ). Need to then convert that number back into an actual Date. I could write a reference tho `const date = new Date()`?

src/mixins/date-mixin.ts
```diff
@@ -86,4 +87,24 @@ export default class DateMixin extends Vue {
...
     return Math.round((dateLocalMs - todayLocalMs) / this.MS_IN_A_DAY)
   }
+
+  /** Validate the DateTime is within the allowed range */
+  isValidDateTime (dateToValidate: Date, ignoreTime: boolean = false): boolean {
+    if (dateToValidate) {
+      const startDate = new Date()
+
+      // Ignore the time difference if the user has input a Date before time.
+      if (ignoreTime && dateToValidate.getDate() === startDate.getDate()) {
```
severinbeauvais, line 97: If you are checking "before" something, shouldn't you have a <= instead of === ?

src/components/ReviewConfirm/IncorporationDateTime.vue
cameron-freshworks: For sure, this little section had to be custom because it wasn't tied to an individual input field, instead a group. When we use the built in validations, it is tied to its component, where i wanted this message to appear under all the selectors as a group. Maybe a better way but this wasn't overly complicated either :)

src/components/ReviewConfirm/IncorporationDateTime.vue
cameron-freshworks: I will check, :re the call amounts!

src/mixins/date-mixin.ts
cameron-freshworks: This is comparing two numbers ( 1-31 ) where i want it to be true IF the Date input is the same day as today, just specifically want to look at the days only here.

severinbeauvais: Hmm, wasn't clear (still isn't clear, though I understand it now). How about explaining it as above in the code comments? :)

cameron-freshworks: Yes sir, 100%. Great idea!

src/components/ReviewConfirm/IncorporationDateTime.vue
lekshmimallika-aot: let maxDate = new Date ()
maxDate.setDate(maxDate.getDate() + 10) or something like this

cameron-freshworks: absolutely, can do!

src/components/ReviewConfirm/IncorporationDateTime.vue
```diff
@@ -297,6 +277,15 @@ export default class IncorporationDateTime extends Mixins(DateMixin) {
       .toLocaleTimeString('en-US', { hour: '2-digit', minute: '2-digit', hour12: true })
   }
 
+  private get isUnderTime (): boolean {
+    if (this.incorporationDateTime.effectiveDate) {
+      const startDate = new Date()
+      const effectiveDate = this.incorporationDateTime.effectiveDate
+      const diff = (effectiveDate.getTime() - startDate.getTime()) / 1000
```
ochiu: I believe since you are just using getTime() you can use `Date.now()` instead of `startDate.getTime()` so it doesn't have to instantiate a date object.

cameron-freshworks: Interesting, i'll try this out!

src/components/ReviewConfirm/IncorporationDateTime.vue
ochiu: Mentioned this in a previous comment, but `Date.now()` may help if you are interested in the milliseconds out the gate without creating a new date object. A little bit more efficient, especially if there are certain functions that get evaluated multiple times for things like validation.

src/components/ReviewConfirm/IncorporationDateTime.vue
cameron-freshworks: Cool, i like it 👍 Thanks!

src/components/ReviewConfirm/IncorporationDateTime.vue
cameron-freshworks: Updated!

tests/unit/IncorporationDateTime.spec.ts
```diff
@@ -222,7 +222,7 @@ describe('Incorporation Date Time', () => {
     const radioIsFutureEffective = radioInput.at(1)
     await radioIsFutureEffective.trigger('click')
 
-    const minTime = wrapper.vm.minTime
+    const minTime = wrapper.vm.minTime()
```
severinbeauvais, line 225: This passes TS checks? I've always had to cast vm to `any` because TS could never find the method or property on vm. If there's a setting that allows this to work correctly then we should port that setting to Filings UI.

severinbeauvais: I wonder if this works because you are using wrapperFactory() ?

severinbeauvais: Yes, that's why. You declared wrapperFactory as `any` :)

src/mixins/date-mixin.ts
severinbeauvais: Can you please mention this is in the comment?
> i want it to be true IF the Date input is the same day as today, just specifically want to look at the days only here

tests/unit/IncorporationDateTime.spec.ts
severinbeauvais: Cam, go ahead and resolve this conversation once you've read these comments.

cameron-freshworks: Interesting! I'm seeing no errors, nor did i change any of the TS configurations 🤔

cameron-freshworks: Could be because the factory method is described as `any`?

cameron-freshworks: We must have been commenting at the same time :)

src/mixins/date-mixin.ts
cameron-freshworks: For sure

PR #111, opened 2020-05-06
3608 - handle token refresher failure

ochiu:
*Issue #:* /bcgov/entity#3608

*Description of changes:*
- added logic to catch error for token refresher failure and refresh the page ( an error is thrown if the refresh token has expired)
- updated keycloak config proxy url for local dev

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -127,7 +127,10 @@ import { AccountAuthorizationDialog, NameRequestInvalidErrorDialog, ConfirmDialo
 import { DateMixin, FilingTemplateMixin, LegalApiMixin, NameRequestMixin } from '@/mixins'
 import { FilingDataIF, ActionBindingIF, ConfirmDialogType } from '@/interfaces'
 import { CertifyStatementResource } from '@/resources'
+
+// Enums and Constants
 import { EntityTypes, FilingCodes, RouteNames, NameRequestStates } from '@/enums'
+import { SessionStorageKeys } from 'sbc-common-components/src/util/constants'
 
 @Component({
   components: {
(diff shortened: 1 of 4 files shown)


vue.config.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     proxy: {
       // this is needed to prevent a CORS error when running locally
       '/local-keycloak-config-url/*': {
-        target: 'https://dev.bcregistry.ca/cooperatives/auth/config/kc/',
+        target: 'https://business-create-dev.pathfinder.gov.bc.ca/businesses/create/config/kc/',
```
severinbeauvais, line 13: 👍

src/App.vue
```diff
@@ -301,11 +304,32 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
...
+  /**
+   * Clears Keycloak token information from session storage
+   */
+  private clearKeycloakSession () : void {
+    sessionStorage.removeItem(SessionStorageKeys.KeyCloakToken)
+    sessionStorage.removeItem(SessionStorageKeys.KeyCloakIdToken)
+    sessionStorage.removeItem(SessionStorageKeys.KeyCloakRefreshToken)
+    sessionStorage.removeItem(SessionStorageKeys.UserFullName)
+    sessionStorage.removeItem(SessionStorageKeys.UserKcId)
+    sessionStorage.removeItem(SessionStorageKeys.UserAccountType)
```
severinbeauvais, line 332: Does one of these also clear our app's token? It's used by isAuthenticated() in a couple of places.

`sessionStorage.getItem('KEYCLOAK_TOKEN')`

ochiu: Yes, `SessionStorageKeys.KeyCloakToken`. Should I move away from the constants within the sbc-components?

severinbeauvais: OK I get it.

Actually, since "we" aren't the ones setting these tokens, it's better to use "their" keys to reference them. (Too bad they don't have a util method to clear these eh?)

Would you mind changing over our other references to KEYCLOAK_TOKEN to their keys? (except in the unit tests 😄 )

ochiu: Done!

PR #117, opened 2020-05-27
3489 add numbered ia

peter-freshworks:
*Issue #:* /bcgov/entity#3489

*Description of changes:*
Changed the Incorporation Application to use temporary registration Id, still need to refactor NameRequest reliant components (tickets created for this)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
     "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
     "lint": "vue-cli-service lint",
-    "test:unit": "vue-cli-service test:unit --testPathPattern"
+    "test:unit": "vue-cli-service test:unit --testPathPattern",
+    "test:debug": "node --inspect-brk node_modules/.bin/vue-cli-service test:unit --testPathPattern --no-cache --watch --runInBand"
   },
   "dependencies": {
     "@babel/compat-data": "^7.8.6",
--- a/src/App.vue
+++ b/src/App.vue
@@ -209,10 +209,6 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
   private get tempId (): string {
     return this.$route.query.id as string
   }
-  /** The Name Request Number from the route. */
-  private get nrNumber (): string {
-    return this.$route.query.nrNumber as string
-  }
 
   /** True if an error dialog is displayed. */
   private get isErrorDialog (): boolean {
@@ -278,7 +274,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     if (!this.haveChanges || force) {
       // redirect to dashboard
       const dashboardUrl = sessionStorage.getItem('DASHBOARD_URL')
-      window.location.assign(dashboardUrl + this.nrNumber)
+      // TODO: window.location.assign(dashboardUrl + this.nrNumber)
       return
     }
 
@@ -302,7 +298,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
       this.setHaveChanges(false)
       // redirect to dashboard
       const dashboardUrl = sessionStorage.getItem('DASHBOARD_URL')
-      window.location.assign(dashboardUrl + this.nrNumber)
+      // TODO: window.location.assign(dashboardUrl + this.nrNumber)
     })
   }
 
@@ -365,47 +361,17 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
         throw error // go to catch()
       })
 
-      /*
-      // fetch NR data
-      const nrResponse = await this.fetchNameRequest(this.tempId).catch(error => {
-        console.log('NR error =', error) // eslint-disable-line no-console
-        // this.nameRequestInvalidErrorDialog = true
-        // throw error // go to catch()
-        return {}
-      })
-
-      // ensure NR was found
-      debugger
-      if (!nrResponse) {
-        this.nameRequestInvalidType = NameRequestStates.NOT_FOUND
-        this.nameRequestInvalidErrorDialog = true
-        return // go to finally()
-      }
-
-      // ensure NR is valid
-      if (!this.isNrValid(nrResponse)) {
-        this.nameRequestInvalidType = NameRequestStates.INVALID
-        this.nameRequestInvalidErrorDialog = true
-        return // go to finally()
-      }
-
-      // ensure NR is consumable
-      const state = this.getNrState(nrResponse)
-      if (!state || state !== NameRequestStates.APPROVED) {
-        this.nameRequestInvalidType = state || NameRequestStates.INVALID
-        this.nameRequestInvalidErrorDialog = true
-        return // go to finally()
-      }
-
-      // if we get this far, the NR is good to go!
-      nrResponse.isConsumable = true
-
-      this.setNameRequestState(this.generateNameRequestState(nrResponse, null))
-      */
       try {
         this.setTemporaryId(this.tempId)
         // Retrieve draft filing if it exists for the NR Number specified
-        const draftFiling = await this.fetchDraft()
+        let draftFiling = await this.fetchDraft()
+        let emptyFiling = this.buildEmptyFiling()
+        if (draftFiling?.incorporationApplication?.nameRequest) {
+          await this.populateNameRequest(draftFiling)
+          emptyFiling = this.buildFiling()
+        }
+
+        draftFiling = { ...draftFiling, ...emptyFiling.filing }
 
         // Parse the draft and update NR data into the store if it exists
         if (draftFiling) {
@@ -433,8 +399,47 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     }
   }
 
-  private populateNameRequest (filing: any): void {
+  private async populateNameRequest (filing: any): Promise<void> {
+    try {
+      this.setCurrentDate(this.dateToUsableString(new Date()))
+      let nameRequest = filing.incorporationApplication.nameRequest
+      // ensure we have a NR number
 
+      if (!nameRequest.nrNumber) {
+        this.nameRequestInvalidType = NameRequestStates.NOT_FOUND
+        this.nameRequestInvalidErrorDialog = true
+        return // go to finally()
+      }
+      // fetch NR data
+      const nrResponse = await this.fetchNameRequest(nameRequest.nrNumber).catch(error => {
+        console.log('NR error =', error) // eslint-disable-line no-console
+        this.nameRequestInvalidErrorDialog = true
+        throw error // go to catch()
+      })
+      // ensure NR was found
+      if (!nrResponse) {
+        this.nameRequestInvalidType = NameRequestStates.NOT_FOUND
+        this.nameRequestInvalidErrorDialog = true
+        return // go to finally()
+      }
+      // ensure NR is valid
+      if (!this.isNrValid(nrResponse)) {
+        this.nameRequestInvalidType = NameRequestStates.INVALID
+        this.nameRequestInvalidErrorDialog = true
+        return // go to finally()
+      }
+      // ensure NR is consumable
+      const state = this.getNrState(nrResponse)
+      if (!state || state !== NameRequestStates.APPROVED) {
+        this.nameRequestInvalidType = state || NameRequestStates.INVALID
+        this.nameRequestInvalidErrorDialog = true
+        return // go to finally()
+      }
+      // if we get this far, the NR is good to go!
+      nrResponse.isConsumable = true
+      this.setNameRequestState(this.generateNameRequestState(nrResponse, filing.Id))
+    } catch (ex) {
+    }
   }
 
   /** Resets all error flags/states. */
--- a/src/App.vue
+++ b/src/App.vue
@@ -367,8 +367,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
         let draftFiling = await this.fetchDraft()
         let emptyFiling = this.buildFiling()
         if (draftFiling?.incorporationApplication?.nameRequest) {
-          await this.populateNameRequest(draftFiling)
-          emptyFiling = this.buildFiling()
+          await this.processNameRequest(draftFiling)
         }
 
         draftFiling = { ...emptyFiling.filing, ...draftFiling }
@@ -376,7 +375,6 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
         // Parse the draft and update NR data into the store if it exists
         if (draftFiling) {
           this.parseDraft(draftFiling)
-          // this.setNameRequestState(this.generateNameRequestState(nrResponse, +draftFiling.header.filingId))
         }
 
         // Initialize Fee Summary
@@ -399,9 +397,8 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     }
   }
 
-  private async populateNameRequest (filing: any): Promise<void> {
+  private async processNameRequest (filing: any): Promise<void> {
     try {
-      this.setCurrentDate(this.dateToUsableString(new Date()))
       let nameRequest = filing.incorporationApplication.nameRequest
       // ensure we have a NR number
 
--- a/src/App.vue
+++ b/src/App.vue
@@ -113,7 +113,7 @@ import { Component, Vue, Watch, Mixins } from 'vue-property-decorator'
 import { State, Action, Getter } from 'vuex-class'
 import TokenService from 'sbc-common-components/src/services/token.services'
 import { BAD_REQUEST, PAYMENT_REQUIRED, FORBIDDEN } from 'http-status-codes'
-
+import { merge } from 'lodash'
 // Components
 import SbcHeader from 'sbc-common-components/src/components/SbcHeader.vue'
 import SbcFooter from 'sbc-common-components/src/components/SbcFooter.vue'
@@ -274,7 +274,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     if (!this.haveChanges || force) {
       // redirect to dashboard
       const dashboardUrl = sessionStorage.getItem('DASHBOARD_URL')
-      // TODO: window.location.assign(dashboardUrl + this.nrNumber)
+      window.location.assign(dashboardUrl + this.getBusinessIdentifier)
       return
     }
 
@@ -298,7 +298,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
       this.setHaveChanges(false)
       // redirect to dashboard
       const dashboardUrl = sessionStorage.getItem('DASHBOARD_URL')
-      // TODO: window.location.assign(dashboardUrl + this.nrNumber)
+      window.location.assign(dashboardUrl + this.getBusinessIdentifier)
     })
   }
 
@@ -418,7 +418,6 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
         return {}
       })
       // ensure NR was found
-      debugger
       if (!nrResponse) {
         this.nameRequestInvalidType = NameRequestStates.NOT_FOUND
         this.nameRequestInvalidErrorDialog = true
--- a/src/App.vue
+++ b/src/App.vue
@@ -164,7 +164,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
   // Global getters
   @Getter haveChanges!: boolean
   @Getter getSteps!: Array<any>
-
+  @Getter getTempId!: string
   // Global actions
   @Action setCurrentStep!: ActionBindingIF
   @Action setCurrentDate!: ActionBindingIF
@@ -176,7 +176,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
   @Action setCreateShareStructureStepValidity!: ActionBindingIF
   @Action setHaveChanges!: ActionBindingIF
   @Action setAccountInformation!: ActionBindingIF
-  @Action setTemporaryId!: ActionBindingIF
+  @Action setTempId!: ActionBindingIF
 
   // Local Properties
   private filingData: Array<FilingDataIF> = []
@@ -202,11 +202,6 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
   /** The URL of the Pay API. */
   private get payApiUrl (): string {
     return sessionStorage.getItem('PAY_API_URL')
-  }
-
-  /** The Temporary Registration Id from the route */
-  private get tempId (): string {
-    return this.$route.query.id as string
   }
 
   /** True if an error dialog is displayed. */
@@ -273,7 +268,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     if (!this.haveChanges || force) {
       // redirect to dashboard
       const dashboardUrl = sessionStorage.getItem('DASHBOARD_URL')
-      window.location.assign(dashboardUrl + this.getBusinessIdentifier)
+      window.location.assign(dashboardUrl + this.getTempId)
       return
     }
 
@@ -297,7 +292,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
       this.setHaveChanges(false)
       // redirect to dashboard
       const dashboardUrl = sessionStorage.getItem('DASHBOARD_URL')
-      window.location.assign(dashboardUrl + this.getBusinessIdentifier)
+      window.location.assign(dashboardUrl + this.getTempId)
     })
   }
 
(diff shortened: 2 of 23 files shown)


src/App.vue
```diff
@@ -272,7 +287,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     if (!this.haveChanges || force) {
       // redirect to dashboard
       const dashboardUrl = sessionStorage.getItem('DASHBOARD_URL')
-      window.location.assign(dashboardUrl + this.nrNumber)
+      // TODO: window.location.assign(dashboardUrl + this.nrNumber)
```
severinbeauvais, line 290: The only change here is to use the TRN instead of the NRN.
Eg, https://dev.bcregistry.ca/cooperatives/T123456789

src/App.vue
```diff
@@ -358,48 +359,22 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
...
         // Retrieve draft filing if it exists for the NR Number specified
-        const draftFiling = await this.fetchDraft()
+        let draftFiling = await this.fetchDraft()
+        let emptyFiling = this.buildEmptyFiling()
+        if (draftFiling?.incorporationApplication?.nameRequest) {
+          await this.populateNameRequest(draftFiling)
+          emptyFiling = this.buildFiling()
+        }
+
+        draftFiling = { ...emptyFiling.filing, ...draftFiling }
```
severinbeauvais, line 372: 👍

severinbeauvais: Although, it would be nice if emptyFiling was filing (instead of containing it). Possible?

src/App.vue
```diff
@@ -358,48 +359,22 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
...
 
         // Parse the draft and update NR data into the store if it exists
         if (draftFiling) {
           this.parseDraft(draftFiling)
-          this.setNameRequestState(this.generateNameRequestState(nrResponse, +draftFiling.header.filingId))
+          // this.setNameRequestState(this.generateNameRequestState(nrResponse, +draftFiling.header.filingId))
```
severinbeauvais, line 377: Future?

src/App.vue
```diff
@@ -422,6 +397,51 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     }
   }
 
+  private async populateNameRequest (filing: any): Promise<void> {
+    try {
+      this.setCurrentDate(this.dateToUsableString(new Date()))
```
severinbeauvais, line 402: This belongs somewhere else (ie, called always -- not only if we have a NR).

src/App.vue
```diff
@@ -422,6 +397,51 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     }
   }
 
+  private async populateNameRequest (filing: any): Promise<void> {
```
severinbeauvais, line 400: Method name made me think it was doing something else. Suggest "processNameRequest", or ...?

src/components/common/Actions.vue
```diff
@@ -195,7 +196,11 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Lega
 
       // assume Pay URL is always reachable
       // otherwise user will have to retry payment later
-      window.location.assign(payUrl)
+      if (!paymentCompleted) {
```
severinbeauvais, line 199: Which ticket is this for?

src/components/common/Actions.vue
```diff
@@ -195,7 +196,11 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Lega
...
+        window.location.assign(payUrl)
+      } else {
+        window.location.assign(dashboardUrl + nrNumber)
```
severinbeauvais, line 202: nrNumber -> tempId

severinbeauvais: Also on lines 193-194 above.

src/mixins/filing-template-mixin.ts
```diff
@@ -19,8 +21,8 @@ export default class FilingTemplateMixin extends Vue {
   // Global Getters
   @Getter isTypeBcomp!: GetterIF
   @Getter getApprovedName!: string
-
-  // Global Actions
+  @Getter getBusinessIdentifier!: string
```
severinbeauvais, line 24: Pls restore blank line for nice code readability :)

src/mixins/legal-api-mixin.ts
```diff
@@ -52,17 +52,15 @@ export default class LegalApiMixin extends Vue {
...
   async fetchDraft (): Promise<any> {
     // get the draft filing from the tasks endpoint
-    const url = `businesses/${this.getBusinessIdentifier}/tasks`
+    const url = `businesses/${this.getBusinessIdentifier}/filings`
     return axios.get(url)
       .then(response => {
         // look at only the first task
-        const todoName = response?.data?.tasks[0]?.task?.todo?.header?.name
-        const filing = response?.data?.tasks[0]?.task?.filing
+        const todoName = response?.data?.filing?.header?.name
```
severinbeauvais, line 59: obsolete

src/mixins/legal-api-mixin.ts
```diff
@@ -52,17 +52,15 @@ export default class LegalApiMixin extends Vue {
...
+        const filing = response?.data?.filing
```
severinbeauvais, line 60: Might want to check below that this isn't falsy.

src/mixins/name-request-mixin.ts
```diff
@@ -19,7 +19,7 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
 
     return {
       nrNumber: nr.nrNum,
-      // TODO: Update entityType to use nr.requestTypeCd when namex supports our entity types
+      // TODO: Remove?
```
severinbeauvais, line 22: No, we should get the entityType from the NR (and the NR is still saying it's a CORP not a BCOMP) so we will still need to fix this later.

src/mixins/filing-template-mixin.ts
```diff
@@ -36,7 +38,52 @@ export default class FilingTemplateMixin extends Vue {
...
+    let filing:IncorporationFilingIF = {
+      filing: {
+        header: {
+          name: INCORPORATION_APPLICATION,
+          certifiedBy: this.stateModel?.certifyState?.certifiedBy,
+          email: this.stateModel?.defineCompanyStep?.businessContact?.email,
+          date: this.stateModel?.currentDate
+        },
+        business: {
+          legalType: this.stateModel?.entityType,
```
severinbeauvais, line 55: Where do we actually set entity type?

src/mixins/filing-template-mixin.ts
```diff
@@ -45,24 +92,24 @@ export default class FilingTemplateMixin extends Vue {
...
         header: {
           name: INCORPORATION_APPLICATION,
-          certifiedBy: this.stateModel.certifyState.certifiedBy,
-          email: this.stateModel.defineCompanyStep.businessContact.email,
+          certifiedBy: '',
+          email: '',
           date: this.stateModel.currentDate
         },
+        business: {
+          legalType: EntityTypes.BCOMP,
```
severinbeauvais, line 100: This should come from the IA.

Having this default value concerns me. I think we should not default it and throw an error if the IA doesn't contain this value.

src/store/getters/state-getters.ts
```diff
@@ -119,7 +118,7 @@ export const isShowBackBtn = (state: any): boolean => {
  * Whether Review and Confirm button should be displayed.
  */
 export const isShowReviewConfirmBtn = (state: any, getters: any): boolean => {
-  return (!!state.stateModel.nameRequest.entityType && state.stateModel.currentStep < getters.getMaxStep)
+  return (!!state.stateModel.entityType && state.stateModel.currentStep < getters.getMaxStep)
```
severinbeauvais, line 121: Any idea why we need entity type to be defined to show this button? Seems to me it's always defined (and if so then we can remove this check).
Maybe ask Cam or Lekshmi.

src/components/common/Actions.vue
severinbeauvais: Ah, this is the left over BCOL PR stuff.
https://github.com/bcgov/bcrs-business-create-ui/pull/114

src/mixins/filing-template-mixin.ts
peter-freshworks: got rid of this method any way, so should be better

src/mixins/name-request-mixin.ts
peter-freshworks: okie dokie

src/mixins/legal-api-mixin.ts
peter-freshworks: done

src/mixins/legal-api-mixin.ts
peter-freshworks: it is

src/mixins/filing-template-mixin.ts
peter-freshworks: okie dokie

src/components/common/Actions.vue
peter-freshworks: gotcha

src/components/common/Actions.vue
peter-freshworks: a different one :D I'll revert

src/App.vue
peter-freshworks: done

src/App.vue
peter-freshworks: turns out this is a duplicated line, and it is always called. removed

src/App.vue
peter-freshworks: it is

src/App.vue
peter-freshworks: because buildFiling() is called elsewhere I'm a little hesitant to do this

src/App.vue
peter-freshworks: done

src/App.vue
severinbeauvais: OK, nvm

package.json
```diff
@@ -6,7 +6,8 @@
     "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
     "lint": "vue-cli-service lint",
-    "test:unit": "vue-cli-service test:unit --testPathPattern"
+    "test:unit": "vue-cli-service test:unit --testPathPattern",
+    "test:debug": "node --inspect-brk node_modules/.bin/vue-cli-service test:unit --testPathPattern --no-cache --watch --runInBand"
```
severinbeauvais, line 10: Could you give a quick tech demo on how to use this? Thx!

src/App.vue
```diff
@@ -272,7 +272,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     if (!this.haveChanges || force) {
       // redirect to dashboard
       const dashboardUrl = sessionStorage.getItem('DASHBOARD_URL')
-      window.location.assign(dashboardUrl + this.nrNumber)
+      window.location.assign(dashboardUrl + this.getBusinessIdentifier)
```
severinbeauvais, line 275: Is "businessIdentified = "tempId" = "temporaryId?
Looking at this diff, I wonder why we have multiple properties for the same thing?

src/App.vue
```diff
@@ -344,81 +344,97 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
...
-      }
-    } catch (e) {
-      // errors should be handled above
-      // just fall through to finally()
-    } finally {
-      this.haveData = true
-      // wait for things to stabilize, then reset flag
-      Vue.nextTick(() => this.setHaveChanges(false))
+      this.setNameRequestState(this.generateNameRequestState(nrResponse, filing.Id))
+    } catch (ex) {
```
severinbeauvais, line 437: Could there be an exception here? Should it be handled? (If not, please add a comment so it doesn't look like missing code.)

severinbeauvais: Also, if lint complains about "unused parameter 'ex'", you can name it "e" (which lint doesn't complain about).

src/App.vue
```diff
@@ -203,9 +203,9 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
     return sessionStorage.getItem('PAY_API_URL')
   }
 
-  /** The Name Request Number from the route. */
-  private get nrNumber (): string {
-    return this.$route.query.nrNumber as string
+  /** The Temporary Registration Id from the route */
+  private get tempId (): string {
+    return this.$route.query.id as string
```
severinbeauvais, line 208: I think you only use this in 1 place. Having it as a getter is cool but may encourage other devs to use it, whereas they should be using TemporaryId. (If I understand correctly.)

src/components/common/EntityInfo.vue
```diff
@@ -26,7 +26,7 @@
           <v-list-item-subtitle class="business-info">
             <dl>
               <dt>{{ entityTitle() }}</dt>
-              <dd>Name Request No: <span id="entity-nr-number">{{ getBusinessIdentifier }}</span></dd>
+              <dd>Name Request No: <span id="entity-nr-number">{{ getNameRequestNumber }}</span></dd>
```
severinbeauvais, line 29: Does this work if there's no NR? (ie, test with Numbered Company)

src/App.vue
severinbeauvais: I don't see where this is called elsewhere...?

src/components/common/Actions.vue
```diff
@@ -189,8 +189,7 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Lega
       // redirect to pay and return to the dashboard
       const authUrl = sessionStorage.getItem('AUTH_URL')
       const dashboardUrl = sessionStorage.getItem('DASHBOARD_URL')
-      const nrNumber = this.$route.query.nrNumber as string
-      const returnUrl = encodeURIComponent(dashboardUrl + nrNumber)
+      const returnUrl = encodeURIComponent(dashboardUrl + this.getBusinessIdentifier)
```
severinbeauvais, line 192: So are we calling the temp reg number "tempId", "temporarId" or "BusinessIdentifier"?

src/mixins/filing-template-mixin.ts
```diff
@@ -45,27 +48,36 @@ export default class FilingTemplateMixin extends Vue {
       filing: {
         header: {
           name: INCORPORATION_APPLICATION,
-          certifiedBy: this.stateModel.certifyState.certifiedBy,
-          email: this.stateModel.defineCompanyStep.businessContact.email,
-          date: this.stateModel.currentDate
+          certifiedBy: this.stateModel?.certifyState?.certifiedBy ?? '',
```
severinbeauvais, line 51: Why use ?? instead of || ?

src/mixins/filing-template-mixin.ts
```diff
@@ -74,23 +86,20 @@ export default class FilingTemplateMixin extends Vue {
...
 
   /**
-   * Parses a fetched draft filing into the store.
-   * @param draftFiling the draft filing body to be parsed
+   * Method to parse a received draft filing into the store
+   * @param draftFiling The draft filing body to be parsed and assigned to store
    */
   parseDraft (draftFiling: any): void {
-    // Set Office Addresses
+    this.setEntityType(draftFiling.business.legalType)
```
severinbeauvais, line 93: This makes sense but please restore the previous comment :)

src/store/mutations/mutations-model.ts
```diff
@@ -3,6 +3,10 @@ import {
   BusinessContactIF, OrgPersonIF, ShareClassIF
 } from '@/interfaces'
 
+export const mutateTemporaryId = (state: any, tempId: string) => {
+  state.stateModel.tempId = tempId
```
severinbeauvais, line 7: If possible, can we name the property and its methods similarly? Eg, mutateTempId or state.stateModel.temporaryId ?

tests/unit/Actions.spec.ts
```diff
@@ -247,15 +252,18 @@ describe('Actions component - Filing Functionality', () => {
...
     const localVue = createLocalVue()
     localVue.use(VueRouter)
     const router = mockRouter.mock()
-    router.push({ name: 'define-company', query: { nrNumber: 'NR 1234567' } })
+    router.push({ name: 'define-company', query: { id: 'T1234567' } })
     wrapper = shallowMount(Actions, { localVue, store, router, vuetify })
 
     // Mock the function calls that may used by saveFiling below
-    jest.spyOn(wrapper.vm, 'createFiling').mockImplementation()
+    // jest.spyOn(wrapper.vm, 'createFiling').mockImplementation()
```
severinbeauvais, line 266: Yep, just delete the obsolete code (and unit test if applicable).

src/App.vue
peter-freshworks: are you referring to "this.setCurrentDate(this.dateToUsableString(new Date()))" or to the method itself?

package.json
peter-freshworks: sure, post-stand up on Monday?

src/App.vue
severinbeauvais: Is setCurrentDate() called somewhere? I didn't see it.

If you're looking for a spot, may I recommend the top of fetchData()?

package.json
severinbeauvais: That would be great? Or maybe right before sprint retro (like we used to do)? Ask Shahriar.

src/App.vue
peter-freshworks: done

src/App.vue
peter-freshworks: that is where it's called...well, in the top of the try block in "fetchData"

src/App.vue
peter-freshworks: yeah I was just transplanting logic from the previous code, but I have removed it as it is only used to set the registration Id

src/components/common/Actions.vue
peter-freshworks: I'll go with TempId :)

src/mixins/filing-template-mixin.ts
peter-freshworks: Checking for null or undefined...falsy values are cool here...but I suppose I could use || as well...do we have a strong preference?

peter-freshworks: also...I'm a C# dev historically :p

src/store/mutations/mutations-model.ts
peter-freshworks: this has been done

src/components/common/EntityInfo.vue
peter-freshworks: well, it doesn't crash, but it will have no number. If this isn't handled in a future ticket I can add a conditional? Thoughts?

src/App.vue
severinbeauvais: 🤦 How did I miss this?

src/mixins/filing-template-mixin.ts
severinbeauvais: Technically those are OK here but there's little usage of ?? so far and it comes with a gotcha (doesn't work for some falsy values) so my preference is ||.

src/App.vue
peter-freshworks: should be good now (copied the old comment as it makes sense here)

src/components/common/EntityInfo.vue
severinbeauvais: Will be handled by  /bcgov/entity#2548 .

severinbeauvais:
Ah dang... you have conflicts to resolve before we can merge.

src/store/getters/state-getters.ts
cameron-freshworks: Good point, do we ever not have a defined entityType? This could have been something implemented very early. Might want to look into removing the entityType boolean.

src/App.vue
```diff
@@ -345,82 +340,99 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
...
+        this.fetchErrorDialog = true
+        throw error // go to catch()
+      }
+    } catch (e) {
+      // errors should be handled above
+      // just fall through to finally()
+    } finally {
+      this.haveData = true
+      // wait for things to stabilize, then reset flag
+      Vue.nextTick(() => this.setHaveChanges(false))
```
cameron-freshworks, line 390: what is `haveChanges` represent? Seems a little vague.

src/App.vue
```diff
@@ -345,82 +340,99 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
...
+    }
+  }
+
+  /** Set up any required fields/ validation if a Name request is present */
+  private async processNameRequest (filing: any): Promise<void> {
```
cameron-freshworks, line 395: I think we have interfaces the define the filing structure, could that potentially be used here to define the type?

src/App.vue
```diff
@@ -345,82 +340,99 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
...
+        return // go to finally()
+      }
       // fetch NR data
-      const nrResponse = await this.fetchNameRequest(this.nrNumber).catch(error => {
+      const nrResponse = await this.fetchNameRequest(nameRequest.nrNumber).catch(error => {
         console.log('NR error =', error) // eslint-disable-line no-console
-        this.nameRequestInvalidErrorDialog = true
-        throw error // go to catch()
+        // this.nameRequestInvalidErrorDialog = true
+        // throw error // go to catch()
```
cameron-freshworks, line 409: Are these comments meant to be left in for future?

src/mixins/filing-template-mixin.ts
cameron-freshworks: I think for consistency, we should stick with `||` as well. 
I couldn't tell you if one is better than the other and if there is a good reason to use `??` over `||` then i'm all for the change, but if its both relatively the same, i'd try to be as congruent as possible.

src/mixins/legal-api-mixin.ts
```diff
@@ -41,8 +41,8 @@ export default class LegalApiMixin extends Vue {
     if (this.getFilingId && this.getFilingId > 0) {
       return this.updateFiling(filing, isDraft)
     } else {
-      // otherwise create a new filing
-      return this.createFiling(filing, isDraft)
+      // This should never happen. Filing Id should always be present
+      throw new Error('Invalid filing Id')
```
cameron-freshworks, line 45: Do we need this else block if we do nothing with it?

src/mixins/legal-api-mixin.ts
```diff
@@ -52,57 +52,29 @@ export default class LegalApiMixin extends Vue {
    */
   async fetchDraft (): Promise<any> {
     // get the draft filing from the tasks endpoint
-    const url = `businesses/${this.getBusinessIdentifier}/tasks`
+    const url = `businesses/${this.getTempId}/filings`
```
cameron-freshworks, line 55: This is a good change, didn't feel great fetching tasks.

src/mixins/legal-api-mixin.ts
```diff
@@ -165,4 +137,35 @@ export default class LegalApiMixin extends Vue {
...
+
+  /**
+    * Ensure consisent object structure for an incorporation application
+    * whether it contains a Name Request or not, and whether it is an initial
+    * draft or it has been previously saved. Object merging does not
+    * work very well otherwise (due to nested properties)
+    * @param filing The filing fetched from legal-api
+    * @returns the filing in safe-empty state if applicable
+  */
+  formatEmptyFiling (filing: any): IncorporationFilingIF {
```
cameron-freshworks, line 149: Would the back end return empty drafts in different structures or wouldn't return the properties at all? I'm just curious about this methods purpose?

src/mixins/filing-template-mixin.ts
severinbeauvais: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

src/App.vue
severinbeauvais: I coded this a while back. It is the global that keeps track of whether the user has changed anything. It's part of the logic to prompt if they try to leave the page without saving.

src/App.vue
severinbeauvais: I almost fell into this ~~rat~~rabbit-hole recently. As I recall, the filing isn't exactly the same as the NR, so we use any here.
If you feel strongly that this should be fixed, please create a new ticket.

src/App.vue
peter-freshworks: what he said!

src/App.vue
peter-freshworks: good catch...weird

src/mixins/filing-template-mixin.ts
peter-freshworks: changed

src/mixins/legal-api-mixin.ts
peter-freshworks: an "empty" draft does not have the same format as a FE filing object...this is a way of ensuring if it is missing necessary fields, that we have a safe-empty object to work with (this would only be an issue for new drafts, otherwise this method passes through the original filing that was retrieved)

src/App.vue
cameron-freshworks: I don't feel strongly about this :) lol

src/mixins/filing-template-mixin.ts
cameron-freshworks: oh nice, thanks for the reference. That is a pretty good explanation!

src/mixins/legal-api-mixin.ts
cameron-freshworks: I see i see, fair enough. Thanks!

PR #118, opened 2020-05-27
3764-folio-number

ochiu:
*Issue #:* /bcgov/entity#3764

*Description of changes:*
- add folio number to step 1
- updated state model
- updated filing template mixin

### DIFF
--- a/src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
+++ b/src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
@@ -7,14 +7,6 @@ export interface NameRequestIF {
   filingId: number | null
 }
 
-// Account information interface
-export interface AccountInformationIF {
-  accountType: string
-  id: number
-  label: string
-  type: string
-}
-
 // Name request response details interface
 export interface NameRequestDetailsIF {
   approvedName: string
--- a/src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
+++ b/src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
@@ -7,6 +7,14 @@ export interface NameRequestIF {
   filingId: number | null
 }
 
+// Account information interface
+export interface AccountInformationIF {
+  accountType: string
+  id: number
+  label: string
+  type: string
+}
+
 // Name request response details interface
 export interface NameRequestDetailsIF {
   approvedName: string
--- a/src/views/DefineCompany.vue
+++ b/src/views/DefineCompany.vue
@@ -49,6 +49,7 @@
       <v-card flat class="step-container">
           <FolioNumber
             :initialValue="folioNumber"
+            :isEditing="true"
             @folioNumberChange="onFolioNumberChange($event)"
           />
       </v-card>
@@ -95,6 +96,7 @@ export default class DefineCompany extends Mixins(EntityFilterMixin) {
 
   // Global getters
   @Getter isEntityType!: GetterIF
+  @Getter isPremiumAccount!: GetterIF
 
   // Global actions
   @Action setEntityType!: ActionBindingIF
@@ -183,11 +185,6 @@ export default class DefineCompany extends Mixins(EntityFilterMixin) {
 
   private get showErrors (): boolean {
     return Boolean(this.$route.query.showErrors)
-  }
-
-  private get isPremiumAccount (): boolean {
-    // TODO: update to check session when available in common components
-    return Boolean(sessionStorage.getItem('CURRENT_ACCOUNT') === 'PREMIUM')
   }
 }
 </script>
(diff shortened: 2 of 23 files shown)


src/views/DefineCompany.vue
```diff
@@ -158,9 +177,18 @@ export default class DefineCompany extends Mixins(EntityFilterMixin) {
...
+    this.setFolioNumber(folioNumber)
+  }
+
   private get showErrors (): boolean {
     return Boolean(this.$route.query.showErrors)
   }
+
+  private get isPremiumAccount (): boolean {
+    // TODO: update to check session when available in common components
+    return Boolean(sessionStorage.getItem('CURRENT_ACCOUNT') === 'PREMIUM')
```
severinbeauvais, line 190: In Filings UI we use a session storage keys enum from sbc-common-components. That might be good to use here.

severinbeauvais: Also it would be nice if there was a constant (or enum) for 'PREMIUM'.

src/views/DefineCompany.vue
```diff
@@ -39,6 +39,20 @@
...
+      </header>
+      <v-card flat class="step-container">
+          <FolioNumber
+            :initialValue="folioNumber"
+            @folioNumberChange="onFolioNumberChange($event)"
+          />
+      </v-card>
+    </section>
   </div>
 </template>
```
severinbeauvais, line 58: Is there a read-only (or isEditing=false) version of this (for review page 5)?

src/views/DefineCompany.vue
ochiu: Just got an update for sbc-common-components, will clean this up, it isn't just an account type string, it actually gives a block of account information. Will add this to the state so it can be used elsewhere

src/views/DefineCompany.vue
ochiu: There is some text in the mock up mentioning this for step 5 that I missed. I will add this.

src/views/DefineCompany.vue
severinbeauvais: Please also add the dependency to the release ticket that contains this feature. (Create UI Rls 0.4 or 0.5 ???)

ochiu: - Updated to use common component keys. 
- I also removed this specific getter and created a global one. 
- The account information is now in the store to make it accessible for other components.
- Account type enum created

src/views/DefineCompany.vue
ochiu: read only version implemented and added relevant unit tests

src/views/DefineCompany.vue
ochiu: > Please also add the dependency to the release ticket that contains this feature. (Create UI Rls 0.4 or 0.5 ???)

Updated ticket with the dependency. I think 0.4 should be fine.

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
```diff
@@ -7,6 +7,14 @@ export interface NameRequestIF {
   filingId: number | null
 }
 
+// Account information interface
```
severinbeauvais, line 10: Please move this to a new file. THIS file contains NR-specific IFs.

tests/unit/DefineCompany.spec.ts
```diff
@@ -77,7 +77,7 @@ describe('Define Company view', () => {
 
   it('displays folio number when it is a premium account', () => {
     store.state.stateModel.nameRequest.entityType = 'BC'
-    sessionStorage.setItem('CURRENT_ACCOUNT', 'PREMIUM')
+    store.state.stateModel.accountInformation.accountType = 'PREMIUM'
```
severinbeauvais, line 80: Please use the new account type enum :)

Also below on line 91.

src/views/DefineCompany.vue
severinbeauvais: So this ticket needs Legal API Rls 2.10, as outlined in /bcgov/entity#3682 ?

tests/unit/DefineCompany.spec.ts
ochiu: Did we not want to use the strings for unit tests to catch changes in the account type?

I thought we did this for other things, but maybe I am thinking something else.

severinbeauvais: Oops, this is a unit test file. You are correct!

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
ochiu: oops... fixed

src/views/DefineCompany.vue
ochiu: Yes, the api ticket /bcgov/entity#3766 is in 2.10 now.

PR #121, opened 2020-06-02
Step 4 UI + summary

lekshmimallika-aot:
*Issue #:* /bcgov/entity#3716

*Description of changes:*
Step 4 UI
Framework changes for step 4
Summary in step 5
Unit tests


severinbeauvais:
So it was decided that we wouldn't store/link to the Corporation Agreement (PDF) in a CMS or other web page?

src/components/IncorporationAgreement/AgreementType.vue
```diff
@@ -0,0 +1,150 @@
...
+        <!-- Summary Content -->
+        <div v-else class="summary-desc">
+            <div><v-icon color="green" class="agreement-valid-icon">mdi-check</v-icon></div>
+             <div>
+                {{ selectedAgreementDescription }}
+            </div>
+        </div>
+      </v-card>
+    </div>
+    <div v-else style="margin-left:1rem;">
```
severinbeauvais, line 30: To prevent inline styles, you can use class="ml-4" instead.

https://vuetifyjs.com/en/styles/spacing/

src/components/IncorporationAgreement/AgreementType.vue
```diff
@@ -0,0 +1,150 @@
...
+import { ActionBindingIF } from '@/interfaces'
+import { AgreementTypeResource } from '@/resources'
+
+@Component
+export default class AgreementType extends Vue {
+   private incorporationAgreementTypes = AgreementTypeResource
+
+  // State
+  @State(state => state.stateModel.incorporationAgreementStep.agreementType)
+  readonly agreementTypeState: string | null
```
severinbeauvais, line 60: This looks strange...
On line 56, types = resource
On this line, state = type

Would it make sense to rename some of these so they agree with each other?

severinbeauvais: Also on line 78  below.

lekshmimallika-aot:
> So it was decided that we wouldn't store/link to the Corporation Agreement (PDF) in a CMS or other web page?

Couple of issues here.

I tried giving the cms url directly. It is in a different domain and because of potential security issues download attribute on hyperlink does not download the file to the system. It opens the file in a new tab. User has to click download button in the new tab to download it. The requirement specifically states that the file should get downloaded. To try this out for reviewers, I have put the download URL as a getter.

2. The existing mechanism of file download in other UIs wont work here because it is not an API call.

After trying out several options and going through multiple articles, making it as a static content was the option that looked feasible to me. But if there is a better known way to do this, I can make that change

src/components/IncorporationAgreement/AgreementType.vue
```diff
@@ -0,0 +1,150 @@
...
+    padding-right: 0.5rem;
+}
+
+.agreement-option-list {
+  padding: 1.5rem;
+}
+
+.agreement-option {
+  padding-top: 1rem;
+  color: #495057;
```
severinbeauvais, line 148: Please use variable `$gray7` for this.

src/components/IncorporationAgreement/AgreementType.vue
```diff
@@ -0,0 +1,150 @@
...
+  private changeAgreementType (): void {
+    this.setIncorporationAgreementStepData({
+      valid: !!this.agreementType,
+      agreementType: this.agreementType
+    })
+  }
+
+  private get selectedAgreementDescription () : string {
+    if (this.agreementTypeState) {
+      return this.incorporationAgreementTypes.find(item => item.code === this.agreementTypeState)
```
severinbeauvais, line 101: If the code is not found, this will return 'undefined'. Would it be better to return '' ?

src/components/ReviewConfirm/Summary.vue
```diff
@@ -1,8 +1,9 @@
 <template>
   <div class="summary-container">
-    <SummaryDefineCompany />
+    <SummaryDefineCompany/>
     <ListPeopleAndRoles :personList="orgPersonList" :isSummary="true" :showErrorSummary="!step2Valid" />
     <ListShareClass :shareClasses="shareClasses" :isSummary="true" :showErrorSummary="!step3Valid"/>
+    <AgreementType :isSummary="true" :showErrorSummary="!step4Valid"/>
```
severinbeauvais, line 6: I *think* you can just specify the attribute when it's "true", eg,
`<AgreementType isSummary :showErrorSummary="!step4Valid" />`

If so, it would be a bit cleaner for all 3 summaries here.

src/store/getters/state-getters.ts
```diff
@@ -155,11 +155,11 @@ export const isShowFilePayBtn = (state: any, getters: any): boolean => {
...
  * Whether File and Pay button should be enabled.
  */
-export const isEnableFilePayBtn = (state: any): boolean => {
+export const isEnableFilePayBtn = (state: any, getters: any): boolean => {
   const step1Valid = state.stateModel.defineCompanyStep.valid
   const step2Valid = state.stateModel.addPeopleAndRoleStep.valid
-  const step3Valid = true // FOR TESTING ONLY
-  const step4Valid = true // FOR TESTING ONLY
+  const step3Valid = getters.isTypeBcomp ? state.stateModel.createShareStructureStep.valid : false
+  const step4Valid = getters.isTypeBcomp ? state.stateModel.incorporationAgreementStep.valid : false
```
severinbeauvais, line 162: 👍

src/mixins/filing-template-mixin.ts
```diff
@@ -65,7 +66,10 @@ export default class FilingTemplateMixin extends Vue {
             extension: this.stateModel?.defineCompanyStep?.businessContact?.extension || ''
           },
           parties: this.stateModel?.addPeopleAndRoleStep?.orgPeople || [],
-          shareClasses: this.stateModel?.createShareStructureStep?.shareClasses || []
+          shareClasses: this.stateModel?.createShareStructureStep?.shareClasses || [],
+          incorporationAgreement: {
+            agreementType: this.stateModel.incorporationAgreementStep.agreementType
```
severinbeauvais, line 71: Do you need a fallback value here if the user saves a draft without setting the agreement type?

src/mixins/filing-template-mixin.ts
```diff
@@ -113,6 +117,12 @@ export default class FilingTemplateMixin extends Vue {
     // Set Share Structure
     this.setShareClasses(draftFiling.incorporationApplication.shareClasses)
 
+    // Set Incorporation Agreement
+    this.setIncorporationAgreementStepData({
+      valid: !!draftFiling.incorporationApplication.incorporationAgreement.agreementType,
```
severinbeauvais, line 122: I think we don't set 'valid' for the other data because each page validates itself when it is created. Can we leave this out here similary?

src/views/IncorporationAgreement.vue
```diff
@@ -0,0 +1,316 @@
...
+  font-weight: bold;
+  padding-bottom: 0.5rem;
+}
+
+.preview-image {
+  width: 200px;
+  height: 259px;
+}
+
+.tool-tip {
```
severinbeauvais, line 308: Are the other tooltips in this project supposed to have this style too?

severinbeauvais: (If yes then it makes sense to move this CSS to base.scss.)

src/views/IncorporationAgreement.vue
```diff
@@ -0,0 +1,316 @@
...
+            incorporators and the first shareholders of the corporation.
+          </span>
+        </v-tooltip>
+        , and a set of
+        <v-tooltip top color="primary">
+          <template v-slot:activator="{ on }">
+            <span v-on="on" class="tool-tip"> Benefit Company Articles</span>
+          </template>
+          <span>
+            The Articles for a benefit company must state the benefits the company intends to provide to society, as
```
severinbeauvais, line 24: Check with Scott or Linda... I think this should be "Benefit Company" (proper noun).

src/views/IncorporationAgreement.vue
```diff
@@ -0,0 +1,316 @@
...
+            well as outlining the rules and procedures for corporate matters such as holding meetings, issuing and
+            transferring shares, and duties of directors and officers.
+          </span>
+        </v-tooltip>
+        containing a
+        <v-tooltip top color="primary">
+          <template v-slot:activator="{ on }">
+            <span v-on="on" class="tool-tip"> benefit provision</span>
```
severinbeauvais, line 32: Benefit Provision ?

src/views/IncorporationAgreement.vue
severinbeauvais: Also on line 35 below.

src/views/IncorporationAgreement.vue
```diff
@@ -0,0 +1,316 @@
...
+              There are no special rights or restrictions attached to any class or series of shares in the corporation’s
+              authorized share structure
+            </div>
+            <div>AND</div>
+          </li>
+          <li>
+            You do not wish to change any of the standard provisions in the sample Articles.
+          </li>
+        </ul>
+        <h4>
```
severinbeauvais, line 90: As the previous heading element is an H2, this should be an H3.
https://www.w3.org/WAI/tutorials/page-structure/headings/#heading-ranks

We did this on the CreateShareStructure page as well and I'd like to see this fixed. I realize we're using the H4 elements to get different styling (font size). So what we should do is change our H4s to H3s and add CSS in base.scss to make the H3s look like we want. Thanks!

src/views/IncorporationAgreement.vue
severinbeauvais: And "articles" should maybe be lower case.

src/views/IncorporationAgreement.vue
```diff
@@ -0,0 +1,316 @@
...
+          outline these special rights or restrictions in the Articles. We recommend seeking professional assistance
+          from a lawyer or accountant to help you prepare your Incorporation Agreement and Articles.
+        </div>
+
+        <h4>Retain the signed Incorporation Agreement and Benefit Company Articles</h4>
+        <ul>
+          <li>
+            The company is required to keep signed copies of the Incorporation Agreement and Articles in the company’s
+            record book. For a complete list of records a company is required to keep please see section 42 of the
+            Business Corporations Act.
```
severinbeauvais, line 117: Can we provide a link here? Ask Scott or Linda.

src/views/IncorporationAgreement.vue
```diff
@@ -0,0 +1,316 @@
...
+@Component({
+  components: {
+    AgreementType
+  }
+})
+export default class IncorporationAgreement extends Vue {
+  private helpToggle: boolean = false
+  private readMoreFlag: boolean = false
+
+  private get documentURL (): string {
```
severinbeauvais, line 198: Is this used anywhere?

src/views/IncorporationAgreement.vue
```diff
@@ -0,0 +1,316 @@
...
+  margin-top: 1.2rem
+}
+
+p{
+  padding-top: 0.5rem;
+}
+
+a {
+  text-decoration: none;
+  color: #1976d2;
```
severinbeauvais, line 232: This colour value is used in a number of places. I suggest adding a variable for it to theme.scss and replacing occurrences of it with its variable (wherever possible).

src/views/IncorporationAgreement.vue
```diff
@@ -0,0 +1,316 @@
...
+  margin-top: 1.5rem;
+  align-items: center;
+}
+
+.download-link-container {
+  padding-top: 1rem;
+  padding-bottom: 1rem;
+  margin-left: 1rem;
+  border-top: solid 1px #DEE2E6;
+  border-bottom: solid 1px #DEE2E6;
```
severinbeauvais, line 284: Please use $gray3 for these 2 colours.

tests/unit/AgreementType.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+      showErrorSummary: showErrorSummary,
+      isSummary: isSummary
+    },
+    router,
+    store,
+    vuetify
+  })
+}
+
+store.state.stateModel.nameRequest.entityType = 'BCOMP'
```
severinbeauvais, line 56: Actually, this should be 'BC'.
Please also fix in OrgPerson.spec.ts and ShareStructure.spec.ts.

tests/unit/AgreementType.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+store.state.stateModel.currentDate = '2020-03-30'
+
+describe('Share Structure component', () => {
+  it('Loads the component in edit mode and both agreement types are not selected', async () => {
```
severinbeauvais, line 60: Doesn't need to be async.

tests/unit/AgreementType.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  })
+
+  it('Selects the radio button if the value is set in the store', async () => {
+    store.state.stateModel.incorporationAgreementStep.agreementType = 'sample'
+    const wrapper: Wrapper<AgreementType> = createComponent()
+    await waitForUpdate(wrapper)
+    expect(wrapper.find(sampleTypeSelector).attributes('aria-checked')).toBe('true')
+    expect(wrapper.find(customTypeSelector).attributes('aria-checked')).toBe('false')
+    store.state.stateModel.incorporationAgreementStep.agreementType = null
+    await waitForUpdate(wrapper)
```
severinbeauvais, line 74: Not needed here since you're done?

severinbeauvais:
> > So it was decided that we wouldn't store/link to the Corporation Agreement (PDF) in a CMS or other web page?
> 
> Couple of issues here.
> 
> 1. I tried giving the cms url directly. It is in a different domain and because of potential security issues download attribute on hyperlink does not download the file to the system. It opens the file in a new tab. User has to click download button in the new tab to download it. The requirement specifically states that the file should get downloaded. To try this out for reviewers, I have put the download URL as a getter.

Yup, too bad this isn't implemented yet: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin

> 2. The existing mechanism of file download in other UIs wont work here because it is not an API call.

Isn't it just a simple GET? Might still work. Try it but don't spend too much time on it.

> After trying out several options and going through multiple articles, making it as a static content was the option that looked feasible to me. But if there is a better known way to do this, I can make that change

The issue with having this as a static file is now we have a second copy, which could get out of sync with the "managed" copy on the other site.

Have you read these?
https://www.nuxeo.com/blog/give-stuff-authenticated-file-download-in-javascript/
https://stackoverflow.com/questions/26921024/html5-download-attribute-not-working-cross-domain-can-it-be-made-to-work (proxy idea)

src/views/IncorporationAgreement.vue
lekshmimallika-aot: No, as I mentioned I put it there for reviewers to reproduce the scenario I mentioned about document download.  I will be removing that

src/mixins/filing-template-mixin.ts
lekshmimallika-aot: The default value is initialized as null. The schema checks against an enum. For draft filing if it is not selected it will be null and api will not validate. For submission the API will throw an error if it is null or not in the enum that has values sample, custom. So this code needs no change

src/mixins/filing-template-mixin.ts
lekshmimallika-aot: Yes, I forgot to remove that. I will remove this

src/views/IncorporationAgreement.vue
lekshmimallika-aot: All these texts are taken as such from the design and wireframes. I will ask Scott to go through those again. I will hold off those changes till I get a confirmation from them

src/views/IncorporationAgreement.vue
lekshmimallika-aot: Not sure. Will be checking with Linda/Scott

src/views/IncorporationAgreement.vue
lekshmimallika-aot: I am not sure about it. I see that people and roles has a similar styling

src/components/IncorporationAgreement/AgreementType.vue
lekshmimallika-aot: I dont think the above situation will occur at all

src/components/ReviewConfirm/Summary.vue
lekshmimallika-aot: It dint work that way. It shows the component in edit mode. So  I am leaving it as it is now.

tests/unit/AgreementType.spec.ts
lekshmimallika-aot: This is working as e xpected. You can see the assertions in the test and those assertions passes only if the component is initialized

severinbeauvais: I believe you.
And it might be fine here.
But in other tests, some checks for null/false/hidden _appeared_ to be working, but actually the component had not finished being mounted. After waiting until the next Vue tick, some of those conditions failed. That's why I now like to wait a tick after mount() to stabilitize before checking for the absence of things.

src/views/IncorporationAgreement.vue
severinbeauvais: I think Scott's intent is for all tooltips to appear as dotted-underline text. That's why I think the tooltip in ListPeopleAndRoles should be updated.

src/components/IncorporationAgreement/AgreementType.vue
severinbeauvais: You're right, it should never happen. But I always encourage safe code (esp since JS just exits a method when it encounters an error so the user might not know the page is incomplete).

Not a deal-breaker. I will accept this PR without this.

src/components/IncorporationAgreement/AgreementType.vue
lekshmimallika-aot: Not sure what you meant here. I have renamed it to incorporationAgreementTypeResource

src/components/IncorporationAgreement/AgreementType.vue
```diff
@@ -0,0 +1,150 @@
...
+        <!-- Summary Content -->
+        <div v-else class="summary-desc">
+            <div><v-icon color="green" class="agreement-valid-icon">mdi-check</v-icon></div>
+             <div>
+                {{ selectedAgreementDescription }}
+            </div>
+        </div>
+      </v-card>
+    </div>
+    <div v-else class="m1-1">
```
severinbeauvais, line 30: I think you want "ml-4" (margin-left) here. See prev link for mapping to px.

lekshmimallika-aot: ml-4 looks weird. I tried that. This one looks aligned

src/components/ReviewConfirm/Summary.vue
severinbeauvais: Darn. It works that way for Vuetify components.

src/views/IncorporationAgreement.vue
lekshmimallika-aot: There the tool tip is for a person/org with a very long name in the table. I dont think we should be using a dotted line there.

src/components/IncorporationAgreement/AgreementType.vue
severinbeauvais: Weird.
You had `style="margin-left:1rem;"` before. The Vuetify class is exactly that, along with "!important".
"m1-1" is not a valid class.
Is it possible you don't need any margin style on this div?

src/components/IncorporationAgreement/AgreementType.vue
severinbeauvais: Thanks for fixing that one.

Here's an example that I found confusing:
```
@State(state => state.stateModel.incorporationAgreementStep.agreementType)
  readonly agreementTypeState: string | null
```
The property in state is "agreementType" but in this component you rename it "agreementTypeState".

Why not map it with the same name?
```
@State(state => state.stateModel.incorporationAgreementStep.agreementType)
  readonly agreementType: string | null
```

severinbeauvais: OK, I completely missed that you have 2 things you want to name the same (the global and the local), so you used extended names to differentiate them. (I was confused.)

What you have now is fine.

src/views/IncorporationAgreement.vue
severinbeauvais: Oh, OK. I agree with you.

src/views/IncorporationAgreement.vue
severinbeauvais: This is Vuetify's primary button colour. If you can't find a name for it, you could call it "primary-blue" 😁

severinbeauvais: It's named "blue darken-2".

lekshmimallika-aot: Ok. will do

src/views/IncorporationAgreement.vue
lekshmimallika-aot: Checked with Linda and Scott. It is benefit provision

src/views/IncorporationAgreement.vue
lekshmimallika-aot: All entity type names are in lower case

src/views/IncorporationAgreement.vue
lekshmimallika-aot: Although it is missing from UX pin, there is a link in requirements docs. I have updated it to that link

lekshmimallika-aot: Linda just confirmed not to put in step 4 and they have decided to put it under Legal Responsibilities section. 

Another change is 'sample' is converted to lowercase throughout

src/views/IncorporationAgreement.vue
lekshmimallika-aot: Update. For consistency, for this release , it can be kept as Benefit Company. After that it needs to be changed to benefit company

src/views/IncorporationAgreement.vue
lekshmimallika-aot: Is it ok to style h3 as h4? or do you want a div with the required style?

src/components/IncorporationAgreement/AgreementType.vue
lekshmimallika-aot: You are right. No need of margin there

src/views/IncorporationAgreement.vue
severinbeauvais: Sure, just style it globally. (h3 will look like h4)

severinbeauvais: It's important to maintain heading rank, per link above. It's a matter of using the right element (h3).
You can style the h3 using any mechanism you want, but as this is really a project-wide style, doing it in base.scss is good.

severinbeauvais:
> > > So it was decided that we wouldn't store/link to the Corporation Agreement (PDF) in a CMS or other web page?

I just asked Kaine/Linda/et al whether a local copy of the PDF is OK for now, and they agree, so don't spend additional time on this now. Please create a backlog ticket to ultimately download the PDFs from the GDX CMS (for future development maybe).

tests/unit/AgreementType.spec.ts
lekshmimallika-aot: Hmm ok.

PR #123, opened 2020-06-03
Fixing the doc download path issue in dev

lekshmimallika-aot:
*Issue #:* /bcgov/entity#3716

*Description of changes:*
The current doc path does not work in dev as the vue process path is different. So changing the doc url to leverage the base url value in session storage

### DIFF
--- a/src/views/IncorporationAgreement.vue
+++ b/src/views/IncorporationAgreement.vue
@@ -169,7 +169,7 @@
             <div class="download-link-container">
               <span>
                 <v-icon color="blue">mdi-file-pdf-outline</v-icon>
-                <a href="/files/benefit_company__corporation_agreement.pdf" download>
+                <a :href="documentURL" download>
                   Download the sample Incorporation Agreement and Benefit Company Articles
                 </a>
               </span>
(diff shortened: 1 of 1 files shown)


src/views/IncorporationAgreement.vue
```diff
@@ -200,6 +200,11 @@ import { AgreementType } from '@/components/IncorporationAgreement'
 export default class IncorporationAgreement extends Vue {
   private helpToggle: boolean = false
   private readMoreFlag: boolean = false
+
+  private get documentURL ():string {
+    return sessionStorage.getItem('BASE_URL') +
+    'files/benefit_company__corporation_agreement.pdf'
```
severinbeauvais, line 206: Are you sure this will work?
Sample Base URL: https://dev.bcregistry.ca/cooperatives/CP0000841/

lekshmimallika-aot: The base url that I see in my session is https://business-create-dev.pathfinder.gov.bc.ca/businesses/create/

Isnt that right?

lekshmimallika-aot: I think the one that you mentioned is for filings UI right?

severinbeauvais: Yes, sorry, wrong project :)

PR #128, opened 2020-06-09
handle nameRequest object for numbered business

severinbeauvais:
*Issue #:* /bcgov/entity#3962

*Description of changes:*
- in a draft, allow nameRequest object without nrNumber
- always save nameRequest object with legalType
- make nrNumber and legalName optional
- removed unneeded fallback values
- added some handy getters
- update unit tests
- misc cleanup

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -381,18 +381,28 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
       })
 
       try {
-        // Retrieve draft filing if it exists for the IA Number specified
+        // fetch draft filing
         let draftFiling = await this.fetchDraft()
-        let emptyFiling = this.buildFiling()
-        if (draftFiling?.incorporationApplication?.nameRequest) {
-          await this.processNameRequest(draftFiling)
-        }
 
+        // merge draft properties into empty filing so all properties are initialized
+        const emptyFiling = this.buildFiling()
         draftFiling = { ...emptyFiling.filing, ...draftFiling }
 
-        // Parse the draft and update Incorporation data into the store if it exists
+        // parse draft filing into the store
         if (draftFiling) {
           this.parseDraft(draftFiling)
+        }
+
+        // verify nameRequest object
+        const nameRequest = draftFiling?.incorporationApplication?.nameRequest
+        if (!nameRequest) {
+          // there should be a nameRequest object but continue even if it doesn't exist
+          // TODO: after MVP, add a real check here
+          console.log('Filing should have nameRequest object') // eslint-disable-line no-console
+        }
+        // if we have a NR number, fetch the NR
+        if (nameRequest?.nrNumber) {
+          await this.processNameRequest(draftFiling)
         }
 
         // Initialize Fee Summary
@@ -415,34 +425,33 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
     }
   }
 
-  /** Set up any required fields/ validation if a Name request is present */
+  /** Fetches NR and validates it. */
   private async processNameRequest (filing: any): Promise<void> {
     try {
-      const nameRequest = filing.incorporationApplication.nameRequest
+      const nrNumber = filing.incorporationApplication.nameRequest.nrNumber
 
-      // ensure we have an NR number
-      if (!nameRequest.nrNumber) {
-        this.nameRequestInvalidType = NameRequestStates.NOT_FOUND
-        this.nameRequestInvalidErrorDialog = true
-        return
-      }
       // fetch NR data
-      const nrResponse = await this.fetchNameRequest(nameRequest.nrNumber).catch(error => {
+      const nrResponse = await this.fetchNameRequest(nrNumber).catch(error => {
         console.log('NR error =', error) // eslint-disable-line no-console
         this.nameRequestInvalidErrorDialog = true
       })
+
       // ensure NR was found
       if (!nrResponse) {
         this.nameRequestInvalidType = NameRequestStates.NOT_FOUND
         this.nameRequestInvalidErrorDialog = true
         return
       }
+
       // ensure NR is valid
       if (!this.isNrValid(nrResponse)) {
         this.nameRequestInvalidType = NameRequestStates.INVALID
         this.nameRequestInvalidErrorDialog = true
         return
       }
+
+      // TODO: verify that nrResponse.entityTypeCd === business.legalType
+
       // ensure NR is consumable
       const state = this.getNrState(nrResponse)
       if (state !== NameRequestStates.APPROVED && state !== NameRequestStates.CONDITIONAL) {
@@ -451,8 +460,8 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
         return
       }
       // if we get this far, the NR is good to go!
-      nrResponse.isConsumable = true
-      this.setNameRequestState(this.generateNameRequestState(nrResponse, filing.Id))
+      const nameRequestState = this.generateNameRequestState(nrResponse, filing.Id)
+      this.setNameRequestState(nameRequestState)
     } catch (e) {
       // errors should be handled above
     }
--- a/src/interfaces/filing-interfaces/filing-interface.ts
+++ b/src/interfaces/filing-interfaces/filing-interface.ts
@@ -1,6 +1,7 @@
 import { IncorporationAddressIf } from '@/interfaces/stepper-interfaces/DefineCompany/address-interface'
 import { OrgPersonIF, ShareClassIF } from '@/interfaces'
 
+/** Interface for incorporation filing data saved to the Legal API. */
 export interface IncorporationFilingIF {
   filing: {
     header: {
@@ -17,10 +18,11 @@ export interface IncorporationFilingIF {
       identifier: string
     },
     incorporationApplication: {
-      nameRequest?: {
-        nrNumber: string
+      // NB: nameRequest must match schema
+      nameRequest: {
         legalType: string
-        legalName: string
+        nrNumber?: string // only set when there is an NR
+        legalName?: string // only set when there is an NR
       },
       offices: IncorporationAddressIf | {},
       contactPoint: {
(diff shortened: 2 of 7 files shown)


src/App.vue
```diff
@@ -381,20 +381,30 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
...
-        // Retrieve draft filing if it exists for the IA Number specified
+        // fetch draft filing
         let draftFiling = await this.fetchDraft()
-        let emptyFiling = this.buildFiling()
-        if (draftFiling?.incorporationApplication?.nameRequest) {
-          await this.processNameRequest(draftFiling)
-        }
 
+        // merge draft properties into empty filing so all properties are initialized
+        const emptyFiling = this.buildFiling()
```
severinbeauvais, line 388: I just moved this above the name request code as it made more sense to me.

src/App.vue
```diff
@@ -415,34 +425,33 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
...
 
-  /** Set up any required fields/ validation if a Name request is present */
+  /** Fetches NR and validates it. */
   private async processNameRequest (filing: any): Promise<void> {
     try {
-      const nameRequest = filing.incorporationApplication.nameRequest
+      const nrNumber = filing.incorporationApplication.nameRequest.nrNumber
 
-      // ensure we have an NR number
-      if (!nameRequest.nrNumber) {
```
severinbeauvais, line 424: Two reasons we don't need this: (1) In the case of a numbered business IA, we now have a nameRequest object without nrNumber. (2) When this method is called, we know we have a NR to fetch.

Note that this nameRequest object (ie, from draft filing) is **not** the name request data from namex/NRO. This object matches the name_request schema that the Legal API requires. See buildFiling() for how this object is saved. See generateNameRequestState() for how we load the store with NR data.

src/App.vue
```diff
@@ -451,8 +460,8 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
         return
       }
       // if we get this far, the NR is good to go!
-      nrResponse.isConsumable = true
-      this.setNameRequestState(this.generateNameRequestState(nrResponse, filing.Id))
+      const nameRequestState = this.generateNameRequestState(nrResponse, filing.Id)
+      this.setNameRequestState(nameRequestState)
```
severinbeauvais, line 464: I did this for temporary testing, and then it seemed cleaner to leave it like this :)

src/mixins/filing-template-mixin.ts
```diff
@@ -30,58 +30,57 @@ export default class FilingTemplateMixin extends Vue {
...
   @Action setEffectiveDate!: ActionBindingIF
   @Action setIsFutureEffective!: ActionBindingIF
   @Action setFolioNumber!: ActionBindingIF
   @Action setIncorporationAgreementStepData!: ActionBindingIF
 
-  /** Constructs a filing body, used when saving a filing. */
+  /** Constructs a filing body from store data. Used when saving a filing. */
   buildFiling (): IncorporationFilingIF {
-    // Format DateTime for Filing
-    const effectiveDate = this.stateModel?.incorporationDateTime?.effectiveDate || ''
```
severinbeauvais, line 44: We don't need all the fallback values because the initial state properties are always declared (see state-model.ts).

src/mixins/filing-template-mixin.ts
```diff
@@ -30,58 +30,57 @@ export default class FilingTemplateMixin extends Vue {
...
-          legalType: this.stateModel?.entityType || '',
-          identifier: this.getTempId || ''
+          legalType: this.stateModel.entityType,
+          identifier: this.getTempId
         },
         incorporationApplication: {
-          offices: this.stateModel?.defineCompanyStep?.officeAddresses || {},
+          nameRequest: {
+            legalType: this.stateModel.entityType
+          },
```
severinbeauvais, line 64: Well, this is the big change: always save nameRequest with legalType property.

tests/unit/App.spec.ts
```diff
@@ -289,7 +278,20 @@ describe('Numbered company setup', () => {
...
+            header: {
+              name: 'incorporationApplication',
+              filingId: 54321,
+              status: 'draft'
+            },
+            business: {
+              identifier: 'T7654321',
+              legalType: 'BC'
+            },
+            incorporationApplication: {
```
severinbeauvais, line 290: This will be needed after MVP, when/if we add a check that this object exists.

src/App.vue
```diff
@@ -381,20 +381,30 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
       })
 
       try {
-        // Retrieve draft filing if it exists for the IA Number specified
+        // fetch draft filing
```
cameron-freshworks, line 384: This comment has gone full circle, back to its roots :)

src/interfaces/filing-interfaces/filing-interface.ts
```diff
@@ -17,10 +18,11 @@ export interface IncorporationFilingIF {
       identifier: string
     },
     incorporationApplication: {
-      nameRequest?: {
-        nrNumber: string
+      // NB: nameRequest must match schema
+      nameRequest: {
```
cameron-freshworks, line 22: is `nameRequest` required, even as an empty object?

cameron-freshworks: Scratch that, i see `legalType` is required.

src/mixins/filing-template-mixin.ts
cameron-freshworks: cool, so this property is ok to be a passed as `null` in the event that a user selects an immediate filing? The backend/schema has changed a bit throughout development on this property, might have initially been an empty string to pass validators potentially?

src/interfaces/filing-interfaces/filing-interface.ts
severinbeauvais: Yes. This "nameRequest" object is required by the Legal API and must conform to the schema (https://github.com/bcgov/business-schemas/blob/master/src/registry_schemas/schemas/name_request.json).
It's a bit confusing because this is **not** the name request data that we get from namex/nro.

src/mixins/filing-template-mixin.ts
severinbeauvais: Not quite.
This method builds a filing object from state data. I'm saying the state data is always declared so we don't need to provide the fallback ( "||" ) values -- which were the same as the initialized values anyways.
If I did this right, the final filing object is the same now as before (except for the new bit).

src/interfaces/filing-interfaces/filing-interface.ts
ochiu: ahh okay, I thought only the business legal type was required. Interesting.

severinbeauvais:
> Looks good. Thanks for the clean up as well!

Thank you for reviewing promptly. And you're welcome :)

src/mixins/filing-template-mixin.ts
```diff
@@ -30,58 +30,57 @@ export default class FilingTemplateMixin extends Vue {
...
-            email: this.stateModel?.defineCompanyStep?.businessContact?.email || '',
-            phone: this.stateModel?.defineCompanyStep?.businessContact?.phone || '',
-            extension: this.stateModel?.defineCompanyStep?.businessContact?.extension || ''
+            email: this.stateModel.defineCompanyStep.businessContact.email,
+            phone: this.stateModel.defineCompanyStep.businessContact.phone,
+            extension: this.stateModel.defineCompanyStep.businessContact.extension
           },
-          parties: this.stateModel?.addPeopleAndRoleStep?.orgPeople || [],
-          shareClasses: this.stateModel?.createShareStructureStep?.shareClasses || [],
+          parties: this.stateModel.addPeopleAndRoleStep.orgPeople,
```
peter-freshworks, line 71: did we test this with an empty draft IA for numbered company? I know I was getting issues here because contact point/offices/parties aren't defined until at least a first save

src/interfaces/filing-interfaces/filing-interface.ts
severinbeauvais: Looks like we don't need `business.legalType` at all (since we can use the value from `incorporationApplication.nameRequest`, but it seems risky to remove it from all 3 UIs right now so I didn't :)

src/mixins/filing-template-mixin.ts
severinbeauvais: Yes, I created a new named IA and navigated to this UI and verified that the draft loaded OK and there were no errors.
In this method, a filing object is created from store data. That store data always exists (since that's how we declare the stateModel object). So, de-referencing its properties should always succeed.
I further confirmed that all stateModel properties/subobjects are already initialized in the store (to the same fallback values we had here in fact).
Do you remember how to reproduce the issues you saw?

src/mixins/filing-template-mixin.ts
cameron-freshworks: ahh fair enough! In this instance, i believe the declared store value for an `effectiveDate` is null, which conflicts a little with the removed fallback value. So good change! 
If the backend doesn't like null ( i'm can't recall what it expects ), the value could be declared as an empty string in store, either way good clean up!

src/mixins/filing-template-mixin.ts
cameron-freshworks: afaik there is zero validation on draft filings now. Was this an issue you encountered recently?

src/mixins/filing-template-mixin.ts
severinbeauvais:
```
    const effectiveDate = this.stateModel.incorporationDateTime.effectiveDate
    const formattedDateTime = effectiveDate &&
      (effectiveDate.toISOString()).replace('Z', '+00:00')
```
effectiveDate is only used in the next line of code (which I didn't change) IF this value is truthy. Both null and '' are falsy, so formattedDateTime isn't assigned (ie, remains undefined). So this is the same functionality as before.

src/mixins/filing-template-mixin.ts
severinbeauvais: No, this is just cleanup while I was already working on this method. But it did bug me that it looked like were providing useful fallbacks, but this code was unnecessary. I guess this principle is sort of DRY and sort of "anti decoy pattern" :)

peter-freshworks: I do not remember...or at least I do remember a numbered company empty draft causing that issue. Before adding fallbacks I had used a method called "buildEmptyFiling()" for testing (if you remember). But, if you say it doesn't happen, I believe you!

PR #130, opened 2020-06-12
#3927 Breadcrumb for Create UI

cameron-freshworks:
*Issue #:* /bcgov/entity#3927

*Description of changes:*

* Implemented Breadcrumb for Create UI

* Updated EntityInfo Tests

*NOTES:*

* Navigation is rather limited in this UI resulting in a rather static breadcrumb with a few computed values.

* Adding the individual steps to the breadcrumb seemed too granular. Feel free to comment if you think it should be otherwise.

* If we determine we wanted something more dynamic based on the routes, we can add the breadcrumb meta data to the individual step routes. 

### DIFF
--- a/src/components/common/EntityInfo.vue
+++ b/src/components/common/EntityInfo.vue
@@ -6,7 +6,6 @@
           slot="item"
           slot-scope="{ item }"
           exact
-          :to="item.to"
           :href="item.href">
           {{ item.text }}
         </v-breadcrumbs-item>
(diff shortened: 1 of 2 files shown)


severinbeauvais:
> * Adding the individual steps to the breadcrumb seemed too granular. Feel free to comment if you think it should be otherwise.
> * If we determine we wanted something more dynamic based on the routes, we can add the breadcrumb meta data to the individual step routes.

IMHO, I don't think this is needed.

cameron-freshworks:
> > * Adding the individual steps to the breadcrumb seemed too granular. Feel free to comment if you think it should be otherwise.
> > * If we determine we wanted something more dynamic based on the routes, we can add the breadcrumb meta data to the individual step routes.
> 
> IMHO, I don't think this is needed.

Definitely agree.

cameron-freshworks:
Have 3 approvals, will merge!

PR #133, opened 2020-06-16
Bug fixes - Share Structure

lekshmimallika-aot:
*Issue #:*  /bcgov/entity#4030,  /bcgov/entity#4051, /bcgov/entity#3882

*Description of changes:*
Bug fix
Added a test for the issue 4051


src/views/CreateShareStructure.vue
```diff
@@ -26,7 +26,7 @@
         of shares, each class must be assigned an identifying name such as Class A, Class B, etc.
       </p>
       <div id="share-structure-sample">
-        <ListShareClass :shareClasses="sharesHelpSample"/>
+        <ListShareClass :shareClasses="sharesHelpSample" :isSummary="true"/>
```
cameron-freshworks, line 29: Just curious, if you just passed `isSummary`, would it work the same? I know this is a Prop that we created, so i'm very curious if work the same as some of the attributes built into Vuetify, that are booleans.  No real action required here :)

severinbeauvais, line 29: I noticed this component, even with isSummary=true, still highlights rows when you mouse over them. You may need to disable cursor events in this case.

lekshmimallika-aot: I tried that in Summary. It dint work like the built in attribute and the component was displayed in the edit mode

tests/unit/ShareStructure.spec.ts
```diff
@@ -379,4 +379,22 @@ describe('Share Structure component', () => {
     await waitForUpdate(wrapper)
     expect(wrapper.find('.v-select').text()).toContain('United States dollar (USD)')
   })
+
+  it('Shows error if maximum shares of class is changed to lower value than series max shares', async () => {
```
severinbeauvais, line 383: Can you please include a test for bug /bcgov/entity#4030 ?

src/views/CreateShareStructure.vue
severinbeauvais: Yeah, Vuetify is probably checking something on its props that we're not, which is why it works for their components but not ours.
This can be someone's pet project to figure out (hint-hint Cameron).

cameron-freshworks: Very cool, thanks for exploring that at least :) 

I'd love too, just need more timmmmeee 😩

tests/unit/ShareStructure.spec.ts
severinbeauvais: Fixed, thanks!

src/views/CreateShareStructure.vue
lekshmimallika-aot: That is the data table feature and I am not sure whether it is a bug or it should be disabled when  it is is summary mode. If it needs to be done, I will have to know a bit more about data table and the deep selectors that can be used in data table I think. Probably in another ticket?

severinbeauvais: It's already described in /bcgov/entity#3882 :)

lekshmimallika-aot: Yes I saw that. All other points in the ticket are resolved as part of this PR and this PR has a P1 fix as well. So doing more rearch on data table should not block this PR as such

severinbeauvais: Fair enough. How about updating that ticket's text and putting it back into the backlog?

lekshmimallika-aot: I prefer to create another ticket as that one has to look into how to disable fetaures like hover styles, pointer style etc. Before that I would like to confirm with PO whether this is actually an issue

severinbeauvais: OK.

lekshmimallika-aot: Checked with Linda. She confirmed that the data fix and the edit issues need to be resolved and the highlighting is not an issue

PR #136, opened 2020-06-17
4046 - file and pay updates

ochiu:
*Issue #:* /bcgov/entity#4046

*Description of changes:*
- Moved the setter for isFilingPaying so it is cleared on error so the file and pay button keeps the loading indicator and actions disabled until redirects happen.


ochiu:
I was able to test this against
1) error on saving filing (402 payment required, error creating invoice)
2) when redirecting to make payment
3) payment was complete (premium account) and redirected to dashboard

PR #140, opened 2020-06-19
fix 3685 and 4093  - business create

peter-freshworks:
*Issue #:* /bcgov/entity#3685
*Issue #:* /bcgov/entity#4093
*Description of changes:*

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -262,6 +262,7 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
       // process errors/warnings
       switch (error?.response?.status) {
         case PAYMENT_REQUIRED:
+          this.haveChanges = false
           const errObj = await this.getErrorObj(this.getErrorCode(error))
           if (errObj) {
             this.bcolObj = errObj
--- a/src/App.vue
+++ b/src/App.vue
@@ -262,6 +262,7 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
       // process errors/warnings
       switch (error?.response?.status) {
         case PAYMENT_REQUIRED:
+          // Changes were saved if a 402 is received. haveChanges flag is cleared.
           this.haveChanges = false
           const errObj = await this.getErrorObj(this.getErrorCode(error))
           if (errObj) {
--- a/src/mixins/bcol-mixin.ts
+++ b/src/mixins/bcol-mixin.ts
@@ -8,7 +8,7 @@ export default class BcolMixin extends Vue {
       const fetchUrl: string = this.payApi + 'codes/errors/' + errCode
       // Currently no desirable way to handle errors during this request,
       // null is returned in any error situation regardless.
-      const errObj: any = await axios.get(fetchUrl).catch()
+      const errObj: any = await axios.get(fetchUrl)
       if (errObj?.data) {
         return errObj.data
       }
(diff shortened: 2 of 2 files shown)


src/mixins/bcol-mixin.ts
```diff
@@ -4,24 +4,28 @@ import axios from '@/utils/axios-auth'
...
+      const fetchUrl: string = this.payApi + 'codes/errors/' + errCode
+      // Currently no desirable way to handle errors during this request,
+      // null is returned in any error situation regardless.
+      const errObj: any = await axios.get(fetchUrl).catch()
+      if (errObj?.data) {
+        return errObj.data
+      }
+      return null
+    } catch (error) {
+      return null
```
cameron-freshworks, line 17: Just curious if you need the try catch if `null` is return regardless. Are you able to `return errObj?.data || null` or are you expecting a potential error on the `fetchUrl`?

src/mixins/bcol-mixin.ts
```diff
@@ -4,24 +4,28 @@ import axios from '@/utils/axios-auth'
...
     }
-    return null
   }
 
   getErrorCode (error: any): string {
     if (error?.response?.data?.errors) {
-      const msgCode = error.response.data.errors.find(x => x.payment_error_type.startsWith('BCOL'))
+      const msgCode = error.response.data.errors.find(x => x.payment_error_type && x.payment_error_type.startsWith('BCOL'))
```
cameron-freshworks, line 23: Does optional chaining work here?

src/App.vue
```diff
@@ -262,6 +262,7 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
       // process errors/warnings
       switch (error?.response?.status) {
         case PAYMENT_REQUIRED:
+          this.haveChanges = false
```
severinbeauvais, line 265: Please add a comment why it's OK to clear this flag in this situation.

severinbeauvais: (The principle I'm following is that code paths are often symmetric/similar, and in this case they're not. As a reviewer, I wonder why they're not. Totally fair scenario, just document it.)

src/mixins/bcol-mixin.ts
severinbeauvais: I'm wondering the same thing.
Also, if you get a lint error about "error not used" then you can change this to the special form `catch (e)` (where it ignores the fact that you're not using e).

src/mixins/bcol-mixin.ts
peter-freshworks: it does

src/App.vue
severinbeauvais: Perfect, thanks!

src/mixins/bcol-mixin.ts
```diff
@@ -8,7 +8,7 @@ export default class BcolMixin extends Vue {
       const fetchUrl: string = this.payApi + 'codes/errors/' + errCode
       // Currently no desirable way to handle errors during this request,
       // null is returned in any error situation regardless.
-      const errObj: any = await axios.get(fetchUrl).catch()
+      const errObj: any = await axios.get(fetchUrl)
```
severinbeauvais, line 11: Have you tested the scenario where this get fails (exception)? Easy to do by corrupting the fetchUrl string above. This just to make sure the error is handled correctly.

src/mixins/bcol-mixin.ts
```diff
@@ -20,7 +20,9 @@ export default class BcolMixin extends Vue {
 
   getErrorCode (error: any): string {
     if (error?.response?.data?.errors) {
-      const msgCode = error.response.data.errors.find(x => x.payment_error_type && x.payment_error_type.startsWith('BCOL'))
+      const msgCode = error.response.data.errors.find(x =>
+        x?.payment_error_type?.startsWith('BCOL')) // eslint-disable-line camelcase
```
severinbeauvais, line 24: FYI, another way to add the lint exception for a long line is to specify is before the line with the command "eslint-disable-next-line ...".

severinbeauvais:
Although the code could be improved, I don't see any issue with merging this anyway.

PR #141, opened 2020-06-19
NR validation before file and pay

lekshmimallika-aot:
*Issue #:* /bcgov/entity#4037

*Description of changes:*
Added an error dialog
Modified file and pay method to validate NR for a named company
Added tests

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -290,12 +290,12 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
     })
 
     this.$root.$on('name-request-invalid-error', async error => {
-      console.log('NR error during File and Pay =', error)
+      console.log('NR error during File and Pay =', error) // eslint-disable-line no-console
       this.fileAndPayInvalidNameRequestDialog = true
     })
 
     this.$root.$on('name-request-retrieve-error', async () => {
-      console.log('Error while retrieving NR during File and Pay')
+      console.log('Error while retrieving NR during File and Pay') // eslint-disable-line no-console
       this.nameRequestInvalidErrorDialog = true
     })
   }
--- a/src/App.vue
+++ b/src/App.vue
@@ -442,7 +442,10 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
           // TODO: after MVP, add a real check here
           console.log('Filing should have nameRequest object') // eslint-disable-line no-console
         }
-        // if we have a NR number, fetch the NR
+        /** Fetches and validates the NR and sets the data to the store. This method is different
+         * from the validateNameRequest method in Actions.vue. This method sets the data to
+         * the store shows a specific message for different invalid states and redirection is to the
+         * Filings Dashboard */
         if (nameRequest?.nrNumber) {
           await this.processNameRequest(draftFiling)
         }
--- a/src/App.vue
+++ b/src/App.vue
@@ -239,7 +239,8 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
       this.fetchErrorDialog ||
       this.invalidIncorporationApplicationDialog ||
       this.paymentErrorDialog ||
-      this.saveErrorDialog
+      this.saveErrorDialog ||
+      this.fileAndPayInvalidNameRequestDialog
     )
   }
 
@@ -310,7 +311,7 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
   private goToManageBusinessDashboard () : void {
     this.fileAndPayInvalidNameRequestDialog = false
     const manageBusinessUrl = `${sessionStorage.getItem('AUTH_URL')}business`
-    window.onbeforeunload = null
+    this.setHaveChanges(false)
     window.location.assign(manageBusinessUrl)
   }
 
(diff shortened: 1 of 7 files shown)


src/App.vue
```diff
@@ -279,12 +287,31 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
           this.saveErrorDialog = true
       }
     })
+
+    this.$root.$on('name-request-invalid-error', async error => {
+      console.log('NR error during File and Pay =', error)
```
severinbeauvais, line 292: If this console.log message is needed then you'll have to disable the lint error.
Also below.

src/App.vue
```diff
@@ -279,12 +287,31 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
...
     // stop listening for save error event
     this.$root.$off('save-error-event')
+    this.$root.$off('name-request-invalid-errort')
+    this.$root.$off('name-request-retrieve-error')
+  }
+
+  private goToManageBusinessDashboard () : void {
+    this.fileAndPayInvalidNameRequestDialog = false
+    const manageBusinessUrl = `${sessionStorage.getItem('AUTH_URL')}business`
+    window.onbeforeunload = null
```
severinbeauvais, line 313: You should clear the 'haveChanges' flag instead.

src/App.vue
```diff
@@ -207,6 +214,7 @@ export default class App extends Mixins(BcolMixin, DateMixin, FilingTemplateMixi
   private haveData: boolean = false
   private saveErrors: Array<object> = []
   private saveWarnings: Array<object> = []
+  private fileAndPayInvalidNameRequestDialog: boolean = false
```
severinbeauvais, line 217: You may have to add this flag to isErrorDialog() and resetFlags().

src/components/common/Actions.vue
```diff
@@ -207,6 +221,26 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Lega
     }
   }
 
+  /** Fetches NR and validates it. */
+  private async validateNameRequest (nrNumber: string): Promise<void> {
```
severinbeauvais, line 225: I noticed this is similar to processNameRequest() in App.vue. Can we name both functions the same? (I like this function's name better.)

src/App.vue
lekshmimallika-aot: This is not for the custom popup. This is to disable the browsers default behavior

src/components/common/Actions.vue
```diff
@@ -171,6 +176,15 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Lega
       window.scrollTo({ top: 1250, behavior: 'smooth' })
       return
     }
+    // If it is a named company IA, validate NR before filing submission
+    if (this.isNamedBusiness && this.getNameRequestNumber) {
```
severinbeauvais, line 180: You don't need both of these. See how they're declared.

src/components/common/Actions.vue
lekshmimallika-aot: The other one has more things to it like setting the data to the store and all. This is just validating the NR

tests/unit/FileAndPayInvalidNameRequestDialog.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+    expect(wrapper.findAll('p').at(0).text())
+      .toContain('The Name Request NR 1234567 and the Incorporation Application for')
+    expect(wrapper.findAll('p').at(0).text())
+      .toContain('My Name Request Inc. are no longer valid.')
+    expect(wrapper.findAll('p').at(1).text())
+      .toContain('If you still wish to incorporate a Benefit Company, please contact Registry Staff ' +
+        'as soon as possible.')
+    expect(wrapper.findAll('p').at(2).text())
+      .toContain('Registries contact information:')
+    expect(wrapper.findAll('div > div').at(0).text()).toContain('IMPORTANT:')
```
severinbeauvais, line 35: These selectors are fragile and suggest that a classname should be used.

src/components/common/Actions.vue
lekshmimallika-aot: True. Will change it

src/App.vue
lekshmimallika-aot: I am not sure whether I understood this. I dont see any lint error locally and in PR checks

src/App.vue
severinbeauvais: What default behaviour?

src/App.vue
severinbeauvais: In GH for this PR, go to Files Changed tab and look through the code. There are 2 lines with a GHA warning.

src/components/common/Actions.vue
severinbeauvais: I was hoping to reconcile the 2 functions that do something very similar, by naming them the same. (I realize they do slightly different things so they can be merged.) It just doesn't feel very DRY at this point.
How about adding a comment to each function saying it's similar to the other one?

src/App.vue
lekshmimallika-aot: Nvm. I got it

src/App.vue
lekshmimallika-aot: Done

src/App.vue
lekshmimallika-aot: Done

src/components/common/Actions.vue
lekshmimallika-aot: Done

tests/unit/FileAndPayInvalidNameRequestDialog.spec.ts
lekshmimallika-aot: Done

src/App.vue
lekshmimallika-aot: ok

PR #143, opened 2020-06-22
4089 - update office addresses 'same as' logic

ochiu:
*Issue #:* /bcgov/entity#4089

*Description of changes:*
- version increment
- add logic to reset address to default when 'same as' checkboxes unchecked
- added a fix where records delivery address is not properly updated in some cases

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bcrs-business-create-ui",
-  "version": "0.5.0",
+  "version": "0.6.0",
   "private": true,
   "scripts": {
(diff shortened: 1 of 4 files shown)


ochiu:
> Code looks good. Please add a couple of unit tests.

Will do

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bcrs-business-create-ui",
-  "version": "0.5.0",
+  "version": "0.6.0",
```
cameron-freshworks, line 3: Oh nice, thanks!

PR #146, opened 2020-06-23
Stepper update

cameron-freshworks:
*Issue #:* /bcgov/entity#

*Description of changes:*

* Enabled stepper activation on wrapper element

* Added hover and current step selected styling

* Added tabbing through steps for accessibility. 

### DIFF
--- a/src/components/common/Stepper.vue
+++ b/src/components/common/Stepper.vue
@@ -1,7 +1,7 @@
 <template>
   <v-container id="step-buttons-container">
     <template v-for="(step, index) in getSteps">
-      <div class="step" :key="index" @click="goTo(step)">
+      <div class="step" :key="index" @click="goTo(step)" v-on:keyup.tab="goTo(step)">
         <div class="step__indicator">
           <div class="step__line"></div>
           <v-btn
--- a/src/components/common/Stepper.vue
+++ b/src/components/common/Stepper.vue
@@ -28,6 +28,7 @@
 </template>
 
 <script lang="ts">
+// Libraries
 import { Component, Vue } from 'vue-property-decorator'
 import { Getter } from 'vuex-class'
 
--- a/src/components/common/Stepper.vue
+++ b/src/components/common/Stepper.vue
@@ -4,8 +4,16 @@
       <div class="step" :key="index" @click="goTo(step)">
         <div class="step__indicator">
           <div class="step__line"></div>
-          <v-btn class="step__btn" :id=step.id outlined fab color="primary" :disabled=step.disabled :ripple="false" :class="{ 'filled': false }">
-            <v-icon class="step__icon">{{ step.icon }}</v-icon>
+          <v-btn
+            class="step__btn"
+            :id=step.id
+            outlined
+            fab
+            color="primary"
+            :disabled=step.disabled
+            :ripple="false"
+            :class="{ 'selected': isCurrentStep(step) }">
+            <v-icon class="step__icon" :class="{ 'selected': isCurrentStep(step) }">{{ step.icon }}</v-icon>
           </v-btn>
           <v-icon class="step__btn2" size="30" color="green darken-1" v-show=step.valid>
             mdi-check-circle
@@ -20,7 +28,6 @@
 </template>
 
 <script lang="ts">
-// Libraries
 import { Component, Vue } from 'vue-property-decorator'
 import { Getter } from 'vuex-class'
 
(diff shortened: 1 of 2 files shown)


src/components/common/Stepper.vue
```diff
@@ -52,6 +72,24 @@ export default class Stepper extends Vue {
   justify-content: center;
 }
 
+.step:hover {
+  cursor: pointer;
+
+  .step__btn {
+    background-color: #1976d2;
+    color: $BCgovInputBG;
```
severinbeauvais, line 80: There's something not quite right here. If you select the Incorporation Agreement icon (handshake), you will notice that the icon has a square around it in a different hue than the fab button.

src/components/common/Stepper.vue
```diff
@@ -52,6 +72,24 @@ export default class Stepper extends Vue {
...
+  }
+
+  .step__icon {
+    color: $BCgovInputBG;
+  }
+}
+
+.selected {
+  background-color: #1976d2 !important;
+  color: $BCgovInputBG !important;
```
severinbeauvais, line 90: There's something a bit odd here as well. I expect that when I hover over a button, I can tell the difference between that one and the selected one. But in this case, they look the same. Try this and compare with the previous stepper implementation.

src/components/common/Stepper.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <v-container id="step-buttons-container">
     <template v-for="(step, index) in getSteps">
-      <div class="step" :key="index" @click="goTo(step)">
+      <div class="step" :key="index" @click="goTo(step)" v-on:keyup.tab="goTo(step)">
```
ochiu, line 4: Not sure if this was intended, but I was trying out the tabbing and it requires tabbing twice to get to the next step for me.

src/components/common/Stepper.vue
cameron-freshworks: hmmm.. I'll look into this!

src/components/common/Stepper.vue
cameron-freshworks: Definitely! In the other stepper the colour was the same except it had an opacity of .8 on mouseover. Will try that here :)

src/components/common/Stepper.vue
cameron-freshworks: I noticed that too actually, i'll see if there is a way to fix this.

severinbeauvais: PS You can remove elements from the tab order by specifying `tabindex="-1"`.

src/components/common/Stepper.vue
cameron-freshworks: Updated in next commit!

src/components/common/Stepper.vue
cameron-freshworks: Updated in next commit!

src/components/common/Stepper.vue
cameron-freshworks: Thanks for the tip! Updated 👍

cameron-freshworks:
> This is an improvement in click area but there's a bit of weirdness. See if you can figure this out. You're close but for now I prefer the hover/selection look of the original stepper.

I took one more crack at the hover/selection :). That being said, i can 100% set it back to how it was, no questions asked if that is still what you prefer!

src/components/common/Stepper.vue
ochiu: Almost! looks like the check mark icon may also need a -1 tabIndex, when I try to tab through any of the steps that are valid it takes 2 tabs.

src/components/common/Stepper.vue
ochiu: Almost! looks like the check mark icon may also need a -1 tabIndex, when I try to tab through any of the steps that are valid it takes 2 tabs.

cameron-freshworks: Ahh!! haha thanks :)

cameron-freshworks: Hmm... weird. The first tabbing once clicked on the page requires two, but seems to cycle through the steps fine, whether valid or invalid. 🤔  Could you give me a sanity check and try again with a few of the steps valid?

ochiu: You are right, seems to be the first tab once clicked, other tabs work fine whether it is valid or not.

I think the main issue is resolved, if you think you know what is up go ahead and fix it, otherwise it doesn't seem to be a big problem. Resolve this if you are satisfied.

cameron-freshworks: Thanks sir! 

Seems like a universal thing, no matter where you click on the page the first tab just initializes and the second starts to cycle through the page. I didn't really notice until i cycled through every element on the page, when i got back to the stepper it cycled right through each step normally. 
Great observation though, thanks!

severinbeauvais: Tabbing seems fine to me.

severinbeauvais:
> I took one more crack at the hover/selection :). That being said, i can 100% set it back to how it was, no questions asked if that is still what you prefer!

LGTM!

PR #147, opened 2020-06-24
Share Structure name field should not allow numbers

lekshmimallika-aot:
*Issue #:* /bcgov/entity#4054

*Description of changes:*
Added rule for not allowing numbers in the share structure name field
Added a unit test

### DIFF
--- a/src/components/CreateShareStructure/ShareStructure.vue
+++ b/src/components/CreateShareStructure/ShareStructure.vue
@@ -184,13 +184,13 @@ export default class ShareStructure extends Mixins(CurrencyLookupMixin) {
 
   private excludedWordsListForClass: string [] = ['share', 'shares', 'value']
   private excludedWordsListForSeries: string [] = ['share', 'shares']
-
   // Rules
   private getNameRule (): Array<Function> {
     let rules: Array<Function> = [
       v => !!v || 'A name is required',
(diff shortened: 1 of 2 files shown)


lekshmimallika-aot:
> I see this PR rejects numbers in the share series name, but this not quite what the ticket description states...
> 
> `Requirement: Field should allow letters and common symbols such as ( , ) . " /`
> 
> I think the code would be better to _include_ only the valid characters. Check with Linda if it's not clear which symbols are acceptable.

I had checked with Riyaz and he said this ticket is about numbers. And I dont see '-', & etc. I think if we are not sure of the list of the characters to allow, special characters should be left as it is since this is not a person name

lekshmimallika-aot:
> I see this PR rejects numbers in the share series name, but this not quite what the ticket description states...
> 
> `Requirement: Field should allow letters and common symbols such as ( , ) . " /`
> 
> I think the code would be better to _include_ only the valid characters. Check with Linda if it's not clear which symbols are acceptable.

I dont see any such requirement in the document https://docs.google.com/document/d/1bzXWvMCFisgC4Wb4xzrX6pIniVDiPlN0ic2cWcQ2qEk/edit#heading=h.42ifq0li1eyl

severinbeauvais:
I converted this PR to a draft so we don't accidentally merge it. The requirements are still in analysis, but the code changes in this PR might be useful.

PR #149, opened 2020-06-25
4156, 4151

ochiu:
*Issue #:* /bcgov/entity#4156, /bcgov/entity#4151

*Description of changes:*
Fixes and updates to step 1 so form validation errors show when a user resumes an application

*Office Addresses*
- added references to address components (sbc common component BaseAddress) so the form validate() function is accessible
- add checks on mounted to compare the default address state to what is loaded so re-validation can occur to show errors (e.g. non BC province)
- updated formValid() check to fix a bug where the validity for all addresses were not correct

*Business Contact*
- Added a default business contact object for comparison to re-evaluate validation (e.g. invalid email)
- Added check on mount to re-validate form if it is not the same as the default form state

### DIFF
--- a/src/components/DefineCompany/OfficeAddresses.vue
+++ b/src/components/DefineCompany/OfficeAddresses.vue
@@ -90,7 +90,7 @@
               </div>
               <div class="address-wrapper"
                    v-if="!isSame(mailingAddress, deliveryAddress, 'actions') || !inheritMailingAddress">
-                <delivery-address ref='regDeliveryAddress'
+                <delivery-address ref="regDeliveryAddress"
                   id="address-registered-delivery"
                   v-if="!inheritMailingAddress"
                   :address="deliveryAddress"
--- a/src/components/DefineCompany/OfficeAddresses.vue
+++ b/src/components/DefineCompany/OfficeAddresses.vue
@@ -63,7 +63,7 @@
             <label>Mailing Address</label>
             <div class="meta-container__inner">
               <div class="address-wrapper">
-                <mailing-address
+                <mailing-address ref="regMailingAddress"
                   id="address-registered-mailing"
                   :address="mailingAddress"
                   :editing="true"
@@ -90,7 +90,7 @@
               </div>
               <div class="address-wrapper"
                    v-if="!isSame(mailingAddress, deliveryAddress, 'actions') || !inheritMailingAddress">
-                <delivery-address
+                <delivery-address ref='regDeliveryAddress'
                   id="address-registered-delivery"
                   v-if="!inheritMailingAddress"
                   :address="deliveryAddress"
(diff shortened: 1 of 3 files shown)


ochiu:
This ended up being a little tricky as the address components are the common components, so I had to add some references to access the validation. We also didn't want errors to show on a fresh form so for step 1 forms I added some checks against what we consider a default state before re-validating.

The other steps don't seem to be as much as issue since we open an edit form and the user hits done.

src/components/DefineCompany/OfficeAddresses.vue
```diff
@@ -90,7 +90,7 @@
               </div>
               <div class="address-wrapper"
                    v-if="!isSame(mailingAddress, deliveryAddress, 'actions') || !inheritMailingAddress">
-                <delivery-address
+                <delivery-address ref='regDeliveryAddress'
```
severinbeauvais, line 93: Please use consistent (double) quotes here.

ochiu:
> That's actually a pretty clean solution. Well done!
> 
> Obviously we were missing unit tests to verify various scenarios. Could you please add some to validate your changes (as well as catch any future regressions)? Thanks.

I looked as business contacts and it does have invalid type checks. I am thinking it must have to do more with how the whole app life cycle works and when validation is triggered.

Also, the address components are from the sbc-common-components. I can definitely try to dig into this, just not sure if I should invest the time now as it could be a time sink.

severinbeauvais:
> > Obviously we were missing unit tests to verify various scenarios. Could you please add some to validate your changes (as well as catch any future regressions)? Thanks.
> 
> I looked as business contacts and it does have invalid type checks. I am thinking it must have to do more with how the whole app life cycle works and when validation is triggered.
> Also, the address components are from the sbc-common-components. I can definitely try to dig into this, just not sure if I should invest the time now as it could be a time sink.

It's not the BaseAddress component that I want to test, but the OfficeAddresses component... just some simple tests to make sure that validation is done when the component is mounted, and passes/fails according to the page content.

If you're out of time then just add some "TODO" comments in the appropriate unit test file with details on what still needs to be tested (eg, which validation scenarios). Thx.

ochiu:
> > > Obviously we were missing unit tests to verify various scenarios. Could you please add some to validate your changes (as well as catch any future regressions)? Thanks.
> > 
> > 
> > I looked as business contacts and it does have invalid type checks. I am thinking it must have to do more with how the whole app life cycle works and when validation is triggered.
> > Also, the address components are from the sbc-common-components. I can definitely try to dig into this, just not sure if I should invest the time now as it could be a time sink.
> 
> It's not the BaseAddress component that I want to test, but the OfficeAddresses component... just some simple tests to make sure that validation is done when the component is mounted, and passes/fails according to the page content.
> 
> If you're out of time then just add some "TODO" comments in the appropriate unit test file with details on what still needs to be tested (eg, which validation scenarios). Thx.

Ahh okay, makes sense. Added some tests for office addresses.

severinbeauvais:
Love the unit tests - nice and simple and to the point. Thanks!

PR #151, opened 2020-07-08
validated FolioNumber + fixed name length validations

severinbeauvais:
*Issue #:* /bcgov/entity#4061

*Description of changes:*

Commit 1:
- fixed OrgPerson length validation
- added FolioNumber rules
- updated nextTick await code
- updated FolioNumber unit tests
- updated OrgPerson unit tests

Commit 2:
- added missing file error code (found while testing #4155)
- removed extra div around label
- fixed layout issue on review page
- deleted obsolete action reference
- misc code indenting fixes


tests/unit/FolioNumber.spec.ts
```diff
@@ -1,45 +1,98 @@
...
-import { shallowMount } from '@vue/test-utils'
 
 Vue.use(Vuetify)
-
 const vuetify = new Vuetify({})
-const store = getVuexStore()
 
 describe('Folio number component', () => {
   it('displays editing view', () => {
-    store.state.stateModel.nameRequest.entityType = 'BC'
```
severinbeauvais, line 19: Store / state is not needed by this component.

tests/unit/FolioNumber.spec.ts
```diff
@@ -1,45 +1,98 @@
...
-    store.state.stateModel.accountInformation.accountType = 'PREMIUM'
-    const wrapper = shallowMount(FolioNumber, {
```
severinbeauvais, line 21: Deep mount is needed so that v-text-field is fully rendered (instead of just stubbed).

tests/unit/FolioNumber.spec.ts
```diff
@@ -1,45 +1,98 @@
...
+    const wrapper = mount(FolioNumber, {
```
severinbeauvais, line 13: Deep mount is needed so that v-text-field is fully rendered (instead of just stubbed).

src/components/DefineCompany/FolioNumber.vue
```diff
@@ -1,28 +1,28 @@
...
-        </v-flex>
-        <v-flex md4>
-          <div><label><strong>Folio Number</strong></label></div>
-          <div id="lbl-folio-number">{{ !!folioNumber ? folioNumber : 'Not entered' }}</div>
-        </v-flex>
+      <v-flex md4>
+        <label><strong>Folio Information</strong></label>
+      </v-flex>
+      <v-flex md4>
+        <div><label><strong>Folio Number</strong></label></div>
```
cameron-freshworks, line 8: Just curious, does it need to be a <div> if it is a <label>?

src/components/DefineCompany/FolioNumber.vue
```diff
@@ -37,13 +37,19 @@ import { Component, Emit, Prop, Vue, Watch } from 'vue-property-decorator'
...
+  private isEditing: boolean
 
   // Data variables
-  private folioNumber:string = null
+  private folioNumber: string = null
+
+  // Validation rules
+  private readonly rules: Array<Function> = [
+    v => /^[0-9A-Za-z]*$/.test(v) || 'Invalid character', // numbers and letters only
+    v => (!v || v.length <= 50) || 'Cannot exceed 50 characters' // maximum character count
```
cameron-freshworks, line 51: Cool, i'm gonna use this, thanks!

tests/unit/Actions.spec.ts
```diff
@@ -94,10 +94,10 @@ describe('Actions component', () => {
...
     store.state.stateModel.incorporationAgreementStep = { valid: true }
     store.state.stateModel.incorporationDateTime = { valid: true }
-    await Vue.nextTick(() => {
-      // verify File and Pay button state
-      expect(wrapper.find('#file-pay-btn').attributes('disabled')).toBeUndefined()
-    })
+    await Vue.nextTick()
+
+    // verify File and Pay button state
+    expect(wrapper.find('#file-pay-btn').attributes('disabled')).toBeUndefined()
```
cameron-freshworks, line 100: Good change, i found sometimes when wrapping the test in a nextTick callback it would artificially pass the test but throw the errors to the terminal.

tests/unit/FolioNumber.spec.ts
cameron-freshworks: Found this a lot as well. I know Jest recommends tests in isolation ( shallowmount , but it is increasingly difficult to test Vuetify components in a shallowmount 🤷‍♂️ .

src/components/DefineCompany/FolioNumber.vue
severinbeauvais: Nope! Existing code that I simply de-indented. Will fix in next commit.

src/components/DefineCompany/FolioNumber.vue
severinbeauvais: Natch. Just note that it's `!v || v.length <= 50` if v is optional, or `v?.length <= 50` if v is required.

tests/unit/Actions.spec.ts
severinbeauvais: Right. The way around that is to specify `done` as a callback method for the test (instead of `()`), and then you have to call `done()` at the end of the arrow function. But if the test fails you just get "test timed out" instead of the actual expect error, so async is way better.

PR #152, opened 2020-07-13
#2464 UI Name Translations

cameron-freshworks:
*Issue #:* /bcgov/entity#2464

*Description of changes:*

* Implements UI for the handling of Optional Name Translations.

* Adds 2 new sub components - Adding/editing Name Translations & Functional List of Name Translations.

* Added to Name Translations to the summary page, filings and parsing from drafts.

* Implemented Unit tests.

### DIFF
--- a/src/components/DefineCompany/AddNameTranslation.vue
+++ b/src/components/DefineCompany/AddNameTranslation.vue
@@ -16,9 +16,16 @@
       </v-row>
       <v-row>
         <v-col class="form__btns pt-0">
-          <v-btn large color="error" id='btn-remove' disabled>Remove</v-btn>
-          <v-btn large class="form-primary-btn" color="primary" id='btn-done' :disabled="!nameTranslationForm"
-               @click="addTranslation()">Done</v-btn>
+          <v-btn large color="error" id='btn-remove'
+            :disabled="!editNameTranslation"
+            @click="removeTranslation">
+              Remove
+          </v-btn>
+          <v-btn large color="primary" id='btn-done' class="form-primary-btn"
+            :disabled="!nameTranslationForm"
+            @click="addTranslation">
+              Done
+          </v-btn>
           <v-btn large class="form-cancel-btn" id='btn-cancel' @click="cancelTranslation">Cancel</v-btn>
         </v-col>
       </v-row>
@@ -28,7 +35,7 @@
 
 <script lang="ts">
 // Libraries
-import { Component, Emit, Prop, Vue, Watch } from 'vue-property-decorator'
+import { Component, Emit, Prop, Vue } from 'vue-property-decorator'
 
 @Component({})
 export default class NameTranslation extends Vue {
@@ -51,12 +58,15 @@ export default class NameTranslation extends Vue {
 
   // Events
   @Emit('addTranslation')
-  private addTranslation (index = null): string {
+  private addTranslation (): string {
     return this.nameTranslation
   }
 
   @Emit('cancelTranslation')
   private cancelTranslation (): void {}
+
+  @Emit('removeTranslation')
+  private removeTranslation (): void {}
 }
 </script>
 
--- a/src/components/DefineCompany/AddNameTranslation.vue
+++ b/src/components/DefineCompany/AddNameTranslation.vue
@@ -1,0 +1,64 @@
+<template>
+  <div id="name-translation">
+    <!-- Name Translation form -->
+    <v-form v-model="nameTranslationForm" ref="ntForm" name="business-contact-form">
+      <v-row>
+        <v-col class="pb-0">
+          <v-text-field
+            filled
+            persistent-hint
+            label="Enter Name Translation"
+            v-model="nameTranslation"
+            :rules="nameTranslationRules"
+            >
+          </v-text-field>
+        </v-col>
+      </v-row>
+      <v-row>
+        <v-col class="form__btns pt-0">
+          <v-btn large color="error" id='btn-remove' disabled>Remove</v-btn>
+          <v-btn large class="form-primary-btn" color="primary" id='btn-done' :disabled="!nameTranslationForm"
+               @click="addTranslation()">Done</v-btn>
+          <v-btn large class="form-cancel-btn" id='btn-cancel' @click="cancelTranslation">Cancel</v-btn>
+        </v-col>
+      </v-row>
+    </v-form>
+  </div>
+</template>
+
+<script lang="ts">
+// Libraries
+import { Component, Emit, Prop, Vue, Watch } from 'vue-property-decorator'
+
+@Component({})
+export default class NameTranslation extends Vue {
+  @Prop({ default: '' })
+  private editNameTranslation: string
+
+  // Local Properties
+  private nameTranslationForm: boolean = false
+  private nameTranslation: string = ''
+
+  // Validation Rules
+  private readonly nameTranslationRules: Array<Function> = [
+    v => !!v || 'A name is required', // is not empty
+    v => /^[a-z|0-9A-Z]+(?: [a-z|0-9A-Z]+)*$/.test(v) || 'Invalid character', // numbers, letters and single spaces only
+    v => (!v || v.length <= 50) || 'Cannot exceed 50 characters' // maximum character count
+  ]
+  mounted () {
+    if (this.editNameTranslation) this.nameTranslation = this.editNameTranslation
+  }
+
+  // Events
+  @Emit('addTranslation')
+  private addTranslation (index = null): string {
+    return this.nameTranslation
+  }
+
+  @Emit('cancelTranslation')
+  private cancelTranslation (): void {}
+}
+</script>
+
+<style lang="scss" scoped>
+</style>
--- a/src/components/DefineCompany/AddNameTranslation.vue
+++ b/src/components/DefineCompany/AddNameTranslation.vue
@@ -9,8 +9,7 @@
             persistent-hint
             label="Enter Name Translation"
             v-model="nameTranslation"
-            :rules="nameTranslationRules"
-            >
+            :rules="nameTranslationRules">
           </v-text-field>
         </v-col>
       </v-row>
@@ -26,7 +25,10 @@
             @click="addTranslation">
               Done
           </v-btn>
-          <v-btn large class="form-cancel-btn" id='btn-cancel' @click="cancelTranslation">Cancel</v-btn>
+          <v-btn large class="form-cancel-btn" id='btn-cancel'
+            @click="cancelTranslation">
+              Cancel
+          </v-btn>
         </v-col>
       </v-row>
     </v-form>
@@ -52,7 +54,9 @@ export default class NameTranslation extends Vue {
     v => /^[a-z|0-9A-Z]+(?: [a-z|0-9A-Z]+)*$/.test(v) || 'Invalid character', // numbers, letters and single spaces only
     v => (!v || v.length <= 50) || 'Cannot exceed 50 characters' // maximum character count
   ]
+
   mounted () {
+    // Editing an existing name translation
     if (this.editNameTranslation) this.nameTranslation = this.editNameTranslation
   }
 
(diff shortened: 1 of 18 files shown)


src/components/DefineCompany/AddNameTranslation.vue
```diff
@@ -0,0 +1,78 @@
+<template>
+  <div id="name-translation">
+    <!-- Name Translation form -->
+    <v-form v-model="nameTranslationForm" ref="ntForm" name="business-contact-form">
```
lekshmimallika-aot, line 4: form name needs to change?

cameron-freshworks: Good catch, absolutely

src/components/DefineCompany/ListNameTranslations.vue
```diff
@@ -0,0 +1,142 @@
...
+
+// Mixins
+import { CommonMixin, EntityFilterMixin } from '@/mixins'
+
+@Component({
+  components: {
+    ConfirmRemoveDialog
+  }
+})
+export default class ListNameTranslations extends Mixins(CommonMixin, EntityFilterMixin) {
```
lekshmimallika-aot, line 80: Are these mixins needed?

src/components/DefineCompany/AddNameTranslation.vue
cameron-freshworks: Updated in next commit!

src/components/DefineCompany/AddNameTranslation.vue
```diff
@@ -0,0 +1,78 @@
...
+  private editNameTranslation: string
+
+  // Local Properties
+  private nameTranslationForm: boolean = false
+  private nameTranslation: string = ''
+
+  // Validation Rules
+  private readonly nameTranslationRules: Array<Function> = [
+    v => !!v || 'A name is required', // is not empty
+    v => /^[a-z|0-9A-Z]+(?: [a-z|0-9A-Z]+)*$/.test(v) || 'Invalid character', // numbers, letters and single spaces only
```
lekshmimallika-aot, line 54: Does this allow French characters? Also I think numbers are not allowed

src/components/DefineCompany/ListNameTranslations.vue
cameron-freshworks: Nope, they are not, Updated in next commit!

src/components/DefineCompany/AddNameTranslation.vue
cameron-freshworks: I'll look into this, looking at design details now..

cameron-freshworks: Yes to French!
Actually wants to allow special characters, but not numbers as well. 
Will update in next commit!

src/components/DefineCompany/AddNameTranslation.vue
```diff
@@ -0,0 +1,78 @@
...
+            persistent-hint
+            label="Enter Name Translation"
+            v-model="nameTranslation"
+            :rules="nameTranslationRules">
+          </v-text-field>
+        </v-col>
+      </v-row>
+      <v-row>
+        <v-col class="form__btns pt-0">
+          <v-btn large color="error" id='btn-remove'
```
severinbeauvais, line 18: Watch for consistent use of quotes. Typically we use " in the template and ' in the script.

src/components/DefineCompany/AddNameTranslation.vue
```diff
@@ -0,0 +1,78 @@
...
+    <!-- Name Translation form -->
+    <v-form v-model="nameTranslationForm" ref="ntForm" name="name-translation-form">
+      <v-row>
+        <v-col class="pb-0">
+          <v-text-field
+            filled
+            persistent-hint
+            label="Enter Name Translation"
+            v-model="nameTranslation"
+            :rules="nameTranslationRules">
```
severinbeauvais, line 12: Would be good practise to give this text field an id (or class), for testing at least.

src/components/DefineCompany/AddNameTranslation.vue
```diff
@@ -0,0 +1,78 @@
+<template>
+  <div id="name-translation">
+    <!-- Name Translation form -->
+    <v-form v-model="nameTranslationForm" ref="ntForm" name="name-translation-form">
```
severinbeauvais, line 4: You don't need the ref unless you're going to reference the form (eg, for validation or reset).

src/components/DefineCompany/ListNameTranslations.vue
```diff
@@ -0,0 +1,139 @@
...
+<script lang="ts">
+// Libraries
+import { Component, Prop, Vue, Emit } from 'vue-property-decorator'
+
+// Dialogs
+import { ConfirmRemoveDialog } from '@/components/dialogs'
+
+@Component({
+  components: {
+    ConfirmRemoveDialog
```
severinbeauvais, line 74: I don't see where this component is used...?

src/components/Summary/SummaryDefineCompany.vue
```diff
@@ -27,6 +27,14 @@
           </div>
         </v-flex>
       </v-layout>
+      <v-layout row v-if="getNameTranslations && getNameTranslations.length" class="mt-3">
```
severinbeauvais, line 30: Would optional chaining work here?

src/components/common/NameRequestInfo.vue
```diff
@@ -97,6 +179,14 @@ export default class NameRequestInfo extends Mixins(DateMixin) {
   @Getter getNameRequestNumber!: GetterIF;
   @Getter getNameRequestDetails!: NameRequestDetailsIF;
   @Getter getNameRequestApplicant!: NameRequestApplicantIF;
+  @Getter getNameTranslations!: Array<string>;
+
+  mounted () {
+    if (this.nameTranslations?.length) {
```
severinbeauvais, line 185: I suggest being explicit here and checking that length > 0. (Instead of checking that length is truthy.)

src/components/DefineCompany/AddNameTranslation.vue
cameron-freshworks: oops, good catch thanks! Updated in next commit

src/components/DefineCompany/AddNameTranslation.vue
cameron-freshworks: You read my mind, implemented while i was writing the tests :) Will be in next commit!

src/components/DefineCompany/AddNameTranslation.vue
cameron-freshworks: ahh thanks, will remove!

src/components/common/NameRequestInfo.vue
```diff
@@ -61,34 +62,115 @@
...
+  }
+
   // Template Enums
   readonly NameRequestStates = NameRequestStates
 
+  // Local Properties
+  private hasNameTranslation = false
+  private isAddingNameTranslation = true
+  private editingNameTranslation = ''
+  private editIndex: number
```
severinbeauvais, line 160: Should this be initialized? As-is it's undefined, but maybe -1 is a better initial value? Just wondering.

src/components/DefineCompany/ListNameTranslations.vue
cameron-freshworks: It's copypasta error 😓. Removed in next commit!

src/components/Summary/SummaryDefineCompany.vue
cameron-freshworks: Optional chaining doesn't seem to work in the template but i will dig into this a bit more and see what the options are!

src/components/common/NameRequestInfo.vue
cameron-freshworks: Absolutely, will update this in the places i checked against `length`

src/components/common/NameRequestInfo.vue
```diff
@@ -196,4 +383,13 @@ ul.numbered-company-list-items {
     margin-left: -1.5em;
   }
 }
+
+#name-translation-info {
+  display: flex;
+  align-items: baseline;
+
+  .v-input--selection-controls .v-input__slot > .v-label, .v-input--selection-controls .v-radio > .v-label {
```
severinbeauvais, line 391: To make this easier to read, I suggest putting the 2 expression on separate lines :)

Also I wonder if an id or class on the label would be cleaner?

src/components/common/NameRequestInfo.vue
cameron-freshworks: That makes sense for sure. I set it -1 before its used anyways, might as well initialize it!

src/interfaces/store-interfaces/state-interface.ts
```diff
@@ -6,6 +6,7 @@ export interface StateModelIF {
   tombstone: TombStoneIF
   accountInformation: AccountInformationIF
   nameRequest: NameRequestIF
+  nameTranslations?: Array<string>
```
severinbeauvais, line 9: This always exists in state, doesn't it? So doesn't need to be optional.

tests/unit/Actions.spec.ts
```diff
@@ -180,7 +180,7 @@ describe('Emits error event if NR validation fails in file and pay', () => {
   })
 })
 
-describe('Actions component - Filing Functionality', () => {
+describe.only('Actions component - Filing Functionality', () => {
```
severinbeauvais, line 183: Don't commit this.

src/components/common/NameRequestInfo.vue
severinbeauvais: Great. Both codes will work, but one will be clearer. (Great code is easy to review ;) )

src/components/Summary/SummaryDefineCompany.vue
severinbeauvais: OK, I was unsure. Might be because template code is JS while script is TS.

Not worth spending a lot of time on this.

src/components/common/NameRequestInfo.vue
cameron-freshworks: 😆 Sorry, i forgot to update this. I was testing to find the element, will update this!

src/interfaces/store-interfaces/state-interface.ts
cameron-freshworks: ah, thanks! Will update in next commit

src/components/DefineCompany/AddNameTranslation.vue
severinbeauvais: Yeah. You only need a ref to reference the form in script, like this:
`this.$refs.addPersonOrgForm.reset()`
or this
`this.$refs.dateTimeForm.validate()`

tests/unit/Actions.spec.ts
cameron-freshworks: You mean i can't get all the unit tests to pass by not running them!? Fine....

Will fix this in next commit 😆

src/components/common/NameRequestInfo.vue
severinbeauvais: If you're wondering, as I scan/review code, this line jumps out :)

Basically you write great code though -- easy to understand and review! Well done 👍

src/components/DefineCompany/AddNameTranslation.vue
cameron-freshworks: Oh nice, yea that seems really helpful when required!

src/components/common/NameRequestInfo.vue
cameron-freshworks: haha... yea, it certainly does 😆 

Thanks a lot!

src/components/DefineCompany/AddNameTranslation.vue
severinbeauvais: I don't think this regex works for French characters. (I tested it with Séverin at https://regex101.com/).

Ref: http://www.asciitable.com/

cameron-freshworks: You bet, this is an outdated commit. It has been updated in the latest to allow French chars :)

PR #175, opened 2020-10-27
updated common component dependency

severinbeauvais:
*Issue #:* /bcgov/entity#5244

*Description of changes:*
- updated sbc-common-components to 2.4.2
- updated version to 1.4.0
- updated cached AddressComplete files
- fixed app title
- fixed exception when there is no user contact/email
- fixed errors when running locally with no LD key


public/config/configuration.json
```diff
@@ -7,7 +7,7 @@
   "SBC_CONFIG_AUTH_API_URL": "https://auth-api-dev.pathfinder.gov.bc.ca/api/v1",
   "PAY_API_URL": "https://pay-api-dev.pathfinder.gov.bc.ca/api/v1/",
   "KEYCLOAK_CONFIG_PATH": "/local-keycloak-config-url/keycloak.json",
-  "ADDRESS_COMPLETE_KEY": "_secret_",
-  "LD_CLIENT_ID": "_secret_",
+  "ADDRESS_COMPLETE_KEY": "",
```
severinbeauvais, line 10: It was previously trying to load AddressComplete with a bum key.

public/index.html
```diff
@@ -4,15 +4,18 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
-    <title>Incorporation Application</title>
+    <title>BC Business Registry</title>
```
severinbeauvais, line 7: The user experience should be the same between our apps (they should not know they're switching apps).

src/main.ts
```diff
@@ -50,7 +50,9 @@ async function start () {
   })
 
   // initialize Launch Darkly
-  await initLDClient()
+  if (window['ldClientId']) {
```
severinbeauvais, line 53: This is to avoid console errors if there is no LD key.

severinbeauvais: (Same code as Filings UI.)

vysakh-menon-aot (approved):

PR #176, opened 2020-11-02
misc cleanup

severinbeauvais:
*Issue #:* None

*Description of changes:*
- removed unneeded icon font preload
- set section icons to blue
- updated isSame() same as other projects
- delete obsolete common mixin methods
- updated package versions to match lock file
- updated checkmarks to the designed colour


package.json
```diff
@@ -12,57 +12,57 @@
     "test:debug": "node --inspect-brk node_modules/.bin/vue-cli-service test:unit --testPathPattern --no-cache --watch --runInBand"
   },
   "dependencies": {
-    "@babel/compat-data": "^7.8.6",
-    "@mdi/font": "^5.0.45",
-    "@sentry/browser": "^5.15.4",
-    "@sentry/integrations": "^5.15.4",
+    "@babel/compat-data": "^7.11.0",
```
severinbeauvais, line 15: These are the versions actually used (per the lock file).
Updating these has no impact.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -12,7 +12,7 @@
     <div id="people-roles-summary" v-if="isSummary">
       <!-- Summary Header -->
       <div class="people-roles-summary-header" >
-        <v-icon>mdi-account-multiple-plus</v-icon>
+        <v-icon color="#38598A">mdi-account-multiple-plus</v-icon>
```
severinbeauvais, line 15: Same as Edit UI.

lekshmimallika-aot (approved):
Nice!

src/views/CreateShareStructure.vue
```diff
@@ -45,8 +45,8 @@
     <p>Your application must include the following:</p>
     <ul>
       <li>
-        <v-icon color="blue" v-if="shareClasses.length > 0" class="cp-valid">mdi-check</v-icon>
-        <span v-else><v-icon color="red" v-if="showErrors" class="cp-invalid">mdi-close</v-icon></span>
+        <v-icon v-if="shareClasses.length > 0" color="green darken-2" class="cp-valid">mdi-check</v-icon>
+        <v-icon v-else :color="showErrors ? 'red': 'transparent'" class="cp-invalid">mdi-close</v-icon>
```
severinbeauvais, line 49: This is the same fix I did for People and Roles so the icon takes up space but doesn't display anything.

src/components/IncorporationAgreement/AgreementType.vue
```diff
@@ -4,27 +4,27 @@
       <v-card flat>
         <!-- Summary Header -->
         <div class="agreement-summary-header">
-            <v-icon>mdi-handshake</v-icon>
-            <label class="agreement-summary-title">
-                <strong>Incorporation Agreement and Benefit Company Articles</strong>
-            </label>
+          <v-icon color="#38598A">mdi-handshake</v-icon>
```
cameron-freshworks, line 7: If we were to set the icon colour in the base style for v-icons would it inherit throughout? 
I know Vuetify was tricky sometimes about inheritance, if color was a prop for instance.

cameron-freshworks (approved):
Looks good, just one question. Merge away when ready 👍

src/components/IncorporationAgreement/AgreementType.vue
severinbeauvais: That's clever.

There are many interesting things on the following page but I don't see any about the icons :/  Also we've been using some v-icons with conditional colours, which we would have to keep doing.

https://vuetifyjs.com/en/features/theme/#customizing

However, I am aware of some "style consistency" tickets in the backlog. When we work on them, we should look into this a bit deeper. If the Vuetify theme won't help us directly, I think we should consider a common set of classes, eg "v-icon-check-ok", "v-icon-header", "v-btn-remove", etc.

I can only find a couple of these "style" tickets atm (#5368, #5470) so we'll have to remember to do this ^^ when we work on them.

PR #197, opened 2021-04-15
Update CICD flow.

pwei1018:
*Issue #:* /bcgov/entity###

*Description of changes:*
1. Migrated the environments to 1password;
2. Modify CD flow on makefile;
3. Modify CI/CD GHA scripts;

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -20,26 +20,32 @@ jobs:
       - name: Set env by input
         run: |
           echo "TAG_NAME=${{ github.event.inputs.environement }}" >> $GITHUB_ENV
-
       - name: Set env by main branch
         if: env.TAG_NAME == '' && github.ref == 'refs/heads/main'
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/cd.yml
```diff
@@ -12,34 +12,40 @@ on:
         default: "dev"
 
 jobs:
-  business-create-ui-cd:
+  business-edit-ui-cd:
```
severinbeauvais, line 15: This is the business-create-ui repo, so this looks wrong.

severinbeauvais (commented):
Looks fine.

What do you (and other reviewers) think about having a "prompt" in the Prod deployment script, as a second line of defence in case of running the Prod deploy pipeline accidentally? (Previously we had a prompt in some scripts, "input required", which we to accept in the Jenkins log.)

severinbeauvais (approved):

thorwolpert (approved):

PR #200, opened 2021-05-12
Added validation for number of shares = 0

severinbeauvais:
*Issue #:* /bcgov/entity#7505

*Description of changes:*
- added rule for required number of shares
- added rule for number of shares > 0
- added unit tests
- updated app version

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "1.5.7",
+  "version": "1.5.8",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


severinbeauvais:
This implementation is similar to the one in bcrs-shared-components but not identical, so I applied the same fix here.

In the future, we should update this UI to use the shared component... I will create a new ticket.

cameron-freshworks (approved):
✅ ✅ ✅ ✅ ✅ ✅ ✅

cameron-freshworks:
> This implementation is similar to the one in bcrs-shared-components but not identical, so I applied the same fix here.
> 
> In the future, we should update this UI to use the shared component... I will create a new ticket.

I believe there is a ticket to import the shared component already, will most likely come in with the new Create Filings?

severinbeauvais:
> I believe there is a ticket to import the shared component already, will most likely come in with the new Create Filings?

Yup, thanks for the hint. I found:

https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/7397

PR #201, opened 2021-05-13
Prevent too large a number of shares

severinbeauvais:
*Issue #:* /bcgov/entity#7488

*Description of changes:*
- added shares max digits rule (as per db schema)
- updated unit tests
- updated app version

### DIFF
--- a/src/components/CreateShareStructure/ShareStructure.vue
+++ b/src/components/CreateShareStructure/ShareStructure.vue
@@ -217,9 +217,10 @@ export default class ShareStructure extends Mixins(CurrencyLookupMixin) {
     let rules: Array<Function> = []
     if (!this.hasNoMaximumShares) {
       rules = [
-        (v: string) => (v !== null && v !== '') || 'Number of shares is required',
+        (v: string) => !!v || 'Number of shares is required',
         (v: string) => /^-?\d+$/.test(v) || 'Must be a whole number',
-        (v: string) => +v > 0 || 'Number must be greater than 0'
+        (v: string) => (+v > 0) || 'Number must be greater than 0',
+        (v: string) => (v && v.toString().length < 16) || 'Number must be less than 16 digits'
       ]
       // To prevent changing share class value to a lower value after adding series.
       if (this.isClass && this.activeIndex !== -1 && !this.hasNoMaximumShares &&
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "1.5.8",
+  "version": "1.5.9",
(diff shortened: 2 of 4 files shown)


src/components/CreateShareStructure/ShareStructure.vue
```diff
@@ -217,9 +217,10 @@ export default class ShareStructure extends Mixins(CurrencyLookupMixin) {
     let rules: Array<Function> = []
     if (!this.hasNoMaximumShares) {
       rules = [
-        (v: string) => (v !== null && v !== '') || 'Number of shares is required',
+        (v: string) => !!v || 'Number of shares is required',
```
severinbeauvais, line 220: It turns out `!!v` works fine if v is a string (ie, "0" is not falsey).

src/components/CreateShareStructure/ShareStructure.vue
```diff
@@ -217,9 +217,10 @@ export default class ShareStructure extends Mixins(CurrencyLookupMixin) {
...
         (v: string) => /^-?\d+$/.test(v) || 'Must be a whole number',
-        (v: string) => +v > 0 || 'Number must be greater than 0'
+        (v: string) => (+v > 0) || 'Number must be greater than 0',
+        (v: string) => (v && v.toString().length < 16) || 'Number must be less than 16 digits'
```
severinbeauvais, line 223: I had to cast v to a string as we sometimes set the value to a number in the unit tests.

cameron-freshworks (approved):
Looks good sir! 
Is this something already implemented in the Shared Components? 
Any concerns of this component being out of sync with the other?

andrepestana-aot (approved):
LGTM!

severinbeauvais:
> Looks good sir!
> Is this something already implemented in the Shared Components?
> Any concerns of this component being out of sync with the other?

I will implement this in the shared component next. (Yesterday I did it the other way around, but this way is easier as I have an app I can play in.)

This component is mutually exclusive with the one in the shared library. Actually they do the same thing but are not a drop-in replacement for each other... so duplication today, /bcgov/entity#7397 tomorrow!

cameron-freshworks:
> > Looks good sir!
> > Is this something already implemented in the Shared Components?
> > Any concerns of this component being out of sync with the other?
> 
> I will implement this in the shared component next. (Yesterday I did it the other way around, but this way is easier as I have an app I can play in.)
> 
> This component is mutually exclusive with the one in the shared library. Actually they do the same thing but are not a drop-in replacement for each other... so duplication today, /[bcgov/entity#7397](https://github.com/bcgov/entity/issues/7397) tomorrow!

Sounds good 😁

PR #203, opened 2021-05-20
Added flag to display filings filed by Registry Staff

severinbeauvais:
*Issue #:* /bcgov/entity#6814

*Description of changes:*
- added isStaff flag to IA filing
- updated interface accordingly
- updated unit test
- updated app patch release

_Yes, I know IAs are not filed by staff, but this is both future-proofing this app and also sets the flag correctly for this filing type  (ie, False) even though it's optional in the Filings History List._

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "1.5.9",
+  "version": "1.5.10",
   "lockfileVersion": 1,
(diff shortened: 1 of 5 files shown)


severinbeauvais:
Converted to Draft as the implementation is under discussion.

severinbeauvais:
The API will set the role on filing. Close this PR as not needed.

PR #204, opened 2021-05-26
Initial Commit: Store update and resource config refactor

cameron-freshworks:
*Issue #:* /bcgov/entity#6953

*Description of changes:*

* Store refactor.
* Created configuration file.
* Removed/refactored older resources after integrating them into config object.
* refactored component usage of resources, where applicable.
* Updated / parameterize unit tests, see working examples.
* Apply steps/components to configuration.


src/interfaces/resource-interfaces/resource-interface.ts
```diff
@@ -1,6 +1,25 @@
...
+    countMinimum: number
+  }
+  incorporationAgreement: Array<IncorporationAgreementTypeIF>,
+  reviewAndConfirm: {
+    completingPartyStatement: {
+      certifyStatementHeader: string
+      certifyStatements: Array<string>
+      certifyClause: string
+    }
+  }
```
severinbeauvais, line 24: OK, yes I like that all differences are identified here and none are optional. This will force us to specify all these properties in the resource files, which will avoid inherited/default resources, which I think is a good thing.

For properties that are not used in a specific resource file, I think we should assign it a null/0 value and add a comment that it's not used for the subject entity type. I'll add an example below.

src/resources/companyResources.ts
```diff
@@ -0,0 +1,254 @@
+import { ResourceIF } from '@/interfaces'
```
severinbeauvais, line 1: How about having this file importing sub-files for all different entity types?

src/resources/companyResources.ts
```diff
@@ -0,0 +1,254 @@
...
+  {
+    entityType: 'BEN',
+    displayName: 'BC Benefit Company',
+    title: 'Benefit Company Statement',
+    description: `This company is a benefit company and, as such, has purposes that include conducting its business
+          in a responsible and sustainable manner and promoting one or more public benefits.`,
+    statement: '',
+    nameRequestType: 'BC',
+    directors: {
+      countMinimum: 1
```
severinbeauvais, line 13: Example: if `countMinimum` is not used for this entity type then we write this:
```
   countMinimum: 0 // not used by this entity type
```

src/interfaces/resource-interfaces/resource-interface.ts
```diff
@@ -1,6 +1,25 @@
-import { CertifyStatementIF } from '@/interfaces'
-
 // Interface to define the resource model example
+import { IncorporationAgreementTypeIF } from '@/interfaces'
```
severinbeauvais, line 2: Do you think the steps should be imported here as well?

src/resources/companyResources.ts
```diff
@@ -0,0 +1,254 @@
+import { ResourceIF } from '@/interfaces'
+
+export const CompanyResources: Array<ResourceIF> = [
+  {
+    entityType: 'BEN',
```
severinbeauvais, line 5: Should use an enum :)

Also maybe we can fetch the displayname below from shared corp-type-module?

src/resources/companyResources.ts
severinbeauvais: Yes, I'd like to see a simpler declaration here (ie, hide details in the imports).

src/components/IncorporationAgreement/AgreementType.vue
```diff
@@ -24,15 +24,15 @@
         <div v-else class="summary-desc">
           <div><v-icon color="green darken-2" class="agreement-valid-icon">mdi-check</v-icon></div>
             <div>
-              {{ selectedAgreementDescription }}
+              {{ getIncorporationAgreement.description }}
```
severinbeauvais, line 27: OK, I see you're reusing this component for the ULC/LTD/CCC companies...

We need to be careful how these components are organized (folder-wise) in this app. They're currently in folders according to step name, but those steps will change for different companies...
1. I think we should organize these components by component name instead (right now they might match so no change required?)
2. I think we should flatten unnecessary folders, for example any step folder with only 1 component in it

I'm not exactly sure what future changes we will need to make... I am trying to avoid painting ourselves into a corner where a component is in a folder that doesn't make sense.

src/App.vue
```diff
@@ -448,7 +448,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
         this.initEntityFees()
 
         // Set the resources
-        this.setCertifyStatementResource(CertifyStatementResource.find(x => x.entityType === this.entityType))
+        this.setCompanyResources(CompanyResources.find(x => x.entityType === this.entityType))
       } catch (error) {
```
severinbeauvais, line 452: Make a note to review the error handling here. It looks gross (might be my old code), and I'm not certain if it works correctly if, say, an unknown entity type is used.

src/store/index.ts
```diff
@@ -5,37 +5,10 @@ import Vuex from 'vuex'
...
-  setBusinessContact, setDefineCompanyStepValidity, setNameRequestState, setFilingId, setFolioNumber,
-  setOfficeAddresses, setOrgPersonList, setAddPeopleAndRoleStepValidity, setShareClasses,
-  setCreateShareStructureStepValidity, setIgnoreChanges, setHaveChanges, setIsFutureEffective,
-  setEffectiveDate, setIsIncorporationDateTimeValid, setAccountInformation, setTempId, setEntityType,
-  setIncorporationAgreementStepData, setNameTranslationState
-} from './actions'
+// Store modules
+import * as Getters from './getters'
+import * as Mutations from './mutations'
+import * as Actions from './actions'
```
severinbeauvais, line 11: Oh thank God for this :)

src/store/state/resource-model.ts
```diff
@@ -1,5 +1,24 @@
 import { ResourceIF } from '@/interfaces'
 
 export const resourceModel: ResourceIF = {
-  certifyStatementResource: null
+  entityType: '',
```
severinbeauvais, line 4: I know '' is falsey just as null is for a string, but do you think we should assign null to these empty strings to really denote that they have no value?

tests/unit/Certify.spec.ts
```diff
@@ -70,8 +70,6 @@ function createComponent (
   certifiedBy: string | undefined = undefined,
   date: string = defaultDate,
   certifyStatementResource: CertifyStatementIF = {
```
severinbeauvais, line 72: Type doesn't exist any more.

The fields below weren't used right?

src/components/IncorporationAgreement/AgreementType.vue
severinbeauvais: (Also need to ensure state isn't organized by step number. Maybe it already isn't -- I haven't checked -- busy with 1M things atm, sorry.)

src/interfaces/resource-interfaces/resource-interface.ts
cameron-freshworks: Will do for sure!

src/resources/companyResources.ts
cameron-freshworks: For sure, will break up each entity!

src/interfaces/resource-interfaces/resource-interface.ts
cameron-freshworks: You bet, as soon as we can confidently identify changes to the steps.

src/resources/companyResources.ts
cameron-freshworks: Absolutely!

src/components/IncorporationAgreement/AgreementType.vue
cameron-freshworks: For sure, i don't disagree here. This seems like it falls into the realm of the store refactor that we discussed earlier though doesn't it? 
- Refactor the store and component directories to be separate of the steps, would fall into the framework no?

The current iteration of these resources are separate of the steps within the store 👍

src/App.vue
cameron-freshworks: Ok, will do 👍

src/store/index.ts
cameron-freshworks: Yesss!

src/store/state/resource-model.ts
cameron-freshworks: Totally, makes sense!

tests/unit/Certify.spec.ts
cameron-freshworks: Sorry, no it doesn't. Will be working in the unit tests more wholly when the configuration is complete.

src/resources/companyResources.ts
cameron-freshworks: In this scenario we don't have any rules that aren't applicable to the current entities we are trying to set a baseline for, but will 100% follow this approach when we do!

After looking at your RFC, i like the idea to apply the rules within here as well! Will do that 👍

src/resources/companyResources.ts
cameron-freshworks: Updated!

src/App.vue
cameron-freshworks: Put in some error handling 👍

src/store/state/resource-model.ts
cameron-freshworks: Updated 👍

src/enums/nameRequestTypes.ts
```diff
@@ -0,0 +1,6 @@
+export enum NameRequestTypes {
+  BENEFIT_COMPANY = 'BC',
+  BC_COMPANY = 'CR',
+  BC_ULC_COMPANY = 'UL',
+  BC_CCC = 'CC',
+}
```
severinbeauvais, line 6: Can we sync this list with the corp type codes (from shared corp-type-module)?

For example,
```
   BENEFIT_COMPANY = CorpTypeCd.BC_COMPANY,
   BC_COMPANY = CorpTypeCd.BC_CORPORATION,
   etc
```

Ref: https://github.com/bcgov/bcrs-shared-components/blob/972765945d09d93368753987ea697f62cfef31ff/src/modules/corp-type-module/corp-type-module.ts#L9

src/resources/index.ts
```diff
@@ -1,4 +1,14 @@
...
-export * from './certifyStatementResource'
-export * from './agreementTypeResource'
-export * from './benefitCompanyStatementResource'
+export const CompanyResources: Array<ResourceIF> = [
+  BenefitCompanyResource,
+  LimitedCompanyResource,
+  UnlimitedCompanyResource,
+  CommunityContributionCompanyResource,
+  CooperativeResource
+]
```
severinbeauvais, line 14: Looks good!

src/enums/nameRequestTypes.ts
cameron-freshworks: I'm not sure we can, for all of them at least. Some of the codes don't exist in our `corp-type-module`, `UL` for example.

If you prefer, we can for the ones tho DO exist in both enums?

severinbeauvais: I think (but not 100%) that, in the future, these will either all match, or should all have an equivalent.

Can you check with Thor about how we'll handle NR/corp type differences?

I'm actually trying to reduce confusion 😛 I think we should only have 1 type list, but we're not there yet and this change would be a step in that direction.

cameron-freshworks: For sure, will inquire 100%. 
That being said, if the types were in Sync we would have no need for this enum but until then, we cannot do this here.

severinbeauvais: I believe that the CorpTypeCd enum contains all entity types as well as all NR types.

What do you mean, "cannot do this here"?

I'm hoping you can keep NameRequestTypes here but show (as per my example above) how they sync with the CorpTypeCd.

src/resources/unlimitedCompany.ts
```diff
@@ -0,0 +1,91 @@
...
+import { NameRequestTypes, RouteNames } from '@/enums'
+import { CorpTypeCd, GetCorpFullDescription } from '@bcrs-shared-components/corp-type-module'
+
+export const UnlimitedCompanyResource: ResourceIF = {
+  entityType: CorpTypeCd.BC_ULC_COMPANY,
+  displayName: GetCorpFullDescription(CorpTypeCd.BC_ULC_COMPANY),
+  title: 'Unlimited Company Statement',
+  description: null,
+  statement: null,
+  nameRequestType: NameRequestTypes.CC,
```
cameron-freshworks, line 11: This is a mistake. Should be `UL` have updated locally, will be in next commit.

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -248,9 +255,11 @@ export default class PeopleAndRoles extends Mixins(EntityFilterMixin) {
     const numOfPeopleWithNoRoles = this.orgPersonList.filter(people => people.roles.length === 0).length
 
     if (this.entityFilter(EntityTypes.BCOMP)) {
-      return numOfCompletingParty === 1 && numOfIncorporator >= 1 && numOfDirector >= 1 && numOfPeopleWithNoRoles === 0
+      return numOfCompletingParty === 1 && numOfIncorporator >= 1 && numOfDirector >= this.getMinimumDirectorCount &&
+        numOfPeopleWithNoRoles === 0
     } else if (this.entityFilter(EntityTypes.COOP)) {
-      return numOfCompletingParty === 1 && numOfIncorporator >= 3 && numOfDirector >= 3 && numOfPeopleWithNoRoles === 0
+      return numOfCompletingParty === 1 && numOfIncorporator >= 3 && numOfDirector >= this.getMinimumDirectorCount &&
+        numOfPeopleWithNoRoles === 0
```
severinbeauvais, line 262: I am in favour of these rules being here as this component is (for now) specific to BEN/ULC/etc businesses.

I am thinking that we would have a different version of this component for future, significantly-different businesses.

For now this is good 👍

src/components/common/Stepper.vue
```diff
@@ -10,12 +10,12 @@
...
+            :disabled=isBusySaving
             :ripple="false"
             :class="{ 'selected-btn': isCurrentStep(step) }">
             <v-icon class="step__icon" :class="{ 'selected-icon': isCurrentStep(step) }">{{ step.icon }}</v-icon>
           </v-btn>
-          <v-icon class="step__btn2" size="30" color="green darken-1" v-show=step.valid>
+          <v-icon class="step__btn2" size="30" color="green darken-1" v-show=isValid(step.to)>
             mdi-check-circle
           </v-icon>
         </div>
```
severinbeauvais, line 21: While you're here, you might want to implement "isInvalid" (with a X circle icon).

src/components/common/Stepper.vue
```diff
@@ -33,11 +33,40 @@ import { Component, Vue } from 'vue-property-decorator'
...
+
+      case RouteNames.INCORPORATION_AGREEMENT:
+        return this.isIncorporationAgreementValid
+
+      case RouteNames.REVIEW_CONFIRM:
+        return this.isApplicationValid
+
+      default:
+        return false
+    }
```
severinbeauvais, line 68: So this is an abstraction to handle any possible future steps/pages? Sounds good.

src/enums/nameRequestTypes.ts
```diff
@@ -0,0 +1,11 @@
+import { CorpTypeCd } from '@bcrs-shared-components/corp-type-module'
+
+/** Enum to show that NR types are a subset of corp types. */
+/** Enum to map NR types to corp types. */
+/** Enum to sync NR types with corp types. */
+export enum NameRequestTypes {
```
severinbeauvais, line 6: Haha, pick just one comment :)

src/resources/benefitCompany.ts
```diff
@@ -0,0 +1,100 @@
...
+  statement: null,
+  nameRequestType: NameRequestTypes.BC,
+  steps: [
+    {
+      id: 'step-1-btn',
+      step: 1,
+      icon: 'mdi-domain',
+      text: 'Define Your\nCompany',
+      to: RouteNames.DEFINE_COMPANY,
+      component: 'define-company'
```
severinbeauvais, line 20: I know these are strings. Where do we get these? From an import? Should we have an enum? 🤷

src/resources/communityContributionCompany.ts
```diff
@@ -0,0 +1,91 @@
...
+import { CorpTypeCd, GetCorpFullDescription } from '@bcrs-shared-components/corp-type-module'
+
+export const CommunityContributionCompanyResource: ResourceIF = {
+  entityType: CorpTypeCd.BC_CCC,
+  displayName: GetCorpFullDescription(CorpTypeCd.BC_CCC),
+  title: 'Community Contribution Company Statement',
+  description: null,
+  statement: null,
+  nameRequestType: NameRequestTypes.CC,
+  steps: [
```
severinbeauvais, line 12: What do you think of extracting these steps into a common file used by BEN/ULC/etc ?

Is it better to be verbose/explicit here, or is it better to keep common things together?

src/components/common/Stepper.vue
cameron-freshworks: Was my thoughts as well! 
Although we currently only have two states `true/false` so would need to apply this validation at some point prompted by the user to follow our current pattern, surely not display an X all the time eh?

src/resources/unlimitedCompany.ts
severinbeauvais, line 11: Should be UL.

src/components/common/Stepper.vue
cameron-freshworks: This is just for the purposes of validation on each step but yessir you betcha. 
All the rest is contained within the configuration but i think validation will need to be something computed ( obviously. )

src/enums/nameRequestTypes.ts
cameron-freshworks: Can't we have them all?!

src/resources/limitedCompany.ts
```diff
@@ -0,0 +1,91 @@
...
+import { NameRequestTypes, RouteNames } from '@/enums'
+import { CorpTypeCd, GetCorpFullDescription } from '@bcrs-shared-components/corp-type-module'
+
+export const LimitedCompanyResource: ResourceIF = {
+  entityType: CorpTypeCd.BC_COMPANY,
+  displayName: GetCorpFullDescription(CorpTypeCd.BC_COMPANY),
+  title: 'Limited Company Statement',
+  description: null,
+  statement: null,
+  nameRequestType: NameRequestTypes.CR,
```
severinbeauvais, line 11: CR? Or BC? 🤷

src/resources/benefitCompany.ts
cameron-freshworks: Probably should! I don't mind doing it now.

src/resources/communityContributionCompany.ts
cameron-freshworks: I considered templating them but wasn't sure. 

If we can say this is default template then so be it, but i wasn't really sure what the future holds, so i just defined them explicitly on the entity config, in the event something does change during development here.

src/store/getters/state-getters.ts
```diff
@@ -190,6 +194,26 @@ export const isBusySaving = (state: any): boolean => {
...
+}
+
+/** Is true when the step is valid. */
+export const isCreateShareStructureValid = (state: any): boolean => {
+  return state.stateModel.createShareStructureStep.valid
+}
+
+/** Is true when the step is valid. */
+export const isIncorporationAgreementValid = (state: any): boolean => {
+  return state.stateModel.incorporationAgreementStep.valid
```
severinbeauvais, line 214: Can we change the store IF to this?

```
stateModel: {
   validFlags: {
      incorporationAgreementStep: boolean,
      createShareStructureSte: boolean,
      etc
   }
}
```

... In order to get away from "step-based data" and move towards "component-based data". Plus it will be easier to extend with more components.

src/resources/communityContributionCompany.ts
cameron-freshworks: But i agree, why don't we start from a template and break them out if required. 👍 
Will update.

src/resources/unlimitedCompany.ts
cameron-freshworks: You betcha, commented on this above (somewhere)

src/store/getters/state-getters.ts
cameron-freshworks: For sure, lets do this now.

src/enums/nameRequestTypes.ts
severinbeauvais: A comment can be multi-line for sure. But I think if you hover on NameRequestTypes you will only see 1 of the 3 comments. Merge them if you want, lol.

src/resources/benefitCompany.ts
severinbeauvais: Well, where did these strings come from?

src/resources/limitedCompany.ts
cameron-freshworks: CR right? 
Screen Shot 2021-05-27 at 1 35 47 PM

src/resources/benefitCompany.ts
cameron-freshworks: The route names, which can be re used here.

src/resources/communityContributionCompany.ts
severinbeauvais: I like that there are separate resource files for each business type. I'm just wondering if we should pull out the steps (which are common here, right) for these companies only?

One possible downside is that this introduces the concept of "default" or "fallback" steps, and we might eventually be tempted to say "use defaults except for ...", and that will get complicated fast.

Maybe with a good name for the steps common to BEN/ULC/etc, this will make sense.

I think a lot will make sense as we work through this. As long as we're intentional about keeping complexity low (at the cost of duplication if necessary), we should be OK 🤞

src/store/getters/state-getters.ts
cameron-freshworks: On second thought, we should do this in the framework ticket. We will be refactoring the store for this exact purpose and this change falls perfectly in with that.

src/components/common/Stepper.vue
severinbeauvais: I believe there are 3 states:
- initial (no validation applied)
- valid when complete/valid
- invalid when incomplete and validation has been applied (via route param I think)

Fortunately, JS booleans have at least 3 states. (joke but also true)

src/resources/communityContributionCompany.ts
cameron-freshworks: Thats the general idea with the duplication, was to keep it simple to start.
My initial gut feeling was to create a template for these 4 entities (BEN, CCC, ULC, BC), who all share the same steps.
I guess i don't trust that design won't change something along the way, so this way we can update them individually, although i think i will go with the template to start and if changes happen, we can break them out into there own.

src/components/common/Stepper.vue
cameron-freshworks: I think due to the time being already a day over without including Unit Test Updates, will need to push back on adding this. 
Mainly due to no UX input here, we would only be guessing what to show and when, right?

src/store/getters/state-getters.ts
severinbeauvais: You're right. I agree!

src/components/common/Stepper.vue
severinbeauvais: I saw the "invalid X" in a comp somewhere.

Just add a todo checkbox for this item to another ticket then.

src/resources/communityContributionCompany.ts
severinbeauvais: OK, we're on the same page.

Go whichever way you want here.

src/components/common/Stepper.vue
cameron-freshworks: Will do. 

Perhaps this was in the Coops designs we just looked at? As thats the only designs in progress with regards to Create UI afaik.

severinbeauvais: Haha, I think you're right on both counts!

src/resources/benefitCompany.ts
cameron-freshworks: Updated!

src/resources/communityContributionCompany.ts
cameron-freshworks: kk, Updated to a single template for now! 

I couldn't think of a better name than BaseStepsTemplate, but i did add to the comments which entities it applies too.

severinbeauvais: OK but let's rename that file/structure soon. It is not a template or base steps (as I don't want us to ever extend it). It is the common steps for BEN/BC/ULC/CCC.

src/resources/stepTemplates.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Currently used for BEN, BC, ULC and CCC.
+ */
+export const BaseStepsTemplate = [
+  {
+    id: 'step-1-btn',
+    step: 1,
+    icon: 'mdi-domain',
+    text: 'Define Your\nCompany',
+    to: RouteNames.DEFINE_COMPANY,
+    component: RouteNames.DEFINE_COMPANY
```
severinbeauvais, line 13: Hmm, this doesn't seem right (that we would have 2 properties with the same value).

I'm thinking "component" maps to the actual component name somehow.

cameron-freshworks: I believe it does yea... 🤔  
Just so happens the names match the routes i suppose. Should we create a duplicate enum with a different name?

severinbeauvais: Very good idea.

severinbeauvais: Hmm, sec.

Do we understand where we get the component name from? Is it like declaring components and Vue automatically converts to `<sub-component>` or `<SubComponent>` ? 

I just think this needs to be clear. 🤔 I'm confused.

cameron-freshworks: I'll dig into it 👍

cameron-freshworks: So yea, our `:component` property which we pass into `<component :is=componentName>` is just like using the component in the template. ie the name of the component. 
For example, the share structure step would work if we passed in `create-share-structure` or `CreateShareStructure` and so on. 
In this case, our routes are aligned with the component names.

severinbeauvais: I suggest using a different enum, to make it clear it's a different thing (even though it's the same values).

Or going back to strings (with a comment on the first one saying what it is, per your comment above).

package.json
```diff
@@ -8,11 +8,12 @@
     "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
     "lint": "vue-cli-service lint",
-    "test:unit": "vue-cli-service test:unit --testPathPattern --coverage",
+    "test:unit": "vue-cli-service test:unit --testPathPattern",
```
severinbeauvais, line 11: I _think_ the GHA CI script will still run the coverage step (but please confirm) and I'm totally OK with removing this step here.

src/resources/stepTemplates.ts
cameron-freshworks: Updated

package.json
cameron-freshworks: ah darn. I always remove this when working on the tests extensively. 
I dunno, should we remove it? Does it bother you as well, when working locally?

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -21,7 +21,7 @@
         <span>
           <v-icon color="blue darken-2">mdi-information-outline</v-icon>
           This step is not complete.
-          <router-link id="router-link" :to="{ path: '/add-people-roles', query: { showErrors: true } }">
+          <router-link id="router-link" :to="{ path: '/add-people-and-roles', query: { showErrors: true } }">
```
severinbeauvais, line 24: All the more reason to have a (separate) enum for this ;)

cameron-freshworks: Created!

src/resources/stepTemplates.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Currently used for BEN, BC, ULC and CCC.
+ */
+export const BaseStepsTemplate = [
+  {
+    id: 'step-1-btn',
+    step: 1,
+    icon: 'mdi-domain',
+    text: 'Define Your\nCompany',
+    to: RouteNames.DEFINE_COMPANY,
+    component: ViewComponentNames.DEFINE_COMPANY
```
severinbeauvais, line 13: 👍

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
severinbeauvais: Oh wait, we do now have 2 enums. Does it make sense to use RouteNames.XXX here?

package.json
cameron-freshworks: It has a report upload step either way, but i'm not entirely sure if it does coverage report if we don't leave in the command. I will put this back in 👍

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Yea, the route enum existed. I'm sure it just wasn't applied in this instance and most likely other `path` references.
It does for sure, but these "clean up refactors" can also wait until the framework ticket. I believe this ticket has already gone outside it's scope 😛

package.json
severinbeauvais: Wait. Maybe we should leave this out? I personally don't like its verbosity very much when I'm running the tests locally.

andrepestana-aot: Yeah, we can have another `run` command: one with coverage and another without it :)

cameron-freshworks: I'm with ya!

severinbeauvais: I'm OK with either approach (remove, keep, or remove but create another package,json entry).

severinbeauvais (approved):
Looks great. Woot!

cameron-freshworks:
> Looks great. Woot!

woo hoo! Thanks sir. Your input here was extremely valuable!

src/components/IncorporationAgreement/AgreementType.vue
```diff
@@ -68,12 +63,19 @@ export default class AgreementType extends Vue {
...
+  @Getter getIncorporationAgreement!: Array<IncorporationAgreementTypeIF>
+
   // Actions
   @Action setIncorporationAgreementStepData!: ActionBindingIF
   @Action setIgnoreChanges!: ActionBindingIF
 
   private agreementType: string | null = null
 
+  private get agreementTypeDescription (): string {
+    return this.getIncorporationAgreement.find(x => x.code === this.agreementTypeState)?.description
```
andrepestana-aot, line 76: Can this `this.getIncorporationAgreement` have more than one true for this condition: `x.code === this.agreementTypeState` ?
If yes, this will get the first one in the array...

andrepestana-aot (approved):
LGTM. Just some comments. Great job!

src/components/IncorporationAgreement/AgreementType.vue
cameron-freshworks: Currently no, as the code is unique between the options in the `incorporationAgreement`. 
The `agreementTypeState` and `code` certainly can have the same value across the various resources but we only load in 1 resource based on the entity type, so there would only ever be the options present int the resource file. 
resource example: 
Screen Shot 2021-05-31 at 1 32 13 PM

PR #205, opened 2021-06-01
Update sbc-common-component version in package.json

eve-git:
*Issue #:* bcgov/entity#6682

*Description of changes:*
Fix bug and disable browser autocomplete when entering address

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2670,66 +2670,6 @@
         "tslint": "^5.20.1",
         "webpack": "^4.0.0",
         "yorkie": "^2.0.0"
-      },
-      "dependencies": {
-        "chalk": {
-          "version": "4.1.1",
-          "resolved": "https://registry.npmjs.org/chalk/-/chalk-4.1.1.tgz",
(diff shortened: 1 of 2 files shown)


severinbeauvais (approved):
1. have the changes in the library been tested in this app?
2. I am OK with not updating the app version (since the "i" icon also displays the sbc-common-components version)
~3. we need to be careful about whether to merge this right now due to releases we're currently trying to push through~ _see response below_

severinbeauvais:
> 3. we need to be careful about whether to merge this right now due to releases we're currently trying to push through

I think we can merge in this PR when ready. However, it will have to be in the NEXT release (not the [current release](https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/reports/release?release=60b6671fcddcd1a30b622f39)). Please add the correct release label to the subject ticket (follow up with James and Linda if needed).

severinbeauvais:
> > 1. we need to be careful about whether to merge this right now due to releases we're currently trying to push through
> 
> I think we can merge in this PR when ready. However, it will have to be in the NEXT release (not the [current release](https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/reports/release?release=60b6671fcddcd1a30b622f39)). Please add the correct release label to the subject ticket (follow up with James and Linda if needed).

As this commit follows the "Initial Commit: Store update and resource config refactor" commit, these need to be in the NEXT release (whatever that is).

PR #210, opened 2021-06-15
1password Keys Integration.

pwei1018:
*Issue #:*
https://github.com/bcgov/entity/issues/7849
https://github.com/bcgov/entity/issues/6801

*Description of changes:*
1. change the key new to match new configuration of 1password.
2. change to read the session keys directly instead of read from auth-api-config session key.
3. Update sbc-common-components from 2.4.18 to 2.4.19


package.json
```diff
@@ -24,7 +24,7 @@
     "launchdarkly-js-client-sdk": "^2.18.0",
     "lodash": "^4.17.20",
     "regenerator-runtime": "^0.13.7",
-    "sbc-common-components": "2.4.18",
+    "sbc-common-components": "2.4.19",
```
severinbeauvais, line 27: Please explicitly identify the changes between 18 and 19 in the tickets for QA to verify.

src/utils/config-helper.ts
```diff
@@ -39,43 +39,36 @@ export async function fetchConfig (): Promise<any> {
...
-  const authUrl: string = response.data['AUTH_URL']
-  sessionStorage.setItem('AUTH_URL', authUrl)
-  console.log('Set Auth URL to: ' + authUrl)
-
   const dashboardUrl: string = response.data['DASHBOARD_URL']
   sessionStorage.setItem('DASHBOARD_URL', dashboardUrl)
   console.log('Set Dashboard URL to: ' + dashboardUrl)
 
-  const legalApiUrl: string = response.data['LEGAL_API_URL']
+  const legalApiUrl: string = response.data['LEGAL_API_URL'] + response.data['LEGAL_API_VERSION'] + '/'
```
severinbeauvais, line 54: OK, so the config keys don't include the trailing /, and you add it here. 👍 I like it.

src/utils/config-helper.ts
```diff
@@ -39,43 +39,36 @@ export async function fetchConfig (): Promise<any> {
...
   console.log('Set Auth API URL to: ' + authApiUrl)
 
-  const payApiUrl: string = response.data['PAY_API_URL']
+  const payApiUrl: string = response.data['PAY_API_URL'] + response.data['PAY_API_VERSION'] + '/'
   sessionStorage.setItem('PAY_API_URL', payApiUrl)
   console.log('Set Pay API URL to: ' + payApiUrl)
 
+  // for system alert banner (sbc-common-components)
+  const statusApiUrl: string = response.data['STATUS_API_URL'] + response.data['STATUS_API_VERSION']
+  sessionStorage.setItem('STATUS_API_URL', statusApiUrl)
```
severinbeauvais, line 69: So all apps still need to set the values for components they import (ie, SbcHeader) ? I was hoping the apps wouldn't have to worry about this (ie, the imported components would worry about their own values). Oh well.

src/utils/config-helper.ts
```diff
@@ -84,10 +77,14 @@ export async function fetchConfig (): Promise<any> {
...
   console.info('Set Address Complete Key.')
 
-  const ldClientId: string = response.data['LD_CLIENT_ID'];
+  const ldClientId: string = response.data['BUSINESS_CREATE_LD_CLIENT_ID'];
   (<any>window).ldClientId = ldClientId
   console.info('Set Launch Darkly Client ID.')
 
+  const sentryEnable = response.data['SENTRY_ENABLE'];
+  (<any>window).sentryEnable = sentryEnable
+  console.log('Set Sentry DSN Enable.')
```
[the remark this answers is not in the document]
pwei1018: What do you mean? We want to keep the sentry dsn value in 1password. However, I added a switch to control it whether we want to use it or not.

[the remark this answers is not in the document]
severinbeauvais: I'm referring to the console.log statement.

It might be useful to display whether the value is True or False. Or maybe we don't want to display anything. What do you think?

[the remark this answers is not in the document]
severinbeauvais: (My preference would be to display the value.)

src/utils/config-helper.ts
pwei1018: yes, I think the common components should not have set value logic in it. Only can read the values from project config.

package.json
pwei1018: done.

src/utils/config-helper.ts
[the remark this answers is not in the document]
pwei1018: oh, ok. I will remove the log.

package.json
severinbeauvais: Where is this documented?

The problem is that it's hard to tell what's new in a new version of sbc-common-components (so that we know what to test or watch for).

The solution is to explicitly document, in the ticket, what changes exist between the previous version and the new version we are importing.

src/utils/config-helper.ts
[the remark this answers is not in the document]
severinbeauvais: > (My preference would be to display the value.)

severinbeauvais (commented):
Hey Patrick, when you make changes, could you just add commits (instead of squashing and force-pushing just one commit)? ..... so that reviewers only have to review the new changes instead of having to review EVERYTHING again. Thanks!

pwei1018:
> Hey Patrick, when you make changes to the PR code, could you just add commits (instead of squashing and force-pushing just one commit)? ..... so that reviewers only have to review the new changes instead of having to review EVERYTHING again. Thanks!

Will you guys do the squash merge of the PR?

package.json
pwei1018: You may need to look at the PR's comment of sbc-common-components.

severinbeauvais:
> Will you guys do the squash merge of the PR?

Yes, it's the default action for the Merge button.

package.json
severinbeauvais: ^^ Yes, that is what needs to be done.

Since you are the dev updating the sbc-common-components dependency (and since there is no other way to keep track of the changes between versions), it falls to you to document the changes, in the ticket, for QA to see (please!).

(I would love another way to know about changes -- say, [a release bulletin](https://github.com/bcgov/sbc-common-components/releases) -- so that we could just refer to that instead of having to dive into the commit history every time.) << @sumesh-aot @saravankumarpa FYI

pwei1018: Ok, I may put the change detail into bcgov/entity#6801. Which you can include this ticket into your release plan. Is that ok?

severinbeauvais: Thanks. Will do!

severinbeauvais (approved):

PR #211, opened 2021-06-17
Updates for Certify

cameron-freshworks:
*Issue #:* /bcgov/entity#6863

*Description of changes:*

* Implemented Shared Certify component.

* Updates to store getters to work with new shared component.

* Minor verbiage updates.


src/resources/benefitCompany.ts
```diff
@@ -86,11 +86,11 @@ export const BenefitCompanyResource: ResourceIF = {
   },
   reviewAndConfirm: {
     completingPartyStatement: {
-      certifyStatementHeader: 'the Completing Party, have examined the Benefit Company ' +
+      certifyStatementHeader: 'the Completing Party, have examined the Company ' +
         'Articles and the Incorporation Agreement applicable to the company that is to be ' +
         'incorporated by the filing of this Incorporation Application and confirm that:',
       certifyStatements: [
-        'The Benefit Company Articles and the Incorporation Agreement both contain a signature ' +
+        'The Company Articles and the Incorporation Agreement both contain a signature ' +
```
severinbeauvais, line 93: Weird. So this resource, which is for Benefit Companies, should not call them benefit companies in the text?

src/resources/unlimitedCompany.ts
```diff
@@ -72,7 +72,7 @@ export const UnlimitedCompanyResource: ResourceIF = {
   },
   reviewAndConfirm: {
     completingPartyStatement: {
-      certifyStatementHeader: 'the Completing Party, have examined the Unlimited Company ' +
+      certifyStatementHeader: 'the Completing Party, have examined the Company ' +
```
severinbeauvais, line 75: Wait a minute. Is this the same blurb for all company types now? If so, should the blurb be part of the component instead of in the resource file?

src/store/getters/state-getters.ts
```diff
@@ -190,7 +191,8 @@ export const isEnableFilePay = (state: any, getters: any): boolean => {
   const step2Valid = state.stateModel.addPeopleAndRoleStep.valid
   const step3Valid = state.stateModel.createShareStructureStep.valid
   const step4Valid = state.stateModel.incorporationAgreementStep.valid
-  const step5Valid = state.stateModel.certifyState.valid && state.stateModel.incorporationDateTime.valid
+  const step5Valid = (state.stateModel.certifyState.valid && state.stateModel.certifyState.certifiedBy) &&
+    state.stateModel.incorporationDateTime.valid
```
severinbeauvais, line 195: Hehe, "(A and B) and C" is the same as "(A and B and C)".

src/store/getters/state-getters.ts
```diff
@@ -225,7 +227,8 @@ export const isIncorporationAgreementValid = (state: any): boolean => {
 export const isApplicationValid = (state: any): boolean => {
   return (state.stateModel.defineCompanyStep.valid && state.stateModel.addPeopleAndRoleStep.valid &&
     state.stateModel.createShareStructureStep.valid && state.stateModel.incorporationDateTime.valid &&
-    state.stateModel.incorporationAgreementStep.valid && state.stateModel.certifyState.valid)
+    state.stateModel.incorporationAgreementStep.valid &&
+    (state.stateModel.certifyState.valid && state.stateModel.certifyState.certifiedBy))
```
severinbeauvais, line 231: This looks almost like a duplicate of `isEnableFilePay`...
- should this evaluation check if incorp date time is valid?
- could `isEnableFilePay = return isApplicationValid` ? Or maybe just use a single getter?

src/store/getters/state-getters.ts
```diff
@@ -237,3 +240,24 @@ export const getMaxStep = (state: any, getters: any): number => {
...
 }
+
+/** The current state of the certify section. */
+export const getCertifyState = (state: any): CertifyIF => {
+  return state.stateModel.certifyState
+}
+
+/** Business contact email address. */
+export const getBusinessEmail = (state: any, getters: any): string => {
+  return state.stateModel.defineCompanyStep.businessContact.email
```
severinbeauvais, line 251: Is defineCompanyStep always declared? Is businessContact always declared? If not, use conditional chaining?

src/store/getters/state-getters.ts
```diff
@@ -237,3 +240,24 @@ export const getMaxStep = (state: any, getters: any): number => {
...
+}
+
+/** The Completing Party's email address. */
+export const getCompletingPartyEmail = (state: any): string | null => {
```
severinbeauvais, line 255: Strings can already be null, so you don't need to specify that as a return type.

src/resources/benefitCompany.ts
cameron-freshworks: Yea, according to the update designs. This could potentially be taken out of resources but not all the bullet points were identical. 
It seems to change so much, i though it would be best to leave it in resources and not apply some bullet points conditionally.

src/store/getters/state-getters.ts
```diff
@@ -237,3 +240,24 @@ export const getMaxStep = (state: any, getters: any): number => {
...
+  const completingParty =
+    state.stateModel.addPeopleAndRoleStep.orgPeople.find(person => {
+      return !!person.roles?.some(role => {
```
severinbeauvais, line 258: No need to change, but if the evaluation looks complicated due to `!!...`, you can also do `Boolean(...)`, which can look cleaner.

src/resources/benefitCompany.ts
severinbeauvais: OK, good idea.

src/resources/unlimitedCompany.ts
cameron-freshworks: Some bullets are the same atm yes, but not all.

src/store/getters/state-getters.ts
cameron-freshworks: good point! Will update

src/store/getters/state-getters.ts
cameron-freshworks: Will look at this!

src/store/getters/state-getters.ts
cameron-freshworks: oh nice! Will simplify then :) Updated in next commit

src/store/getters/state-getters.ts
cameron-freshworks: updated in next commit

src/store/getters/state-getters.ts
cameron-freshworks: Yea, this is a good spot for refactoring. I'm going to remove `isEnabledFilePay` to replace with isApplicationValid. The latter makes more sense now that the FilePay btn is always technically enabled.

cameron-freshworks: Updated in next commit.
They were near duplicates.

severinbeauvais: Sounds good!

src/store/getters/state-getters.ts
cameron-freshworks: Since we are dealing with a filing, they properties themselves are present but not the value. 
Either way though, optional chaining would not hurt! Added in next commit.

src/store/getters/state-getters.ts
cameron-freshworks: Updated in next commit

src/store/getters/state-getters.ts
severinbeauvais: Actually, I would rather not use optional chaining if it's not needed (since it can mask errors, such as unit test errors).

severinbeauvais (approved):

src/store/getters/state-getters.ts
cameron-freshworks: Copy that!

cameron-freshworks: Removed 👍

PR #212, opened 2021-06-22
Initial Commit

cameron-freshworks:
*Issue #:* /bcgov/entity#6858

*Description of changes:*

* Added sample articles for other Limited, unlimited companies.

* Updated unit tests.

* Updated Header block and tooltips for non BCOMP entities.

* Updates to include exceptions for CCC's.


severinbeauvais (commented):
Looks good so far.

Is the "sample articles" blurb the same for all entity types so far?

cameron-freshworks:
> Looks good so far.
> 
> Is the "sample articles" blurb the same for all entity types so far?
 
 Very close, Bcomps is a little different :(

src/App.vue
```diff
@@ -436,6 +436,9 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
 
         // parse draft filing into the store
         if (draftFiling) {
+          // const mocktype = 'ULC'
+          // draftFiling.business.legalType = mocktype
+          // draftFiling.incorporationApplication.nameRequest.legalType = mocktype
```
severinbeauvais, line 441: For testing, I presume?
I don't mind if you commit this, but please add a comment we will easily find to remove this later.

src/components/IncorporationAgreement/AgreementType.vue
```diff
@@ -29,15 +29,32 @@
...
-          <v-radio v-for="(item, index) in getIncorporationAgreementDocuments"
-            :key="index" :value="item.code" :id="`agreement-type-${item.code}`"
+        <template v-if="isTypeCC">
+          <v-checkbox
+            v-for="(item, index) in getIncorporationAgreementDocuments"
+            :id="`agreement-type-${item.code}`"
+            class="agreement-option-list"
+            :key="index"
+            v-model="agreementType"
+            :value="item.code"
```
severinbeauvais, line 39: Oh? It's possible to use both v-model and :value ?

severinbeauvais (approved):

src/App.vue
cameron-freshworks: oh no no 😁 Let me take this out! Oopsies haha
this shouldn't be required for much longer hopefully!

src/components/IncorporationAgreement/AgreementType.vue
cameron-freshworks: It is! In this case, as similar to the radio btns, let me look into this a bit tho!

src/App.vue
cameron-freshworks: Removed!

src/resources/communityContributionCompany.ts
```diff
@@ -43,39 +39,20 @@ export const CommunityContributionCompanyResource: ResourceIF = {
...
-          `You wish to change any of the standard provisions in the sample Articles.`
+        header: `Retain the signed Incorporation Agreement and Articles`,
+        helpText: [
+          `The company is required to keep signed copies of the Incorporation Agreement and Articles in the company’s
+          record book. For a complete list of records a company is required to keep please see section 42 of the
+          Business Corporations Act.`
         ]
       }
     ],
+    articleUrl: `Sample_articles.pdf`,
```
andrepestana-aot, line 50: Is it the article url or article file name?

andrepestana-aot (approved):
Just added a comment but looks good to me! Good job! 👍

src/resources/communityContributionCompany.ts
cameron-freshworks: Good point, this is actually the file name. The url to get the file from the public folder is done in the component itself. 
So yea, this is technically just the file name. Will update!

src/components/IncorporationAgreement/AgreementType.vue
severinbeauvais: I wonder if this just happens to work, or if it's a good way to set the v-model initially?

Anyway, as long as this code works as expected, carry on (no change needed).

cameron-freshworks: Thats a good question! 
I'm a little suspicious of it as well, now that you mentioned it because i can't exactly differentiate between to the two values/purposes. 
It does work as expected, as currently the `value` is used locally to determine the the output text for the Summary/Review page and the model is being set to store. 🤷

src/resources/communityContributionCompany.ts
cameron-freshworks: Updated!

src/components/IncorporationAgreement/AgreementType.vue
cameron-freshworks: Ah i see, sorry the model is a boolean and the value is the string(code) used to determine the descriptions. 
Applies to both checkboxes and radio btns in this case.

severinbeauvais: Oh, _that_ value. I confused it with the boolean.
Resolving this conversation then.

PR #214, opened 2021-07-05
Incorporation Application UX Updates

cameron-freshworks:
*Issue #:* /bcgov/entity#6856

*Description of changes:*

* Updated verbiage.
* Removed blank LTD statement.
* Added People and roles btn text for all entities. 

### DIFF
--- a/src/views/DefineCompany.vue
+++ b/src/views/DefineCompany.vue
@@ -2,7 +2,9 @@
   <div>
     <div class="mt-10 company-statement">
       <p>
-        <span class="company-statement-label">{{ getCompanyResources.title }}:</span>
+        <span v-if="getCompanyResources.description" class="company-statement-label">
+          {{ getCompanyResources.title }}:
+        </span>
         {{ getCompanyResources.description }}
       </p>
     </div>
--- a/src/views/DefineCompany.vue
+++ b/src/views/DefineCompany.vue
@@ -2,9 +2,7 @@
   <div>
     <div class="mt-10 company-statement">
       <p>
-        <span v-if="getCompanyResources.description" class="company-statement-label">
-          {{ getCompanyResources.title }}:
-        </span>
+        <span class="company-statement-label">{{ getCompanyResources.title }}</span>
         {{ getCompanyResources.description }}
       </p>
     </div>
--- a/src/views/DefineCompany.vue
+++ b/src/views/DefineCompany.vue
@@ -1,7 +1,7 @@
 <template>
   <div>
     <div class="mt-10 company-statement">
-      <p>
+      <p v-if="getCompanyResources.title">
         <span class="company-statement-label">{{ getCompanyResources.title }}</span>
         {{ getCompanyResources.description }}
       </p>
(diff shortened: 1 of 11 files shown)


src/views/DefineCompany.vue
```diff
@@ -2,7 +2,9 @@
   <div>
     <div class="mt-10 company-statement">
       <p>
-        <span class="company-statement-label">{{ getCompanyResources.title }}:</span>
+        <span v-if="getCompanyResources.description" class="company-statement-label">
+          {{ getCompanyResources.title }}:
```
severinbeauvais, line 6: Wait - if there's a description then display the title and the description? I'm confused maybe.

cameron-freshworks: :) I suppose either property could have been used here. The statement title or the description.
But you are right, in the even there is no description that we want to display (LTD's) there is no title either.

severinbeauvais: Could you clarify the logic in the code pretty please?

cameron-freshworks: Absolutely, i think it makes sense to check the title opposed to the description as well.

cameron-freshworks: This was actually a snafu on my part. 

I had the statement title in the resource file where it wasn't required. 
So opposed to having a condition i just removed the statement title (because there is none for LTD) and moved the colon into the titles.
So no conditions required, just taken from the Resource file verbatim.

src/views/DefineCompany.vue
```diff
@@ -2,7 +2,7 @@
   <div>
     <div class="mt-10 company-statement">
       <p>
-        <span class="company-statement-label">{{ getCompanyResources.title }}:</span>
+        <span class="company-statement-label">{{ getCompanyResources.title }}</span>
```
severinbeauvais, line 5: I think this might display "null" on the page. You might need to add a v-if for title.

cameron-freshworks: It doesn't display anything when `null`, tested!

severinbeauvais: OK, that's good; I just learned something. (Not all frameworks are like this.)

Wouldn't it still be better to not render the span if there's no title?

cameron-freshworks: I'm not sure, is there a downside to displaying the span if it is empty?

severinbeauvais: 🤷

cameron-freshworks: Less (code ) is more right? 😆

cameron-freshworks: Added safety check :)

tests/unit/DefineCompany.spec.ts
```diff
@@ -63,7 +63,7 @@ for (const mock of mockEntityInfo) {
       expect(wrapper.find('#folio-number-header').text()).toContain('Folio / Reference Number (optional)')
     })
 
-    it(`displays company statement`, () => {
+    itIf(mock.entityType !== 'BC')(`displays company statement`, () => {
```
severinbeauvais, line 66: Small typo.

severinbeauvais (approved):
GTG after small fix.

tests/unit/DefineCompany.spec.ts
cameron-freshworks: Which part is the typo 😆  Sorry, i'm missing it!

severinbeauvais: > That's a method wrapper.
> Since we parameterized the tests they don’t all apply to all types, so it’s to exclude the test.

PR #216, opened 2021-08-03
Coops and Global IA Framework Updates

cameron-freshworks:
*Issue #:* /bcgov/entity#7214

*Notes:*
* Unit Test updates in Progress
* Found a duplicate request for account info, Updating

*Description of changes:*

* Adjusted btn locations on the Actions component when on the Review/Summary step.
* Updated Review step validations, message styles and prompts.
* Stepper Design/Ux updates.
* Coops resources and stepTemplate updates.
* Placeholder Components for Rules and Memorandums, alongside baseline validations, routes and views.
* Updated certify with disable prop and applied validations.
* Request for current profile name and applied to store.


src/services/auth.services.ts
```diff
@@ -0,0 +1,29 @@
...
+  /** Fetches specified org's info. */
+  static async fetchOrgInfo (orgId: number): Promise<any> {
+    const url = `${this.authApiUrl}orgs/${orgId}`
+    return axios.get(url)
+      .then(response => {
+        if (response?.data) return response.data
+        throw new Error('Invalid org info')
+      })
+  }
+}
```
severinbeauvais: Do you think we should move legal-api-mixin.ts in this services folder too? Is it as easy as moving the file and a bit of renaming?

cameron-freshworks: Sorry, i didn't realize we already captured these requests in a mixin so i removed my additional code.

src/App.vue
```diff
@@ -459,6 +462,18 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
         const companyResources = CompanyResources.find(x => x.entityType === this.entityType)
         if (companyResources) this.setCompanyResources(companyResources)
         else throw new Error('Invalid Entity Type')
```
severinbeauvais, line 464: I'm wondering if this is working correctly, since I over-wrote the legal type in App.vue, which should result in an invalid type here (since I don't have your CP code yet), but there was not error and I saw this:

image

src/services/auth.services.ts
severinbeauvais: I just noticed that too. OK, nvm :)

src/App.vue
cameron-freshworks: I think this is a type error or error on a component. 

CP is a valid type but it's resource file was just bare bones. So missing data is most likely causing an error rendering some components.
This IS fixed with the update resource file for Coops.

severinbeauvais: Ah, the placeholder data existed for CP. OK, np then.

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
```diff
@@ -15,24 +15,10 @@
...
-          >
-            Return to this step to complete it.
-          </router-link>
-        </span>
-      </div>
     </div>
 
     <!-- List Display Section -->
-    <div id="people-roles-list">
+    <div id="people-roles-list" :class="{ 'invalid-section': showErrorSummary && getValidateSteps }">
```
severinbeauvais, line 21: I'll probably rename "getValidateSteps" at some point because it appears to return the validate steps, but it actually returns a boolean indicating whether the user has tried to submit a filing. Maybe I'll call the state variable "doValidate" or "validateState". Thoughts?

src/components/CreateMemorandum/UploadMemorandum.vue
```diff
@@ -0,0 +1,66 @@
+<template>
+  <v-card flat id="upload-memorandum">
+
+    <!-- Summary Section -->
+    <div id="upload-memorandum-summary" v-if="isSummary">
```
severinbeauvais, line 5: Since this is a v-card, could v-card-text etc be used for some of these blocks? Or would it conflict with the other, similar components?

src/components/CreateMemorandum/UploadMemorandum.vue
```diff
@@ -0,0 +1,66 @@
...
+      <!-- Summary Header -->
+      <div class="upload-memorandum-summary-header" >
+        <v-icon color="#38598A">mdi-text-box-multiple</v-icon>
+        <label class="upload-memorandum-title"><strong>Memorandum</strong></label>
```
severinbeauvais, line 9: Geek alert.
This should actually be "bold" not "strong".
https://www.geeksforgeeks.org/difference-between-strong-and-bold-tag-in-html/
End alert.

src/components/CreateMemorandum/UploadMemorandum.vue
```diff
@@ -0,0 +1,66 @@
...
+          <v-icon color="error">mdi-information-outline</v-icon>
+          <span class="error-text"> This step is not complete. </span>
+          <router-link :to="{ path: `/${RouteNames.CREATE_MEMORANDUM}`, query: { showErrors: true } }">
+            Return to this step to complete it.
+          </router-link>
+        </span>
+      </div>
+    </div>
+
+    <v-card-title>Memorandum Placeholder component</v-card-title>
```
severinbeauvais, line 24: Title at the end? Or title element used for emphasis?

src/components/AddPeopleAndRoles/ListPeopleAndRoles.vue
cameron-freshworks: Yea, was thinking on this as well! I'm fine with anything

src/components/CreateMemorandum/UploadMemorandum.vue
cameron-freshworks: Absolutely, although i'm not even sure what the summary view looks like for these components.

src/components/CreateMemorandum/UploadMemorandum.vue
cameron-freshworks: ahh! Will fix :) 
Perhaps i should just be using the `font-weight-bold` helper class anywho?

src/components/CreateMemorandum/UploadMemorandum.vue
```diff
@@ -0,0 +1,66 @@
...
+import { Component, Prop, Vue } from 'vue-property-decorator'
+import { State } from 'vuex-class'
+
+// Enums
+import { RouteNames } from '@/enums'
+
+@Component({})
+export default class UploadMemorandum extends Vue {
+  // Global state
+  @State(state => state.stateModel.createRulesStep.valid)
```
severinbeauvais, line 38: I'd prefer we use getters instead of accessing state.

I searched and found that accessing state is done a lot in this project (gack!).

Reason: we don't want components to assume (or think they know) what the state structure looks like; we do want components to use an "API" (interface/contract) between system modules (like store).

src/components/CreateMemorandum/UploadMemorandum.vue
cameron-freshworks: Yea, no real significance here :) Just slapped some content in there for display purposes. There are just placeholders, 100% subject to change.

src/components/CreateMemorandum/UploadMemorandum.vue
```diff
@@ -0,0 +1,66 @@
...
+  padding: 1.25rem;
+
+  .upload-memorandum-title {
+    padding-left: 0.5rem;
+  }
+}
+
+.upload-memorandum-error-message {
+  padding-top: 1.25rem;
+  padding-left: 1.25rem;
```
severinbeauvais, line 63: I know much other code in this project specifies layout in the "style" block, but IMHO these should use Vuetify helper classes in the HTML. (Exception: if we implement standard classes for different things in a common style file.)
Don't rewrite this but I'd prefer to see inline layout classes in future code.

src/components/CreateMemorandum/UploadMemorandum.vue
cameron-freshworks: For sure, i agree.
Changing everywhere is the right thing to do, but the scope of that change is a little larger and will require regression testing. 

For this PR, i will update these new/edited instances of state accessors to getters and we should make effort to update these as we go.

src/components/CreateRules/UploadRules.vue
```diff
@@ -0,0 +1,68 @@
+<template>
+  <v-card flat id="upload-rules">
+
+    <!-- Summary Section -->
+    <div id="upload-rules-summary" v-if="isSummary">
+      <!-- Summary Header -->
+      <div class="upload-rules-summary-header" >
+        <v-icon color="#38598A">mdi-format-list-text</v-icon>
```
severinbeauvais, line 8: I suggest we add named colours to our Vuetify themes declaration, instead of using hex colours throughout the code.

src/components/CreateMemorandum/UploadMemorandum.vue
```diff
@@ -0,0 +1,66 @@
...
+  color: $primary-blue;
+}
+</style>
```
severinbeauvais, line 66: All my comments in this file apply to the other new files.

src/components/CreateMemorandum/UploadMemorandum.vue
cameron-freshworks: Ok fair enough, i see your point. 
My thought process was, we needed class for the `color` so i put the stylings in a single place. 
Curious though, i'm not sure if the helper classes are in rem, pretty sure they are set pixels no? Something like 4px per digit?

src/components/CreateRules/UploadRules.vue
cameron-freshworks: Ah yes, i did add them to the theme, perhaps forgot to update them here in the placeholder components.

src/components/CreateMemorandum/UploadMemorandum.vue
severinbeauvais: I was trying to resolve (for myself) the difference between `<b>` and `font-weight: 700`, and best as I can tell, they're equivalent. (Bold assumes 700.) So up to you.

src/components/CreateMemorandum/UploadMemorandum.vue
severinbeauvais: OK, next person to work in these files can fix it.

src/components/CreateMemorandum/UploadMemorandum.vue
severinbeauvais: OK, next person to work in these files can fix it.

src/components/CreateMemorandum/UploadMemorandum.vue
severinbeauvais: Excellent, thanks. Rome wasn't built in a day.

src/components/CreateMemorandum/UploadMemorandum.vue
cameron-freshworks: ok right on!

src/components/CreateMemorandum/UploadMemorandum.vue
cameron-freshworks: At least 2 days!

src/components/CreateMemorandum/UploadMemorandum.vue
severinbeauvais: TBH I'm not hard and fast on this even in my own code (and some classes are only created at run-time so you have to use the style block) (and sometimes you want to do something for only the first, or last, element in a list so you have to use the style block).

Even with the need for a colour style (though you could use a base class to set fg or bg colour ;) ), I try to have layout styles in the HTML.

I try to do what looks clean or feels right. I presume you've done the same so just leave it :)

severinbeauvais: ^^ in other projects we have base classes like "red-text", etc

severinbeauvais: ^^ but I'm not asking you to change it! :)

src/components/CreateRules/UploadRules.vue
severinbeauvais: You didn't add this colour to the Vuetify theme ... pick a nice semantic name for it and then you can use it in Vuetify elements/components.

cameron-freshworks: oh this was a carry over from the pre existing icons in the summary headers! Ok, will update, for ALL of them 👍

src/resources/stepTemplates.ts
```diff
@@ -46,3 +46,48 @@ export const BaseStepsTemplate = [
...
+    component: ViewComponentNames.ADD_PEOPLE_AND_ROLES
+
+  },
+  {
+    id: 'step-3-btn',
+    step: 3,
+    icon: 'mdi-format-list-text',
+    text: 'Upload Rules',
+    to: RouteNames.CREATE_RULES,
+    component: ViewComponentNames.CREATE_RULES
```
severinbeauvais, line 75: Hey, new route and component here, cool!

src/router/routes.ts
```diff
@@ -56,6 +58,16 @@ export const routes = [
       requiresAuth: true
     }
   },
+  {
+    path: '/create-rules',
+    name: RouteNames.CREATE_RULES,
+    component: CreateRules,
+    meta: {
+      step: 3,
+      label: 'Create Rules',
```
severinbeauvais, line 67: Check whether we use the label. I suspect we don't. This was probably meant to be step text and we never cleaned it up.

src/store/getters/state-getters.ts
```diff
@@ -218,14 +218,26 @@ export const isIncorporationAgreementValid = (state: any): boolean => {
 }
 
 /** Whether all the incorporation steps are valid. */
-export const isApplicationValid = (state: any): boolean => {
-  return (
-    state.stateModel.defineCompanyStep.valid &&
-    state.stateModel.addPeopleAndRoleStep.valid &&
+export const isApplicationValid = (state: any, getters: any): boolean => {
+  // Base company steps
+  const isCompanyStepsValid = (
```
severinbeauvais, line 223: I like the comment, but the variable name had me going "hmm" for a bit. Suggest you rename it isBaseStepsValid.

src/views/CreateMemorandum.vue
```diff
@@ -0,0 +1,23 @@
...
+
+// Components
+import { UploadMemorandum } from '@/components/CreateMemorandum'
+
+@Component({
+  components: {
+    UploadMemorandum
+  }
+})
+export default class CreateRules extends Vue {}
```
severinbeauvais, line 19: Please fix class name.

src/components/Summary/SummaryDefineCompany.vue
```diff
@@ -1,53 +1,55 @@
...
+          <span class="error-text"> This step is not complete. </span>
+          <router-link :to="{ path: `/${RouteNames.DEFINE_COMPANY}`, query: { showErrors: true } }">
+            Return to this step to complete it.
+          </router-link>
+        </span>
+      </div>
+      <div class="section-container">
+        <!--TODO: Replace container content with Name Request Summary when it is ready -->
+        <v-layout row>
+          <v-flex md4>
```
severinbeauvais, line 20: Note to all devs: v-layout and v-flex is deprecated. We should be using v-row and v-col from now on.

https://vuetifyjs.com/en/components/grids/#v-col
https://vuetifyjs.com/en/components/grids/#v-row
https://vuetifyjs.com/en/components/grids/#caveats

I'm not asking to change this code right now, but new code should use the newer components, and we should fix the old code "on brick at a time" (re: Rome). TIA.

src/components/common/Stepper.vue
```diff
@@ -1,7 +1,13 @@
 <template>
-  <v-container id="step-buttons-container">
+  <v-card id="step-buttons-container">
```
severinbeauvais, line 2: How is v-card more appropriate than v-container? (I don't disagree, just curious.)

https://vuetifyjs.com/en/components/cards/
https://vuetifyjs.com/en/components/grids/#v-container

tests/unit/DefineCompany.spec.ts
```diff
@@ -63,7 +63,7 @@ for (const mock of mockEntityInfo) {
       expect(wrapper.find('#folio-number-header').text()).toContain('Folio / Reference Number (optional)')
     })
 
-    itIf(mock.entityType !== 'BC')(`displays company statement`, () => {
+    itIf(['BEN', 'ULC', 'CCC'].includes(mock.entityType))(`displays company statement`, () => {
```
severinbeauvais, line 66: Yes! Good change! This includes the specified types instead of assuming that everything not the other type is OK.

src/views/ReviewConfirm.vue
```diff
@@ -9,20 +9,20 @@
 
     <section class="mt-10">
       <header>
-        <h2>Review</h2>
+        <h2>Review and Confirm</h2>
+        <p class="mt-1">Review the information in your application. If you need change or complete anything, return to
+          the step make the necessary change.</p>
```
severinbeauvais, line 14: Typo - missing word or something.

src/views/ReviewConfirm.vue
```diff
@@ -9,20 +9,20 @@
...
       </header>
-      <Summary />
+      <Summary class="mt-6"/>
     </section>
-    <section class="mt-10">
+    <section v-if="!isTypeCoop" class="mt-10">
```
severinbeauvais, line 18: Maybe we should create a new getter to INCLUDE all types that aren't Coop instead of assuming that everything not a Coop is OK ?

src/views/ReviewConfirm.vue
```diff
@@ -106,7 +113,7 @@ export default class ReviewConfirm extends Mixins() {
   }
 
   /** Called when component is mounted. */
-  mounted (): void {
+  async mounted (): Promise<void> {
```
severinbeauvais, line 116: Not async?

src/views/ReviewConfirm.vue
```diff
@@ -33,23 +33,28 @@
...
+          :statements="getCompletingPartyStatement.certifyStatements"
+          :message="getCompletingPartyStatement.certifyClause"
+          :enableMailTo="true"
+          :businessEmail="getBusinessEmail"
+          :completingPartyEmail="getCompletingPartyEmail"
+          :isStaff="isRoleStaff"
+          :firstColumn="3"
+          :secondColumn="9"
+          :invalidSection="isCertifyInvalid"
+          :disableEdit="!isRoleStaff"
```
severinbeauvais, line 53: Does this new prop mean the "isStaff" functionality of Certify is no longer needed? (UX question, I know.)

severinbeauvais (commented):
Lots of great implementation work in here! Just a few comments.

src/components/CreateMemorandum/UploadMemorandum.vue
cameron-freshworks: Updated in next commit 👍

src/components/CreateMemorandum/UploadMemorandum.vue
cameron-freshworks: Updated in next commit 👍

src/components/CreateRules/UploadRules.vue
cameron-freshworks: Updated in next commit 👍

src/resources/stepTemplates.ts
cameron-freshworks: We're growing!

src/router/routes.ts
cameron-freshworks: I don't believe we do, will remove them 👍

src/views/CreateMemorandum.vue
cameron-freshworks: Fixed!

src/components/Summary/SummaryDefineCompany.vue
cameron-freshworks: 100%. Noted 👍

src/views/ReviewConfirm.vue
cameron-freshworks: Referring to the `If you need change`? => `If you need to change` ?

cameron-freshworks: oh maybe `the step make` => `the step to make` as well

cameron-freshworks: Both fixed

src/views/ReviewConfirm.vue
cameron-freshworks: Sounds good!

cameron-freshworks: Updated! In a few places

src/views/ReviewConfirm.vue
cameron-freshworks: Good catch, not anymore. 
at one point was making the profile request in this hook until i found it elsewhere.
Removed!

src/views/ReviewConfirm.vue
cameron-freshworks: I do not know know, potentially though. 
I thought it might be more useful to disable the input and let the app decide vs adding more staff rules into the component itself. 

isStaff seems to just change the statement text very slightly 🤷

src/assets/styles/theme.scss
```diff
@@ -56,3 +56,9 @@ $BCgovInputBG: #ffffff;
 
 // Font Colors
 $BCgovFontColorInverted: #ffffff;
+
+// App Colors
+$app-blue: #1669bb; // same as the Vuetify theme primary
+$app-green: #1a9031; // same as the Vuetify theme success
+$app-red: #d3272c; // same as the Vuetify theme error
+$app-dk-blue: #38598a; // same as the custom theme dkBlue
```
severinbeauvais, line 64: FWIW, this is still a Vuetify theme colour.

src/store/getters/state-getters.ts
```diff
@@ -217,15 +228,37 @@ export const isIncorporationAgreementValid = (state: any): boolean => {
...
-    state.stateModel.defineCompanyStep.valid &&
-    state.stateModel.addPeopleAndRoleStep.valid &&
+export const isApplicationValid = (state: any, getters: any): boolean => {
+  // Base company steps
+  const isBaseStepsValid = (
     state.stateModel.createShareStructureStep.valid &&
     state.stateModel.incorporationDateTime.valid &&
     state.stateModel.incorporationAgreementStep.valid &&
-    state.stateModel.incorporationDateTime.valid &&
+    state.stateModel.incorporationDateTime.valid
```
severinbeauvais, line 248: Duplicate of line 246.

src/assets/styles/theme.scss
cameron-freshworks: Oh interesting, i guess i got confused because the name was custom and not an override of the actual Theme colour name, ie primary. Will update the comment!

src/store/getters/state-getters.ts
cameron-freshworks: Good catch thanks!

vysakh-menon-aot (approved):

src/assets/styles/theme.scss
cameron-freshworks: fixed

src/store/getters/state-getters.ts
cameron-freshworks: fixed

severinbeauvais (approved):

PR #218, opened 2021-08-05
8352 implemented misc TODOs

severinbeauvais:
*Issue #:* /bcgov/entity#8352

*Description of changes:*
- throw error if nameRequest object not in filing
- display errors on unexpected errors
- display useful logs if NR checks fail
- ensure NR type matches entity type
- map NR type to entity type
- added NR validation checks
- updated unit tests
- fixes for date TODOs
- added date mixin methods
- added code exceptions for Jest
- app version: 2.0.6

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -429,11 +429,8 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
 
         // verify nameRequest object
         const nameRequest = draftFiling?.incorporationApplication?.nameRequest
-        if (!nameRequest) {
-          // there should be a nameRequest object but continue even if it doesn't exist
-          // TODO: after MVP, add a real check here
-          console.log('Filing should have nameRequest object') // eslint-disable-line no-console
-        }
+        if (!nameRequest) throw new Error('missing Name Request object')
+
         /** Fetches and validates the NR and sets the data to the store. This method is different
          * from the validateNameRequest method in Actions.vue. This method sets the data to
          * the store shows a specific message for different invalid states and redirection is to the
@@ -446,7 +443,7 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
         // An unknown entity type will need to be handled here
         const companyResources = CompanyResources.find(x => x.entityType === this.getEntityType)
         if (companyResources) this.setCompanyResources(companyResources)
-        else throw new Error('Invalid Entity Type')
+        else throw new Error('invalid Entity Type')
 
         // set current profile name to store for field pre population
         // proceed only if we are not staff
@@ -468,8 +465,9 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
         this.fetchErrorDialog = true
         throw error // go to catch()
       }
-    } catch (e) {
+    } catch (error) {
       // errors should be handled above
+      console.error('Unhandled error in fetchData() =', error) // eslint-disable-line no-console
       // just fall through to finally()
     } finally {
       this.haveData = true
@@ -504,25 +502,35 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
 
       // ensure NR is valid
       if (!this.isNrValid(nrResponse)) {
+        console.log('NR is not valid') // eslint-disable-line no-console
         this.nameRequestInvalidType = NameRequestStates.INVALID
         this.nameRequestInvalidErrorDialog = true
         return
       }
 
-      // TODO: verify that nrResponse.entityTypeCd === business.legalType
+      // ensure types match
+      if (this.nrTypeToEntityType(nrResponse) !== this.getEntityType) {
+        console.log('NR type doesn\'t match entity type') // eslint-disable-line no-console
+        this.nameRequestInvalidType = NameRequestStates.INVALID
+        this.nameRequestInvalidErrorDialog = true
+        return
+      }
 
       // ensure NR is consumable
       const state = this.getNrState(nrResponse)
       if (state !== NameRequestStates.APPROVED && state !== NameRequestStates.CONDITIONAL) {
+        console.log('NR is not consumable') // eslint-disable-line no-console
         this.nameRequestInvalidType = state || NameRequestStates.INVALID
         this.nameRequestInvalidErrorDialog = true
         return
       }
+
       // if we get this far, the NR is good to go!
       const nameRequestState = this.generateNameRequestState(nrResponse, filing.Id)
       this.setNameRequestState(nameRequestState)
-    } catch (e) {
+    } catch (error) {
       // errors should be handled above
+      console.error('Unhandled error in processNameRequest() =', error) // eslint-disable-line no-console
     }
   }
 
--- a/src/App.vue
+++ b/src/App.vue
@@ -413,11 +413,12 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
         // fetch draft filing
         let draftFiling = await this.fetchDraft()
 
-        // If there is an existing filing, check if it is in a valid state to be edited
+        // if there is an existing filing, check if it is in a valid state to be edited
         if (draftFiling) {
           this.invalidIncorporationApplicationDialog = this.hasInvalidFilingState(draftFiling)
           if (this.invalidIncorporationApplicationDialog) return
         }
+
         // merge draft properties into empty filing so all properties are initialized
         const emptyFiling = this.buildFiling()
         draftFiling = { ...emptyFiling.filing, ...draftFiling }
--- a/src/components/ReviewConfirm/IncorporationDateTime.vue
+++ b/src/components/ReviewConfirm/IncorporationDateTime.vue
@@ -192,7 +192,7 @@ export default class IncorporationDateTime extends Mixins(DateMixin) {
 
     // Clear Future Effective from store / fee summary
     this.setIsFutureEffective(false)
-    this.emitDateTime('')
+    this.emitDateTime(null)
   }
 
   /** Construct the Date Object for storage */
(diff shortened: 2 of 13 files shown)


src/App.vue
```diff
@@ -429,11 +429,8 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
 
         // verify nameRequest object
         const nameRequest = draftFiling?.incorporationApplication?.nameRequest
-        if (!nameRequest) {
-          // there should be a nameRequest object but continue even if it doesn't exist
-          // TODO: after MVP, add a real check here
-          console.log('Filing should have nameRequest object') // eslint-disable-line no-console
-        }
+        if (!nameRequest) throw new Error('missing Name Request object')
```
severinbeauvais, line 432: This exception will be caught by line 459.

src/App.vue
```diff
@@ -468,8 +465,9 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
         this.fetchErrorDialog = true
         throw error // go to catch()
       }
-    } catch (e) {
+    } catch (error) {
       // errors should be handled above
+      console.error('Unhandled error in fetchData() =', error) // eslint-disable-line no-console
```
severinbeauvais, line 470: This doesn't affect processing but displays the error (and stack trace) if something above didn't handle it.

src/App.vue
```diff
@@ -504,25 +502,35 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
 
       // ensure NR is valid
       if (!this.isNrValid(nrResponse)) {
+        console.log('NR is not valid') // eslint-disable-line no-console
```
severinbeauvais, line 505: I thought this (and the lines below) should log a bit more info on why the NR is invalid.

src/components/Summary/SummaryDefineCompany.vue
```diff
@@ -17,7 +17,7 @@
         </span>
       </div>
       <div class="section-container">
-        <!--TODO: Replace container content with Name Request Summary when it is ready -->
+        <!-- FUTURE: Replace container content with Name Request Summary when it is ready -->
```
severinbeauvais, line 20: TODO: should be done ASAP
FUTURE: do later, maybe

src/components/common/Actions.vue
```diff
@@ -199,10 +199,7 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Lega
       if (this.isBusySaving) return
       this.setIsFilingPaying(true)
 
-      if (
-        this.getIncorporationDateTime.effectiveDate &&
-        !this.isValidDateTime(this.getIncorporationDateTime.effectiveDate)
-      ) {
+      if (!this.isValidDateTime(this.getIncorporationDateTime.effectiveDate)) {
```
severinbeauvais, line 202: `isValidDateTime()` already handles empty date, so extra check here was not needed.

src/mixins/name-request-mixin.ts
```diff
@@ -53,13 +52,18 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
...
+      nr &&
+      nr.nrNum &&
+      nr.applicants &&
       nr.state &&
       nr.expirationDate &&
       nr.names?.length > 0 &&
-      nr.nrNum &&
-      nr.requestTypeCd)
+      this.nrTypeToEntityType(nr) &&
+      this._getApprovedName(nr)
```
severinbeauvais, line 65: I added a few checks here.

src/mixins/name-request-mixin.ts
```diff
@@ -112,6 +116,21 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
     if (nr.state === NameRequestStates.CONDITIONAL) {
       return nr.names.find(name => name.state === NameRequestStates.CONDITION).name
     }
-    return '' // should never happen
+    return null // should never happen
```
severinbeauvais, line 119: Both '' and null display "nothing" (so this change doesn't break anything), and I think returning null is more precise if this should never happen anyway.
Both '' and null are falsey, so they're interchangeable (except you can use string methods on '' but not on null).
Any thoughts on which is better or if this matters?

src/mixins/name-request-mixin.ts
```diff
@@ -17,8 +17,7 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
   generateNameRequestState (nr: any, filingId: number): NameRequestIF {
     return {
       nrNumber: nr.nrNum,
-      // TODO: Update entityType to use nr.requestTypeCd when namex supports our entity types
-      entityType: CorpTypeCd.BENEFIT_COMPANY,
```
severinbeauvais, line 21: I'm not sure how this code worked for CR/UL/CC previously ... Hmm, everything was not a BEN until now, was it?

src/mixins/name-request-mixin.ts
```diff
@@ -112,6 +116,21 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
...
+   */
+  nrTypeToEntityType (nr: any): CorpTypeCd {
+    switch (nr.entity_type_cd) {
+      case NameRequestTypes.BC: return CorpTypeCd.BENEFIT_COMPANY
+      case NameRequestTypes.CR: return CorpTypeCd.BC_CORPORATION
+      case NameRequestTypes.UL: return CorpTypeCd.BC_UNLIMITED
+      case NameRequestTypes.CC: return CorpTypeCd.BC_CCC
+      case NameRequestTypes.CP: return CorpTypeCd.COOP
+    }
+    return null // unknown entity type code
```
severinbeauvais, line 134: This seems like a trivial method, but it does handle BC -> BEN, and it also returns null for NR entity type codes we don't support yet.

src/mixins/name-request-mixin.ts
severinbeauvais: Oh, I guess the previous code worked because we got the entity type from the draft IA, not from the NR (and we didn't check if they matched).

tests/unit/Actions.spec.ts
```diff
@@ -36,8 +36,8 @@ const nrData = {
   },
   consentFlag: 'R',
   corpNum: null,
+  entity_type_cd: 'BC',
```
severinbeauvais, line 39: The sample NRs I've seen have both `entity_type_cd` and `requestTypeCd` set to the same value.

Does anyone know for sure which one to look at???

tests/unit/App.spec.ts
```diff
@@ -481,7 +481,7 @@ describe('App component', () => {
 
   it('loads a name request into the store', () => {
     // Validate Name Request
-    expect(store.state.stateModel.entityType).toBe(nrData.requestTypeCd)
+    expect(store.state.stateModel.entityType).toBe('BEN')
```
severinbeauvais, line 484: "BC" was mapped to "BEN".

src/components/ReviewConfirm/IncorporationDateTime.vue
```diff
@@ -227,16 +227,13 @@ export default class IncorporationDateTime extends Mixins(DateMixin) {
     if (this.incorporationDateTime) {
       // Set the chosen filing time option
       this.selectDate = this.incorporationDateTime.isFutureEffective ? ISFUTUREEFFECTIVE : ISIMMEDIATE
-      this.isFutureEffective = this.selectDate === ISFUTUREEFFECTIVE
+      this.isFutureEffective = (this.selectDate === ISFUTUREEFFECTIVE)
 
-      // Reference Store and create date object
       const effectiveDate = this.incorporationDateTime.effectiveDate
-      const dateToParse = effectiveDate && new Date(effectiveDate)
```
severinbeauvais, line 234: `effectiveDate` is already a Date object.

src/components/ReviewConfirm/IncorporationDateTime.vue
```diff
@@ -368,13 +364,13 @@ export default class IncorporationDateTime extends Mixins(DateMixin) {
   /** Emit DateTime Valid event. */
   @Emit('valid')
   private emitIsValidDateTime (): boolean {
-    return this.isImmediate ? true
-      : this.isFutureEffective && this.isValidDateTime(this.incorporationDateTime.effectiveDate)
+    if (this.isImmediate) return true
+    return (this.isFutureEffective && this.isValidDateTime(this.incorporationDateTime.effectiveDate))
```
severinbeauvais, line 368: Simplification of ternary expression.

src/mixins/filing-template-mixin.ts
```diff
@@ -48,11 +49,6 @@ export default class FilingTemplateMixin extends Vue {
 
   /** Constructs a filing body from store data. Used when saving a filing. */
   buildFiling (): IncorporationFilingIF {
-    // Format DateTime for filing.
-    const effectiveDate = this.getIncorporationDateTime.effectiveDate
-    const formattedDateTime = effectiveDate &&
-      (effectiveDate.toISOString()).replace('Z', '+00:00')
```
severinbeauvais, line 54: I just moved this stuff down to where it's used.

src/views/ReviewConfirm.vue
```diff
@@ -149,8 +149,7 @@ export default class ReviewConfirm extends Mixins() {
   }
 
   /** Handler for setting DateTime. */
-  // TODO: fix type (should be Date)
-  private setDateTime (val: string): void {
+  private setDateTime (val: Date): void {
```
severinbeauvais, line 152: Easiest. Fix. Ever.

eve-git (approved):
LGTM

src/App.vue
```diff
@@ -504,25 +503,35 @@ export default class App extends Mixins(DateMixin, FilingTemplateMixin, LegalApi
...
         this.nameRequestInvalidType = NameRequestStates.INVALID
         this.nameRequestInvalidErrorDialog = true
         return
       }
 
-      // TODO: verify that nrResponse.entityTypeCd === business.legalType
+      // ensure types match
+      if (this.nrTypeToEntityType(nrResponse) !== this.getEntityType) {
```
cameron-freshworks, line 513: Is this to protect against NR Type changes? 
Thats the only scenario i can think of where this might get out of sync, ie an IA is created under an original nr type which is then edited later?

cameron-freshworks (approved):
Looks good to me!

src/App.vue
severinbeauvais: This was a left-over TODO from very early one. I just figured it was wise to verify that the correct NR type is being used.

Also I think we once said that the NR type would determine the business type, although I think that's now determined earlier in the process (ie, when affiliating the NR or creating the draft IA).

Your use case is a good example of a check we'd want. We'd also possibly that check in the MB table or entity dashboard.

cameron-freshworks: Interesting! Lets leave this comment open but don't let it hold your PR back 👍

PR #227, opened 2021-08-19
Updated Type Names

cameron-freshworks:
*Issue #:* /bcgov/entity#8514

*Description of changes:*
* Updated type names to match requirements update.

### DIFF
--- a/src/enums/coopType.ts
+++ b/src/enums/coopType.ts
@@ -1,5 +1,5 @@
 export enum CoopType {
-  COOPERATIVE = 'Cooperative Without a Special Purpose',
-  HOUSING_COOPERATIVE = 'Special Purpose - Housing Cooperative',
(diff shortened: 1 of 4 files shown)


src/enums/coopType.ts
```diff
@@ -1,5 +1,5 @@
 export enum CoopType {
-  COOPERATIVE = 'Cooperative Without a Special Purpose',
```
severinbeauvais, line 2: Hah, I was wondering if this type of coop was having an existential crisis or something.

severinbeauvais (approved):
Merge when ready.

src/enums/coopType.ts
cameron-freshworks: hahaha... oh man. 
This was the original iteration too! Too bad we couldn't leave it like this 😆

PR #230, opened 2021-08-24
8352 More UI tweaks

severinbeauvais:
*Issue #:* /bcgov/entity#8352

*Description of changes:*
- deleted obsolete colour #1976d2
- fixed typo
- updated blue text to #1669bb on several pages
- updated X icons to #d3272c on several pages
- added bullet when no share classes (and not error)
- fixed gray9 hex code
- created override styles for disabled text fields
- added whitespace above Mailing Address (ie, below Roles)
- added whitespace above Delivery Address (ie, below checkbox)
- app version: 2.0.15


cameron-freshworks (approved):

src/assets/styles/overrides.scss
```diff
@@ -40,6 +40,16 @@
   transform: translateY(-8px) scale(0.75);
 }
 
+// Disabled Text Fields
+.v-text-field.v-input--is-readonly {
+  pointer-events: none;
```
severinbeauvais, line 45: Disabling pointer events means Vuetify doesn't make this look enabled when you click on it.

src/assets/styles/theme.scss
```diff
@@ -28,7 +28,6 @@ $BCgovABlue1: #6e93ff;
 $BCgovABlue2: #3b6cff;
 $BCgovABlue3: #0846ff;
 $BCgovABlue4: #003bee;
-$primary-blue: #1976d2;
```
severinbeauvais, line 31: Obsolete. We need to use `#1669bb` instead (primary or app-blue).

src/components/AddPeopleAndRoles/PeopleAndRoles.vue
```diff
@@ -29,35 +29,35 @@
...
-            <v-icon v-else-if="getShowErrors" color="red" class="cp-invalid">mdi-close</v-icon>
+            <v-icon v-else-if="getShowErrors" color="error" class="cp-invalid">mdi-close</v-icon>
             <v-icon v-else>mdi-circle-small</v-icon>
             <span class="chk-list-item-txt">{{rule.text}}</span>
           </li>
           <li v-if="rule.rule === Rules.NUM_INCORPORATORS" :key="index">
             <v-icon v-if="validMinimumIncorporators" color="green darken-2"
               class="incorp-valid">mdi-check</v-icon>
-            <v-icon v-else-if="getShowErrors" color="red" class="incorp-invalid">mdi-close</v-icon>
+            <v-icon v-else-if="getShowErrors" color="error" class="incorp-invalid">mdi-close</v-icon>
```
severinbeauvais, line 39: "error" is a named Vuetify colour style with the correct hex value.

src/components/common/Stepper.vue
```diff
@@ -135,7 +135,7 @@ export default class Stepper extends Vue {
   }
 
   .v-btn:before {
-    background-color: #1976d2;
+    background-color: $app-blue;
```
severinbeauvais, line 138: I was skeptical about changing this colour (and below), but the stepper looked the same before and after -- I think some other overrides are actually setting the colour here (and/or below). This can be updated in a future ticket if needed.

src/main.ts
```diff
@@ -72,7 +72,7 @@ async function start () {
             success: '#1a9031', // same as $app-green
             error: '#d3272c', // same as $app-red
             gray7: '#495057', // same as $gray7
-            gray9: '#495057' // same as $gray9
+            gray9: '#212529' // same as $gray9
```
severinbeauvais, line 75: Not sure why this was wrong -- and it feels risky to fix it since many pages use gray9 -- but now this is correct and any pages that are now wrong can be updated in a future ticket if needed.

PR #244, opened 2021-09-17
8433 COOP IA upload memorandum - move scrollToComponent to page level, UI text updates

argush3:
*Issue #:* /bcgov/entity#8433

*Description of changes:*

* moved `scrollToComponent` from component to page level for upload rules and upload memorandum components.  Note `scrollToComponent` can be probably moved to the root component(`App`) but have held off on doing this as I think it would make sense that if we decide to do this we would want to move all the other `scrollToComponent` calls to the `App` level.  
* update tooltip text and other component text based off of UXA feedback

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "2.0.23",
+  "version": "2.1.0",
   "private": true,
   "appName": "Create UI",
   "sbcName": "SBC Common Components",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "2.0.22",
+  "version": "2.0.23",
   "private": true,
   "appName": "Create UI",
   "sbcName": "SBC Common Components",
--- a/src/components/CreateRules/UploadRulesSummary.vue
+++ b/src/components/CreateRules/UploadRulesSummary.vue
@@ -5,7 +5,7 @@
       <label class="upload-rules-title pl-2"><strong>Rules</strong></label>
     </div>
 
-    <div v-if="!getCreateRulesStep.valid" class="invalid-section">
+    <div v-if="!getCreateRulesStep.validationDetail.valid" class="invalid-section">
       <div class="upload-rules-error-message">
         <span>
           <v-icon color="error">mdi-information-outline</v-icon>
@@ -17,7 +17,7 @@
       </div>
     </div>
 
-    <div v-if="getCreateRulesStep.valid" class="upload-rules-success-message">
+    <div v-if="getCreateRulesStep.validationDetail.valid" class="upload-rules-success-message">
       <v-row no-gutters>
         <v-col md="1">
           <v-icon class="upload-success-chk">mdi-check</v-icon>
--- a/src/mixins/common-mixin.ts
+++ b/src/mixins/common-mixin.ts
@@ -1,5 +1,6 @@
 import { Component, Vue } from 'vue-property-decorator'
 import { omit, isEqual } from 'lodash'
+import { ValidationItemDetailIF } from '@/interfaces'
 
 /**
  * Mixin that provides some useful common utilities.
@@ -50,5 +51,15 @@ export default class CommonMixin extends Vue {
       return false
     }
     return true
+  }
+
+  buildValidFlags (validationItems: ValidationItemDetailIF[]): object {
+    let result = {}
+    for (const vi of validationItems) { result[vi.name] = vi.valid }
+    return result
+  }
+
+  buildElementIds (validationItems: ValidationItemDetailIF[]): object {
+    return validationItems.map(vi => vi.elementId)
   }
 }
--- a/src/mixins/common-mixin.ts
+++ b/src/mixins/common-mixin.ts
@@ -53,12 +53,22 @@ export default class CommonMixin extends Vue {
     return true
   }
 
+  /**
+   * Extracts validation flags from validation item detail object
+   * @param validation items details object
+   * @return array of validation flags for all validation items
+   */
   buildValidFlags (validationItems: ValidationItemDetailIF[]): object {
     let result = {}
     for (const vi of validationItems) { result[vi.name] = vi.valid }
     return result
   }
 
+  /**
+   * Extracts element ids from validation item detail object
+   * @param validation items details object
+   * @return array of element ids for all validation items
+   */
   buildElementIds (validationItems: ValidationItemDetailIF[]): object {
     return validationItems.map(vi => vi.elementId)
   }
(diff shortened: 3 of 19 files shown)


src/mixins/common-mixin.ts
```diff
@@ -51,4 +52,24 @@ export default class CommonMixin extends Vue {
...
+  }
+
+  /**
+   * Extracts element ids from validation item detail object
+   * @param validation items details object
+   * @return array of element ids for all validation items
+   */
+  buildElementIds (validationItems: ValidationItemDetailIF[]): object {
+    return validationItems.map(vi => vi.elementId)
+  }
```
cameron-freshworks, line 74: Cool!

src/views/CreateRules.vue
```diff
@@ -6,17 +6,55 @@
...
+  /** Called when component is created. */
+  private created (): void {
+    // ignore data changes until page has loaded
+    this.setIgnoreChanges(true)
+    Vue.nextTick(() => {
+      this.setIgnoreChanges(false)
+    })
+  }
+
+  @Watch('$route')
```
cameron-freshworks, line 43: Just to confirm, this was ok? Only being called when you expected it to be etc

cameron-freshworks (approved):
Looks good

vysakh-menon-aot (approved):

src/views/CreateRules.vue
argush3: yes, i believe this is ok and i've done a bunch of testing too.  The scrolling for each page level component should only happen when there are errors(`getShowErrors`) and the route matches the page level component.  Does that answer your question?

cameron-freshworks: For sure! 
I think Severin had concern it was called on any route change due to the way Vue hooks can create/mount components before they are displayed. If you see nothing abnormal than all good!

argush3: yeah, it is probably ideal it is moved to `App` level but I moved it to at least the page level for now.  I made a note in the description.  I do think we want to eventually move it to the root component(`App`) level but it felt like a bigger change that would involve a bit more testing so I held off for now.  If we do it, I think we would have to do some refactoring of other components to match what the upload rules/memorandum components for tracking validation items.

@severinbeauvais are you ok with what i've done so far?  or would you prefer me to try and move everything to the `App` level?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "2.0.22",
+  "version": "2.0.23",
```
severinbeauvais, line 3: Note to all (including me): let's bump this to 2.1 or 3 next time we release something big.

And/or, what would you guys think to change version to yyyy.mm.dd ?

cc @thorwolpert

src/mixins/common-mixin.ts
```diff
@@ -51,4 +52,24 @@ export default class CommonMixin extends Vue {
...
+
+  /**
+   * Extracts validation flags from validation item detail object
+   * @param validation items details object
+   * @return array of validation flags for all validation items
+   */
+  buildValidFlags (validationItems: ValidationItemDetailIF[]): object {
+    let result = {}
+    for (const vi of validationItems) { result[vi.name] = vi.valid }
+    return result
```
severinbeauvais, line 64: This function looks fine except for the return type: it's not a complete object since (I think) it doesn't contain certain object methods. I think it would be better to return type `any`. Thoughts?

severinbeauvais: (Line 62 declares an empty object so I might be wrong here.)

argush3: yes, `any` makes sense.  will get that updated.

argush3: taking another look based on your last comment...

src/mixins/common-mixin.ts
```diff
@@ -51,4 +52,24 @@ export default class CommonMixin extends Vue {
...
+  }
+
+  /**
+   * Extracts element ids from validation item detail object
+   * @param validation items details object
+   * @return array of element ids for all validation items
+   */
+  buildElementIds (validationItems: ValidationItemDetailIF[]): object {
```
severinbeauvais, line 72: Can return type be `string[]` ?

src/components/CreateRules/UploadRulesSummary.vue
```diff
@@ -5,7 +5,7 @@
       <label class="upload-rules-title pl-2"><strong>Rules</strong></label>
     </div>
 
-    <div v-if="!getCreateRulesStep.valid" class="invalid-section">
+    <div v-if="!getCreateRulesStep.validationDetail.valid" class="invalid-section">
```
severinbeauvais, line 8: Are you sure that `validationDetail` is not null? It's initialized to null in Filing Template Mixin so this line of code might error.

Same on line 20 below.

src/store/getters/state-getters.ts
```diff
@@ -158,7 +158,7 @@ export const getCreateRulesStep = (state: StateIF): CreateRulesIF => {
 
 /** Is true when the step is valid. */
 export const isRulesValid = (state: StateIF): boolean => {
-  return getCreateRulesStep(state).valid
+  return getCreateRulesStep(state).validationDetail.valid
```
severinbeauvais, line 161: Are you sure that `validationDetail` is not null? Maybe you can use optional chaining.

severinbeauvais: OK, I see you initialize this in State Model. Looks like it's OK.

src/components/CreateRules/UploadRulesSummary.vue
severinbeauvais: OK, I see you initialize this in State Model. Looks like it's OK.

src/views/CreateRules.vue
severinbeauvais: My initial thought was to have a single route change handler, but actually I think what you've done here is OK: you have a route handler on this page, and another one on the Create Memorandum page. That keeps things localized, which is good.

Less good is that this page (all pages in fact) are always rendered/instantiated, but conditionally shown - see App.vue template. (This is not true of some components, which are conditionally rendered per v-if directives.) This means all route change handlers always run. I guess you haven't seen any issues with this but it might be good to confirm this, and maybe have the handlers of hidden pages do nothing (check "if shown"?).

severinbeauvais: Oh my, I see you've already done this (checking the route)... well done! Sorry, I need to look closer :)

severinbeauvais: So - I'm totally OK with this.

severinbeauvais (approved):
Looks good. I'm leaving any final changes (if needed) with you.

src/mixins/common-mixin.ts
argush3: so it should be a complete object as you its initialized using `let result = {}` as you had noted.

package.json
argush3: updated to `2.1.0`

src/mixins/common-mixin.ts
argush3: updated to have return type of `string[]`

PR #246, opened 2021-09-24
9068 Add encryption/locked validation and component tests to FileUploadPreview

argush3:
*Issue #:* /bcgov/entity#9068

*Description of changes:*

* Tests for covering validation and emitting of events for FileUploadPreview
* Addition of encryption and content lock validation checks


argush3:
Was meaning to write some tests for file upload component but didn't quite get there before 9068 closed off.  But here they are...

src/mixins/document-mixin.ts
```diff
@@ -4,7 +4,9 @@ import { Component, Vue } from 'vue-property-decorator'
 import { DocumentUpload } from '@/interfaces'
 import { PdfPageSize } from '@/enums'
 import pdfjsLib from 'pdfjs-dist/build/pdf'
-pdfjsLib.GlobalWorkerOptions.workerSrc = 'public/js/pdf.worker.min.js'
+import pdfjsWorker from 'pdfjs-dist/build/pdf.worker.entry'
+// Note: needs to be defined this way to work in tests
+pdfjsLib.GlobalWorkerOptions.workerSrc = pdfjsWorker
```
severinbeauvais, line 9: Can `require()` be used instead?

severinbeauvais (approved):
Nice!

andrepestana-aot (approved):
Looks good. Do we have any tests for when a file is encrypted and should be rejected?

cameron-freshworks (approved):
Looks good!

argush3:
> Looks good. Do we have any tests for when a file is encrypted and should be rejected?

The FE doesn't actually currently check for encryption.  i don't think pdf.js has any api methods to check for it either.

src/mixins/document-mixin.ts
argush3: will try

severinbeauvais:
> Looks good. Do we have any tests for when a file is encrypted and should be rejected?

Or locked? (Since we might not be able to edit/stamp it ???)

andrepestana-aot:
> > Looks good. Do we have any tests for when a file is encrypted and should be rejected?
> 
> The FE doesn't actually currently check for encryption. i don't think pdf.js has any api methods to check for it either.

I'm not sure if we can trust in this example but I did some tests and it seems to work: https://stackoverflow.com/q/52081345/16940959

severinbeauvais:
> I'm not sure if we can trust in this example but I did some tests and it seems to work: https://stackoverflow.com/q/52081345/16940959

Do we have regular, encrypted (and also locked) PDFs to test with?

andrepestana-aot:
> > I'm not sure if we can trust in this example but I did some tests and it seems to work: https://stackoverflow.com/q/52081345/16940959
> 
> Do we have regular, encrypted (and also locked) PDFs to test with?

You can create protected pdfs here: https://smallpdf.com/protect-pdf

src/mixins/document-mixin.ts
argush3: updated to use `require()`

jest.config.js
```diff
@@ -7,6 +7,6 @@ module.exports = {
     }
   },
   preset: '@vue/cli-plugin-unit-jest/presets/typescript-and-babel',
-  transformIgnorePatterns: []
-
+  transformIgnorePatterns: [],
+  testTimeout: 30000 // need to extend timeout from default of 5000ms for FileUploadPreview component tests
```
severinbeauvais, line 11: I think there's a way to override the timeout for just your test / test file. That would be preferable.

argush3: ok, yeah that makes sense

src/mixins/document-mixin.ts
```diff
@@ -3,8 +3,8 @@ import axios from '@/utils/axios-auth'
 import { Component, Vue } from 'vue-property-decorator'
 import { DocumentUpload } from '@/interfaces'
 import { PdfPageSize } from '@/enums'
-import pdfjsLib from 'pdfjs-dist/build/pdf'
-pdfjsLib.GlobalWorkerOptions.workerSrc = 'public/js/pdf.worker.min.js'
+const pdfjsLib = require('pdfjs-dist/build/pdf')
+pdfjsLib.GlobalWorkerOptions.workerSrc = require('pdfjs-dist/build/pdf.worker.entry')
```
severinbeauvais, line 7: I'm sure you played with this, but does the import on line 6 need `require`, or can it follow the usual import syntax?

argush3:
add check for encryption and also a test.  but wasn't able to figure out how to deal with locked files

severinbeauvais (commented):
Great improvements!

So there's no need to check for a locked PDF? (Don't we need to edit it, to stamp it for example?)

severinbeauvais:
> wasn't able to figure out how to deal with locked files

Doesn't matter if we don't need to edit them. But I'm just asking the questions - I don't really know :>

argush3:
> > wasn't able to figure out how to deal with locked files
> 
> Doesn't matter if we don't need to edit them. But I'm just asking the questions - I don't really know :>

I think we need to check for locking.  I will do some more digging as we will need to edit the doc.  That was one check we missed on the BE I think.

jest.config.js
argush3: added timeouts to specific tests

src/mixins/document-mixin.ts
argush3: reverted line 6 to back what it was and it seems to work

argush3:
> > > wasn't able to figure out how to deal with locked files
> > 
> > 
> > Doesn't matter if we don't need to edit them. But I'm just asking the questions - I don't really know :>
> 
> I think we need to check for locking. I will do some more digging as we will need to edit the doc. That was one check we missed on the BE I think.

Added content lock check validation.  Did some testing and although the BE doesn't explicitly check for this, it does return a "invalid file" error when the filing endpoint is hit as a result of an exception when trying to parse the file.

PR #247, opened 2021-09-30
Dissolutions framework

cameron-freshworks:
*Issue #:* /bcgov/entity#8838

*Description of changes:*
* Directory restructure to isolate Incorporations and Dissolutions.
* Base config files for supported dissolution entity types.
* Baseline filing template methods to create/parse drafts.  *Needs expanding as we build out filing properties.
* Some code clean up.
* Testing updates.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "2.1.2",
+  "version": "2.1.3",
```
severinbeauvais, line 3: I'm not sure how to go about this, but IMO Coop IAs should increase the major number, and so should Dissolutions.

cameron-freshworks: Very good point, we should developing dissolutions at 2.2.x at the very least eh?

severinbeauvais: Since we're done with Coop IAs (I think), let's just go to 3.0.0 ? 🤷‍♂️

src/App.vue
```diff
@@ -22,6 +22,12 @@
       @retry="fetchData()"
     />
 
+    <invalid-dissolution-dialog
```
severinbeauvais, line 25: I recommend changing our component elements to CamelCase (this one and the others, at least in this file).

src/App.vue
```diff
@@ -125,34 +131,40 @@
...
 
 // Components
 import PaySystemAlert from 'sbc-common-components/src/components/PaySystemAlert.vue'
 import SbcHeader from 'sbc-common-components/src/components/SbcHeader.vue'
 import SbcFooter from 'sbc-common-components/src/components/SbcFooter.vue'
 import SbcFeeSummary from 'sbc-common-components/src/components/SbcFeeSummary.vue'
-import { EntityInfo, Stepper, Actions } from '@/components/common'
-import Views from '@/views'
+import { Actions, EntityInfo, Stepper } from '@/components/common'
+import * as Views from '@/views'
```
severinbeauvais, line 147: Why?

src/App.vue
cameron-freshworks: Will do 👍

src/App.vue
```diff
@@ -252,6 +279,10 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
     return process.env.ABOUT_TEXT
   }
 
+  private get filingTitle (): string {
+    return this.isIncorporationFiling ? FilingNames.INCORPORATION_APPLICATION : FilingNames.DISSOLUTION_FILING
```
severinbeauvais, line 283: Why not change `isIncorporationFiling` to `getFilingName: FilingNames` ?

This would support more types in the future than a boolean.

src/App.vue
```diff
@@ -293,6 +324,16 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
       console.log('Error while retrieving NR during File and Pay') // eslint-disable-line no-console
       this.nameRequestInvalidErrorDialog = true
     })
+
+    // Set identifier to store
+    const id = this.$route.query?.id as string
+    if (id?.startsWith('CP') || id?.startsWith('BC')) {
+      this.setBusinessId(id)
+      this.setFilingType(FilingTypes.DISSOLUTION)
```
severinbeauvais, line 332: Here you call it Filing Type, but above the enum is Filing Name. Is it all the same thing? Does this need both store properties?

src/App.vue
```diff
@@ -380,18 +421,20 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
     // reset errors in case this method is invoked more than once (ie, retry)
     this.resetFlags()
 
+    // Feature flag safety check
+    let supportedFilings = await getFeatureFlag('supported-filings')
+    if (!supportedFilings?.includes(this.$route.meta.filingType)) {
+      this.accountAuthorizationDialog = true
```
severinbeauvais, line 427: This is not an account authorization problem.

See https://github.com/bcgov/business-edit-ui/blob/c17dd44ff39cb6062b0ec47925475fac74dadb8d/src/views/Alteration.vue#L268

src/App.vue
```diff
@@ -380,18 +421,20 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
...
+    }
+
+    // Protect the routes not associated with this filing type.
+    if (this.$route.meta.filingType !== this.getFilingType) {
+      this.accountAuthorizationDialog = true
```
severinbeauvais, line 432: Again, not an account auth problem.

Maybe you could store a "reason text" so that you could reuse this dialog? Or create a new dialog.

src/App.vue
```diff
@@ -419,61 +462,100 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
...
-        }
-
-        // merge draft properties into empty filing so all properties are initialized
-        const emptyFiling = this.buildFiling()
-        draftFiling = { ...emptyFiling.filing, ...draftFiling }
-
-        // parse draft filing into the store
-        if (draftFiling) {
-          this.parseDraft(draftFiling)
+        let draftFiling, resources
```
severinbeauvais, line 466: This code looks OK except it's pretty big. I suggest making 2 sub-methods from it (and the common stuff can stay here).

src/enums/routeNames.ts
```diff
@@ -5,11 +5,19 @@
...
     INCORPORATION_AGREEMENT = 'incorporation-agreement',
-    REVIEW_CONFIRM = 'review-confirm'
+    REVIEW_CONFIRM = 'review-confirm',
+
+    // Dissolution Route Names
+    DEFINE_DISSOLUTION = 'define-dissolution',
+    UPLOAD_RESOLUTION = 'upload-special-resolution',
+    UPLOAD_AFFIDAVIT = 'upload-affidavit',
+    REVIEW_CONFIRM_DISSOLUTION = 'review-confirm-dissolution'
 }
```
severinbeauvais, line 23: What do you think of renaming these...
INCORP_XXX
INCORP_YYY
ETC
and
DISSOLUTION_XXX
DISSOLUTION_YYY
ETC
... to avoid possible duplication of values and to make is easier to know right away what the route it for?

src/enums/viewComponentNames.ts
```diff
@@ -5,11 +5,19 @@
...
   INCORPORATION_AGREEMENT = 'incorporation-agreement',
-  REVIEW_CONFIRM = 'review-confirm'
+  REVIEW_CONFIRM = 'review-confirm',
+
+  // Dissolution View Components
+  DEFINE_DISSOLUTION = 'define-dissolution',
+  UPLOAD_SPECIAL_RESOLUTION = 'upload-special-resolution',
+  UPLOAD_AFFIDAVIT = 'upload-affidavit',
+  REVIEW_CONFIRM_DISSOLUTION = 'review-confirm-dissolution'
 }
```
severinbeauvais, line 23: This looks like the same list as Route Names. Could we have just a single list? Perhaps with a comment somewhere explaining that the route names are the same as the component names.

src/interfaces/filing-interfaces/filing-interface.ts
```diff
@@ -59,3 +59,23 @@ export interface IncorporationFilingIF {
...
+      name: string
+      certifiedBy: string
+      date: string
+      effectiveDate?: string // Optional and should be set only for future effective filings
+      filingId?: number // Optional as this is not required when building a filing - causes an error for new filings
+      folioNumber?: string // Optional to the user and only displayed for certain account types
+      isFutureEffective: boolean
+    },
+    business: {
+      legalType: string,
```
severinbeauvais, line 76: Do we have a type for this?

src/interfaces/store-interfaces/state-interface.ts
```diff
@@ -45,5 +45,5 @@ export interface StateModelIF {
 
 export interface StateIF {
   stateModel: StateModelIF
-  resourceModel: ResourceIF
+  resourceModel: IncorporationResourceIF
```
severinbeauvais, line 48: Should this be type `IncorporationResourceIF | DissolutionResourceIF` ?

src/mixins/auth-api-mixin.ts
```diff
@@ -0,0 +1,27 @@
...
+   * Fetches authorizations.
+   * @param businessIdentifier the business identifier (eg, BC1219948)
+   * @returns a promise to return the authorizations object
+   */
+  async fetchAuthorizations (businessIdentifier: string): Promise<any> {
+    if (!businessIdentifier) throw new Error('Invalid parameter \'businessIdentifier\'')
+
+    const url = `entities/${businessIdentifier}/authorizations`
+    const authUrl = sessionStorage.getItem('AUTH_API_URL')
+    const config = { baseURL: authUrl }
```
severinbeauvais, line 23: If this gets complicated in unit testing, we can just make `authURL` part of `url` and not need to specify baseURL.

src/mixins/legal-api-mixin.ts
```diff
@@ -23,7 +23,7 @@ export default class LegalApiMixin extends Vue {
    * Fetches a draft filing.
    * @returns a promise to return the draft filing, or null if not found
    */
-  async fetchDraft (): Promise<any> {
+  async fetchDraftIA (): Promise<any> {
     // get the draft filing from the tasks endpoint
     const url = `businesses/${this.getTempId}/filings`
```
severinbeauvais, line 28: Uh, this isn't the tasks endpoint (per comment above)...?

I'm confused (i) why this works, and (ii) if this works, why?

src/mixins/legal-api-mixin.ts
```diff
@@ -48,19 +48,49 @@ export default class LegalApiMixin extends Vue {
...
+          throw new Error('Invalid API response')
+        }
+
+        // save Filing ID from the header
+        this.setFilingId(filingId)
+        return filing
+      })
+      .catch((error) => {
+        if (error?.response?.status === NOT_FOUND) {
+          return null // Dissolution not found
```
severinbeauvais, line 74: Add a comment why you need this (ie, a dissolution may or may not exist and you need to handle both cases).

src/App.vue
cameron-freshworks: Because we now have multiple directories in Views, with the root of `views` exporting multiple directories from index. 
We need to assign these multiple exports a reference in App.vue before we can `spread` it, this is what worked at least!

src/App.vue
cameron-freshworks: Good idea, will do 👍

src/App.vue
cameron-freshworks: Two different properties and values. 
 Filing Name for instance `Incorporation Application` 
 Filing type as lear defines it `incorporationApplication`
 and so on. 
 So the filing name is most often used for display purposes and the type is the data.

src/App.vue
cameron-freshworks: Fair enough, i used this one as it was generic enough without giving a user too much information. Not sure we want to explicitly state 'What' isn't available and 'why' though.

src/store/actions/actions-model.ts
```diff
@@ -1,17 +1,25 @@
...
+  commit('mutateFilingType', filingType)
+}
+
 export const setEntityType = ({ commit }, entityType: CorpTypeCd): void => {
   commit('mutateEntityType', entityType)
 }
 
-export const setCompanyResources = ({ commit }, companyResources: ResourceIF): void => {
-  commit('mutateCompanyResources', companyResources)
+export const setResources = ({ commit }, companyResources: IncorporationResourceIF): void => {
```
severinbeauvais, line 21: Again, you probably need a union type here.

src/App.vue
cameron-freshworks: My initial pass was as you suggest, ended up combining them. Can break them out though, no problem :)

src/enums/routeNames.ts
cameron-freshworks: I like that the route names match the steps, and subsequently what the user sees on screen. Imo creating a disconnect between route name and the steps would make it tougher to know right away what the route is for no?

src/enums/viewComponentNames.ts
cameron-freshworks: This came up before, it was your recommendation we have separate enums despite the values being the same, as you didn't like the idea of using the same enum for separate properties in the routes, you are keeping me on my toes here haha

src/App.vue
severinbeauvais: Ah, it's because you're (re)exporting 3 things in src/views/index.ts instead of just one thing as before.

Another way to do this is to import all 3 things and then export them as one.

src/enums/viewComponentNames.ts
cameron-freshworks: Screen Shot 2021-09-30 at 2 21 27 PM

src/interfaces/filing-interfaces/filing-interface.ts
cameron-freshworks: Ah yes, we should

src/interfaces/store-interfaces/state-interface.ts
cameron-freshworks: Yes, good catch!

src/mixins/auth-api-mixin.ts
cameron-freshworks: kk

tests/unit/DefineCompany.spec.ts
```diff
@@ -1,5 +1,5 @@
 import { shallowWrapperFactory } from '../jest-wrapper-factory'
-import DefineCompany from '@/views/DefineCompany.vue'
+import DefineCompany from '@/views/Incorporations/DefineCompany.vue'
```
severinbeauvais, line 2: You might be able to just import from the index file, eg,
`import { DefineCompany } from '@/views'`

src/mixins/legal-api-mixin.ts
cameron-freshworks: I was also confused. I deleted the comment but must have restored it when i created a new endpoint altogether. 
Not sure why the comment said tasks unless this was refactored at some point. 

But yea, will remove the comment. It's a unique request to `/filings` where there is only ever 1 response,
ie `filing` where a normal request will return a `filings` array with all completed/paid filings. 

The divergence is a result of the temp reg number as the identifier, it takes a new flow in the api!

src/store/actions/actions-model.ts
cameron-freshworks: 100%, thanks!

tests/unit/DefineCompany.spec.ts
cameron-freshworks: Will try!

src/App.vue
severinbeauvais: That's fair. How about asking PO/UX?

package.json
cameron-freshworks: Sure, lets go!

src/App.vue
cameron-freshworks: Cool, will look at it!

src/App.vue
severinbeauvais: It will make it easier to review :)

If I was looking at the code in an IDE then I'd just fold the sub-sections for the same effect.

cameron-freshworks: Totally, i 100% agree with you. I can improve this 👍

src/enums/viewComponentNames.ts
severinbeauvais: Haha, ok, nvm.

src/App.vue
cameron-freshworks: Good point, maybe we should ask. PO and designers. 
What 'do' we want to tell users trying to manually enter in routes that are not available yet 😆

cameron-freshworks: Suspicious characters out there..

src/interfaces/filing-interfaces/filing-interface.ts
severinbeauvais: CorpTypeCd ?

src/interfaces/store-interfaces/state-interface.ts
severinbeauvais: Finally a good catch! All my other comments were fluff 😛

src/mixins/legal-api-mixin.ts
severinbeauvais: Ah, gotcha.

Come to think of it, Filings UI does the same thing: https://github.com/bcgov/business-filings-ui/blob/c3d2a9c87604a25e64d9cac0fd773e83c0e7a7bd/src/App.vue#L514

I generally like comments. It might be worth keeping the comment and explaining this unique case. (I will do this in Filings UI.)

severinbeauvais: image

package.json
severinbeauvais: What this means is that the final version of a completed feature might be, say, 2.1.3 (or later if we have to bugfix it after deployment), while the initial (draft/alpha) version of a new feature might be, say, 3.0.0. That seems OK to me :)

src/App.vue
severinbeauvais: It would take a savvy character but our code IS public so ... yeah, you make a good point. 😁

src/App.vue
cameron-freshworks: Updated in next commit 👍

src/App.vue
cameron-freshworks: Good fix here, refactored and applied in `EntityInfo` as well, in next commit.

src/App.vue
cameron-freshworks: For the sake of consistency, i've followed the same pattern you went with in Alterations :)
Thanks for pointing this out!

src/App.vue
cameron-freshworks: Updated with new dialog in next commit.

src/App.vue
cameron-freshworks: updated in next commit!

src/interfaces/filing-interfaces/filing-interface.ts
cameron-freshworks: Indeed! Updated

src/interfaces/store-interfaces/state-interface.ts
cameron-freshworks: haha psh, hardly. This was a good reminded of some pr tech debt i had left behind!

src/mixins/legal-api-mixin.ts
cameron-freshworks: Perfect! Updated

src/mixins/legal-api-mixin.ts
cameron-freshworks: I don't think it's needed. I followed the pattern of the Temp REG request, but this error is handled if any of the operations failed inside the `.then()` and bubbles up to the try/catch in App.vue
Removed 👍

src/store/actions/actions-model.ts
cameron-freshworks: Still working through this one, same as above. Right now typescript is only accepting one and throwing errors on any resource getters that pulls on a value not in both interfaces.

src/mixins/legal-api-mixin.ts
severinbeauvais: OK. As you mentioned this morning, this UI always starts with a draft Disso.

src/store/actions/actions-model.ts
severinbeauvais: You might to explicitly cast props or return values. If you give me a specific example then I can try to figure it out. (I think I had to do something similar recently.)

severinbeauvais: Here's what I had. Basically I couldn't directly do `item.comments = comments` due to the return value of fetchComments. My solution was the ternary on line 797.

image

cameron-freshworks: Updated in next commit!

tests/unit/DefineCompany.spec.ts
cameron-freshworks: This works! Updated

src/App.vue
```diff
@@ -79,7 +91,7 @@
           <v-row>
             <v-col cols="12" lg="9">
               <header>
-                <h1>Incorporation Application</h1>
+                <h1>{{ getFilingName }}</h1>
```
severinbeauvais, line 94: Is other UIs, `name` is actually the filing type.
Here, is it actually the title? Maybe need to clean up some naming.

severinbeauvais: Meh, maybe this is OK. I see that `FilingNames` is an enum with the user-friendly filing name, just as this getter suggests.
The problem lies with the API field `name` that is actually the type. I dunno how to solve this atm. Just beware of this for now.

src/enums/routeNames.ts
severinbeauvais: nvm

src/mixins/legal-api-mixin.ts
severinbeauvais: This code is now gone.

src/interfaces/filing-interfaces/filing-interface.ts
```diff
@@ -59,3 +60,23 @@ export interface IncorporationFilingIF {
...
+      effectiveDate?: string // Optional and should be set only for future effective filings
+      filingId?: number // Optional as this is not required when building a filing - causes an error for new filings
+      folioNumber?: string // Optional to the user and only displayed for certain account types
+      isFutureEffective: boolean
+    },
+    business: {
+      legalType: CorpTypeCd,
+      identifier: string
+    },
+    dissolution: {}
```
severinbeauvais, line 80: If this works, great! If not, you can use `any` and a comment indicating this is a untyped object. Or something like that.

src/mixins/legal-api-mixin.ts
```diff
@@ -47,20 +48,65 @@ export default class LegalApiMixin extends Vue {
         throw error
       })
   }
+  private mockResponse = [
```
severinbeauvais, line 51: Make sure you delete this before your final commit :)

src/mixins/legal-api-mixin.ts
```diff
@@ -47,20 +48,65 @@ export default class LegalApiMixin extends Vue {
...
+  /**
+   * Fetches a draft dissolution filing.
+   * @returns a promise to return the draft filing, or null if not found
+   */
+  async fetchDraftDissolution (): Promise<any> {
+    // get the draft filing from the tasks endpoint
+    const url = `businesses/${this.getBusinessId}/tasks`
+    return axios.get(url)
+      .then(response => {
+        response.data.tasks = this.mockResponse
```
severinbeauvais, line 81: And this.

src/store/getters/state-getters.ts
```diff
@@ -25,6 +24,31 @@ import { getMaxStep } from './resource-getters'
...
+
+/** The current filing name. */
+export const getFilingName = (state: StateIF): FilingNames => {
+  switch (getFilingType(state)) {
+    case FilingTypes.INCORPORATION_APPLICATION:
+      return FilingNames.INCORPORATION_APPLICATION
+    case FilingTypes.DISSOLUTION:
+      return FilingNames.DISSOLUTION_FILING
+  }
+}
```
severinbeauvais, line 50: Add a default case in the switch or a fallback return here.

I guess this didn't cause a lint error because it's smart enough to know there are only 2 cases, and you handle them both. Still, the code can be more robust here.

src/mixins/legal-api-mixin.ts
cameron-freshworks: haha yes :) Thanks!

src/mixins/legal-api-mixin.ts
cameron-freshworks: on it, thank you!

src/store/getters/state-getters.ts
cameron-freshworks: K, will add a default / fall back to the enum, otherwise we will have to modify the return type. Unless the latter is more appealing, thoughts?

src/interfaces/filing-interfaces/filing-interface.ts
cameron-freshworks: It works, and is going to be built on as soon as we being developing the filing itself. 
I'm not too worried about this at the moment :)

src/App.vue
cameron-freshworks: For sure, i just followed the api value in this case. I do agree

src/store/getters/state-getters.ts
severinbeauvais: Don't change the enum. Just return null (which is, handily, falsy).

cameron-freshworks: oh right on!

severinbeauvais: Perfect.

severinbeauvais (approved):
Just one comment left but either way I approve merging these changes. (Would be nice to have another reviews though.)

src/mixins/legal-api-mixin.ts
```diff
@@ -6,34 +6,35 @@ import { NOT_FOUND } from 'http-status-codes'
...
+import { FilingTypes } from '@/enums'
 
 /**
  * Mixin that provides integration with the Legal API.
  */
 @Component({})
 export default class LegalApiMixin extends Vue {
-  readonly INCORPORATION_APPLICATION = 'incorporationApplication'
-
+  @Getter getBusinessId!: string
```
argush3, line 16: I believe business number is a number in the db.  is there an issue if it isn't a string?

andrepestana-aot (approved):
Looks really good! I like the resources thing to keep some pre baked objects!

src/mixins/legal-api-mixin.ts
cameron-freshworks: This property is referring to the business identifier, which is a string in the response or request schemas. 
ie `BC1234567`

cameron-freshworks: Screen Shot 2021-10-04 at 9 24 03 AM

cameron-freshworks: Sorry, crummy screenshot there. Doesn't give much context 😆 
But yes, straight from schema for the business.

argush3: oh ok, yeah that needs a string.

cameron-freshworks:
Just going to add a new icon, provided from Tracey. Hold off on merge 😁

cameron-freshworks:
> Just going to add a new icon, provided from Tracey. Hold off on merge 😁

Edit* It wasn't read, so never mind. Merging :)

PR #273, opened 2021-10-29
9051 Finished Effective Date Time integration

severinbeauvais:
*Issue #:* /bcgov/entity#9051

*Description of changes:*

This still has the component local to the project. The next step is to move the component to the bcrs-shared-components repo and then import it here.

- improved fee summary affix layout
- cleaned up App classes
- updated Dissolution Date and Time validation
- added prop to parse initial FED
- added support for saving/restoring null FED option
- added create methods to ignore loading changes
- app version: 3.0.15

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -81,23 +81,26 @@
       </div>
     </transition>
 
-    <sbc-header />
+    <SbcHeader />
+
     <PaySystemAlert />
+
     <div class="app-body">
       <main v-if="!isErrorDialog">
-        <entity-info />
+        <EntityInfo />
 
-        <v-container class="view-container pt-8">
+        <v-container id="container-main" class="py-8">
           <v-row>
             <v-col cols="12" lg="9">
               <header>
                 <h1>{{ getFilingName }}</h1>
               </header>
 
-              <stepper class="mt-10" />
+              <Stepper class="mt-10" />
+
               <!-- Sign in and sign out components -->
-              <sign-in v-if="isRouteName(RouteNames.SIGN_IN)" />
-              <sign-out v-if="isRouteName(RouteNames.SIGN_OUT)" />
+              <SignIn v-if="isRouteName(RouteNames.SIGN_IN)" />
+              <SignOut v-if="isRouteName(RouteNames.SIGN_OUT)" />
 
               <!-- Only render when data is ready, or validation can't be properly evaluated. -->
               <template v-if="haveData">
@@ -111,10 +114,10 @@
               </template>
             </v-col>
 
-            <v-col cols="12" lg="3" style="position: relative" class="mt-2">
+            <v-col cols="12" lg="3">
               <aside>
-                <affix relative-element-selector=".col-lg-9" :offset="{ top: 86, bottom: 12 }">
-                  <sbc-fee-summary
+                <affix relative-element-selector=".col-lg-9" :offset="{ top: 100, bottom: -100 }">
+                  <SbcFeeSummary
                     :filingData="feeFilingData"
                     :payURL="payApiUrl"
                   />
@@ -124,14 +127,14 @@
           </v-row>
         </v-container>
 
-        <actions
+        <Actions
           :key="$route.path"
           @goToDashboard="goToDashboard()"
         />
       </main>
     </div>
 
-    <sbc-footer :aboutText=aboutText />
+    <SbcFooter :aboutText=aboutText />
   </v-app>
 </template>
 
@@ -868,8 +871,8 @@ export default class App extends Mixins(
 </script>
 
 <style lang="scss" scoped>
-// place app header on top of dialogs (and therefore still usable)
-.app-header {
-  z-index: 1000;
+aside {
+  position: relative;
+  z-index: 10; // on top of stepper
 }
 </style>
--- a/src/assets/styles/layout.scss
+++ b/src/assets/styles/layout.scss
@@ -9,7 +9,9 @@
   position: sticky;
   top: 0;
   width: 100%;
-  z-index: 2;
+  // place app header on top of dialogs
+  // and therefore still usable when a dialog is displayed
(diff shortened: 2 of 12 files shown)


src/assets/styles/layout.scss
```diff
@@ -9,7 +9,9 @@
   position: sticky;
   top: 0;
   width: 100%;
-  z-index: 2;
+  // place app header on top of dialogs
+  // and therefore still usable when a dialog is displayed
+  z-index: 1000;
```
severinbeauvais, line 14: I moved this here because this app-header block already existed (no need to have both global and local css config).

src/assets/styles/layout.scss
```diff
@@ -36,22 +38,6 @@
   max-width: 1360px; // should match auth-web, etc
 }
 
-// Page Layout
-.view-container {
```
severinbeauvais, line 40: This wasn't actually used.

src/mixins/legal-api-mixin.ts
```diff
@@ -92,10 +92,12 @@ export default class LegalApiMixin extends Vue {
 
     return axios.put(url, filing).then(response => {
       const filing = response?.data?.filing
-      const filingId = +filing?.header?.filingId
+      const filingId = +filing?.header?.filingId || 0
```
severinbeauvais, line 95: I like 0 better than undefined :)

This is also what the other methods in this class do.

src/App.vue
```diff
@@ -81,23 +81,26 @@
...
+
     <PaySystemAlert />
+
     <div class="app-body">
       <main v-if="!isErrorDialog">
-        <entity-info />
+        <EntityInfo />
 
-        <v-container class="view-container pt-8">
+        <v-container id="container-main" class="py-8">
```
severinbeauvais, line 92: The class wasn't actually used but I added the id in case we need to find this element.

src/App.vue
```diff
@@ -111,10 +114,10 @@
               </template>
             </v-col>
 
-            <v-col cols="12" lg="3" style="position: relative" class="mt-2">
+            <v-col cols="12" lg="3">
               <aside>
-                <affix relative-element-selector=".col-lg-9" :offset="{ top: 86, bottom: 12 }">
-                  <sbc-fee-summary
+                <affix relative-element-selector=".col-lg-9" :offset="{ top: 100, bottom: -100 }">
```
severinbeauvais, line 119: The offset is a bit "voodoo". Through trial and error, I found that these values work better for most screen sized (except very narrow/short). There's an [existing ticket](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/9809) to look into this more closely.

src/App.vue
```diff
@@ -868,8 +871,8 @@ export default class App extends Mixins(
 </script>
 
 <style lang="scss" scoped>
-// place app header on top of dialogs (and therefore still usable)
-.app-header {
-  z-index: 1000;
+aside {
```
severinbeauvais, line 874: This is so the aside (Fee Summary) is never below the stepper icons.

src/mixins/filing-template-mixin.ts
```diff
@@ -272,7 +272,7 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
           name: FilingTypes.DISSOLUTION,
           certifiedBy: this.getCertifyState.certifiedBy,
           date: this.getCurrentDate,
-          isFutureEffective: this.getEffectiveDateTime.isFutureEffective
+          isFutureEffective: null
```
severinbeauvais, line 275: The filing now supports 3 states:
- null: no option selected
- true: user selected future effective
- false: user selected immediate

src/views/Dissolutions/DefineDissolution.vue
```diff
@@ -59,12 +59,24 @@ import { EntityFilterMixin, EnumMixin } from '@/mixins'
...
+  // Global actions
+  @Action setIgnoreChanges!: ActionBindingIF
+
+  /** Called when component is created. */
+  private created (): void {
+    // ignore data changes until page has loaded
+    this.setIgnoreChanges(true)
+    Vue.nextTick(() => {
+      this.setIgnoreChanges(false)
+    })
```
severinbeauvais, line 77: This was needed to ignore the initial changes on load. The test for this is:
1. load the filing
2. reload the filing, and watch for no "you have unsaved changes" alert

src/mixins/legal-api-mixin.ts
cameron-freshworks: Is it to prevent a typeError and allow the method to handle an invalid response?

tests/unit/ReviewConfirmDissolution.spec.ts
```diff
@@ -5,33 +5,41 @@ import { ReviewConfirmDissolution } from '@/views/Dissolutions'
...
-    hasEffectiveDateTime: false
+    isStaff: true
+  },
+  {
+    entityType: 'CP',
+    isStaff: false
   },
   {
     entityType: 'BEN',
-    headerTitle: 'Review and Confirm',
```
cameron-freshworks, line 13: haha i think this was more of an 'example' than anything 😛

cameron-freshworks (approved):
Looks good!

src/mixins/legal-api-mixin.ts
severinbeauvais: No, I just prefer filingId to be a value instead of no value. I think the previous code worked just as well.

tests/unit/ReviewConfirmDissolution.spec.ts
severinbeauvais: Well, now it's evolved a bit :)

andrepestana-aot (approved):
LGTM

PR #274, opened 2021-11-03
9051 Imported shared Effective Date Time components

severinbeauvais:
*Issue #:* /bcgov/entity#9051

*Description of changes:*
- app version: 3.0.17
- imported shared DatePicker and EffectiveDateTime components
- imported new enum
- imported new interface
- deleted local components/ enum/ interface/ unit test
- commented out code to restore Future Effective data (per PO)
- misc cleanup


package.json
```diff
@@ -17,9 +17,10 @@
     "@bcrs-shared-components/contact-info": "1.0.18",
     "@bcrs-shared-components/corp-type-module": "1.0.4",
     "@bcrs-shared-components/court-order-poa": "1.0.11",
-    "@bcrs-shared-components/date-picker": "1.1.2",
-    "@bcrs-shared-components/enums": "1.0.13",
-    "@bcrs-shared-components/interfaces": "1.0.23",
+    "@bcrs-shared-components/date-picker": "1.1.5",
```
severinbeauvais, line 20: DatePicker didn't change (other than its package.json) so there's no harm in updating it here.

EffectiveDateTime, below, is a new import (the gist of this PR).

Enums and Interfaces, below, contain new items.

src/components/index.ts
```diff
@@ -0,0 +1,4 @@
+// external components
+export { CourtOrderPoa } from '@bcrs-shared-components/court-order-poa'
+export { DatePicker } from '@bcrs-shared-components/date-picker'
+export { EffectiveDateTime } from '@bcrs-shared-components/effective-date-time'
```
severinbeauvais, line 4: This is so that components in this app can import shared components the same was as local components. This is particulaly so that EffectiveDateTime can import DatePicker, both in the shared repo (ie, for tests + Storybook) and in this app.

src/views/Dissolutions/ReviewConfirmDissolution.vue
```diff
@@ -232,8 +231,8 @@ import { DateMixin } from '@/mixins'
 
 // Components
 import { AssociationDetails, DissolutionStatement } from '@/components/DefineDissolution'
-import { EffectiveDateTime, StaffPayment } from '@/components/common'
-import { CourtOrderPoa } from '@bcrs-shared-components/court-order-poa'
+import { CourtOrderPoa, EffectiveDateTime } from '@/components'
```
severinbeauvais, line 234: Example of loading shared components. In fact, this component doesn't know whether it's a shared component or not (which is good).

I didn't change over all the other shared imports, but at some point I think we should do that. It's not a huge improvement (so no rush) but it's a bit of cleanup.

The idea behind this: components should know as little as possible about "the world"; they should only know what they need to know. For example, this component knows that it needs these imports, but it shouldn't care where they come from.

cameron-freshworks (approved):
🌟

src/views/Dissolutions/ReviewConfirmDissolution.vue
argush3: cool

argush3 (approved):

severinbeauvais:
I'm looking into a weird new build error...

eve-git (approved):
Nice!

package.json
```diff
@@ -34,7 +35,7 @@
     "regenerator-runtime": "^0.13.7",
     "sbc-common-components": "2.4.25",
     "uuid": "^8.3.1",
-    "vue": "^2.6.11",
+    "vue": "2.6.12",
```
severinbeauvais, line 38: @cameron-freshworks @eve-git @argush3 @andrepestana-aot @vysakh-menon-aot FYI!!!

Something happened after Vue 2.6.12 that results in numerous `npm run build` errors regarding our decorators (eg, `@Component`).

I looked at the Vue release notes and didn't see any breaking changes 🤷‍♂️ 

Note that the Vue Template Compiler, below, should be the same version as Vue, or you get a build warning.

This change is not a permanent fix -- some day we will want to update our Vue version -- but that can be another ticket.

severinbeauvais:
Ignore stupid code coverage report above -- it's because I removed a component and its unit test file and instead imported the shared component 🙄

PR #277, opened 2021-11-05
9288 - Fee Summary - filing codes

andrepestana-aot:
*Issue #:* /bcgov/entity#9288

*Description of changes:*

- Configured the filingCodes to COOP dissolution;
- Added tests;
- Bumped version


src/interfaces/payment-interfaces/fees-interface.ts
```diff
@@ -16,7 +16,7 @@ export interface TaxesIF {
   gst: number
 }
 
-export const EmptyFees: FeesIF = {
+export const EmptyFees: Array<FeesIF> = [{
```
severinbeauvais, line 19: Since FeesIF is a single object, I think EmptyFees should be a single object too. You can move the array declaration to where this object is assigned.

src/App.vue
```diff
@@ -287,12 +296,27 @@ export default class App extends Mixins(
...
-    return this.getFilingData
-      ? [{ ...this.getFilingData,
-        futureEffective: this.getEffectiveDateTime.isFutureEffective,
-        priority: this.getStaffPaymentStep.staffPayment.isPriority,
-        waiveFees: (this.getStaffPaymentStep.staffPayment.option === StaffPaymentOptions.NO_FEE) }]
-      : []
+    let filingData: Array<FilingDataIF> = []
+    if (this.getFilingData) {
+      filingData = cloneDeep(this.getFilingData)
+      if (this.getEntityType === CorpTypeCd.COOP && this.getFilingData.length) {
```
severinbeauvais, line 302: Change to `length > 0` to be explicit.

Explanation: you are not checking for general validity (truthiness) but specifically if there is at least 1 array element.

src/App.vue
```diff
@@ -287,12 +296,27 @@ export default class App extends Mixins(
...
+        // Only set Future Effective and Priority to Special Resolution Fee
+        const filteredFilingData = filingData.filter(x => x.filingTypeCode === FilingCodes.SPECIAL_RESOLUTION)
```
severinbeauvais, line 304: Try using `filingData.find()` to return an element.

src/App.vue
```diff
@@ -287,12 +296,27 @@ export default class App extends Mixins(
...
+        if (filteredFilingData[0]) {
+          const specialResolutionFilingData = filteredFilingData[0]
+          specialResolutionFilingData.futureEffective = this.getEffectiveDateTime.isFutureEffective
+          specialResolutionFilingData.priority = this.getStaffPaymentStep.staffPayment.isPriority
+        }
+        filingData.forEach(x => {
```
severinbeauvais, line 310: Try using `filingData.some()` here.

andrepestana-aot: I'm not sure if I understood. In this case I need to set waiveFees for all elements. Why should we use `some()`?

src/App.vue
andrepestana-aot: makes sense. Will do it!

src/App.vue
andrepestana-aot: ok!

src/interfaces/payment-interfaces/fees-interface.ts
andrepestana-aot: oh, yes. You're right. Thanks!

src/App.vue
severinbeauvais: My mistake. No change needed.

severinbeauvais (approved):

severinbeauvais:
@andrepestana-aot Did you run the unit tests locally? Were there a lot of warnings? If so, they need to be fixed soon please.

andrepestana-aot:
> @andrepestana-aot Did you run the unit tests locally? Were there a lot of warnings? If so, they need to be fixed soon please.

There were some. Let me check. I'll fix them

PR #285, opened 2021-11-05
Updated DateMixin to be compatible with other projects

severinbeauvais:
*Issue #:* None

*Description of changes:*
- DateMixin.dateToPacificDate() is now compatible with the same method in our other projects
- This PR impacts nothing atm, but makes things in sync for possible future changes.
- This PR doesn't update the app version because it doesn't matter from a UI perspective.

### DIFF
--- a/src/mixins/date-mixin.ts
+++ b/src/mixins/date-mixin.ts
@@ -81,31 +81,20 @@ export default class DateMixin extends Mixins(CommonMixin) {
    * @example "2021-01-01 07:00:00 GMT" -> "Dec 31, 2020"
    * @example "2021-01-01 08:00:00 GMT" -> "Jan 1, 2021"
    */
-  dateToPacificDate (date: Date, long = false): string {
+  dateToPacificDate (date: Date, longMonth = false, showWeekday = false): string {
     // safety check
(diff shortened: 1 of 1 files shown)


src/mixins/date-mixin.ts
```diff
@@ -81,31 +81,20 @@ export default class DateMixin extends Mixins(CommonMixin) {
    * @example "2021-01-01 07:00:00 GMT" -> "Dec 31, 2020"
    * @example "2021-01-01 08:00:00 GMT" -> "Jan 1, 2021"
    */
-  dateToPacificDate (date: Date, long = false): string {
+  dateToPacificDate (date: Date, longMonth = false, showWeekday = false): string {
```
severinbeauvais, line 84: I separated out long month and whether to show the weekday. This makes this method compatible with our other projects.
Eg1, normal usage: `this.dateToPacificDate(date)`
Eg2, called from dateToPacificDateTime:  `this.dateToPacificDate(date, true)`
Eg3, special case where you also want the weekday:  `this.dateToPacificDate(date, true, true)`

cameron-freshworks (approved):

PR #286, opened 2021-11-08
9288 - Removed unit test warnings from App.spec.ts

andrepestana-aot:
*Issue #:* /bcgov/entity#9288

*Description of changes:*

- Removed unit test warnings from App.spec.ts

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -303,16 +303,23 @@ export default class App extends Mixins(
         // Only set Future Effective and Priority to Special Resolution Fee
         const specialResolutionFilingData = filingData.find(x => x.filingTypeCode === FilingCodes.SPECIAL_RESOLUTION)
         if (specialResolutionFilingData) {
-          specialResolutionFilingData.futureEffective = this.getEffectiveDateTime.isFutureEffective
-          specialResolutionFilingData.priority = this.getStaffPaymentStep.staffPayment.isPriority
+          if (this.getStaffPaymentStep.staffPayment.option === StaffPaymentOptions.NO_FEE) {
+            filingData.forEach(x => {
+              x.waiveFees = true
+            })
+          } else {
+            specialResolutionFilingData.futureEffective = this.getEffectiveDateTime.isFutureEffective
+            specialResolutionFilingData.priority = this.getStaffPaymentStep.staffPayment.isPriority
+          }
         }
-        filingData.forEach(x => {
-          x.waiveFees = (this.getStaffPaymentStep.staffPayment.option === StaffPaymentOptions.NO_FEE)
-        })
       } else if (this.getFilingData[0]) {
-        filingData[0].futureEffective = this.getEffectiveDateTime.isFutureEffective
-        filingData[0].priority = this.getStaffPaymentStep.staffPayment.isPriority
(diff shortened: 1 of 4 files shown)


src/App.vue
```diff
@@ -303,16 +303,23 @@ export default class App extends Mixins(
...
-        filingData[0].futureEffective = this.getEffectiveDateTime.isFutureEffective
-        filingData[0].priority = this.getStaffPaymentStep.staffPayment.isPriority
-        filingData[0].waiveFees = (this.getStaffPaymentStep.staffPayment.option === StaffPaymentOptions.NO_FEE)
+        // Avoid waiveFee with priority or futureEffective in the same request
+        if (this.getStaffPaymentStep.staffPayment.option === StaffPaymentOptions.NO_FEE) {
+          filingData[0].waiveFees = true
+        } else {
+          filingData[0].futureEffective = this.getEffectiveDateTime.isFutureEffective
+          filingData[0].priority = this.getStaffPaymentStep.staffPayment.isPriority
+        }
```
andrepestana-aot, line 322: I realized that the UI was sending `waiveFees` and `priority` flags at the same time to  the API, then I changed it.

severinbeauvais: Good catch! I, too, find things like this thanks to the unit tests. 👍

severinbeauvais: OK, I see the warnings still display for unknown custom element affix. I found that you can fix this warning by "using" Affix just like Vuetify... but this uncovers a new warning/error (which I didn't chase down).
Eg,
```
import Affix from 'vue-affix'
Vue.use(Affix)
...

severinbeauvais: ^^ the offsetHeight error is because affix can't find the relative element.

https://github.com/eddiemf/vue-affix/blob/12d7dc22ee8bc17945067cf6c2781847f8a1d254/src/affix.vue#L207

severinbeauvais: ^^ the affix component correctly gets the relativeElement prop in the tests, but this call returns null:
`document.querySelector(this.relativeElementSelector);`

👀

andrepestana-aot:
> Thanks for looking into this. The unit tests are certainly more complete now.
> 
> I think we should time-box this work -- in other words, don't spend more than, say, 1/2 day on this. If there are warnings left after that, so be it. But I don't think we should completely hide console errors (please revert that).

The only console errors left were the affix ones. I've spent too much time on Friday trying to solve it.

src/App.vue
severinbeauvais: I saw your comment about Friday... closing this conversation.

andrepestana-aot:
>Maybe some "failed to mount" errors could be fixed by mounting (instead of shallowMounting) components?

If we shallowMount it, it doesn't mount the affix's child SbcFeeSummary that we're trying to test.

severinbeauvais:
OK, just revert your console error change and then this PR is ok by me.

andrepestana-aot:
>I note that we stub Affix: true, which I think previously worked to resolve this warning. So something changed (maybe the version of jest/vue-test-utils that we use).

I need to change it. ` Affix: true` means we want to stub it but SbcFeeSummary is the affix's child. Then we cannot stub it.

severinbeauvais (approved):

PR #287, opened 2021-11-09
8910 Dissolution Custodians

cameron-freshworks:
*Issue #:* /bcgov/entity#8910

*Description of changes:*
* Implements Dissolution Custodian components and related items.
* Some code refactoring to utilize common rules and enums.
* Implemented Define Dissolutions validations.
* Component / form level validations.
* updated unit testing.


src/rules/officer-rules.ts
```diff
@@ -0,0 +1,35 @@
...
+]
+
+export const lastNameRules: Array<Function> = [
+  v => !!v || 'A last name is required',
+  v => !/^\s/g.test(v) || 'Invalid spaces', // leading spaces
+  v => !/\s$/g.test(v) || 'Invalid spaces', // trailing spaces
+  v => (v?.length <= 30) || 'Cannot exceed 30 characters' // maximum character count
+]
+
+export const orgNameRules: Array<Function> = [
```
severinbeauvais, line 22: How about moving this to a new file, and also emailRules below?

severinbeauvais, line 22: How about moving this to a new file, and also emailRules below?

src/store/getters/state-getters.ts
```diff
@@ -458,9 +460,19 @@ export const getCourtOrderStep = (state: StateIF): CourtOrderStepIF => {
 
 /** The custodian email. */
 export const getCustodianEmail = (state: StateIF): string => {
-  return 'hardcoded.custodian@email.com' // TODO: Change this once custodian details are available in state
+  return getCustodian(state)?.officer.email || '(Not entered)'
```
severinbeauvais, line 463: I think the fallback should be in the UI component.

src/mixins/filing-template-mixin.ts
```diff
@@ -375,6 +378,9 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
       dissolutionStatementType: draftFiling.dissolution?.dissolutionStatementType
     })
 
+    // // Take the first custodian, as there is only a singular custodian in a dissolution filing.
+    if (draftFiling.dissolution.parties) this.setCustodianOfRecords({ ...draftFiling.dissolution.parties[0] })
```
severinbeauvais, line 382: This looks complicated. Would this work?
```
if (draftFiling.dissolution.parties) this.setCustodianOfRecords(draftFiling.dissolution.parties[0])
```

src/components/DefineDissolution/CareAndCustodySelect.vue
```diff
@@ -0,0 +1,100 @@
...
+    // Todo: Apply option to store when required
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+.section-container {
+  padding: 1.5rem 2rem;
```
severinbeauvais, line 67: This conflicts with the Vuetify whitespace helper classes. I recommend using those classes as much as possible as I have found it makes layout tweaks easier.

severinbeauvais (commented):
Looks good so far.

src/rules/officer-rules.ts
cameron-freshworks: This is a new file :) 
I put them here as they are rules on properties associated with our definition of the `officer`
I don't mind re organizing this, just not sure why for email or org vs the name rules?

src/store/getters/state-getters.ts
cameron-freshworks: Ah, definitely agree. Will fix!

src/mixins/filing-template-mixin.ts
cameron-freshworks: oo yea, i think that simplifies it a bit. Will try and update.

src/components/DefineDissolution/CareAndCustodySelect.vue
cameron-freshworks: Definitely, can update this 👍

src/rules/officer-rules.ts
severinbeauvais: It just seems that orgNameRules and emailRules don't belong in a file named officer-rules ...?

src/mixins/filing-template-mixin.ts
severinbeauvais: The theoretical advantage of destructuring and then objectifying is that the properties are copied (instead of the object being assigned) but I think the latter is OK.

src/rules/officer-rules.ts
cameron-freshworks: Yea, the email rule could be pulled out but in the context of our apps an organization is a `PartyType `. A property defined on `officers` only, i think*. 
Yea, it doesn't seem totally intuitive, i see what you are saying for sure.

cameron-freshworks: I had an epiphany this morning... it's revolutionary. 
Will extract these rules out into two files, one called `name-rules` and the other called `contact-rules` 😁 
Updating!

andrepestana-aot (commented):
Loooks good so far!

vysakh-menon-aot (approved):

src/rules/officer-rules.ts
severinbeauvais: Vive this kind of revolution!
(Or could just call it `rules.ts`, then you only need to import 1 file.)

severinbeauvais: I just looked at your changes re: this. They look good too, so I'm closing this conversation.

severinbeauvais (approved):
Looks good so far. What about the unit tests?

PR #292, opened 2021-11-15
Dissolutions ux, validation and filing json updates

cameron-freshworks:
*Issue #:* /bcgov/entity#8910

*Description of changes:*
* Updates Scroll to validations on DefineDissolutions
* Various UX updates for Custodians as seen here: https://github.com/bcgov/entity/issues/8910
* Applied current date to dissolution and role appointment date
* Updated v-models for text input to `trim` values instead of present user with leading or trailing white space errors. 
* Updated rules for names to catch white spaces with no characters
* Updated unit testing

*Note:* - All Ux Requests

### DIFF
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -13,7 +13,7 @@ import {
 
 // Constants and enums
 import { INCORPORATION_APPLICATION } from '@/constants'
-import { CorpTypeCd, DissolutionTypes, EffectOfOrders, FilingTypes, StaffPaymentOptions } from '@/enums'
+import { CorpTypeCd, DissolutionTypes, EffectOfOrders, FilingTypes, RoleTypes, StaffPaymentOptions } from '@/enums'
 
 /**
  * Mixin that provides the integration with the Legal API.
@@ -289,10 +289,19 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
           legalName: this.getBusinessLegalName
         },
         dissolution: {
+          dissolutionDate: this.getCurrentDate,
           affidavitConfirmed: this.getAffidavitStep.validationDetail.validationItemDetails[0]?.valid || false,
           custodialOffice: this.getBusiness.officeAddress,
           dissolutionType: this.getDissolutionType,
-          parties: [this.getCustodian]
+          parties: [{
+            ...this.getCustodian,
+            roles: [
+              {
+                roleType: RoleTypes.CUSTODIAN,
+                appointmentDate: this.getCurrentDate
+              }
+            ]
+          }]
         }
       }
     }
--- a/src/resources/Dissolutions/cooperative.ts
+++ b/src/resources/Dissolutions/cooperative.ts
@@ -39,7 +39,7 @@ export const CooperativeDissolutionResource: DissolutionResourceIF = {
     }
   ],
   custodialRecords: {
-    sectionSubtitle: `Enter the full name, email address and mailing and delivery address of the Custodian of
+    sectionSubtitle: `Enter the full name, email address, and mailing and delivery address of the Custodian of
       Records. Addresses must be located in Canada.`,
     custodianTitle: `Custodian of Records`,
     helpSection: {
--- a/src/resources/Dissolutions/cooperative.ts
+++ b/src/resources/Dissolutions/cooperative.ts
@@ -39,8 +39,8 @@ export const CooperativeDissolutionResource: DissolutionResourceIF = {
     }
   ],
   custodialRecords: {
-    sectionSubtitle: `Enter the full name, email address and mailing address of the Custodian of Records. Addresses must
-     be located in Canada.`,
+    sectionSubtitle: `Enter the full name, email address and mailing and delivery address of the Custodian of
+      Records. Addresses must be located in Canada.`,
     custodianTitle: `Custodian of Records`,
     helpSection: {
       helpText: [
--- a/src/rules/first-name-rules.ts
+++ b/src/rules/first-name-rules.ts
@@ -1,5 +1,5 @@
 export const FirstNameRules: Array<Function> = [
   v => !!v || 'First name is required',
-  v => /^(?=.*\d)|(?=.*[a-z])|(?=.*[A-Z])/.test(v) || 'A firm name is required', // Contains at least a single character
+  v => /^(?=.*\d)|(?=.*[a-z])|(?=.*[A-Z])/.test(v) || 'First name is required', // Contains at least a single character
   v => (v?.length <= 30) || 'Cannot exceed 30 characters' // maximum character count
 ]
--- a/src/rules/first-name-rules.ts
+++ b/src/rules/first-name-rules.ts
@@ -1,5 +1,4 @@
 export const FirstNameRules: Array<Function> = [
-  v => !!v || 'First name is required',
-  v => /^(?=.*\d)|(?=.*[a-z])|(?=.*[A-Z])/.test(v) || 'First name is required', // Contains at least a single character
+  v => !!v.trim() || 'First name is required',
   v => (v?.length <= 30) || 'Cannot exceed 30 characters' // maximum character count
 ]
--- a/src/rules/first-name-rules.ts
+++ b/src/rules/first-name-rules.ts
@@ -1,4 +1,4 @@
 export const FirstNameRules: Array<Function> = [
-  v => !!v.trim() || 'First name is required',
+  v => !!v?.trim() || 'First name is required',
   v => (v?.length <= 30) || 'Cannot exceed 30 characters' // maximum character count
 ]
--- a/src/rules/first-name-rules.ts
+++ b/src/rules/first-name-rules.ts
@@ -1,6 +1,5 @@
 export const FirstNameRules: Array<Function> = [
-  v => !!v || 'A first name is required',
-  v => !/^\s/g.test(v) || 'Invalid spaces', // leading spaces
-  v => !/\s$/g.test(v) || 'Invalid spaces', // trailing spaces
+  v => !!v || 'First name is required',
+  v => /^(?=.*\d)|(?=.*[a-z])|(?=.*[A-Z])/.test(v) || 'A firm name is required', // Contains at least a single character
   v => (v?.length <= 30) || 'Cannot exceed 30 characters' // maximum character count
 ]
--- a/src/rules/last-name-rules.ts
+++ b/src/rules/last-name-rules.ts
@@ -1,4 +1,4 @@
 export const LastNameRules: Array<Function> = [
-  v => !!v.trim() || 'Last name is required',
+  v => !!v?.trim() || 'Last name is required',
   v => (v?.length <= 30) || 'Cannot exceed 30 characters' // maximum character count
 ]
--- a/src/rules/last-name-rules.ts
+++ b/src/rules/last-name-rules.ts
@@ -1,5 +1,5 @@
 export const LastNameRules: Array<Function> = [
   v => !!v || 'Last name is required',
-  v => /^(?=.*\d)|(?=.*[a-z])|(?=.*[A-Z])/.test(v) || 'A firm name is required', // Contains at least a single character
+  v => /^(?=.*\d)|(?=.*[a-z])|(?=.*[A-Z])/.test(v) || 'Last name is required', // Contains at least a single character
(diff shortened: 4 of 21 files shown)


src/mixins/filing-template-mixin.ts
```diff
@@ -289,10 +289,19 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
           legalName: this.getBusinessLegalName
         },
         dissolution: {
+          dissolutionDate: this.getCurrentDate,
```
severinbeauvais, line 292: This is YYYY-MM-DD in Pacific tz. Please verify that the API converts this to UTC.

PS In the case of `header.date` (filing date), it's not used for anything in the UI, and the API will overwrite it anyway (eg, "2021-11-15T22:22:02.470955+00:00").

severinbeauvais: (I am not sure if the API keeps track of dissolution time. If so then it needs the correct tz offset in the db.)

src/mixins/filing-template-mixin.ts
```diff
@@ -289,10 +289,19 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
...
           affidavitConfirmed: this.getAffidavitStep.validationDetail.validationItemDetails[0]?.valid || false,
           custodialOffice: this.getBusiness.officeAddress,
           dissolutionType: this.getDissolutionType,
-          parties: [this.getCustodian]
+          parties: [{
+            ...this.getCustodian,
+            roles: [
+              {
+                roleType: RoleTypes.CUSTODIAN,
+                appointmentDate: this.getCurrentDate
```
severinbeauvais, line 301: I just checked. For a COD filing, this is YYYY-MM-DD Pacific.

https://github.com/bcgov/business-filings-ui/blob/a3ecbcbca80c5421cb4dacad3a33999c047583c7/src/components/common/Directors.vue#L891

src/resources/Dissolutions/cooperative.ts
```diff
@@ -39,8 +39,8 @@ export const CooperativeDissolutionResource: DissolutionResourceIF = {
     }
   ],
   custodialRecords: {
-    sectionSubtitle: `Enter the full name, email address and mailing address of the Custodian of Records. Addresses must
-     be located in Canada.`,
+    sectionSubtitle: `Enter the full name, email address and mailing and delivery address of the Custodian of
```
severinbeauvais, line 42: I recommend a comma between "email address" and "and mailing".

src/mixins/filing-template-mixin.ts
cameron-freshworks: > This is YYYY-MM-DD in Pacific tz. Please verify that the API converts this to UTC.
> 
> PS In the case of `header.date` (filing date), it's not used for anything in the UI, and the API will overwrite it anyway (eg, "2021-11-15T22:22:02.470955+00:00").

I'm not sure what the API does with this date but according to the schema it is required, and in this format (YYYY-MM-DD)

src/rules/last-name-rules.ts
```diff
@@ -1,6 +1,5 @@
 export const LastNameRules: Array<Function> = [
-  v => !!v || 'A last name is required',
-  v => !/^\s/g.test(v) || 'Invalid spaces', // leading spaces
-  v => !/\s$/g.test(v) || 'Invalid spaces', // trailing spaces
+  v => !!v || 'Last name is required',
+  v => /^(?=.*\d)|(?=.*[a-z])|(?=.*[A-Z])/.test(v) || 'A firm name is required', // Contains at least a single character
```
severinbeauvais, line 3: Aren't the name rules used in other components (where it may not be a firm name)?

src/resources/Dissolutions/cooperative.ts
```diff
@@ -39,8 +39,8 @@ export const CooperativeDissolutionResource: DissolutionResourceIF = {
...
+      Records. Addresses must be located in Canada.`,
```
severinbeauvais, line 43: Does region have to be BC (which you initialize in a few places)?

src/resources/Dissolutions/cooperative.ts
cameron-freshworks: Will add!

src/mixins/filing-template-mixin.ts
cameron-freshworks: The format of `getCurrentDate` is also YYYY-MM-DD or is there is a separate concern here?

src/rules/last-name-rules.ts
cameron-freshworks: gah, copy pasta mistake, will fix for all these. good catch thanks!

cameron-freshworks: Updated in next commit

src/mixins/filing-template-mixin.ts
severinbeauvais: Where did you see this in the schema? I only see
```
                "dissolutionDate": {
                    "type": "string",
                    "format": "date"
                },
```
which doesn't tell me the format, really. 🤷‍♂️

src/resources/Dissolutions/cooperative.ts
cameron-freshworks: It does yes

cameron-freshworks: It is curious why the designs reference that for Corps but Canada for Coops.. 🤔

src/mixins/filing-template-mixin.ts
severinbeauvais: I just wanted to confirm that the appointment date is the same as in the COD filing.

It's confirmed. No further action needed.

src/mixins/filing-template-mixin.ts
cameron-freshworks: In the example data, is the only other reference.
As well as the ability to file with the address in that format.

src/resources/Dissolutions/cooperative.ts
severinbeauvais: Can you confirm the inconsistency with the designers?

src/mixins/filing-template-mixin.ts
cameron-freshworks: ahh i see, okie doke!

src/mixins/filing-template-mixin.ts
severinbeauvais: ok

src/resources/Dissolutions/cooperative.ts
cameron-freshworks: 100%, i'm now concerned this alludes to differing rules for both the UI and Validators to check..

src/mixins/filing-template-mixin.ts
severinbeauvais: You keep using that word. I do not think it means what you think it means.

https://www.urbandictionary.com/define.php?term=Okie-Doke

cameron-freshworks: 😆 
I know! We have discussed this haha

I first heard this from "Ned Flanders" on the Simpsons... i wonder if Matt Groening knew this when writing those episodes 😁

src/rules/first-name-rules.ts
```diff
@@ -1,6 +1,5 @@
 export const FirstNameRules: Array<Function> = [
-  v => !!v || 'A first name is required',
-  v => !/^\s/g.test(v) || 'Invalid spaces', // leading spaces
-  v => !/\s$/g.test(v) || 'Invalid spaces', // trailing spaces
+  v => !!v || 'First name is required',
+  v => /^(?=.*\d)|(?=.*[a-z])|(?=.*[A-Z])/.test(v) || 'First name is required', // Contains at least a single character
```
severinbeauvais, line 3: Isn't this obsolete, since the value is trimmed so it must be something?
Or is this to limit the valid character set?

src/mixins/filing-template-mixin.ts
severinbeauvais: FWIW, the Great Vizzini kept using his word, too 😁

src/rules/first-name-rules.ts
cameron-freshworks: Trim will remove the space before/after the strings input but if a user copy pasted a big block of white space in, it wouldn't be caught by our initial rule. So yea, this is to confirm there is an actual character in there. 
Seems like white space is considered a form of character.

src/mixins/filing-template-mixin.ts
cameron-freshworks: hahah who is that? I had to google.
Is this a Princess Bride reference ??

src/rules/first-name-rules.ts
severinbeauvais: Can you give me some test steps to reproduce this? I am doing something similar for Transactional Folio Number and trim seems to work fine 🤷‍♂️

cameron-freshworks: I found i could do it just by hitting space repeatedly.

lekshmimallika-aot (approved):
Looks good!

src/rules/first-name-rules.ts
cameron-freshworks: Same implementation?
ie `v-model.trim="someStringValue"` ?

severinbeauvais: I think it's a Vuetify bug! I was able to paste enough spaces to exceed my 50 chars, but still the correct value (with no spaces) is mutated in the store.

severinbeauvais: This fixes it for me:
```
export const FolioNumberRules: Array<Function> = [
  v => (!v || v.trim().length <= 50) || 'Cannot exceed 50 characters' // maximum character count
]
```

severinbeauvais: > Same implementation? ie `v-model.trim="someStringValue"` ?

Yeah, I thought you copied that from me. Great minds think alike!

cameron-freshworks: Ahh yea, most likely! 
I wonder where/when the trim occurs.

severinbeauvais: Based on my testing, the v-model is trimmed before being updated, but the component still processes its rules against the internal value (that is in the input field).

cameron-freshworks: I'll try updating my rules to match yours, it's cleaner than way if it works all the same!

src/resources/Dissolutions/cooperative.ts
cameron-freshworks: Sounds like this is a missed rule for both UI and API to verify. Going to take more time than i have today to finish. 
Alas.. i tried 😁

src/rules/first-name-rules.ts
cameron-freshworks: updated, much better, thank you!

src/resources/Dissolutions/cooperative.ts
cameron-freshworks: How does it sound if i complete this in a follow up commit so this work doesn't get stale and we have what we need for Demo?

severinbeauvais: Sounds good.

src/rules/first-name-rules.ts
```diff
@@ -1,5 +1,4 @@
 export const FirstNameRules: Array<Function> = [
-  v => !!v || 'First name is required',
-  v => /^(?=.*\d)|(?=.*[a-z])|(?=.*[A-Z])/.test(v) || 'First name is required', // Contains at least a single character
+  v => !!v.trim() || 'First name is required',
```
severinbeauvais: You need to verify that v is not null before you call trim on its instance.

severinbeauvais: Idea:
```
v => (!v || v.trim().length < 1) || 'First name is required`
```

cameron-freshworks: Optional chaining seems to work

severinbeauvais: I think your new rules works, too.

severinbeauvais (approved):

src/rules/first-name-rules.ts
cameron-freshworks: Sorry, i had two quick commits. I put in the optional chaining immediately after :)

PR #294, opened 2021-11-17
8930 Updated unit tests

severinbeauvais:
*Issue #:* /bcgov/entity#8930

*Description of changes:*
- updated unit tests
- no app version change needed (no functional changes)

### DIFF
--- a/src/components/common/TransactionalFolioNumber.vue
+++ b/src/components/common/TransactionalFolioNumber.vue
@@ -54,7 +54,7 @@ export default class TransactionalFolioNumber extends Mixins(CommonMixin) {
 
   /** True if this component is valid. */
   get isValid (): boolean {
-    return (!this.doValidate || this.$refs.folioNumberInput?.valid)
+    return (!this.doValidate || !!this.$refs.folioNumberInput?.valid)
   }
 
   /** Called when user has changed the Local Folio Number. */
--- a/tests/unit/ReviewConfirmDissolution.spec.ts
+++ b/tests/unit/ReviewConfirmDissolution.spec.ts
@@ -9,24 +9,40 @@ const store = getVuexStore()
 const reviewConfirmTestCases = [
   {
     entityType: 'CP',
+    isPremium: false,
     isStaff: true
   },
   {
     entityType: 'CP',
+    isPremium: true,
+    isStaff: false
+  },
+  {
+    entityType: 'CP',
+    isPremium: false,
     isStaff: false
   },
   {
     entityType: 'BEN',
+    isPremium: false,
     isStaff: true
   },
   {
     entityType: 'BEN',
+    isPremium: true,
+    isStaff: false
+  },
+  {
+    entityType: 'BEN',
+    isPremium: false,
     isStaff: false
   }
 ]
 
 for (const test of reviewConfirmTestCases) {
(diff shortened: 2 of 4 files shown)


src/components/common/TransactionalFolioNumber.vue
```diff
@@ -54,7 +54,7 @@ export default class TransactionalFolioNumber extends Mixins(CommonMixin) {
 
   /** True if this component is valid. */
   get isValid (): boolean {
-    return (!this.doValidate || this.$refs.folioNumberInput?.valid)
+    return (!this.doValidate || !!this.$refs.folioNumberInput?.valid)
```
severinbeauvais, line 57: This is just to change the response to a boolean (instead of, sometimes, undefined, which is falsy anyway... but testing).

tests/unit/TransactionalFolioNumber.spec.ts
```diff
@@ -23,42 +25,122 @@ const tests = [
...
+      expect(wrapper.vm.localFolioNumber).toBe(test.props.transactionalFolioNumber)
+    })
+
+    itIf(!test.props.doValidate)('is valid when validation is disabled', () => {
+      expect(wrapper.vm.isValid).toBe(true)
+      expect(wrapper.find('#transactional-folio-number-container').classes('invalid-section')).toBe(false)
+      expect(wrapper.find('label').classes('error-text')).toBe(false)
+    })
+
+    // FUTURE: fix this; for some reason, the component under test can't read `this.$refs.folioNumberInput`
```
severinbeauvais, line 98: @cameron-freshworks Have you ever seen this??? The getter in the component always sees `undefined` for the ref. I tried it as a method as well, no change. This looks the same as other test files where we seem to do this successfully, but no luck for me today.

tests/unit/ReviewConfirmDissolution.spec.ts
```diff
@@ -9,24 +9,40 @@ const store = getVuexStore()
...
+    isPremium: false,
     isStaff: false
   }
 ]
 
 for (const test of reviewConfirmTestCases) {
-  describe(`Review Confirm view for a ${test.entityType} as a ${test.isStaff ? 'staff' : 'regular'} user`, () => {
+  const type = test.isPremium ? 'premium' : test.isStaff ? 'staff' : 'regular'
+
+  describe(`Review Confirm view for a ${test.entityType} as a ${type} user`, () => {
```
cameron-freshworks, line 45: cool!

tests/unit/TransactionalFolioNumber.spec.ts
cameron-freshworks: hmm, i'm not sure i've seen this issue specifically but inside the jest test wouldn't the reference to the `ref` be something like `wrapper.vm.$refs.refNameDefinedOnTheComponent`?

My first thought was to also use a regular mount (which you've already done of course :) ) followed by using the defined ref on the component. Other than that i'm not sure 🤷

severinbeauvais: The test can actually read `wrapper.vm.$refs.refNameDefinedOnTheComponent`. It's the actual code that can't (so the getter returns the wrong value).
🤷🤷🤷

PR #296, opened 2021-11-18
Disable Certify for Coops Only

cameron-freshworks:
*Issue #:* /bcgov/entity#10147

*Description of changes:*
* Added check to allow Corps to edit the Certify field.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "3.0.3",
+  "version": "3.0.3b",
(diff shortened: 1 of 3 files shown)


cameron-freshworks:
Not exactly what we discussed with regards to the resourcing but considering the nature of this being a hot fix, opted to keep it as simple as possible. 
We can always adjust in the future when the need arises for a different configuration here.

severinbeauvais (approved):
I had to think about that logic a bit but it looks right 😁

PR #300, opened 2021-11-24
8913 Update special resolution step for COOP voluntary dissolutions

argush3:
*Issue #:* /bcgov/entity#8913

*Description of changes:*

* Update special resolutions component according to changes in design
  * add resolution date section
  * add resolution text section
  * add resolution signature section
*  Update coop resource file
* Renamed `UploadResolution` component to `CompleteResolution`


src/views/Dissolutions/ReviewConfirmDissolution.vue
```diff
@@ -160,18 +173,66 @@
...
+          >Return to this step to finish it</router-link>
+        </span>
+        <v-row class="ml-2 mt-6" no-gutters>
+          <v-col md="3">
+            <strong>Resolution Date</strong>
           </v-col>
-          <v-col v-if="isTypeCoop" md="11" id="file-name-col">
-            <span>FILE_NAME_PLACEHOLDER</span>
+          <v-col md="9" class="ml-n1">
+            {{resolutionDate}}
```
argush3, line 193: wrong date format

src/views/Dissolutions/ReviewConfirmDissolution.vue
```diff
@@ -160,18 +173,66 @@
...
+          <v-col md="9" class="ml-n1">
+            {{signingParty}}
+          </v-col>
+        </v-row>
+        <v-row class="ml-2 mt-5" no-gutters>
+          <v-col md="3">
+            <strong>Date Signed</strong>
+          </v-col>
+          <v-col md="9" class="ml-n1">
+            {{signingDate}}
```
argush3, line 227: wrong date format

src/views/Dissolutions/ReviewConfirmDissolution.vue
severinbeauvais: What does this mean? Does someone have to fix this yet, or can you do it in this ticket?

src/components/CreateResolution/CompleteResolution.vue
```diff
@@ -0,0 +1,828 @@
...
+          <v-card flat id="resolution-text-card" class="pt-7 pb-5 pl-6 pr-6">
+            <v-row>
+              <v-col id="resolution-text-card-left-col" cols="2" class="pt-6" >
+                <v-card-title class="resolution-text-vcard-title pl-1 pr-0">
+                  <div>{{getCreateResolutionResource.resolutionTextSection.textLabel}}</div>
+                </v-card-title>
+              </v-col>
+              <v-col id="resolution-text-card-right-col" cols="10" class="pt-6 pl-6 pr-5">
+                <v-form ref="resolutionTextFormRef">
+                  <v-textarea ref="resolutionTextRef"
```
argush3, line 146: There is an issue I'm having with the autogrow funcitonality in this `v-textarea`.  It autogrows fine when typing inputting text but on resumin of a draft it doesn't autogrow properly.  Instead I believe it just grows to the height that is defined by the `rows` prop value.  I tried to wrap it with a `<v-lazy>` and that works but introduces other issues as it doesn't load the component until a user actually views the step that the component.

src/views/Dissolutions/ReviewConfirmDissolution.vue
argush3: i will be fixing it.  i just noted it.  it's a note for myself.

src/components/CreateResolution/CompleteResolution.vue
severinbeauvais: Can you tell how many text rows are in this component? Could you use that to resize it on draft resume?

If all else fails, check with UX, but IMO you can size this component "tall enough" initially (eg, 3-4 rows) and then don't worry about draft resume...

cameron-freshworks: I've not! Looking at this now though. 
The issue is that the extended field is hidden when returning from a draft where the text exceeds the initial area?

cameron-freshworks: Definitely agree with Severin, it's probably not worth digging into too too much, but it could be that components size is reactive based on the `value:` prop. You could try applying the model to the value, if there is an initial value, that could potentially be picked up by the component.
Just spitballing, but that might be the prop it watches, not sure!

argush3: i'll try fiddling with resize a bit and talk to UX if i don't get anywhere with it.

src/components/CreateResolution/CompleteResolution.vue
```diff
@@ -0,0 +1,828 @@
...
+        <u class="help-btn" @click="helpToggle = !helpToggle"><small>Hide Help</small></u>
+      </section>
+    </div>
+
+    <section id="sample-resolution-section" v-if="entityFilter(CorpTypeCd.COOP)" class="mt-10">
+      <header id="sample-resolution-header">
+        <h2>{{getCreateResolutionResource.sampleFormSection.header}}</h2>
+      </header>
+      <p class="mt-2" v-html="getCreateResolutionResource.sampleFormSection.text">
+      </p>
```
cameron-freshworks, line 75: Could probably be 1 line

src/views/Dissolutions/ReviewConfirmDissolution.vue
argush3: updated to expected date format

src/views/Dissolutions/ReviewConfirmDissolution.vue
argush3: updated to expected date format

src/components/CreateResolution/CompleteResolution.vue
argush3: done

src/mixins/date-mixin.ts
```diff
@@ -99,6 +113,30 @@ export default class DateMixin extends Mixins(CommonMixin) {
...
   }
 
+  /**
+   * Converts a date string of YYYY-mm-dd format to a date string with more display options in Pacific timezone.
+   * @example "2021-01-01" -> "Dec 31, 2020"
+   */
+  dateStrToPacificDate (dateStr: string, longMonth = false, showWeekday = false): string {
+    // safety checks
+    if (!dateStr) return null
+    const date = this.createUtcDateFromStr(dateStr)
```
severinbeauvais, line 123: I'm glad you copied this from another project, but please make it exactly the same, so that it can be moved to the shared components repo at some point.

src/components/CreateResolution/CompleteResolution.vue
```diff
@@ -0,0 +1,828 @@
...
+      <v-card flat id="sample-resolution-card" class="pt-7 pb-5 pl-6">
+        <v-row>
+          <v-col id="sample-resolution-card-left-col" cols="1" class="pt-6" >
+          </v-col>
+          <v-col id="sample-resolution-card-center-col" cols="3" class="pl-8 pt-6">
+            <img :src="previewImageSource()" class="preview-image" />
+          </v-col>
+          <v-col id="sample-resolution-card-right-col" cols="6" class="pt-6 pr-15">
```
cameron-freshworks, line 82: No action required here, as we are not building for mobile.
Just keep in mind that defining cols as such would result in all of these trying to jam into 1 line no matter what the screen size. Vue's grid system handles this pretty easily by having a prop for each size (xs, sm, lg.. etc) and cols being the default when none are defined.
Works totally fine here though, just sharing my thoughts :)

src/components/CreateResolution/CompleteResolution.vue
cameron-freshworks: Also, the props tab spacing seems a little different than the spacing used elsewhere in this component.

cameron-freshworks: See `date-picker` line 112

src/components/CreateResolution/CompleteResolution.vue
```diff
@@ -0,0 +1,827 @@
...
+                    {{item.value}}
+                  </v-col>
+                </v-row>
+                <v-row v-if="item.type === ItemTypes.PARTIAL_ITEMS" no-gutters>
+                  <v-col cols="auto" class="pt-1 mr-2"><v-icon>mdi-circle-small</v-icon></v-col>
+                  <v-col cols="11" class="pt-1 pb-1" no-gutters>
+                    <span v-for="(partialItem, index) in item.value" :key="index">
+                      <span v-if="partialItem.type === ItemTypes.TEXT" v-html="partialItem.value"></span>
+                      <span v-if="partialItem.type === ItemTypes.PLACEHOLDER
+                                && partialItem.value === PLACEHOLDER_LEGAL_NAME_INLINE">
```
cameron-freshworks, line 268: Maybe preference, not sure, but in the past the `&&` operator would be on the initial line before the line break. 
Super small thing

src/components/CreateResolution/CompleteResolution.vue
```diff
@@ -0,0 +1,827 @@
...
+    return docUrl
+  }
+
+  private previewImageSource (): string {
+    // Note: the image file path did not resolve correctly when using the require function directly.  In order
+    // to get the image path resolving correctly, needed to get the image context first and use that to build
+    // the final image file path
+    const images = require.context('@/assets/images/', false, /\.png$/)
+    return images('./' + this.getCreateResolutionResource.sampleFormSection.previewImagePath)
+  }
```
cameron-freshworks, line 381: This is cool, i fiddled with something similar and couldn't figure it out 👍

src/components/CreateResolution/CompleteResolution.vue
severinbeauvais: What Cam says is enforced by our linter in script code.

src/mixins/date-mixin.ts
argush3: I'm not sure what you exactly mean by copied from another project.  Most of the functions in this file were already in place.  I had specific use case which is why I needed to create this function.  Although I realize now that I can simplify this function a lot by calling `dateToPacificDate`.

src/components/CreateResolution/CompleteResolution.vue
argush3: yup, i can fix it.

src/components/CreateResolution/CompleteResolution.vue
```diff
@@ -0,0 +1,827 @@
...
+  private get confirmLabel (): string {
+    const confirmTextItems = this.getCreateResolutionResource.confirmSection.text
+    const result = confirmTextItems.map(item => {
+      if (item.type === ItemTypes.PLACEHOLDER && item.value === this.PLACEHOLDER_LEGAL_NAME) {
+        return `<strong>${this.getBusinessLegalName}</strong>`
+      }
+      return item.value
+    }).join(' ')
+    return result
+  }
```
cameron-freshworks, line 392: Seems like a lot for a dynamic label, are the `value`'s of the `confirmTextItems` much different than the html template returned there?
Is the main purpose to provide the `strong` tag in the event the conditions are met?

src/components/CreateResolution/CompleteResolution.vue
```diff
@@ -0,0 +1,827 @@
...
+    return [
+      (v: string) => !!v || 'Select date',
+      (v: string) => expectedDateFormat.test(v) || 'Date format should be YYYY-MM-DD',
+      (v: string) =>
+        this.validateIsBetweenDates(this.resolutionDateMin,
+          this.resolutionDateMax,
+          this.createUtcDateFromStr(v)) ||
+        `Date should be between ${this.resolutionDateMinStr} (incorporation date) and ${this.resolutionDateMaxStr}`
+    ]
+  }
```
cameron-freshworks, line 406: I like it! 
We've recently introduced some `rule` files, would it make sense to relocate this rule set to be reusable from that location, or are they too specific you think?

argush3: yeah, great idea.  The name ones were really handy.  I think some of these can go into a common file for sure.  Will look into it.

cameron-freshworks: Also, if these are to stay in this component, should they be `private` as well?

src/mixins/filing-template-mixin.ts
```diff
@@ -413,9 +440,15 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
         validationItemDetails: []
       },
       resolutionConfirmed:
-        draftFiling.dissolution?.specialResolution?.resolutionConfirmed ||
+        draftFiling?.specialResolution?.resolutionConfirmed ||
```
severinbeauvais, line 443: You don't need conditional chaining on `draftFiling` (since no other code in the method uses it).

src/mixins/filing-template-mixin.ts
```diff
@@ -413,9 +440,15 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
...
         draftFiling.dissolution?.resolution?.resolutionConfirmed || false
     }
+    if (draftFiling.specialResolution) {
+      createResolution.resolutionDate = draftFiling.specialResolution?.resolutionDate
+      createResolution.resolutionText = draftFiling.specialResolution?.resolution
+      createResolution.signingPerson = draftFiling.specialResolution?.signatory
+      createResolution.signingDate = draftFiling.specialResolution?.signingDate
```
severinbeauvais, line 450: You don't need conditional chaining here since you verified that the object exists on line 446.

I wonder if restructuring would simplify this code? But I'm OK with this too.

argush3: will take a look.

src/components/CreateResolution/CompleteResolution.vue
argush3: yes, these should stay `private`.  good call.

src/views/Dissolutions/ReviewConfirmDissolution.vue
```diff
@@ -160,18 +173,66 @@
...
+        </v-row>
+        <v-row class="ml-2 mt-3" no-gutters>
+          <v-col md="3">
+            <strong>Signing Party</strong>
+          </v-col>
+          <v-col md="9" class="ml-n1">
+            {{signingParty}}
+          </v-col>
+        </v-row>
+        <v-row class="ml-2 mt-5" no-gutters>
```
severinbeauvais, line 222: This block of code is large enough, and uses enough getters, that I would prefer to see it in its own sub-component -- that would keep THIS component (the review page) cleaner.

Doesn't have to be done in this PR if you're crunched for time.

src/views/Dissolutions/ReviewConfirmDissolution.vue
```diff
@@ -668,4 +766,22 @@ export default class ReviewConfirmDissolution extends Mixins(DateMixin) {
     }
   }
 }
+
+// styles specific to resolution summary section
```
severinbeauvais, line 770: It's fair to have this here atm, but this should go into a sub-component if possible.

It's not possible to move the other specific CSS into their components since those are shared components.

src/views/Dissolutions/ReviewConfirmDissolution.vue
argush3: ok, sounds good.  I'll see how it goes.  Might be in a separate PR.

src/mixins/date-mixin.ts
severinbeauvais: Sorry, I didn't realize you cloned an existing local method. No offense intended!

src/views/Dissolutions/ReviewConfirmDissolution.vue
argush3: will move into sub-component when I create the resolution summary component.

src/components/CreateResolution/CompleteResolution.vue
argush3: should it pick it up when I run `npm run lint`?  doesn't pick it up currently.  no big deal anyways.  was just curious.

severinbeauvais (approved):
Looks good so far. I'm approving as I don't have any show-stopper comments and it looks like you're dealing with the other comments already.

src/components/CreateResolution/CompleteResolution.vue
severinbeauvais: Sadly (or thankfully), our linter doesn't much look at the template code. I think it identifies long lines (sometimes) and that's about it. 🤷‍♂️

argush3: probably thankfully...haha

argush3: updated

lekshmimallika-aot (approved):
Looks good

src/components/CreateResolution/CompleteResolution.vue
cameron-freshworks: Feel free to resolve this convo at your discretion!

cameron-freshworks (approved):
Looks good!
As discussed there is a follow up pr for tests, so if there is anything minor pending it can be touched up there 👍

src/components/CreateResolution/CompleteResolution.vue
argush3: the purpose of this is not just the strong tag but to allow swapping of the placeholder value as well.

src/mixins/filing-template-mixin.ts
argush3: this is actually req'd.  i tested creating a new coop voluntary dissolution filing and it errors out.  i've kept this in for now but I do plan on taking a look at restructuring the code in this block on the PR where I will create the resolution summary component.  

i.e this block of code
```
    // restore Resolution data
    const createResolution: CreateResolutionIF = {
      validationDetail: {
        valid: false,
        validationItemDetails: []
      },
      resolutionConfirmed:
        draftFiling?.specialResolution?.resolutionConfirmed ||
        draftFiling.dissolution?.resolution?.resolutionConfirmed || false
    }
    if (draftFiling.specialResolution) {
      createResolution.resolutionDate = draftFiling.specialResolution.resolutionDate
      createResolution.resolutionText = draftFiling.specialResolution.resolution
      createResolution.signingPerson = draftFiling.specialResolution.signatory
      createResolution.signingDate = draftFiling.specialResolution.signingDate
    }
    this.setResolution(createResolution)
```

src/mixins/filing-template-mixin.ts
argush3: have updated to remove conditional chaining but as mentioned above, will be taking a second look at code around here in subsequent PR.

argush3:
Merging code here so it can start get some UXA.  I will address outstanding issues in mentioned in this ticket as well as unit tests in a subsequent PR.

src/mixins/filing-template-mixin.ts
severinbeauvais: I'm referring to `draftFiling?` ... tons of other code in this method dereferences it, so I think it's safe to assume `draftFiling` exists.

argush3: oh ic.  k will fix that up in subsequent work.

PR #303, opened 2021-11-29
8913 vd special resolution uxa feedback updates

argush3:
*Issue #:* /bcgov/entity#8913

*Description of changes:*

* UXA feedback updates
* Added tests for `CompleteResolution` component
* misc cleanup

### DIFF
--- a/src/assets/styles/base.scss
+++ b/src/assets/styles/base.scss
@@ -142,6 +142,10 @@ p {
   font-weight: normal;
 }
 
+.tool-tip {
+  cursor: default
+}
+
 // Loading Indicator
 .loading-container {
   display: flex;
--- a/src/components/Affidavit/Affidavit.vue
+++ b/src/components/Affidavit/Affidavit.vue
@@ -452,4 +452,11 @@ ul {
     text-decoration: none;
   }
 }
+
+#upload-affidavit-card {
+    // text area word count counter
+    ::v-deep label,v-label,theme--light {
+      color: $gray7 !important;
+    }
+}
 </style>
--- a/src/components/Affidavit/Affidavit.vue
+++ b/src/components/Affidavit/Affidavit.vue
@@ -454,8 +454,8 @@ ul {
 }
 
 #upload-affidavit-card {
-    // text area word count counter
-    ::v-deep label,v-label,theme--light {
+    // file upload input field
+    ::v-deep label.v-label.theme--light {
       color: $gray7 !important;
     }
 }
--- a/src/components/Affidavit/Affidavit.vue
+++ b/src/components/Affidavit/Affidavit.vue
@@ -6,7 +6,7 @@
         <p>Before submitting the voluntary dissolution you must <b>complete and sign</b> an
           <v-tooltip top max-width="15rem" content-class="top-tooltip" transition="fade-transition">
             <template v-slot:activator="{ on }">
-              <span v-on="on" class="dotted-underline"> affidavit </span>
+              <span v-on="on" class="tool-tip dotted-underline"> affidavit </span>
             </template>
             <span>
               Affidavit – A declaration signed by a Commissioner for Taking Oaths.
--- a/src/components/CreateResolution/CompleteResolution.vue
+++ b/src/components/CreateResolution/CompleteResolution.vue
@@ -105,7 +105,7 @@
           <v-row>
             <v-col id="resolution-date-card-left-col" cols="2" class="pt-6" >
               <v-card-title class="resolution-date-vcard-title pl-1 pr-0">
-                <div>Resolution Date</div>
+                Resolution Date
               </v-card-title>
             </v-col>
             <v-col id="resolution-date-card-right-col" cols="10" class="pt-6 pl-6 pr-5">
@@ -139,7 +139,7 @@
             <v-row>
               <v-col id="resolution-text-card-left-col" cols="2" class="pt-6" >
                 <v-card-title class="resolution-text-vcard-title pl-1 pr-0">
-                  <div>{{getCreateResolutionResource.resolutionTextSection.textLabel}}</div>
+                  {{getCreateResolutionResource.resolutionTextSection.textLabel}}
                 </v-card-title>
               </v-col>
               <v-col id="resolution-text-card-right-col" cols="10" class="pt-6 pl-6 pr-5">
@@ -163,15 +163,15 @@
       </div>
     </section>
 
-    <section id="resolution-signatory-info-section" v-if="entityFilter(CorpTypeCd.COOP)" class="mt-10">
-      <header id="resolution-signatory-info-header">
+    <section id="resolution-signature-info-section" v-if="entityFilter(CorpTypeCd.COOP)" class="mt-10">
+      <header id="resolution-signature-info-header">
         <h2>{{getCreateResolutionResource.resolutionSignatureSection.header}}</h2>
       </header>
       <p class="mt-2">
         {{getCreateResolutionResource.resolutionSignatureSection.description}}
       </p>
       <div :class="{ 'invalid-section': getShowErrors && (!this.isSigningPersonValid || !this.isSigningDateValid) }">
-        <v-card flat id="-resolution-signature-card" class="pt-7 pb-5 pl-6 pr-8">
+        <v-card flat id="resolution-signature-card" class="pt-7 pb-5 pl-6 pr-8">
           <v-row no-gutters>
             <v-col id="resolution-signature-card-left-col" cols="2" class="pt-7" >
               <v-card-title class="resolution-signature-vcard-title pl-2 pr-0">
@@ -547,12 +547,12 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin, E
             {
               name: 'signingPersonValid',
               valid: this.isSigningPersonValid,
-              elementId: 'resolution-signatory-info-header'
+              elementId: 'resolution-signature-info-header'
             },
             {
               name: 'signingDateValid',
               valid: this.isSigningDateValid,
-              elementId: 'resolution-signatory-info-header'
+              elementId: 'resolution-signature-info-header'
             },
             {
               name: 'resolutionConfirmed',
@@ -706,6 +706,12 @@ ul {
   color: $gray7;
 }
 
+// Used for title/placeholder text of input fields.  This was required as some input fields require
+// were not using the expected color of $gray7
+::v-deep label,.v-label,theme--light {
+  color: $gray7 !important;
+}
+
 .chk-list-item-txt {
   margin-left: 0.5rem;
 }
@@ -754,6 +760,13 @@ ul {
 
   a {
     text-decoration: none;
+  }
+}
+
+#resolution-text-section {
+  // text area word count counter
+  ::v-deep .v-counter {
+    color: $gray7 !important;
   }
 }
 
@@ -844,6 +857,7 @@ ul {
   padding-top: 1px;
   font-size: 16px;
   font-weight: bold;
+  color: $gray9 !important;
 }
 
 </style>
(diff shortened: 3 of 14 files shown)


src/assets/styles/base.scss
```diff
@@ -142,6 +142,10 @@ p {
   font-weight: normal;
 }
 
+.tool-tip {
+  cursor: default
```
severinbeauvais, line 146: Did UX ask for all tooltips to be an arrow instead of a pointing hand?

May I recommend the "help" cursor for tooltips?

argush3: this is what was requested by UX:

>don't change the mouse cursor to the text tool on hover for tool tips (keep as regular pointer)

src/components/Affidavit/Affidavit.vue
```diff
@@ -452,4 +452,11 @@ ul {
     text-decoration: none;
   }
 }
+
+#upload-affidavit-card {
+    // text area word count counter
+    ::v-deep label,v-label,theme--light {
```
severinbeauvais, line 458: This looks like incorrect syntax. I think maybe you mean:
`::v-deep label.v-label.theme--light`

severinbeauvais: "v-label" and "theme-light" are not HTML elements; they are classes, so they need a period in front of them.

This CSS targets all "label"s. Is it working correctly?

src/components/CreateResolution/CompleteResolution.vue
```diff
@@ -96,15 +97,15 @@
...
-      <p class="mt-2">
+      <p class="section-description mt-2">
         {{getCreateResolutionResource.resolutionDateSection.description}}
       </p>
       <div :class="{ 'invalid-section': getShowErrors && !this.isResolutionDateValid }">
         <v-card flat id="resolution-date-card" class="pt-7 pb-5 pl-6 pr-6">
           <v-row>
             <v-col id="resolution-date-card-left-col" cols="2" class="pt-6" >
               <v-card-title class="resolution-date-vcard-title pl-1 pr-0">
-                <div>Resolution Date</div>
```
severinbeauvais, line 107: I agree; these divs are needed.

No need to change, but sometimes I wrap title text or button text in spans (eg, when they are on multiple lines) so that we can select the text directly in the unit tests and then use `toBe` instead of `toContain` (due to extra line breaks in the latter).

severinbeauvais: ^^ oops, typo. Added "not".

src/components/Affidavit/Affidavit.vue
argush3: yeah, it seems to be working properly but only because we probably aren't using any other labels outside of the use case I am looking at.  I'll take another look at this.

src/components/CreateResolution/CompleteResolution.vue
```diff
@@ -367,6 +369,11 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin, E
   // Validation Rules
   readonly Rules = Rules
 
+  /** Is True if Jest is running the code. */
+  get isJestRunning (): boolean {
```
severinbeauvais, line 373: This is already declared in Common Mixin.

src/assets/styles/base.scss
severinbeauvais: OK.

Can you ask them if the "help" cursor would be better?

src/components/CreateResolution/CompleteResolution.vue
```diff
@@ -690,6 +714,12 @@ ul {
   color: $gray7;
 }
 
+// Used for title/placeholder text of input fields.  This was required as some input fields require
+// were not using the expected color of $gray7
+::v-deep label,.v-label,theme--light {
```
severinbeauvais, line 719: Check classes again.

src/components/CreateResolution/CompleteResolution.vue
```diff
@@ -653,6 +662,21 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin, E
     this.updateResolutionStepValidationDetail()
   }
 
+  private async reRenderResolutionText (): Promise<void> {
```
severinbeauvais, line 665: Not async. Does the method within need to be awaited?

argush3: good catch, it used to be awaited.

src/components/CreateResolution/CompleteResolutionSummary.vue
```diff
@@ -155,6 +168,22 @@ export default class CompleteResolutionSummary extends Mixins(DateMixin) {
...
+    this.resolutionTextHeightUpdateRequired = false
+    this.$refs.resolutionTextRef.calculateInputHeight()
+  }
+
+  // Previously, the resolution text area would not render to the appropriate height relative to the amount of content
+  // when navigating from another step.  In hooking into the visibility change event on the resolution text area via the
+  // v-observe-visibility property, we are able to force a re-calculation of the text area height when a user navigates
+  // to the complete resolution summary step from another step for the first time. This results in the text area being
+  // rendered to the appropriate height.
+  private async onResolutionVisibilityChanged (isVisible, entry) {
```
severinbeauvais, line 182: Also not async :)

I'm curious to see this in action. Looks neat.

src/components/CreateResolution/CompleteResolution.vue
```diff
@@ -653,6 +662,21 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin, E
...
+    this.$refs.resolutionTextRef.calculateInputHeight()
+  }
+
+  // Previously, the resolution text area would not render to the appropriate height relative to the amount of content
+  // when navigating from another step.  In hooking into the visibility change event on the resolution text area via the
+  // v-observe-visibility property, we are able to force a re-calculation of the text area height when a user navigates
+  // to the complete resolution step from another step for the first time. This results in the text area being rendered
+  // to the appropriate height.
```
severinbeauvais, line 673: Thanks for the comment. This is exactly the kind of thing that needs a comment 👍 

So the component resizes when the user lands on it, or when the page renders?

argush3: it resizes when the component is actually on the complete resolution step and the component is visible in the users view port

src/components/CreateResolution/CompleteResolution.vue
severinbeauvais: Also `onResolutionVisibilityChanged()` below.

src/components/CreateResolution/CompleteResolutionSummary.vue
```diff
@@ -129,6 +137,11 @@ export default class CompleteResolutionSummary extends Mixins(DateMixin) {
     return '(Not Entered)'
   }
 
+  @Watch('getCreateResolutionStep.resolutionText')
+  private onResolutionTextChanged (): void {
```
severinbeauvais, line 141: Can you just watch `this.resolutionText`, instead of the field on the getter?

severinbeauvais: Or could you add a method to the `@change` event of resolutionTextRef?

argush3: true, let me test that.

src/components/CreateResolution/CompleteResolution.vue
```diff
@@ -653,6 +662,21 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin, E
     this.updateResolutionStepValidationDetail()
   }
 
+  private async reRenderResolutionText (): Promise<void> {
+    this.$refs.resolutionTextRef.calculateInputHeight()
+  }
```
severinbeauvais, line 667: This is such a small method, and only called from 1 place. Could its body just be moved to the caller?

(And same on CompleteResolutionSummary page.)

tests/jest-wrapper-factory.ts
```diff
@@ -11,13 +13,19 @@ import { getVuexStore } from '@/store'
 // Utils
 import { createLocalVue, mount, shallowMount } from '@vue/test-utils'
 import { IncorporationResources } from '@/resources'
+// This polyfill(intersection-observer) was installed as a dev dependency as without it components being tested
+// that use the IntersectionObserver API results in warnings that browser does not support this API.
+// e.g. IntersectionObserver API is not available in your browser.
+//      Please install this polyfill: https://github.com/w3c/IntersectionObserver/tree/master/polyfill
+require('intersection-observer')
```
severinbeauvais, line 20: Thanks for the comment. (It answers the question "why is this here?")

Why use "require"? Does "import" work as well?

src/components/CreateResolution/CompleteResolution.vue
argush3: yes

tests/jest-wrapper-factory.ts
argush3: i think i was just looking at the documentation and just used what was provided.  Will try importing like we do normally.

src/assets/styles/base.scss
argush3: sure, will ask

argush3: reply from Tracey:

> Tracey Lebedovich @tracey.lebedovich
I think just keep it simple and don’t change the cursor.
aka - keep it as a regular pointer cursor
(not text cursor, not help icon)

severinbeauvais: ok

src/components/CreateResolution/CompleteResolution.vue
argush3: updated

src/components/CreateResolution/CompleteResolutionSummary.vue
argush3: updated to watch `resolutionText`.  Tried `@change` but things weren't updating properly.

tests/jest-wrapper-factory.ts
argush3: updated to use `import` from `require`

src/components/CreateResolution/CompleteResolution.vue
argush3: removed this function and moved logic into onResolutionVisibilityChanged

src/components/CreateResolution/CompleteResolutionSummary.vue
argush3: removed `async`.  yeah, it's kinda cool...this directive.  I could see it coming in handy for other stuff.

src/components/CreateResolution/CompleteResolution.vue
argush3: Removed both instances of `reRenderResolutionText` function and moved logic into `onResolutionVisibilityChanged` function.

src/components/CreateResolution/CompleteResolutionSummary.vue
severinbeauvais: Could you do a small tech talk on it?

argush3: sure, i'll talk about it a bit at the BCROS developer demo thing.

src/components/Affidavit/Affidavit.vue
argush3: updated to use correct syntax and tested to be applying styles correctly.

src/components/CreateResolution/CompleteResolution.vue
argush3: updated to use correct syntax and tested to be applying styles correctly.

severinbeauvais (approved):
Looks good,

PR #310, opened 2021-12-06
Validation and Label fixes

cameron-freshworks:
*Issue #:* /bcgov/entity#10394 && /bcgov/entity#10395

*Description of changes:*
* Fixed logic on when to apply EffectiveDate validations, removed Staff Only clause.
* Update v-card-title with word break to override base vuetify styling of `break all`.

### DIFF
--- a/src/store/getters/state-getters.ts
+++ b/src/store/getters/state-getters.ts
@@ -433,7 +433,7 @@ export const isDissolutionValid = (state: StateIF): boolean => {
   const isCourtOrderValid = isStaff ? getCourtOrderStep(state).valid : true
   const isStaffPaymentValid = isStaff ? getStaffPaymentStep(state).valid : true
 
-  const isEffectiveDateTimeValid = (isBaseCompany(state) && isStaff)
+  const isEffectiveDateTimeValid = isBaseCompany(state)
     ? getEffectiveDateTime(state).valid
     : true
 
--- a/tests/unit/ReviewConfirmDissolution.spec.ts
+++ b/tests/unit/ReviewConfirmDissolution.spec.ts
@@ -58,7 +58,7 @@ for (const test of reviewConfirmTestCases) {
       expect(wrapper.find('h2').text()).toBe('Review and Confirm')
     })
 
-    it('displays Effective Date Time section only for corp and staff', () => {
+    it('displays Effective Date Time section for corp as both staff and non-staff', () => {
       wrapper = shallowWrapperFactory(
         ReviewConfirmDissolution,
         null,
--- a/tests/unit/ReviewConfirmDissolution.spec.ts
+++ b/tests/unit/ReviewConfirmDissolution.spec.ts
@@ -58,6 +58,21 @@ for (const test of reviewConfirmTestCases) {
       expect(wrapper.find('h2').text()).toBe('Review and Confirm')
     })
 
+    it('displays Effective Date Time section for corp', () => {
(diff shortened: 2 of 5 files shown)


lekshmimallika-aot (approved):

argush3 (approved):

src/store/getters/state-getters.ts
```diff
@@ -433,7 +433,7 @@ export const isDissolutionValid = (state: StateIF): boolean => {
   const isCourtOrderValid = isStaff ? getCourtOrderStep(state).valid : true
   const isStaffPaymentValid = isStaff ? getStaffPaymentStep(state).valid : true
 
-  const isEffectiveDateTimeValid = (isBaseCompany(state) && isStaff)
+  const isEffectiveDateTimeValid = isBaseCompany(state)
```
severinbeauvais, line 436: Oh right, this WAS changed since FED was implemented.

Could you please update the corresponding unit test?
https://github.com/bcgov/business-create-ui/blob/49b35ce721bceb3f7f85a1be2f62a230b9324da9/tests/unit/ReviewConfirmDissolution.spec.ts#L61

cameron-freshworks: You bet, done and done!

cameron-freshworks: Original implementation was Staff only i believe, so this definitely made sense then!

severinbeauvais: I was thinking you'd remove the staff code from the unit test function, since it's no longer relevant.

severinbeauvais: Glad you fixed the function description though :)

cameron-freshworks: Ah i see, well it doesn't hurt either :) Not a bad thing to test for both roles right!

cameron-freshworks: We already know it renders from other tests, so i figured this was better than just removing the test.
But we can certainly do that if you prefer. I'm good with anything!

severinbeauvais: No, it doesn't hurt, but it's an example of the undesirable "decoy pattern"... where code looks like it's doing something but it's a red herring.
(Enough metaphors in here for you? haha)

cameron-freshworks: Will remove! 😁

tests/unit/ReviewConfirmDissolution.spec.ts
```diff
@@ -58,22 +58,6 @@ for (const test of reviewConfirmTestCases) {
...
-        {
-          entityType: test.entityType,
-          tombstone: { authRoles: test.isStaff ? ['staff'] : [] }
-        },
-        null,
-        DissolutionResources
-      )
-
-      const expected = (test.entityType !== 'CP')
-      expect(wrapper.find('#effective-date-time').exists()).toBe(expected)
```
severinbeauvais, line 74: Oh no, don't remove the unit test. It's still valuable for checking the presence of the component depending on the entity type. Just remove the bit that sets the authRoles depending on whether the user is staff or not, since the behaviour is the same in both cases.

cameron-freshworks: Ah i see, i was caught by the decoys. 
I thought the above test was for the EffectiveDateTime as well, but it's testing for the View as a whole. 
Restored the test.

severinbeauvais (approved):

PR #321, opened 2022-01-19
Dissolutions staff support and payment flow

cameron-freshworks:
*Issue #:* /bcgov/entity#9529

*Description of changes:*
* Refactors to support Dissolution Staff Filings
* Add staff payment flow for dissolutions (but not IAs)
* Validation fixes on Step 2 when resuming from save or pay failure
* Fix issue with Court Order / POA not meeting schema requirements
* Updated unit tests 
* Updated theme to match Filings UI
* edit/renamed Contact Info Component

### DIFF
--- a/src/components/CreateResolution/CompleteResolution.vue
+++ b/src/components/CreateResolution/CompleteResolution.vue
@@ -516,7 +516,7 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin, E
   }
 
   private get isConfirmResolutionValid (): boolean {
-    return this.$refs.confirmResolutionChkRef.valid
+    return this.$refs.confirmResolutionChkRef?.valid
   }
 
   private isResolutionValid (): boolean {
@@ -525,9 +525,9 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin, E
         this.$refs.resolutionDatePickerRef.isDateValid() &&
         this.isSigningPersonValid &&
         this.$refs.signatureDatePickerRef.isDateValid() &&
-        this.$refs.confirmResolutionChkRef.valid
+        this.$refs.confirmResolutionChkRef?.valid
     }
-    return this.$refs.confirmResolutionChkRef.valid
+    return this.$refs.confirmResolutionChkRef?.valid
   }
 
   private updateResolutionStepValidationDetail () {
--- a/src/components/common/ContactInfo.vue
+++ b/src/components/common/ContactInfo.vue
@@ -1,18 +1,18 @@
 <template>
-  <ul class="pl-0">
+  <ul class="contact-info pl-0">
     <li class="contact-container">
-      <v-icon small class="contact-icon">mdi-phone</v-icon>
-      <span class="font-weight-bold contact-key">Canada and U.S. Toll Free:</span>
+      <v-icon class="contact-icon" size="18px">mdi-phone</v-icon>
+      <span class="px-2">Canada and U.S. Toll Free:</span>
       <a href="tel:+1-877-526-1526" class="contact-value">1-877-526-1526</a>
     </li>
     <li class="contact-container">
-      <v-icon small class="contact-icon">mdi-phone</v-icon>
-      <span class="font-weight-bold contact-key">Victoria Office:</span>
+      <v-icon class="contact-icon" size="18px">mdi-phone</v-icon>
+      <span class="px-2">Victoria Office:</span>
       <a href="tel:+1-250-952-0568" class="contact-value">250-952-0568</a>
     </li>
     <li class="contact-container">
-      <v-icon small class="contact-icon">mdi-email</v-icon>
-      <span class="font-weight-bold contact-key">BC Registries Email:</span>
+      <v-icon class="contact-icon" size="18px">mdi-email</v-icon>
+      <span class="px-2">Email:</span>
       <a href="mailto:BCRegistries@gov.bc.ca" class="contact-value">BCRegistries@gov.bc.ca</a>
     </li>
   </ul>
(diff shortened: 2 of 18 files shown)


src/mixins/filing-template-mixin.ts
```diff
@@ -377,7 +377,7 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
     if (courtOrder.hasPlanOfArrangement || courtOrder.fileNumber) {
       filing.dissolution.courtOrder = {
         fileNumber: courtOrder.fileNumber,
-        effectOfOrder: courtOrder.hasPlanOfArrangement ? EffectOfOrders.PLAN_OF_ARRANGEMENT : null,
+        effectOfOrder: courtOrder.hasPlanOfArrangement ? EffectOfOrders.PLAN_OF_ARRANGEMENT : '',
```
cameron-freshworks, line 380: This would cause a failure when run against the schema as the schema calls for `string` type

src/components/CreateResolution/CompleteResolution.vue
```diff
@@ -516,7 +516,7 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin, E
   }
 
   private get isConfirmResolutionValid (): boolean {
-    return this.$refs.confirmResolutionChkRef.valid
+    return this.$refs.confirmResolutionChkRef?.valid
```
cameron-freshworks, line 519: When resuming from payment or save failures, Step 2 was seeing a type error which prevent the Checkbox component from re rendering. This and the optional chains below fix that. 
I believe we saw this bug before in other instances, not sure if we had a bug ticket for this, but this will fix it.

src/components/common/ContactInfo.vue
```diff
@@ -1,18 +1,18 @@
 <template>
```
severinbeauvais, line 1: This really needs to be a shared component 😁 

(future)

cameron-freshworks: Agreed!

src/enums/roleTypes.ts
```diff
@@ -4,5 +4,6 @@ export enum RoleTypes {
   DIRECTOR = 'Director',
   SUBSCRIBER = 'Subscriber',
   LIQUIDATOR = 'Liquidator',
-  CUSTODIAN = 'Custodian'
+  CUSTODIAN = 'Custodian',
+  STAFF = 'STAFF'
```
severinbeauvais, line 8: Why a different case for staff?

src/mixins/filing-template-mixin.ts
severinbeauvais: I think string allows null.

severinbeauvais (approved):
Looks good.

So we're NOT opening a Staff Payment dialog after they click File and Pay? The UI/UX folks specifically wanted that in Filings UI. 🤷‍♂️

src/mixins/filing-template-mixin.ts
cameron-freshworks: 🤔  It would resolve to `None` when being checked against the schema, so my assumption is that the backend would coerce `null` -> `None` type somewhere along the line. Potentially when saving the json, not 100% sure. 
Either way, this ensured we met schema requirements as it would fail on `effectOfOrder : None`
Think we should dig deeper to this?

src/enums/roleTypes.ts
cameron-freshworks: Mostly to reference an enum value in the logic vs a string. 
This enum made sense as we refer to `staff` as a role type, although not directly in the context of a specific filing as some of the other roles are.
We could declare it as a local const as well, either or.

cameron-freshworks:
> Looks good.
> 
> So we're NOT opening a Staff Payment dialog after they click File and Pay? The UI/UX folks specifically wanted that in Filings UI. 🤷‍♂️

This was an interesting topic as i brought this up specifically to UX. They held a vote (no joke 😁 ) and they wanted to keep it inline. So i can confirm this is intended!

src/mixins/filing-template-mixin.ts
severinbeauvais: You may be right. I see this, which suggest that we have to explicitly allow null:

https://github.com/bcgov/business-schemas/blob/836b85359b67da49a2fc89c5aa6987806f896cef/src/registry_schemas/schemas/address.json#L34

src/enums/roleTypes.ts
severinbeauvais: Should the other items in this list be all caps then? Or is this enum used for 2 different things?

src/mixins/filing-template-mixin.ts
cameron-freshworks: Ah right, i knew there was something previously that had been updated in order for the nulls to pass through. Thanks for asking/sharing this, this is very good to know!

src/enums/roleTypes.ts
cameron-freshworks: I think it's a matter of what we are comparing it too/from. 
`STAFF` is all caps as thats how the value is represented in the `userInfo` response.
In the other use cases RoleTypes is being used as both value and label. If i had to guess the value might not matter if it's all caps but the Label might not want to be all caps. 
So yea, two different cases. 
Perhaps it makes sense to declare a local const of our `STAFF` role for use in the logic vs put it in this enum? 
The use cases do differ so the value format wouldn't really be cohesive with the others.

lekshmimallika-aot (approved):
Looks good

src/enums/roleTypes.ts
severinbeauvais: I recommend a local const for STAFF.

cameron-freshworks: done and done, ty!

vysakh-menon-aot (approved):

argush3 (approved):

PR #324, opened 2022-01-28
Framework Update - Support for Registration Filings

cameron-freshworks:
*Issue #:* /bcgov/entity#9244

*Description of changes:*
* Implements directories for Registrations Views and Resources 
* Registration Base resources files and step templates for SP and GP including minimum tombstone data
* Updates CorpModule version to include NR only 'FR' type for Sole Prop
* Registration Interfaces
* Save / Resume fling template methods for Registration
* Refactors to support varying entity types as TEMP REG number 
* ^ Utilize the filing type from the draft filing in order to differentiate

* TODO: Add `registration` to our `supported-filings` key in LD for DEV only


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "3.1.6",
+  "version": "3.2.0",
```
severinbeauvais, line 3: Excellent.

While you're here, I think it would be good to update sbc-common-components to the latest version, which supports account id sync. See https://github.com/bcgov/bcregistry/pull/51 for details, or we can chat about it some time.

src/App.vue
```diff
@@ -718,23 +713,37 @@ export default class App extends Mixins(
 
     // fetch draft filing
     // NB: will throw if API error
-    let draftFiling = await this.fetchDraftIA()
+    let draftFiling = await this.fetchDraftApplication()
```
severinbeauvais, line 716: You might need to catch a thrown error here. I see that line 725 checks for a null response, but you might also see other errors, which you should handle gracefully.

src/App.vue
```diff
@@ -718,23 +713,37 @@ export default class App extends Mixins(
...
+    if (!this.hasValidFilingState(draftFiling)) return
 
     // merge draft properties into empty filing so all properties are initialized
-    const emptyFiling = this.buildIncorporationFiling()
-    draftFiling = { ...emptyFiling, ...draftFiling }
-
-    // parse draft filing into the store
-    if (draftFiling) {
-      this.parseIncorporationsDraft(draftFiling)
+    let resources, parseFiling: Function
```
severinbeauvais, line 728: Doesn't this also declare `resources` as a Function? If so, that's not right.

src/components/common/Actions.vue
```diff
@@ -241,11 +261,22 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Lega
...
+            break
+          case FilingTypes.REGISTRATION:
+            filing = await this.buildRegistrationFiling()
+            break
+          case FilingTypes.DISSOLUTION:
+            filing = await this.buildDissolutionFiling()
+            break
+        }
 
+        // Save filing
```
severinbeauvais, line 277: I see a lot of code duplication now. Can it be simplified easily?

src/enums/filingCodes.ts
```diff
@@ -12,5 +12,7 @@ export enum FilingCodes {
   DISSOLUTION_COURT_ORDERED_LIQUIDATION = 'DIS_COLQD',
   DISSOLUTION_RESTORED_IN_ERROR = 'DIS_RSTR',
   SPECIAL_RESOLUTION = 'SPRLN',
-  AFFIDAVIT = 'AFDVT'
+  AFFIDAVIT = 'AFDVT',
+  REGISTRATION_SP = 'FRREG',
+  REGISTRATION_GP = 'FRREG'
```
severinbeauvais, line 17: Does these really have the same code?

package.json
cameron-freshworks: Absolutely, will do this!

src/App.vue
cameron-freshworks: I'm not sure, i didn't think so but i can look closer at this. Will do some testing.

src/enums/filingCodes.ts
cameron-freshworks: Indeed
See ticket: https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/9251

cameron-freshworks: In the comments, Linda posts about her research into these.

src/interfaces/filing-interfaces/filing-interface.ts
```diff
@@ -69,6 +69,38 @@ export interface IncorporationFilingIF {
...
+    identifier: string
+  },
+  registration: {
+    // NB: nameRequest must match schema
+    nameRequest: {
+      legalType: string
+      nrNumber?: string // only set when there is an NR
+      legalName?: string // only set when there is an NR
+    },
+    offices: BaseAddressObjIF | {},
```
severinbeauvais, line 94: I'm not exactly sure how an "empty office" is used, but an empty object seems suspicious: it's easier to check for a null office than an empty office object.

I see this isn't the only object declared as possibly empty in this file so 🤷‍♂️

src/mixins/legal-api-mixin.ts
```diff
@@ -156,4 +164,31 @@ export default class LegalApiMixin extends Vue {
     }
     return toReturn
   }
+
+  /**
+   * Ensure consistent object structure for a registration application,
+   * whether it contains a Name Request or not, and whether it is an initial
+   * draft or it has been previously saved. Object merging does not
+   * work very well otherwise due to nested properties.
```
severinbeauvais, line 172: I wouldn't feel bad using lodash functions to merge objects, if that makes our life easier.

Another possibility that may be easier is to declare an empty object (in the IF file) and then we can assign (deep copy) that when we need it.

severinbeauvais: (if alterantives are not easier then don't worry about it)

severinbeauvais (approved):
Just a few small questions and comments but looks good!

src/enums/filingCodes.ts
severinbeauvais: ok

package.json
cameron-freshworks: So mostly just unit tests breaking, but will work through these 👍

src/App.vue
cameron-freshworks: Confirmed that this syntax doesn't cast the initial definition as well.

src/components/common/Actions.vue
cameron-freshworks: Updated in next commit 👍

src/interfaces/filing-interfaces/filing-interface.ts
cameron-freshworks: hmm yea, in an interface i would expect ONLY the type
Will update this.

src/App.vue
severinbeauvais: OK, I just learned something!

How about putting those on separate lines?

cameron-freshworks: Absolutely! 
Updated in next commit

package.json
cameron-freshworks: I'm a bit of a loss how to resolve these failing tests 😵 
Will keep testing but it's quite odd, as when run individually they can succeed but when run in the full test suite, they fail and/or time out. 
Interesting problem for sure, not one i think we can ignore since we will NEED to update the common components.

src/mixins/legal-api-mixin.ts
cameron-freshworks: I'm ok with the current implementation although if we want to remove these methods we can definitely refactor.
I can't recall the reason for initial implementation, do you recall?

severinbeauvais: Don't refactor unless it simplifies your life.

I think you implemented this originally, and it had to do with merging data into an "empty" object.

src/App.vue
cameron-freshworks: Sorry i forgot to update here.
An error in the request or if the filing is somehow fails to build/null it will be caught in our `fetchData` in our `catch` block starting on line 623 of `app`

severinbeauvais: Oh, right, that's a try-catch block that wraps this function.

PR #326, opened 2022-02-01
Updated CC version + minor refactors

cameron-freshworks:
*Issue #:* /bcgov/entity#10922

*Description of changes:*
* Updated CC version
* Minor refactors to resolve Unit test breakages

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -151,7 +151,8 @@ import * as Sentry from '@sentry/browser'
 import { getFeatureFlag, updateLdUser } from '@/utils'
 
 // Components
-import { PaySystemAlert, SbcFeeSummary, SbcFooter, SbcHeader } from '@/components'
+import SbcHeader from 'sbc-common-components/src/components/SbcHeader.vue'
+import { PaySystemAlert, SbcFeeSummary, SbcFooter } from '@/components'
 import { Actions, BreadCrumb, EntityInfo, Stepper } from '@/components/common'
 import * as Views from '@/views'
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "3.2.0",
+  "version": "3.2.1",
   "lockfileVersion": 1,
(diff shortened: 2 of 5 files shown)


src/App.vue
```diff
@@ -151,7 +151,8 @@ import * as Sentry from '@sentry/browser'
 import { getFeatureFlag, updateLdUser } from '@/utils'
 
 // Components
-import { PaySystemAlert, SbcFeeSummary, SbcFooter, SbcHeader } from '@/components'
+import SbcHeader from 'sbc-common-components/src/components/SbcHeader.vue'
+import { PaySystemAlert, SbcFeeSummary, SbcFooter } from '@/components'
```
severinbeauvais, line 155: Isn't this equivalent to the previous code?

severinbeauvais: NVM - I see what you've done.

Why did you do this?

severinbeauvais (approved):
Must be \*\*magic\*\* because I don't see how this fixed the test breakage...?

src/App.vue
cameron-freshworks: Something has changed on the components side, that impacted how we import/exported out of our common index. 
I imagine it's more to do with how we create a reference for it locally, ie `export ( default as ... )`
I don't know how/why at this time. Will look into this some more though in the morning as per the explanation. 
Perhaps there is a better way to do this!

cameron-freshworks:
> Must be **magic** because I don't see how this fixed the test breakage...?

Magic indeed! 
Something wasn't resolving in the test with regards to the sbc-Header obviously. 
It's not the header component itself but rather it's sub components. 
Perhaps my solution was too heavy handed and i can just import directly for Unit tests, since thats the time we were getting errors. 
Will look into this more! 👍

src/App.vue
cameron-freshworks: Not quite, the import/export reference we created in our index file would no longer work.
I've yet to discover the reason why a direct import resolves but looking closer at this.

PR #327, opened 2022-02-02
10853 Login and logout redirects

severinbeauvais:
*Issue #:* /bcgov/entity#10853

*Description of changes:*
- initialize token service in main instead of App (needed to sync session before routing)
- redirect to BC Registry for login, then return here
- after logout, redirect to login page
- app version = 3.2.2


cameron-freshworks (approved):
Looks good!

PR #330, opened 2022-02-05
10999 Handle account id sync

severinbeauvais:
*Issue #:* /bcgov/entity#10999

*Description of changes:*
- stored account id in session
- created "navigate" function to automatically add accountid parameter
- replaced window.location.assign with new function
- fixed date/time conversions broken by Node 12
- updated breadcrumb resources to append account id
- fixed invalid URLs in unit tests
- deleted Auth API mixin (use Auth Services instead)
- app version = 3.2.3


src/App.vue
```diff
@@ -302,12 +300,18 @@ export default class App extends Mixins(
...
+  }
+
   /** The route breadcrumbs list. */
   private get breadcrumbs (): Array<BreadcrumbIF> {
     const crumbs: Array<BreadcrumbIF> = [
       {
         text: this.legalName || this.getNumberedEntityName,
-        href: `${sessionStorage.getItem('DASHBOARD_URL')}${this.getEntityIdentifier}`
+        href: `${sessionStorage.getItem('DASHBOARD_URL')}${this.getEntityIdentifier}/${this.getParams()}`
       },
```
severinbeauvais, line 315: I'm thinking of moving this breadcrumb to BreadCrumbResources.ts to get rid of the duplicate `getParams()` above, but on the other hand I'd have to pass in legalName, getNumberedEntityName and getEntityIdentifier. Thoughts?

cameron-freshworks: Fine with whatever, should work as long as our resources are localized.
Is it easy to bind Getters in a resource file?

severinbeauvais: It's possible, yes. For example:
```
import { getVuexStore } from '@/store'
const store = getVuexStore()
const getFilingType = store.getters.getFilingType as FilingTypes
```

I'm not sure which I prefer. Or rather, dislike the least 😛

cameron-freshworks: haha i hear ya. 
Your call!

cameron-freshworks (approved):

lekshmimallika-aot (approved):

src/App.vue
severinbeauvais: Well, have a look at my new commit and tell me what you think.

cameron-freshworks: Looks fine to me!

severinbeauvais: OK. If we end up hating this code then we can change it later.

PR #331, opened 2022-02-07
10999 Fixed init bug

severinbeauvais:
*Issue #:* /bcgov/entity#10999

*Description of changes:*

- init app at end of create(), not on initial route change
- no need to update version number since the deployed app is currently broken and it will be obvious when the fix is deployed :)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -435,6 +435,9 @@ export default class App extends Mixins(
       console.log('Error while retrieving NR during File and Pay') // eslint-disable-line no-console
       this.nameRequestInvalidErrorDialog = true
     })
+
+    // init app
+    this.onRouteChanged()
   }
 
   /** Fetches and stores the current JS date. */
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -900,19 +903,13 @@ export default class App extends Mixins(
...
     if (!this.isRouteName(RouteNames.SIGN_IN) && !this.isRouteName(RouteNames.SIGN_OUT)) {
       this.assignIdentifier()
-
       this.setCurrentStep(this.$route.meta?.step || 1)
-
-      // wait a moment for token to be available in session storage
-      // and JS date to be fetched from server
-      await Vue.nextTick()
-
       await this.fetchData(true)
```
severinbeauvais, line 913: What happened here is the route change handler ran before created(), so the initial time/date was not set.

Now, this function is not run "immediately", and I call it at the end of created() instead. Tested and works fine.

cameron-freshworks (approved):

cameron-freshworks:
Merge when ready, so i can pull it in :)

severinbeauvais:
Merged!

PR #334, opened 2022-02-10
9245 Component refactoring and cleanup before further development

severinbeauvais:
*Issue #:* /bcgov/entity#9245

*Description of changes:*

The point of this refactor is to provide some separation between filing types in this project, to support concurrent development efforts. In a nutshell, this was done by moving some components around, renaming some ids/files/routes, and especially getting rid of most `index.ts` files for components. Previously, the index files abstracted component imports, leading to nested imports, conflicting imports, and a general inability to tell which component was using which other component (and vice versa). This refactor also includes a few small fixes.

- fixed isDissolutionFiling logic in fetchData()
- added reroute logic to fetchData()
- replaced duplicate code with checkAuth()
- added validateAndScroll() to Incorporation Agreement component
- added/updated some component IDs
- moved dialogs/ out of components/
- moved misc components to specific sub-folders
- deleted component index files and imported components directly
- pruned filing names enum
- pruned filing types enum
- updated most route names
- updated some component names
- updated most route paths
- renamed all views
- updated unit tests
- enabled date mixin unit test that should now work in GHA
- fixed unit test warnings/errors
- misc cleanup
- updated sbc-common-components to 2.4.39
- app version = 3.3.0


src/components/DefineRegistration/DefineRegistrationSummary.vue
```diff
@@ -61,21 +67,12 @@
 </template>
 
 <script lang="ts">
-// Libraries
```
severinbeauvais, line 64: I think we now know what these are and no longer need these comments.

src/components/common/Actions.vue
```diff
@@ -4,7 +4,7 @@
     <div class="buttons-left">
       <v-btn
         v-if="isSummaryStep"
-        id="app-summary-cancel-btn"
+        id="app-cancel-btn"
```
severinbeauvais, line 7: Yes, there's another button with this same ID on this page, but they're never displayed at the same time, so I think it's easier to have a single ID. Thoughts?

src/enums/filingNames.ts
```diff
@@ -1,9 +1,5 @@
...
-  INCORPORATION_APPLICATION= 'Incorporation Application',
-  VOLUNTARY_DISSOLUTION_FILING = 'Voluntary Dissolution',
-  INVOLUNTARY_DISSOLUTION_FILING = 'Involuntary Dissolution',
-  ADMINISTRATIVE_DISSOLUTION_FILING = 'Administrative Dissolution',
-  VOLUNTARY_LIQUIDATION_FILING = 'Voluntary Liquidation',
-  COURT_ORDERED_LIQUIDATION_FILING = 'Court Ordered Liquidation',
-  REGISTRATION = 'Registration'
+  INCORPORATION_APPLICATION = 'Incorporation Application',
+  REGISTRATION = 'Registration Application',
+  VOLUNTARY_DISSOLUTION = 'Voluntary Dissolution',
```
severinbeauvais, line 4: I think we're not going to need those other filing names (and types) after all.

src/resources/Registrations/stepTemplates.ts
```diff
@@ -9,25 +9,25 @@ export const RegistrationSteps = [
     id: 'step-1-btn',
     step: 1,
     icon: 'mdi-domain',
-    text: 'Define Your \nCompany',
-    to: RouteNames.DEFINE_REGISTRATION,
-    component: ViewComponentNames.DEFINE_REGISTRATION
+    text: 'Define Your \nBusiness',
```
severinbeauvais, line 12: Changed text to match UI design.

src/store/index.ts
```diff
@@ -18,7 +23,12 @@ export function getVuexStore () {
     state: { ...State },
     getters: { ...Getters },
     mutations: { ...Mutations },
-    actions: { ...Actions }
+    actions: { ...Actions },
+    modules: {
+      dissolution: DissolutionModule,
+      incorporation: IncorporationModule,
+      registration: RegistrationModule
```
severinbeauvais, line 30: I think you can see where this is going, muhahaha.

For now, this doesn't seem to hurt anything.

src/store/modules/dissolution.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  @Mutation
+  public setMyVal (val: string) {
+    this.myVal = val
+  }
+
+  @Action({ commit: 'setMyVal', rawError: true })
+  public getMyVal (): string {
+    return this.myVal
+  }
```
severinbeauvais, line 15: To be implemented... and also the other 2 below.

src/store/index.ts
severinbeauvais: It remains to be seen whether it's easier/cleaner to use separate modules to store separate filing data...

src/views/Dissolution/Affidavit.vue
```diff
@@ -18,7 +18,7 @@ import { ActionBindingIF, UploadAffidavitIF } from '@/interfaces'
     Affidavit
   }
 })
-export default class UploadAffidavit extends Mixins(CommonMixin) {
```
severinbeauvais, line 21: The classname is thrown away by the index.ts file (which imports default as XXX, thus assigning the class name at that time).

tests/unit/date-mixin.spec.ts
```diff
@@ -20,8 +20,7 @@ describe('Date Mixin', () => {
     await Vue.nextTick()
   })
 
-  // FUTURE: fix so this works in GH CI action
```
severinbeauvais, line 23: This works now that the Date Mixin uses the lowest common denominator for local ("en-US").

src/assets/styles/base.scss
```diff
@@ -315,5 +309,5 @@ ul.list {
 
 .step-container {
   margin-top: 1rem;
-  padding: 1.25rem;
+  padding: 1.5rem;
```
cameron-freshworks, line 312: This may throw off some alignment on the review summary pages.. can't recall for sure. 
Either way, good to ensure all the step 1 and summaries are still in alignment 👍

src/components/README.md
```diff
@@ -0,0 +1,12 @@
...
+`components/common/` contains components used by multiple filing types
+`components/Dissolution/` contains components used only by dissolution filings
+`components/Incorporation/` contains components used only by incorporation filings
+`components/Registration/` contains components used only by registration filings
+
+# WARNING
+
+To prevent circular dependencies, do not create or use `index.ts` files for components
+in this folder or sub-folders, EXCEPT FOR `@/components/index.ts`, which is used
+exclusively by external components.
```
cameron-freshworks, line 12: Helpful!

cameron-freshworks (approved):
Looks great to me! 
Obviously we will need to give it some good regression testing / layout review etc. 
Would benefit from a few sets of eyes 👀

src/assets/styles/base.scss
severinbeauvais: I think this class isn't used for any review summary pages. I have spot-checked all pages for all filing types and it looks OK, I think. As you mentioned below, a second (and third) set of eyes would be great.

image

severinbeauvais: I see LOTS of inconsistent whitespace in different filings/pages but I'm trying to touch as little as possible this time around!

vysakh-menon-aot (approved):

PR #336, opened 2022-02-15
9245 Implement Business Addresses and more

severinbeauvais:
*Issue #:* /bcgov/entity#9245

*Description of changes:*

- implemented Business Addresses functionality
- implemented Fee Acknowledgment functionality
- implemented initial People and Roles stuff
- added Folio Number to registration flow
- added Staff Payment to registration flow
- added Certify to registration to registration flow
- implemented validation on registration pages
- updated registration interfaces, store, etc
- updated some prop access modifiers
- updated some created/mounted access modifiers
- updated some event names for consistency
- updated Folio Number component
- deleted unneeded created() in views
- declared some "empty" objects for init
- updated SP/GP resources (for P&R)
- excluded SP/GP from name translations
- improved typing in filing template
- fixed Org Person template error
- fixed Staff Payment changes not triggering "have changes" alert
- replaced hard-coded font sizes with CSS variables
- simplified address interfaces
- misc interfaces cleanup
- misc store cleanup
- misc layout tweaks
- added/updated unit tests
- app version = 3.3.3


src/components/Registration/StartDate.vue
```diff
@@ -43,15 +43,11 @@ export default class StartDate extends Mixins(DateMixin) {
...
-  /** Initial date string value. */
-  @Prop({ default: '' })
-  readonly initialValue: string
-
   // Global actions
   @Action setRegistrationStartDate!: ActionBindingIF
 
   // Global getters
-  @Getter getCurrentJsDate!: Date
+  @Getter getRegistration!: RegistrationStateIF
```
severinbeauvais, line 50: I figured, if this component writes to the store (line 47) then it might as well read from the store and save a prop and simplify the parent view.

src/components/common/BusinessContactInfo.vue
```diff
@@ -150,15 +150,15 @@ export default class BusinessContactInfo extends Mixins(CommonMixin) {
...
 
   // Events
   @Emit('contactInfoChange')
-  private emitContactInfo (contactInfo: BusinessContactIF): void { }
+  private emitContactInfo (contactInfo: BusinessContactIF): void {}
 
-  @Emit('contactInfoFormValidityChange')
-  private emitContactFormState (valid: boolean): void { }
+  @Emit('valid')
+  private emitValid (valid: boolean): void {}
```
severinbeauvais, line 161: Just trying to make the event names consistent.

src/interfaces/store-interfaces/state-interfaces/dissolution-state-interface.ts
```diff
@@ -1,10 +1,13 @@
...
+import { OrgPersonIF } from '@/interfaces'
 
 export interface DissolutionStateIF {
   dissolutionType: DissolutionTypes
   dissolutionDate: string
-  dissolutionStatementStep: DissolutionStatementIF
+  dissolutionStatementStep: {
+    valid: boolean
+    dissolutionStatementType?: DissolutionStatementTypes
+  },
```
severinbeauvais, line 10: I removed a nested interface here.

In my opinion, a nested interface is mostly useful if you're going to use the property as a whole. Here, we don't need to do that, so I simplified things.

src/interfaces/store-interfaces/state-interfaces/registration-state-interface.ts
```diff
@@ -1,4 +1,8 @@
+import { BusinessAddressIF } from '@/interfaces/Registration/business-address'
+
 export interface RegistrationStateIF {
-  valid: boolean
+  defineBusinessValid: boolean
+  peopleRolesValid: boolean
   startDate: string
+  businessAddress: BusinessAddressIF | {}
```
severinbeauvais, line 7: Conversely, here, we do use `businessAddress` as a whole, in several places, so it makes sense for it to have its own interface.

src/components/Registration/StartDate.vue
severinbeauvais: FYI, I'm also thinking about the layered architecture here:
- App.vue is the top level
- views (pages) are still high-level and shouldn't deal with "details" (if those details can be handled by components)
- components handle all the details, include store getters/actions if possible
- low-level (and shared/common) components should not read/write to the store, but should communicate with props/events if possible (as a last resort, they may have a callable public method)

The 2 ideas above are:
1. encapsulation of details / each file should know only what it needs to know
2. API to low-level components (which is part of 1. since low-level components should not know about the store)

src/components/Registration/BusinessAddresses.vue
```diff
@@ -0,0 +1,412 @@
...
+  /**
+   * Updates the specified address' data.
+   * @param addressToUpdate the address type to set the data of
+   * @param address the new address data
+   */
+  updateAddress (addressToUpdate: string, address: AddressIF): void {
+    switch (addressToUpdate) {
+      case this.MAILING_ADDRESS:
+        // avoid reactive looping
+        if (isEqual(this.mailingAddress, address)) break
```
severinbeauvais, line 208: I found that BaseAddress component sends an update every time its address prop is updated (which is now reactive), which caused an infinite loop. This code breaks that loop.

src/components/common/AgreementType.vue
```diff
@@ -33,7 +33,7 @@
       </div>
     </div>
 
-    <div v-else-if="isTypeCC">
+    <div v-else-if="isTypeCCC">
```
severinbeauvais, line 36: The idea with this getter (and the others) is to abstract away from its value and instead use a human-understandable label. In this case a "CCC" is a common high level term for this type of business.

src/interfaces/filing-interfaces/filing-interface.ts
```diff
@@ -1,9 +1,8 @@
 import {
-  BaseAddressObjIF,
-  IncorporationAddressIF
-} from '@/interfaces/stepper-interfaces/DefineCompany/address-interface'
```
severinbeauvais, line 4: All interfaces can be imported from `@/interfaces`. It seems there's no issue with nested imports. (Unlike nested components.)

src/interfaces/payment-interfaces/fees-interface.ts
```diff
@@ -1,21 +1,22 @@
...
+  tax?: TaxesIF
+  total?: number
 }
 
 export interface TaxesIF {
   pst: number
   gst: number
 }
 
+// NB: use cloneDeep when assigning EmptyOrgPerson
```
severinbeauvais, line 19: Due to nested object.

src/views/Dissolution/DissolutionAffidavit.vue
```diff
@@ -22,17 +22,6 @@ export default class DissolutionAffidavit extends Mixins(CommonMixin) {
...
-  @Action setIgnoreChanges!: ActionBindingIF
-
-  /** Called when component is created. */
-  private created (): void {
-    // ignore data changes until page has loaded
-    this.setIgnoreChanges(true)
-    Vue.nextTick(() => {
-      this.setIgnoreChanges(false)
-    })
-  }
```
severinbeauvais, line 34: I believe this is no longer needed since the data is now loaded before the page is created.

src/components/common/OfficeAddresses.vue
```diff
@@ -595,13 +593,13 @@ label:first-child {
 }
 
 .inherit-checkbox {
-  margin-top: -3px;
+  margin-top: 0;
```
severinbeauvais, line 596: ^^ So the label and checkbox line up.

src/components/common/OrgPerson.vue
```diff
@@ -363,6 +364,9 @@ export default class OrgPerson extends Mixins(EntityFilterMixin, CommonMixin) {
       return
     }
 
+    // if not already assigned, initialize delivery address to prevent template errors
+    if (!this.inProgressDeliveryAddress) this.inProgressDeliveryAddress = cloneDeep(EmptyAddress)
```
severinbeauvais, line 368: This was an error I saw in the dev console at runtime.

src/views/Dissolution/DissolutionReviewConfirm.vue
```diff
@@ -4,7 +4,7 @@
     <section class="mt-10">
       <header>
         <h2>Review and Confirm</h2>
-        <p id="intro-text" class="mt-2">
+        <p class="mt-4 mb-6">
```
severinbeauvais, line 7: We use these same classes in all headers on this page, so the whitespace now looks good :)

src/views/Dissolution/DissolutionReviewConfirm.vue
```diff
@@ -179,7 +179,7 @@
       <header>
         <h2>Dissolution Documents Delivery</h2>
         <p class="mt-4 mb-6">
-          Copies of the dissolution documents will be sent to the following email addresses listed below:
```
severinbeauvais, line 182: I talked to Tracey about fixing the redundancy here.

src/views/Dissolution/DissolutionReviewConfirm.vue
```diff
@@ -579,6 +567,11 @@ h2::before {
     padding: 0.75rem 0 !important;
   }
 
+  // bring the main label down a bit to line up with text-field
+  .row:first-of-type .col:first-of-type {
+    padding-top: 8px !important;
+  }
```
severinbeauvais, line 573: This is unique to the Certify component, since there is a text-field (box) at the top of the second column. This change looks better (and matches the comps).

cameron-freshworks (approved):
Looks good 👍

lekshmimallika-aot (approved):

PR #337, opened 2022-02-18
Fixed CertifiedBy assignment

cameron-freshworks:
*Issue #:* /bcgov/entity#11237

*Description of changes:*
* Remove redundant assignment of certify state when parsing DRAFT dissolution.

*Notes:* 
- This parsing of the draft filing always happens in the case of Incorporations and Dissolutions. 
- In this parsing, the certify state would be assigned but it was redundant as we assign the certifiedBy state from the Account ALWAYS. 
- Previously iteration must have worked as the draft was MOST LIKELY* assigned prior to our assignment in our `App.vue` but after a rework, the timing might have swapped. Subsequently overriding the properties from the account by the properties in an empty draft filing. 

- Another solution would be to assign the `certifiedBy` into the draft value from Filings UI where the Dissolution is created. 

### DIFF
--- a/src/views/Dissolution/DissolutionReviewConfirm.vue
+++ b/src/views/Dissolution/DissolutionReviewConfirm.vue
@@ -375,16 +375,6 @@ export default class DissolutionReviewConfirm extends Mixins(DateMixin) {
     })
   }
 
-  /** Called when component is mounted. */
-  mounted (): void {
-    this.setCertifyState(
-      {
-        valid: this.getCertifyState.valid,
-        certifiedBy: this.getCertifyState.certifiedBy
-      }
-    )
-  }
-
   /** Is true when the Dissolution Date and Time section is invalid. */
   get isDissolutionDateTimeInvalid (): boolean {
     return (this.getValidateSteps && !this.getEffectiveDateTime.valid)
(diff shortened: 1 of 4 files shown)


src/views/Dissolution/DissolutionReviewConfirm.vue
```diff
@@ -375,6 +375,16 @@ export default class DissolutionReviewConfirm extends Mixins(DateMixin) {
...
 
+  /** Called when component is mounted. */
+  mounted (): void {
+    this.setCertifyState(
+      {
+        valid: this.getCertifyState.valid,
+        certifiedBy: this.getCertifyState.certifiedBy
+      }
+    )
+  }
```
severinbeauvais, line 386: Theoretically, this code does nothing -- the certify state was already set in fetchData().

This code _might_ hide the timing issue, but it's not a reliable solution.

I wonder if it's a reactive problem: the Certify component doesn't realize this object has been updated. But theoretically, all store objects are reactive.

Hmmm...

cameron-freshworks: For sure, can look at it further 👍

cameron-freshworks: Removed this, as well as a reassignment in the parsing of a draft Dissolution.

PR #340, opened 2022-02-18
Updated Router Switch case

cameron-eyds:
*Issue #:* /bcgov/entity#11237

*Description of changes:*
* Updated switch case to break switch case and continue execution.


src/App.vue
```diff
@@ -570,16 +570,16 @@ export default class App extends Mixins(
...
-            return
+            break
           case FilingTypes.REGISTRATION:
             this.$router.push(RouteNames.REGISTRATION_DEFINE_BUSINESS).catch(() => {})
-            return
+            break
           default:
             this.invalidRouteDialog = true
-            return
+            break
```
severinbeauvais, line 582: Ah, good point. I thought it was safe to restart fetchData() but it's probably better to continue.

severinbeauvais (approved):

PR #343, opened 2022-02-23
9245 Fixed some resolution page errors

severinbeauvais:
*Issue #:* /bcgov/entity#9245

*Description of changes:*
- added conditional chaining to date picker refs to avoid dereferencing them before they are created
- app version = 3.3.5

### DIFF
--- a/src/components/Dissolution/CompleteResolution.vue
+++ b/src/components/Dissolution/CompleteResolution.vue
@@ -497,19 +497,19 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin, E
   ]
 
   private get isResolutionDateValid (): boolean {
-    return this.$refs.resolutionDatePickerRef.isDateValid()
+    return this.$refs.resolutionDatePickerRef?.isDateValid()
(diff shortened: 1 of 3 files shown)


cameron-freshworks (approved):

src/components/Dissolution/CompleteResolution.vue
```diff
@@ -497,19 +497,19 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin, E
   ]
 
   private get isResolutionDateValid (): boolean {
-    return this.$refs.resolutionDatePickerRef.isDateValid()
+    return this.$refs.resolutionDatePickerRef?.isDateValid()
```
severinbeauvais, line 500: The "resolution page is invalid but all components are valid" bug did not reproduce for me locally. However, when doing hot-reload of code changes, I saw console errors about "unable to call isDateValid on undefined". So that's what this PR fixes. I don't know why the other component refs don't have a problem. This change can't hurt and might fix the issue in Dev.

PR #344, opened 2022-02-25
9842 Misc UI/UX fixes

severinbeauvais:
*Issue #:* /bcgov/entity#9842

*Description of changes:*

Commit 1:
- tweaked NR info whitespace
- changed APPROVED to Approved
- added some typing
- fixed a string
- updated unsaved changes wording
- added resolution validation check
- hide entity info email/phone for registrations
- fixed reg step 2 icon
- updated FN label
- commented out Fee Acknowledgement
- fixed unit tests
- app version = 3.4.0

Commit 2:
- changed legal API mixin to a service
- updated unit tests

Commit 3:
- updated breadcrumb for SBC Staff to be same as Staff
- implemented delayed getter for current account
- add org info to store
- cleaned up auth service
- added getters for SBC Staff
- updated unit tests

Commit 4:
- updated checkbox disabled styles
- updated company name colour
- added "gutter" to review page addresses
- fixed "same as" bug in address components
- updated filing name
- imported latest common components to get base address changes
- fixed user drop-down menu not on top of stepper or fee summary
- misc cleanup


src/components/Dissolution/CompleteResolution.vue
```diff
@@ -577,7 +577,9 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin, E
           ]
         }
     }
-    this.setResolutionStepValidationDetail(validationDetail)
+    if (validationDetail) {
```
severinbeauvais, line 580: Safety check: we don't want to store null for validation detail; instead, leave the default values.

src/components/Dissolution/CompleteResolution.vue
```diff
@@ -605,6 +607,7 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin, E
   }
 
   async mounted (): Promise<void> {
+    // wait for components to load/stabilize then update validation state in store
```
severinbeauvais, line 610: I saw a lot of `await Vue.nextTick()` in this component, suggesting there is some timing fragility. This may be what we're seeing with the stepper saying the component in incomplete, but on the step itself everything looks OK.

src/components/common/EntityInfo.vue
```diff
@@ -58,32 +60,33 @@ export default class EntityInfo extends Mixins(EnumMixin) {
   @Getter getFilingType!: FilingTypes
   @Getter isEntityType!: boolean
   @Getter isIncorporationFiling!: boolean
+  @Getter isRegistrationFiling!: boolean
 
   /** The entity application title.  */
-  private get entityTitle (): string {
+  get entityTitle (): string {
```
severinbeauvais, line 66: There's no harm in getters being called externally (which we don't do anyway). The upside is that now VS Code doesn't say this getter is not used.

src/components/common/NameRequestInfo.vue
```diff
@@ -364,17 +372,18 @@ p {
 }
 
 ul {
-  font-size: $px-14;
+  font-size: $px-15;
```
severinbeauvais, line 375: This is closer to the UI comp without looking weird (due to other font sizes not being exactly according to the spec). This was an easy fix. Doing all isn't easy.

package.json
```diff
@@ -34,7 +34,7 @@
     "lodash": "^4.17.20",
     "pdfjs-dist": "2.0.943",
     "regenerator-runtime": "^0.13.7",
-    "sbc-common-components": "2.4.39",
+    "sbc-common-components": "2.4.41",
```
severinbeauvais, line 37: For updated Base Address (with classnames on fields and whitespace above delivery instructions).

src/App.vue
```diff
@@ -822,35 +831,51 @@ export default class App extends Mixins(
...
+   * Wait up to 10 sec for current account to be synced (typically by SbcHeader).
+   */
+  private async getCurrentAccount (): Promise<any> {
+    let account: any
+    for (let i = 0; i < 100; i++) {
+      const currentAccount = sessionStorage.getItem(SessionStorageKeys.CurrentAccount)
+      account = JSON.parse(currentAccount)
+      if (account) break
+      await sleep(100)
+    }
```
severinbeauvais, line 845: I haven't seen this count get above 1 in this UI, but I poached this code from BC Registry UI where I sometimes saw delays of 2-3 seconds before CURRENT_ACCOUNT was available in session storage. So I think it's good to have this here.

src/App.vue
```diff
@@ -914,8 +937,17 @@ export default class App extends Mixins(
 </script>
 
 <style lang="scss" scoped>
+// display drop-down menu on top of stepper and fee summary
+::v-deep .app-header {
+  z-index: 3;
```
severinbeauvais, line 942: By default the SBC Header has a z-index of 2. What weird is that the user drop-down has a z-index of 8 but it was not honoured here.

I didn't want to change SBC Header for everyone (ie, update common components) because there are various uses of z-index in common components and I'm not familiar with how they all interact.

src/assets/styles/base.scss
```diff
@@ -288,11 +288,6 @@ p {
   border: 0;
 }
 
-// Vue Affix
-.vue-affix {
-  width: 282px;
```
severinbeauvais, line 293: Moved to the only place where this is used (App.vue).

src/assets/styles/overrides.scss
```diff
@@ -108,14 +108,21 @@
...
-  color: $gray7 !important;
+.v-input--checkbox .v-input__control .v-input__slot {
+  label {
+    color: $gray7;
+  }
 }
 
 // Disabled Checkboxes
-.theme--light.v-input--selection-controls.v-input--is-disabled:not(.v-input--indeterminate) .v-icon {
-  color: $app-blue !important;
```
severinbeauvais, line 117: I remember an explicit UI requirement to make this app-blue, but the requirement has changed.

Actually, I think the new style looks better so not complaining:

image

src/components/Registration/BusinessAddresses.vue
```diff
@@ -4,25 +4,17 @@
     <template v-if="!isEditing">
       <v-row no-gutters id="address-summary">
         <v-col md="3" class="mr-n1"><label><strong>Business Addresses</strong></label></v-col>
-        <v-col md="4">
+        <v-col md="4" class="pr-4">
```
severinbeauvais, line 7: This is to have a 20px "gutter" between the mailing and delivery addresses.

I added this same style to the delivery address below so they are both the same width.

src/components/Registration/BusinessAddresses.vue
```diff
@@ -404,9 +396,7 @@ label {
 // }
 
 // italicize delivery instructions
-::v-deep .base-address {
-  .address-block__info-row:last-of-type {
-    font-style: italic;
```
severinbeauvais, line 409: Oops, this italicized the country if there was no delivery instructions. But only for registrations, which were not deployed yet 😁

src/components/common/OfficeAddresses.vue
```diff
@@ -4,51 +4,35 @@
...
-            :address="recDeliveryAddress"
-            :editing="false"
-          />
-          <div v-else-if="isEmptyAddress(recDeliveryAddress)">(Not entered)</div>
-          <div v-else-if="inheritRegisteredAddress">Same as Registered Office</div>
-          <div v-else>Same as Mailing Address</div>
+          <div v-if="isEmptyAddress(recDeliveryAddress)">(Not entered)</div>
+          <div v-else-if="isSame(deliveryAddress, recDeliveryAddress)">Same as Registered Office</div>
+          <div v-else-if="isSame(recMailingAddress, recDeliveryAddress)">Same as Mailing Address</div>
+          <DeliveryAddress v-else :address="recDeliveryAddress" :editing="false" />
```
severinbeauvais, line 35: ^^^ simplified logic!

I didn't do this frivolously... I was already in this area fixing a bug where the inherit flags were not being updated (since the component on this page is another instance and doesn't have the same class variables as the editable address component). So instead I check if the addresses are the same (which is also what happens if the user types in the same address multiple times then saves/reloads the draft: the checkbox is automatically checked).

src/services/legal.services.ts
```diff
@@ -1,35 +1,22 @@
...
-@Component({})
-export default class LegalApiMixin extends Vue {
-  @Getter getBusinessId!: string
-  @Getter getFilingId!: number
-  @Getter getTempId!: string
-
-  @Action setFilingId!: ActionBindingIF
-  @Action setFilingType!: ActionBindingIF
-
+export default class LegalServices {
```
severinbeauvais, line 10: I changed this from a mixin to a service class. Now it doesn't need store getters!

tests/unit/Actions.spec.ts
```diff
@@ -7,12 +7,16 @@ import sinon from 'sinon'
...
 import mockRouter from './MockRouter'
+import LegalServices from '@/services/legal.services'
 
 Vue.use(Vuetify)
 
 const vuetify = new Vuetify({})
 const store = getVuexStore()
 
+// mock services function
+const mockUpdateFiling = jest.spyOn((LegalServices as any), 'updateFiling').mockImplementation()
```
severinbeauvais, line 18: This is a mock for the `updateFiling` function in Legal Services class.

This also provides a default implementation so it actually "runs".

I just made this a global variable in this test file and "reset" it when I need.

tests/unit/Actions.spec.ts
```diff
@@ -538,7 +536,7 @@ describe('Actions component - Filing Functionality', () => {
 
   it('Emits "Go To Dashboard" event when onClickCancel is called', async () => {
     const mockBuildFiling = jest.spyOn(wrapper.vm, 'buildIncorporationFiling')
-    const mockUpdateFiling = jest.spyOn(wrapper.vm, 'updateFiling')
+    mockUpdateFiling.mockReset().mockImplementation()
```
severinbeauvais, line 539: I specific the default mock implementation again because some tests above set a specific implementation. (But doesn't really matter for the purposes of this test.)

src/views/Registration/RegistrationReviewConfirm.vue
```diff
@@ -36,14 +36,14 @@
...
-          Confirm the legal name of the person authorized to complete and submit this application.
+          Confirm the legal name of the person authorized to complete and submit this registration.
         </p>
       </header>
       <Certify
         :class="{ 'invalid-section': isCertifyInvalid }"
         :currentDate="getCurrentDate"
         :certifiedBy="getCertifyState.certifiedBy"
-        :entityDisplay="getCompletingPartyStatement.entityDisplay"
+        :entityDisplay="getCompletingPartyStatement.entityDisplay + ' applicant'"
```
severinbeauvais, line 46: I have not heard back whether the word "applicant" should follow entity type for all filings, so for now I did this.

Re: https://projects.invisionapp.com/share/3511UPXPMT2P#/screens/459583835/comments

cameron-freshworks (approved):
Looks good!
A lot of little things, deserves some thorough regression testing. 
Perhaps call out the scope of the refactors for Riyaz since the ticket title might not do the scope justice :)

vysakh-menon-aot (approved):

PR #345, opened 2022-03-02
9518 NR legal type refactor + misc cleanup

severinbeauvais:
*Issue #:* /bcgov/entity#9518

*Description of changes:*
- used nr.legalType instead of converting
- fixed whitespace in Entity Info component
- deleted obsolete Entity Filter mixin
- replaced entityFilter() with isXXX getters
- added entity subtype enum for SP
- removed unneeded nameRequestType
- removed "applicant" from certify stmt
- added parentheses around "Not entered"
- updated unit tests
- misc cleanup
- app version = 3.4.1


src/components/common/EntityInfo.vue
```diff
@@ -2,23 +2,25 @@
   <div id="entity-info">
     <v-container>
       <v-row no-gutters class="pt-3 pb-3">
-        <v-col cols="12" md="9">
+        <v-col cols="12" md="9" class="pr-5">
```
severinbeauvais, line 5: I move the gutter padding to the right of this column so that, if the columns wrap, the second column isn't offset on the left side.

src/mixins/name-request-mixin.ts
```diff
@@ -17,7 +17,7 @@ export default class NameRequestMixin extends Mixins(DateMixin) {
   generateNameRequestState (nr: any, filingId: number): NameRequestIF {
     return {
       nrNumber: nr.nrNum,
-      entityType: this.nrTypeToEntityType(nr),
+      entityType: nr.legalType,
```
severinbeauvais, line 20: Vysakh assures me that all NRs fetched via Legal API will have this value, and it will already have been correctly converted from NR type (eg, BC -> BEN).

src/resources/Incorporations/benefitCompany.ts
```diff
@@ -10,7 +10,6 @@ export const BenefitCompanyResource: IncorporationResourceIF = {
   description: 'This company is a benefit company and, as such, has purposes that include conducting its ' +
     'business in a responsible and sustainable manner and promoting one or more public benefits.',
   statement: null,
-  nameRequestType: NameRequestTypes.BC,
```
severinbeauvais, line 13: I'm not sure what the original plan for this field was, but I couldn't find anywhere it's used.

src/views/Registration/RegistrationReviewConfirm.vue
```diff
@@ -43,7 +43,7 @@
         :class="{ 'invalid-section': isCertifyInvalid }"
         :currentDate="getCurrentDate"
         :certifiedBy="getCertifyState.certifiedBy"
-        :entityDisplay="getCompletingPartyStatement.entityDisplay + ' applicant'"
+        :entityDisplay="getCompletingPartyStatement.entityDisplay"
```
severinbeauvais, line 46: UI/UX team changed their mind on this. Easy change 😛

vysakh-menon-aot (approved):

severinbeauvais:
Local test results (to see entity info header and NR data):

image

image

src/App.vue
```diff
@@ -743,8 +743,8 @@ export default class App extends Mixins(
       }
 
       // ensure types match
-      if (this.nrTypeToEntityType(nrResponse) !== this.getEntityType) {
-        console.log('NR type doesn\'t match entity type') // eslint-disable-line no-console
+      if (nrResponse.legalType !== this.getEntityType) {
+        console.log('NR legal type doesn\'t match entity type') // eslint-disable-line no-console
```
cameron-freshworks, line 747: nice! 
so essentially no more local mapping required?

src/enums/entitySubtypes.ts
```diff
@@ -0,0 +1,12 @@
...
+export enum CoopType {
+  COMMUNITY_SERVICE_COOPERATIVE = 'CSC',
+  HOUSING_COOPERATIVE = 'HC',
+  ORDINARY_COOPERATIVE = 'CP',
+}
+
+/** Entity subtype for Sole Prop registration. */
+export enum BusinessType {
+  DBA = 'DBA',
+  SP = 'SP',
```
cameron-freshworks, line 11: Interesting! 
I'm surprised we didn't have this before, or did we in some other form?

src/mixins/name-request-mixin.ts
cameron-freshworks: This is a nice change! 

I'm not aware of the full scope of the api change but will this change require some attention in Auth as well? 
I know we've had to do similar conversions with the entity table and creating draft filings etc, i think*

cameron-freshworks (approved):
Some great clean up!

src/App.vue
severinbeauvais: Correct. It's now done in the API before the UI gets the data. (It's a new field in the NR JSON.)

cameron-freshworks: Very nice!

src/enums/entitySubtypes.ts
severinbeauvais: This is a new subtype field in the filing. As I understand it, it's only valid for Sole Prop filings. (Similar to Association Type that is used as a subtype for Coop IAs.)

I added this in after chatting with Vysakh. We will use this soon I think.

severinbeauvais: Here's the new Business Type field in the registration schema:
https://github.com/bcgov/business-schemas/blob/6eee00ec5035b90ab5fb7b74dc5697ec172e721f/src/registry_schemas/schemas/registration.json#L27

PR #346, opened 2022-03-03
9518 cleanup and attempt to fix timing issues

severinbeauvais:
*Issue #:* /bcgov/entity#9518

*Description of changes:*
- now set haveData in try{} instead of finally{}
- added some logging (temporary)
- replaced Vue.nextTick() with this.$nextTick()
- removed some unneeded "ignore changes"
- moved some "step validation" from created() to mounted()
- app version = 3.4.2

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -142,7 +142,7 @@
 
 <script lang="ts">
 // Libraries
-import { Component, Mixins, Vue, Watch } from 'vue-property-decorator'
+import { Component, Mixins, Watch } from 'vue-property-decorator'
 import { Action, Getter } from 'vuex-class'
 import { PAYMENT_REQUIRED } from 'http-status-codes'
 import { cloneDeep } from 'lodash'
@@ -373,7 +373,7 @@ export default class App extends Mixins(
   private get saveErrorDialogName (): string {
     switch (this.getFilingType) {
       case FilingTypes.INCORPORATION_APPLICATION: return 'Application'
-      case FilingTypes.REGISTRATION: return 'Application'
+      case FilingTypes.REGISTRATION: return 'Registration'
       case FilingTypes.VOLUNTARY_DISSOLUTION: return 'Filing'
     }
   }
@@ -614,14 +614,16 @@ export default class App extends Mixins(
           }
         )
       }
+
+      console.log('*** got data')
+      this.haveData = true
     } catch (error) {
       // errors should be handled above
       console.error('Unhandled error in fetchData() =', error) // eslint-disable-line no-console
       // just fall through to finally()
     } finally {
-      this.haveData = true
       // wait for things to stabilize, then reset flag
-      Vue.nextTick(() => this.setHaveChanges(false))
+      this.$nextTick(() => this.setHaveChanges(false))
     }
   }
 
--- a/src/components/Dissolution/CompleteAffidavit.vue
+++ b/src/components/Dissolution/CompleteAffidavit.vue
@@ -1,5 +1,5 @@
 <template>
-  <div id="dissolution-affidavit">
+  <div id="complete-affidavit">
     <section class="mt-10">
       <header>
         <h2>1. Affidavit</h2>
@@ -190,7 +190,7 @@ import FileUploadPreview from '@/components/common/FileUploadPreview.vue'
     FileUploadPreview
   }
 })
-export default class Affidavit extends Mixins(CommonMixin, DocumentMixin, EnumMixin) {
+export default class CompleteAffidavit extends Mixins(CommonMixin, DocumentMixin, EnumMixin) {
   // Refs
   $refs!: {
     confirmAffidavitChk: FormIF
@@ -301,7 +301,8 @@ export default class Affidavit extends Mixins(CommonMixin, DocumentMixin, EnumMi
   }
 
   private updateAffidavitStepValidity () {
-    const validationDetail:ValidationDetailIF =
+    console.log('*** setting affidavit step validity')
+    const validationDetail: ValidationDetailIF =
       {
         valid: this.hasAffidavitConfirmed && this.hasValidUploadFile,
         validationItemDetails: [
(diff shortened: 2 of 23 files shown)


src/App.vue
```diff
@@ -614,14 +614,16 @@ export default class App extends Mixins(
           }
         )
       }
+
+      console.log('*** got data')
```
severinbeauvais, line 618: The issue that's bugging me atm is that the dissolution Complete Resolution and Complete Affidavit steps don't show a valid checkmark when you resume a draft, even though those pages are valid. Example:

https://business-create-dev.apps.silver.devops.gov.bc.ca/businesses/create/dissolution-define-dissolution?id=BC0870918&accountid=2288

Unfortunately this doesn't happen if you reload the page, or locally. I believe there's a timing issuel possibly the pages render (and decide what's valid) before the data is loaded.

So this is an attempt to verify that we get the data before the pages render. This is temporary code that I remove later.

src/App.vue
```diff
@@ -614,14 +614,16 @@ export default class App extends Mixins(
...
+      this.haveData = true
     } catch (error) {
       // errors should be handled above
       console.error('Unhandled error in fetchData() =', error) // eslint-disable-line no-console
       // just fall through to finally()
     } finally {
-      this.haveData = true
```
severinbeauvais, line 622: One theory I have is that there was an exception above and this finally block sets this flag, which is the gate to let the pages render. By moving this code to the bottom of the try block, it will only be run if everything in the try block succeeds (and will not be run if there was an exception).

NB: "finally" block runs after both try and catch blocks

src/components/Dissolution/CompleteAffidavit.vue
```diff
@@ -336,6 +337,11 @@ export default class Affidavit extends Mixins(CommonMixin, DocumentMixin, EnumMi
     this.affidavitConfirmed = this.getAffidavitStep.affidavitConfirmed
     this.hasValidUploadFile = !!this.uploadAffidavitDocKey
     this.hasAffidavitConfirmed = this.affidavitConfirmed
+  }
+
+  async mounted (): Promise<void> {
+    // wait for components to load/stabilize then update validation state in store
+    await this.$nextTick()
     this.updateAffidavitStepValidity()
```
severinbeauvais, line 345: This is testing another theory. I might revert this later, dunno, shouldn't hurt like this in the long term if I don't.

src/components/Dissolution/CompleteResolution.vue
```diff
@@ -599,16 +600,16 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin) {
     const foundingDate = new Date(this.getBusinessFoundingDate)
     foundingDate.setHours(0, 0, 0, 0)
     this.foundingDate = foundingDate
-    this.resolutionConfirmed = !!this.getCreateResolutionStep.resolutionConfirmed
+    this.resolutionConfirmed = this.getCreateResolutionStep.resolutionConfirmed
     this.resolutionDateText = this.getCreateResolutionStep.resolutionDate
     this.resolutionText = this.getCreateResolutionStep.resolutionText
-    this.signingPerson = this.getCreateResolutionStep.signingPerson || EmptyPerson
+    this.signingPerson = this.getCreateResolutionStep.signingPerson || { ...EmptyPerson }
```
severinbeauvais, line 606: I haven't seen a bug for this but it was incorrect.

src/components/common/AgreementType.vue
```diff
@@ -102,15 +101,9 @@ export default class AgreementType extends Mixins(EnumMixin) {
...
-  // Lifecycle methods
+  /** Called when component is created. */
   created (): void {
-    // temporarily ignore data changes
-    this.setIgnoreChanges(true)
     this.agreementType = this.getIncorporationAgreementStep.agreementType
-    // watch data changes once page has loaded (in next tick)
-    Vue.nextTick(() => {
-      this.setIgnoreChanges(false)
-    })
```
severinbeauvais, line 113: The original reason for this code was to ignore changes to the store when this component started up (so that we would only tell the user they have unsaved changes if they made changes -- not changes due to component init).

However, there are no changes to the store in created() here, so that code is not needed.

src/components/common/Stepper.vue
```diff
@@ -22,7 +22,7 @@
           >
             <v-icon class="step__icon" :class="{ 'selected-icon': isCurrentStep(step) }">{{ step.icon }}</v-icon>
           </v-btn>
-          <v-icon class="step__btn2" size="30" color="success darken-1" v-show=isValid(step.to)>
+          <v-icon class="step__btn2" size="30" color="success darken-1" v-show="isValid(step.to)">
```
severinbeauvais, line 25: The old code worked but looked funny.

src/App.vue
```diff
@@ -614,14 +614,16 @@ export default class App extends Mixins(
...
       // wait for things to stabilize, then reset flag
-      Vue.nextTick(() => this.setHaveChanges(false))
+      this.$nextTick(() => this.setHaveChanges(false))
```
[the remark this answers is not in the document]
cameron-freshworks: Cool

cameron-freshworks (approved):
Code looks good, hopefully this resolves the issue!

src/App.vue
severinbeauvais: I can no longer reproduce this issue in Dev. There are several possibilities:

1. setting haveData in the try block instead of the finally block solved it
2. calling the step validity method from mounted instead of created, and after letting things settle a bit, solved it
3. the timing issue still exists but isn't see atm due to code changes
4. the stars aligned differently right now and the issue still exists but is sporadic.

🤷‍♂️

severinbeauvais: Ah nuts, I just reproduced it in Dev, so it's possibility 3 or 4.

HOWEVER, my logging shows this...

image

... and I know that parseDissolutionDraft() clears (initializes) validationItemDetails[], so that's why you can see the step checkmark for a split second and then disappears (ie, it thinks the page is not valid).

Since fetchData() (which displays "got data") is not being called from an error dialog, it means the route has changed twice. Hmm... 🤔

PR #348, opened 2022-03-04
9518 Removed debug code + fixed App error handling + prep for Sbc Staff registrations

severinbeauvais:
*Issue #:* /bcgov/entity#9518 and /bcgov/entity#11382

*Description of changes:*
- removed previous debugging statements
- enabled validation styles on registration document delivery component
- allowed SbcStaff to edit registration document delivery completing party
- allowed SbcStaff to edit registration staff certified by
- allowed SbcStaff to see registration staff certify statement
- allowed SbcStaff to bypass email and phone  checks on app init
- fixed error handling in App
- app version = 3.4.4

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -74,7 +74,7 @@
 
     <!-- Initial Page Load Transition -->
     <transition name="fade">
-      <div class="loading-container" v-show="!haveData">
+      <div class="loading-container" v-show="!haveData && !isErrorDialog">
         <div class="loading__content">
           <v-progress-circular color="primary" size="50" indeterminate />
           <div class="loading-msg">Loading</div>
@@ -87,6 +87,7 @@
     <PaySystemAlert />
 
     <div class="app-body">
+      <!-- Don't show page if an error dialog is displayed. -->
       <main v-if="!isErrorDialog">
         <BreadCrumb :breadcrumbs="breadcrumbs" />
         <EntityInfo />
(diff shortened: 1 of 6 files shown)


src/App.vue
```diff
@@ -74,7 +74,7 @@
 
     <!-- Initial Page Load Transition -->
     <transition name="fade">
-      <div class="loading-container" v-show="!haveData">
+      <div class="loading-container" v-show="!haveData && !isErrorDialog">
```
severinbeauvais, line 77: If there is a data fetch error then we don't set haveData (so that the pages aren't rendered). But we need to hide the spinner to see any error dialogs.

src/App.vue
```diff
@@ -633,7 +635,7 @@ export default class App extends Mixins(
     await this.checkAuth(this.getBusinessId).catch(error => {
       console.log('Auth error =', error) // eslint-disable-line no-console
       this.accountAuthorizationDialog = true
-      throw error // go to catch()
```
severinbeauvais, line 636: This method doesn't know anything about a catch. (That's in the parent component.)

src/App.vue
```diff
@@ -689,6 +691,8 @@ export default class App extends Mixins(
         resources = RegistrationResources
         parseFiling = this.parseRegistrationDraft
         break
+      default:
+        throw new Error(`Invalid filing type = ${this.getFilingType}`)
```
severinbeauvais, line 695: Sometimes, during development, an update event occurred but filing type was cleared. Without this check, the code would try to parse "nothing" and display an error. This catches the error as early as possible.

cameron-freshworks (approved):
Looks good 👍

PR #357, opened 2022-03-18
Document Delivery: Shared component#11408

shabeeb-aot:
*Issue #:* /bcgov/entity#11408

*Description of changes:*
* Added additional label and additional email field 

### DIFF
--- a/src/components/common/DocumentDelivery.vue
+++ b/src/components/common/DocumentDelivery.vue
@@ -42,7 +42,7 @@
       </v-col>
     </v-row>
 
-     <v-row no-gutters v-if="additionalLabel && additionalEmail" class="pt-3">
+     <v-row no-gutters v-if="additionalLabel" class="pt-3">
       <v-col cols="12" sm="3" class="pr-4">
         <label class="title-label">{{additionalLabel}}</label>
       </v-col>
--- a/src/components/common/DocumentDelivery.vue
+++ b/src/components/common/DocumentDelivery.vue
@@ -3,10 +3,10 @@
     <!-- Registered Office -->
     <v-row no-gutters>
       <v-col cols="12" sm="3" class="pr-4">
-        <label class="title-label">Registered Office</label>
+        <label class="title-label">{{contactLabel}}</label>
       </v-col>
       <v-col cols="12" sm="9">
-        <span id="office-email">{{registeredOfficeEmail || '(Not entered)'}}</span>
+        <span id="office-email">{{contactValue || '(Not entered)'}}</span>
       </v-col>
     </v-row>
 
@@ -47,7 +47,7 @@
         <label class="title-label">{{additionalLabel}}</label>
       </v-col>
       <v-col cols="12" sm="9">
-        <span id="additonal-email">{{additionalEmail || '(Not entered)'}}</span>
+        <span id="additonal-email">{{additionalValue || '(Not entered)'}}</span>
       </v-col>
     </v-row>
   </div>
@@ -59,7 +59,8 @@ import { Component, Emit, Prop, Vue, Watch } from 'vue-property-decorator'
 
 @Component({})
 export default class DocumentDelivery extends Vue {
-  @Prop({ default: null }) readonly registeredOfficeEmail: string
+  @Prop({ required: true }) readonly contactLabel: string
+  @Prop({ default: null }) readonly contactValue: string
   @Prop({ default: null }) readonly custodianEmail: string
   @Prop({ default: null }) readonly completingPartyEmail: string
   @Prop({ default: null }) readonly documentOptionalEmail: string
@@ -67,13 +68,9 @@ export default class DocumentDelivery extends Vue {
   @Prop({ default: false }) readonly editableCompletingParty: boolean
   @Prop({ default: false }) readonly showCustodianEmail: boolean
   @Prop({ default: false }) readonly invalidSection: boolean
-  // additionaLabel can be used for additon fields.
-  //  For SP can be used for proprietor
-  // FOr GP label can be Partners
-  @Prop({ default: null }) readonly additionalLabel: string
-  // additionalEmail  For SP can be used for single email
-  // For GP additionalEmail can be multipel with comma separated
-  @Prop({ default: null }) readonly additionalEmail: string
+
+  @Prop({ default: null }) readonly additionalLabel: string // for additon fields.
+  @Prop({ default: null }) readonly additionalValue: string
 
   // Local properties
   private optionalEmail: string = ''
--- a/src/components/common/DocumentDelivery.vue
+++ b/src/components/common/DocumentDelivery.vue
@@ -41,6 +41,15 @@
         <span id="custodian-email">{{custodianEmail || '(Not entered)'}}</span>
       </v-col>
     </v-row>
+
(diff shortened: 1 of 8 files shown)


src/components/common/DocumentDelivery.vue
```diff
@@ -58,6 +67,13 @@ export default class DocumentDelivery extends Vue {
   @Prop({ default: false }) readonly editableCompletingParty: boolean
   @Prop({ default: false }) readonly showCustodianEmail: boolean
   @Prop({ default: false }) readonly invalidSection: boolean
+  // additionaLabel can be used for additon fields.
+  //  For SP can be used for proprietor
+  // FOr GP label can be Partners
+  @Prop({ default: null }) readonly additionalLabel: string
+  // additionalEmail  For SP can be used for single email
+  // For GP additionalEmail can be multipel with comma separated
+  @Prop({ default: null }) readonly additionalEmail: string
```
shabeeb-aot, line 76: do we need to update ```additionalValue``` thoughts?

src/components/common/DocumentDelivery.vue
```diff
@@ -41,6 +41,15 @@
         <span id="custodian-email">{{custodianEmail || '(Not entered)'}}</span>
       </v-col>
     </v-row>
+
```
severinbeauvais, line 44: Do you need a block for Business Contact also?

Screen Shot 2022-03-18 at 9 51 14 AM

src/components/common/DocumentDelivery.vue
severinbeauvais: What is `additionalValue`? Not used?

All existing rows have a label and value. This is similar so I'm OK with it.

src/components/common/DocumentDelivery.vue
```diff
@@ -58,6 +67,13 @@ export default class DocumentDelivery extends Vue {
   @Prop({ default: false }) readonly editableCompletingParty: boolean
   @Prop({ default: false }) readonly showCustodianEmail: boolean
   @Prop({ default: false }) readonly invalidSection: boolean
+  // additionaLabel can be used for additon fields.
+  //  For SP can be used for proprietor
+  // FOr GP label can be Partners
+  @Prop({ default: null }) readonly additionalLabel: string
```
severinbeauvais, line 73: 1. I think you don't need these comments: Developers will figure out what this can be used for, and now you've made this component filing-specific. Remove them :)

2. If you do need to add comments, use JDoc style so that developers can hover over the method/prop and see the comment as contextual help.

severinbeauvais (commented):
I'm OK with this approach.

Needs a bit of cleanup per my comments.

You should update the app version.

You should add a test case or 2.

src/components/common/DocumentDelivery.vue
shabeeb-aot: You are right , we need it, I was thinking can I use the same registeredOfficeEmail field for value and label convert to prop?
Like ```@Prop({ default: 'Registered Office' }) readonly registeredOfficeLabel: string```
Or do you prefer to add it as an additional prop (may be businessContactEmail) and put v-if ?

src/components/common/DocumentDelivery.vue
shabeeb-aot: So i will leave as it is now :)

src/components/common/DocumentDelivery.vue
cameron-freshworks: Thats a good idea i think!
If you are going to change the label to allow variations, a more generic prop name might be better, even though you'd default to `Registered Office`. `OfficeLabel` or something maybe? We are moving towards variations of offices, ie registeredOffice, recordsOffice, businessOffice etc

src/views/DissolutionFirm/DissolutionFirm.vue
```diff
@@ -38,7 +38,10 @@
         :documentOptionalEmail="getDocumentDelivery.documentOptionalEmail"
         @update:optionalEmail="setDocumentOptionalEmail($event)"
         @valid="setDocumentOptionalEmailValidity($event)"
+        additionalEmail="test@test2.com"
```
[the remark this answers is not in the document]
cameron-freshworks: Sounds good. 

Not required but always welcome to create an additional Story, for this additional Prop/label.
Since this prop is not required, you don't have update the existing stories, just create a new one with the added prop/value. Super ez :)

cameron-freshworks (approved):
Not required but always welcome to create an additional Story, for this additional Prop/label.
Since this prop is not required, you don't have update the existing stories, just create a new one with the added prop/value. Super ez :)

Looks good though, just some minor comments and Severins to address 👍

src/components/common/DocumentDelivery.vue
severinbeauvais: 1. Yes, let's do something generic for the first address. I just noticed that in my feature, I also need to change the first label! Can we call it `contactLabel` and `contactValue` (or contactEmail) ?

1, Don't use a default value as that can lead to cascading properties, which will be hard to figure out in the future. Instead, update everywhere this is used to pass in the desired label.

cameron-freshworks:
> Not required but always welcome to create an additional Story, for this additional Prop/label. Since this prop is not required, you don't have update the existing stories, just create a new one with the added prop/value. Super ez :)
> 
> Looks good though, just some minor comments and Severins to address 👍

*Also after updating the stories, storybook:build before pushing up the last commit!

src/components/common/DocumentDelivery.vue
severinbeauvais: I changed my mind! I think I prefer `additionalValue` (to be consistent with contactLabel and contactValue).

src/components/common/DocumentDelivery.vue
```diff
@@ -41,6 +41,15 @@
...
     </v-row>
+
+     <v-row no-gutters v-if="additionalLabel" class="pt-3">
+      <v-col cols="12" sm="3" class="pr-4">
+        <label class="title-label">{{additionalLabel}}</label>
+      </v-col>
+      <v-col cols="12" sm="9">
+        <span id="additonal-email">{{additionalEmail || '(Not entered)'}}</span>
+      </v-col>
+    </v-row>
```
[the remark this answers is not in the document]
severinbeauvais: FYI, default for `additionalLabel` is null (falsey), but there is no default for contactLabel as it's not conditional -- instead it is "required".

src/views/DissolutionFirm/DissolutionFirm.vue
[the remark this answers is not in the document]
severinbeauvais: This is now deleted (null), so the component will show "not entered".
Closing.

severinbeauvais (approved):

severinbeauvais:
> Not required but always welcome to create an additional Story, for this additional Prop/label. Since this prop is not required, you don't have update the existing stories, just create a new one with the added prop/value. Super ez :)

I agree with Cameron: please create additional unit tests for these changes.

I've merged this now because I know you won't get around to unit tests until the next day and I need to rebase and pick up these changes ASAP.

severinbeauvais:
> *Also after updating the stories, storybook:build before pushing up the last commit!

Ah. To be clear, this PR is for a component shared by different pages in Create UI. It is not a bcrs-shared-component :)

PR #358, opened 2022-03-21
9246 Incremental work on P&R + more fixes and cleanup

severinbeauvais:
*Issue #:* /bcgov/entity#9246

*Description of changes:*

- fixed entity info and footer horiz whitespace
- fixed registration review page validation
- updated fetchDraftApplication to support a list response (_should I leave this in or remove the decoy pattern code?_)
- replaced local Confirm Dialog with shared one
- replaced local BusinessContactIF with shared ContactPointIF (#11519)
- displayed phone extension in Entity Info (#11519)
- fixed "have changes" bug when changing dissolution contact info (#11519)
- mapped extension <-> phoneExtension in Auth Services (#11519)
- replaced ConfirmRemoveDialog with ConfirmDialog
- moved removal confirmation out of ListPeopleRoles
- hid add buttons when rule is satisfied
- added confirmation step to add method
- added confirmation step to remove method
- added some mixin methods
- added unit tests
- misc cleanup
- app version = 3.5.1


src/App.vue
```diff
@@ -90,7 +90,12 @@
       <!-- Don't show page if an error dialog is displayed. -->
       <main v-if="!isErrorDialog">
         <BreadCrumb :breadcrumbs="breadcrumbs" />
-        <EntityInfo class="py-5 px-3" />
+
+        <div id="entity-info-wrapper">
+          <v-container class="py-5">
+            <EntityInfo />
+          </v-container>
+        </div>
```
severinbeauvais, line 98: Container needed to be wrapped in a div so that div can have white background while container auto-sizes according to screen width.

src/components/Dissolution/AssociationDetails.vue
```diff
@@ -140,12 +141,18 @@ export default class AssociationDetails extends Mixins(CommonMixin, EnumMixin) {
   }
 
   /** Event handler for contact information changes. */
-  private async onContactInfoChange (event: BusinessContactIF): Promise<void> {
+  private async onContactInfoChange (event: ContactPointIF): Promise<void> {
+    // temporarily ignore data changes
+    this.setIgnoreChanges(true)
+
```
severinbeauvais, line 147: Don't set the Have Changes flag here (whether user clicked OK or Cancel) because the filing data has not changed -- the contact info change was done in real time in Auth db.

src/components/common/Actions.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <v-container id="action-buttons-container" class="list-item">
+  <div id="action-buttons-container" class="list-item pa-0">
```
severinbeauvais, line 2: Component should be smallest container possible -- parent can set whitespace according to its needs.

severinbeauvais: PS "v-container" is the block in the middle of our apps that doesn't get wider after a certain max screen width.

Here, v-container was moved to the parent.

src/components/common/Stepper.vue
```diff
@@ -79,18 +78,18 @@ export default class Stepper extends Vue {
...
       case RouteNames.DISSOLUTION_AFFIDAVIT: return this.isAffidavitValid
       case RouteNames.DISSOLUTION_DEFINE_DISSOLUTION: return this.isDissolutionDefineDissolutionValid
       case RouteNames.DISSOLUTION_RESOLUTION: return this.isResolutionValid
-      case RouteNames.DISSOLUTION_REVIEW_CONFIRM: return this.isApplicationValid
+      case RouteNames.DISSOLUTION_REVIEW_CONFIRM: return this.isDissolutionValid
 
       case RouteNames.REGISTRATION_DEFINE_BUSINESS: return this.getRegistration.defineBusinessValid
       case RouteNames.REGISTRATION_PEOPLE_ROLES: return this.isAddPeopleAndRolesValid
-      case RouteNames.REGISTRATION_REVIEW_CONFIRM: return this.isApplicationValid
+      case RouteNames.REGISTRATION_REVIEW_CONFIRM: return this.isRegistrationValid
```
severinbeauvais, line 92: This worked before because `isRegistrationValid` would return the correct value for the current filing type, but that's a bit obscure. This is more understandable.

src/services/auth.services.ts
```diff
@@ -38,7 +38,13 @@ export default class AuthServices {
...
       if (response?.data) {
         return {
-          contacts: response.data.contacts,
+          contacts: response.data.contacts
+            // map phoneExtension -> extension in every element
+            .map(contact => ({
+              email: contact.email,
+              phone: contact.phone,
+              extension: contact.phoneExtension
+            })),
```
severinbeauvais, line 47: So this service does the translation between the app's schema and the API's schema.

src/services/legal.services.ts
```diff
@@ -9,18 +9,19 @@ import { FilingTypes } from '@/enums'
...
     const url = `businesses/${id}/filings`
     return axios.get(url)
       .then(response => {
-        // look at only the first filing
-        const filing = response?.data?.filing
+        let filing
+        // if response is a list then look at only the first filing
+        if (Array.isArray(response?.data?.filings)) filing = response.data.filings[0].filing
+        // otherwise expect response to be a single filing
+        if (response?.data?.filing) filing = response.data.filing
```
severinbeauvais, line 24: This is in preparation for some API cleanup that Vysakh is working on.

Actually, this will probably need to call the tasks endpoint, but this can be refactored when needed.

src/store/getters/state-getters.ts
```diff
@@ -513,7 +513,8 @@ export const isIncorporationApplicationValid = (state: StateIF): boolean => {
 /** Whether all the registration steps are valid. */
 export const isRegistrationValid = (state: StateIF): boolean => {
   const isCertifyValid = getCertifyState(state).valid && !!getCertifyState(state).certifiedBy
-  const isFeeAcknowledgementValid = getRegistration(state).feeAcknowledgement
+  // const isFeeAcknowledgementValid = getRegistration(state).feeAcknowledgement
+  const isFeeAcknowledgementValid = true // FUTURE: use line above instead
```
severinbeauvais, line 517: This caused the review page to be invalid, because the component and data are coded but not displayed (so the user can never check the checkbox).

tests/unit/AddEditOrgPerson.spec.ts
```diff
@@ -7,6 +7,9 @@ import { getVuexStore } from '@/store'
 import AddEditOrgPerson from '@/components/common/AddEditOrgPerson.vue'
 import { EmptyOrgPerson } from '@/interfaces'
 
+// mock the console.warn function to hide "[Vuetify] Unable to locate target XXX"
+console.warn = jest.fn()
```
severinbeauvais, line 11: The warning is because the shared ConfirmDialog expected to attach itself to its parent component, but for some reason Jest isn't setting the parent's ID, so the dialog can't find its attachment.

The "attach" prop was no present in the previous, local version of ConfirmDialog, which avoided this issue but unfortunately did not attach the dialog to the desired parent. (Didn't seem to matter in practise but still...)

tests/unit/BusinessContactInfo.spec.ts
```diff
@@ -54,19 +53,17 @@ function getLastEvent (wrapper: Wrapper<BusinessContactInfo>, name: string): any
...
 function createComponent (
-  email: string = '',
-  confirmEmail: string = '',
-  phone: string = '',
-  extension?: number,
-  isEditing: boolean = true
+  email = '',
+  confirmEmail = '',
+  phone = '',
+  extension: number = null,
```
severinbeauvais, line 59: Small semantic fix. Now extension is null (instead of undefined) if omitted as a formal parameter.

tests/unit/legal.services.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+describe('Legal Services', () => {
+  it('fetches a draft application as a single item', async () => {
+    // mock single item response
+    sinon.stub(axios, 'get').withArgs('businesses/123/filings')
+      .returns(new Promise(resolve => resolve({
+        data: { filing: draftRegistration }
+      })))
+
+    // fetch draft and check it
+    const draft: any = await LegalServices.fetchDraftApplication('123')
```
severinbeauvais, line 27: Don't even need to mount a component here -- just call the service directly.

src/components/Registration/RegPeopleAndRoles.vue
```diff
@@ -76,6 +76,7 @@
       </v-btn>
 
       <v-btn
+        v-if="!validNumProprietors"
```
severinbeauvais, line 79: Only show button when rule is not satisfied.

src/components/Registration/RegPeopleAndRoles.vue
```diff
@@ -158,7 +160,19 @@ export default class RegPeopleAndRoles extends Mixins(PeopleRolesMixin) {
   // NB: see mixin for common methods, getters, etc.
   //
 
-  private addOrgPerson (roleType: RoleTypes, partyType: PartyTypes): void {
+  protected async addOrgPerson (roleType: RoleTypes, partyType: PartyTypes): Promise<void> {
```
severinbeauvais, line 163: `protected` means the template (and any object that inherits this class) can see this method, but it can't be seen externally.

Ref: https://www.typescriptlang.org/docs/handbook/2/classes.html#member-visibility

src/mixins/people-roles-mixin.ts
```diff
@@ -197,7 +205,18 @@ export default class PeopleRolesMixin extends Vue {
...
+    const orgPerson = this.orgPersonList[index]
+    const isProprietor = this.isProprietor(orgPerson)
+
+    if (isProprietor && this.isPerson(orgPerson)) {
+      if (!await this.confirmRemoveProprietorPerson()) return
+    } else if (isProprietor && this.isOrganization(orgPerson)) {
+      if (!await this.confirmRemoveProprietorOrganization()) return
+    } else {
+      if (!await this.confirmRemove()) return
+    }
```
severinbeauvais, line 218: In addition to consistency, an advantage of moving the confirmation out of ListPeopleAndRoles and into here is that the default confirmation is prompted if you click the Delete button while editing a person. (Previously there was no confirmation in this case.)

src/components/Registration/RegPeopleAndRoles.vue
cameron-freshworks: Very cool
For my own knowledge what is the purpose?
ie why Protected vs Private or Public?

severinbeauvais: This is a class, i.e., an object-oriented structure. It can contain some things that external code should not be able to see or change.

In this case, I don't want `addOrgPerson` to be publicly accessible, but if I make it private then my IDE's plugins tell me it's not used. It's actually used in the template only. Protected meets both these conditions.

vysakh-menon-aot (approved):

PR #364, opened 2022-03-29
Feature/11413

shabeeb-aot:
*Issue #:* /bcgov/entity#11413

*Description of changes:*
* Added Dissolution Date in DissolutionFirm
* Added unit test case for DissolutionFirm

### DIFF
--- a/src/views/DissolutionFirm/DissolutionFirm.vue
+++ b/src/views/DissolutionFirm/DissolutionFirm.vue
@@ -9,7 +9,6 @@
         </header>
 
         <div class="pb-8" >
-
           <!-- Association Details -->
           <section>
             <AssociationDetails
@@ -21,6 +20,38 @@
             />
           </section>
         </div>
+      </v-card>
+    </section>
+
+    <!-- Dissolution Dissolution Date -->
+    <section id="document-delivery-section" class="mt-10">
+      <header>
+        <h2>Business Dissolution Date</h2>
+        <p class="mt-4 mb-6">
+          Enter the dissolution date of the business. The dissolution date must be after the business start date and registration date. The dissolution date cannot be in the future.
+        </p>
+      </header>
+      <v-card flat class="mt-6">
+
+       <!-- EDIT SECTION -->
+        <v-row no-gutters>
+          <v-col cols="12" sm="3" class="pr-4">
+            <label class="start-date-title title-label">Dissolution Date</label>
+          </v-col>
+          <v-col cols="12" sm="9" class="pt-4 pt-sm-0" id="start-date-selector">
+            <date-picker
+              id="date-picker"
+              ref="startDateRef"
+              title="Dissolution Date"
+              :nudgeRight="40"
+              :nudgeTop="85"
+              :minDate="startDateMinStr"
+              :maxDate="startDateMaxStr"
+              :inputRules="getShowErrors ? startDateRules : []"
+              @emitDateSync="startDateHandler($event)"
+            />
+          </v-col>
+        </v-row>
       </v-card>
     </section>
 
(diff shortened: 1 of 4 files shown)


tests/unit/DissolutionFirm.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+// Test Case Data
+// GP will come soon
+const dissolutionFirmTestCases = [
+  {
+    entityType: 'SP',
+    isPremium: false,
+    isStaff: false
+  }
+  // {
```
shabeeb-aot, line 17: GP will be coming soon. SO made it as an array.

tests/unit/DissolutionFirm.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+          entityType: test.entityType,
+          accountInformation: { accountType: test.isPremium ? 'PREMIUM' : 'BASIC' }
+        },
+        null,
+        DissolutionResources
+      )
+
+      expect(wrapper.find('#folio-number-section').exists()).toBe(test.isPremium)
+    })
+
```
shabeeb-aot, line 85: There may be staff component , will update when we are doing staff components

cameron-freshworks:
Just a heads up, might help to add some PR with some content details. 

Obviously the linked ticket itself should* cover all the task details but more often than not there is additional updates included. Always good to call it out :)
Doesn't have to be too detailed!

src/views/DissolutionFirm/DissolutionFirm.vue
```diff
@@ -24,6 +23,38 @@
       </v-card>
     </section>
 
+    <!-- Dissolution Dissolution Date -->
+    <section id="document-delivery-section" class="mt-10">
+      <header>
+        <h2>Business Dissolution Date</h2>
+        <p class="mt-4 mb-6">
```
severinbeauvais, line 30: I discourage the use of bottom padding or margins for sequential components like this. (There are special cases for panels/cards/misc.) This bottom margin does nothing anyway because the next component specifies an equivalent top margin. This is my preference after much work cleaning up layouts :)  Please update (now or later is ok).

src/views/DissolutionFirm/DissolutionFirm.vue
```diff
@@ -24,6 +23,38 @@
...
+
+       <!-- EDIT SECTION -->
+        <v-row no-gutters>
+          <v-col cols="12" sm="3" class="pr-4">
+            <label class="start-date-title title-label">Dissolution Date</label>
+          </v-col>
+          <v-col cols="12" sm="9" class="pt-4 pt-sm-0" id="start-date-selector">
+            <date-picker
+              id="date-picker"
+              ref="startDateRef"
```
severinbeauvais, line 44: Leave the ref out since you're not using it. (It's typically used to call methods in this component, such as validate.)

severinbeauvais:
> Just a heads up, might help to add some details to the PR.

I agree. It's very helpful to the reviewers and also later when preparing a release or when investigating a bug :)

I usually document in my commit messages what I did, and then I paste those messages in the PR description. It's pretty easy; I recomment it.

cameron-freshworks (approved):
Just some minor stuff that Sev pointed out but looks good! 
👍

shabeeb-aot:
> > Just a heads up, might help to add some details to the PR.
> 
> I agree. It's very helpful to the reviewers and also later when preparing a release or when investigating a bug :)
> 
> I usually document in my commit messages what I did, and then I paste those messages in the PR description. It's pretty easy; I recommend it.

I usually add some comments in Description of changes, 
I missed it , thanks for reminding

severinbeauvais (approved):
Good looking code!

PR #371, opened 2022-04-08
10891 Implemented P&R for General Partnerships

severinbeauvais:
*Issue #:* /bcgov/entity#10891

~This is ready to go except for some UI confirmations (or changes) from Yui.~ No changed needed atm.

*Description of changes:*
- display delivery address for partners in list
- updated org label, blurb and confirm text
- business number is only for proprietor
- added roles checkbox for partner
- prompt delivery address also for partners
- added rule for num partners
- added buttons for adding partners
- added role type and rule for partner
- updated misc logic, getters, etc
- added some classnames for testing
- updated unit tests
- app version = 3.5.13

### DIFF
--- a/src/components/Registration/RegAddEditOrgPerson.vue
+++ b/src/components/Registration/RegAddEditOrgPerson.vue
@@ -89,6 +89,7 @@
             <article v-if="isOrg && isManualAdd">
               <label v-if="activeIndex === -1">Add Business or Corporation Unregistered in B.C.</label>
               <label v-else>Edit Business or Corporation Unregistered in B.C.</label>
+
               <!-- FUTURE -->
               <!-- <a class="lookup-toggle float-right">Business or Corporation Look Up</a> -->
 
@@ -100,7 +101,7 @@
 
               <HelpContactUs class="mt-6" />
 
-              <!-- Confirm checkbox -->
+              <!-- Confirm checkbox (org only) -->
               <v-checkbox
                 class="confirm-checkbox mt-8"
                 hide-details
@@ -117,7 +118,7 @@
                 </template>
               </v-checkbox>
 
-              <!-- Organization Name -->
+              <!-- Organization Name (org only) -->
               <v-text-field
                 filled
                 class="mt-8 mb-n6 org-name"
@@ -131,7 +132,7 @@
                 v-if="isProprietor"
                 filled
                 persistent-hint
-                class="item mt-8 mb-n2"
+                class="item business-number mt-8 mb-n2"
                 label="Business Number (Optional)"
                 hint="First 9 digits of the CRA Business Number"
                 v-model.trim="orgPerson.officer.businessNumber"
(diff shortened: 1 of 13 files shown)


tests/unit/RegPeopleAndRoles.spec.ts
```diff
@@ -65,10 +65,11 @@ function getPersonList (roles = null): any {
   ]
 }
 
-describe('Registration People And Roles component', () => {
+describe('Registration People And Roles component - SP', () => {
```
severinbeauvais, line 68: I'll create a new section for GP :)

vysakh-menon-aot (approved):

tests/unit/RegPeopleAndRoles.spec.ts
severinbeauvais: Done.

src/components/Registration/RegAddEditOrgPerson.vue
```diff
@@ -160,6 +166,16 @@
...
                     />
                   </v-col>
+
+                  <v-col cols="4" v-if="showPartnerRole" class="py-0">
+                    <v-checkbox
+                      class="mt-5"
+                      v-model="selectedRoles"
+                      :value="RoleTypes.PARTNER"
+                      :label="RoleTypes.PARTNER"
+                      :disabled="true"
```
cameron-freshworks, line 176: Always disabled, so readonly?

src/enums/roleTypes.ts
```diff
@@ -4,6 +4,7 @@ export enum RoleTypes {
   CUSTODIAN = 'Custodian',
   INCORPORATOR = 'Incorporator',
   LIQUIDATOR = 'Liquidator',
+  PARTNER = 'Partner',
```
cameron-freshworks, line 7: Did the API update? I'm still seeing lowercase at the moment, but i think it's moving towards this? Can't recall what we settled on.

cameron-freshworks (approved):
Looks good! Just a few questions.

src/components/Registration/RegAddEditOrgPerson.vue
severinbeauvais: Yes but the readonly style is incorrect.

Disabled:

image

Readonly:

image

src/enums/roleTypes.ts
severinbeauvais: @vysakh-menon-aot Will this work or does the API need a change?

You can see here that, currently, a registration saves role types in Title Case:

image

src/components/Registration/RegAddEditOrgPerson.vue
severinbeauvais: I guess, technically, disabled is more correct since sometimes the user can check one of these Roles boxes, just not in this scenario. Whereas readonly is more for presentation-only (IMO).

cameron-freshworks: Sounds good!

src/enums/roleTypes.ts
vysakh-menon-aot: This will work. And in case of edit-ui the parties endpoint change should be in dev now (as lekshmi worked on that ticket today)

PR #373, opened 2022-04-13
11854 Remove "optional" addresses before validation

severinbeauvais:
*Issue #:* /bcgov/entity#11854

*Description of changes:*
- removed conditional address validation for IAs
- removed conditional address validation for Regs
- app version = 3.5.12a


vysakh-menon-aot (approved):

src/components/Registration/RegAddEditOrgPerson.vue
```diff
@@ -206,7 +206,7 @@
                 ref="mailingAddressNew"
                 class="mt-4 mb-n6"
                 :editing="true"
-                :schema="enableRules ? PersonAddressSchema : null"
+                :schema="PersonAddressSchema"
```
severinbeauvais, line 209: There is other code to enable the rules but it doesn't nothing here, now. (This will be reworked in a future ticket.)

cameron-freshworks (approved):
Looks good. 
I imagine we need to do the same for Edit UI? 
Although, it is only STAFF ONLY filings (corrections) in the parties where this might be visible.. 
Thoughts?

severinbeauvais:
> Looks good. I imagine we need to do the same for Edit UI? Although, it is only STAFF ONLY filings (corrections) in the parties where this might be visible.. Thoughts?

Did you implement conditional address validation in Edit UI? If not, no change needed.

cameron-freshworks:
> > Looks good. I imagine we need to do the same for Edit UI? Although, it is only STAFF ONLY filings (corrections) in the parties where this might be visible.. Thoughts?
> 
> Did you implement conditional address validation in Edit UI? If not, no change needed.

We have but for ChangeOfRegistrations (Yet to be released)

PR #374, opened 2022-04-13
Staff View: Dissolution Filing SP/GP#11411 | comments

shabeeb-aot:
*Issue #:* /bcgov/entity#11411

*Description of changes:*
* Added staff comments
* Redirection for SP/GP filing from App.vue


src/components/common/EntityInfo.vue
```diff
@@ -105,11 +123,15 @@ export default class EntityInfo extends Mixins(EnumMixin) {
         return this.getApprovedName
     }
   }
+  /** The Business ID string. */
+  private get businessId (): string {
+    return this.getBusinessId
+  }
```
severinbeauvais, line 129: I think you don't need this. The template can see `getBusinessId` I think.

src/components/common/EntityInfo.vue
```diff
@@ -127,4 +149,7 @@ export default class EntityInfo extends Mixins(EnumMixin) {
   color: $gray9;
   font-weight: bold;
 }
+.staff-comments{
+  margin-left: -12px;
+}
```
severinbeauvais, line 154: Does the alignment of the staff comment component work ok when you change the screen width?

severinbeauvais: I recommend (but don't insist on) just adding class `ml-n3` to the template... IMO, layout changes like this are easier to manage if they are in the template.

src/mixins/index.ts
```diff
@@ -1,3 +1,5 @@
+// import { default as FilingMixin } from './filing-mixin'
+export { FilingMixin } from '@bcrs-shared-components/mixins'
```
severinbeauvais, line 2: If this works then I just learned something :) Thanks!

src/views/DissolutionFirm/DissolutionFirm.vue
```diff
@@ -48,8 +48,9 @@
               :nudgeTop="85"
               :minDate="startDateMinStr"
               :maxDate="startDateMaxStr"
-              :inputRules="getShowErrors ? startDateRules : []"
+              :inputRules="startDateRules || []"
```
severinbeauvais, line 51: You don't need the fallback value since startDateRules is a getter and always has a value.

src/views/DissolutionFirm/DissolutionFirm.vue
```diff
@@ -328,6 +330,10 @@ export default class DissolutionFirm extends Mixins(DateMixin) {
         ${this.dateToPacificDate(this.startDateMax, true)}`
     ]
   }
+
+  private startDateHandler (event: string): void{
+    this.dissolutionDate = event
+  }
```
severinbeauvais, line 336: This is correct code, but if there will be no other code here, you can just do the assignment in the template, eg `@emitDateSync="dissolutionDate = $event"`.

I leave this up to you.

severinbeauvais (approved):
I identified some possible cleanup (up to author's discretion), otherwise LGTM.

I note that this is a partial feature implementation since you are not yet doing anything with the dissolution date value.

severinbeauvais:
Also, you need to fix the 2 failing checks before this is merged,

cameron-freshworks (approved):

PR #376, opened 2022-04-14
9828/11470 Collateral Fix GP Create UI Business Type checkbox/Coop Incorporation - Rules

Lcyc29:
Issue #: /[bcgov/business-create-ui#9828](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/9828)

Description of changes:

- for GP only, add Business Type component with checkbox (see comments below)
   design: https://projects.invisionapp.com/share/3511UPXPMT2P#/screens/459770324

Issue #: /[bcgov/business-create-ui#11470](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/11470)

Description of changes:

- The Cooperative name is identified exactly as follows throughout the `Memorandum` --> `Rules of the Association`

### DIFF
--- a/src/components/Incorporation/UploadRules.vue
+++ b/src/components/Incorporation/UploadRules.vue
@@ -110,7 +110,10 @@
                 <v-row no-gutters>
                   <v-col cols="1"><v-icon>mdi-circle-small</v-icon></v-col>
                   <v-col cols="11">
-                    <p class="mb-0" v-html="ruleMessage"></p>
+                    <p class="mb-0">
+                      The Cooperative name is identified <b>exactly</b> as follows throughout&nbsp;
+                      the Rules of the Association:
+                    </p>
                     <div class="mt-2 mb-0 font-weight-bold">{{getNameRequestDetails.approvedName}}</div>
                   </v-col>
                 </v-row>
@@ -217,7 +220,6 @@ export default class UploadRules extends Mixins(CommonMixin, DocumentMixin) {
   private uploadRulesDocKey: string = null
   private helpToggle = false
   // eslint-disable-next-line max-len
-  private ruleMessage = 'The Cooperative name is identified <b>exactly</b> as follows throughout the Rules of the Association:'
 
   @Getter getShowErrors!: boolean
   @Getter getNameRequestDetails!: NameRequestDetailsIF
--- a/src/components/Incorporation/UploadRules.vue
+++ b/src/components/Incorporation/UploadRules.vue
@@ -111,8 +111,8 @@
                   <v-col cols="1"><v-icon>mdi-circle-small</v-icon></v-col>
                   <v-col cols="11">
                     <p class="mb-0">
-                      The Cooperative name is identified <b>exactly</b> as follows throughout&nbsp;
-                      the Rules of the Association:
+                      The Cooperative name is identified <b>exactly</b> as follows throughout
+                      <span> the</span> Rules of the Association:
                     </p>
                     <div class="mt-2 mb-0 font-weight-bold">{{getNameRequestDetails.approvedName}}</div>
                   </v-col>
@@ -219,7 +219,6 @@ export default class UploadRules extends Mixins(CommonMixin, DocumentMixin) {
   private uploadRulesDoc: File = null
   private uploadRulesDocKey: string = null
   private helpToggle = false
-  // eslint-disable-next-line max-len
 
   @Getter getShowErrors!: boolean
   @Getter getNameRequestDetails!: NameRequestDetailsIF
--- a/src/components/Incorporation/UploadRules.vue
+++ b/src/components/Incorporation/UploadRules.vue
@@ -112,7 +112,7 @@
                   <v-col cols="11">
                     <p class="mb-0">
                       The Cooperative name is identified <b>exactly</b> as follows throughout
-                      <span> the</span> Rules of the Association:
+                      the Rules of the Association:
                     </p>
                     <div class="mt-2 mb-0 font-weight-bold">{{getNameRequestDetails.approvedName}}</div>
                   </v-col>
--- a/src/components/Incorporation/UploadRules.vue
+++ b/src/components/Incorporation/UploadRules.vue
@@ -110,9 +110,7 @@
                 <v-row no-gutters>
                   <v-col cols="1"><v-icon>mdi-circle-small</v-icon></v-col>
                   <v-col cols="11">
-                    <p class="mb-0">
-                      The Cooperative name is identified <b>exactly</b> as follows throughout the Memorandum:
-                    </p>
+                    <p class="mb-0" v-html="ruleMessage"></p>
                     <div class="mt-2 mb-0 font-weight-bold">{{getNameRequestDetails.approvedName}}</div>
                   </v-col>
                 </v-row>
@@ -218,6 +216,8 @@ export default class UploadRules extends Mixins(CommonMixin, DocumentMixin) {
   private uploadRulesDoc: File = null
   private uploadRulesDocKey: string = null
   private helpToggle = false
+  // eslint-disable-next-line max-len
+  private ruleMessage = 'The Cooperative name is identified <b>exactly</b> as follows throughout the Rules of the Association:'
 
   @Getter getShowErrors!: boolean
   @Getter getNameRequestDetails!: NameRequestDetailsIF
--- a/src/components/Registration/BusinessTypeConfirm.vue
+++ b/src/components/Registration/BusinessTypeConfirm.vue
@@ -55,7 +55,7 @@ export default class BusinessTypeConfirm extends Vue {
   }
 
   /** Emits updated business type event. */
-  @Emit('update:BusinessTypeConfirm')
+  @Emit('update:businessTypeConfirm')
   private emitBusinessType (): boolean {
     return this.checked
   }
--- a/src/components/Registration/BusinessTypeConfirm.vue
+++ b/src/components/Registration/BusinessTypeConfirm.vue
@@ -53,7 +53,7 @@ export default class BusinessTypeConfirm extends Vue {
   /** Called when component is mounted. */
   protected mounted (): void {
     // init model variable + validate
-    this.checked = this.businessTypeConfirm
+    this.checked = this.businessTypeConfirm || false
     this.emitValid()
   }
 
--- a/src/components/Registration/BusinessTypeConfirm.vue
+++ b/src/components/Registration/BusinessTypeConfirm.vue
@@ -1,5 +1,5 @@
 <template>
-  <div id="business-type">
+  <div id="business-type-confirm">
     <v-row no-gutters>
       <v-col cols="12" sm="3" class="pr-4">
         <label class="d-block title-label">Business Type</label>
@@ -16,18 +16,17 @@
 
       <v-col cols="12" sm="9">
         <p class="mb-0">General Partnership</p>
-        <div id="business-type-div">
+        <div id="business-check-div">
           <v-checkbox
-            class="business-type-check mt-0"
-            hide-details
+            class="mt-0"
             v-model="checked"
-          />
-          <p class="mb-0" id="business-type-text">
-            I acknowledge that a General Partnership cannot be changed into a
+            hide-details
+            label="I acknowledge that a General Partnership cannot be changed into a
             Sole Proprietorship (including DBA). If this is necessary, a new
             Name Request Number and Statement of Registration (along with
-            associated fees) will be required.
-          </p>
+            associated fees) will be required."
+          >
+          </v-checkbox>
         </div>
       </v-col>
     </v-row>
@@ -82,7 +81,7 @@ export default class BusinessTypeConfirm extends Vue {
 <style lang="scss" scoped>
 @import "@/assets/styles/theme.scss";
 
-#business-type {
+#business-type-confirm {
   .col-sm-3 {
     color: $gray9;
     font-weight: bold;
@@ -93,18 +92,18 @@ export default class BusinessTypeConfirm extends Vue {
   }
 }
 
-#business-type-div {
+#business-check-div {
   display: flex;
   margin-top: 20px;
 }
 
-#business-type-div .business-type-check {
+#business-check-div .mt-0 {
   margin-left: -3px;
   padding-top: 1px;
 }
 
-#business-type-text {
-  font-size: 15px;
+::v-deep .v-input--checkbox .v-input__slot {
+  align-items: flex-start;
 }
 
 .v-icon.mdi-open-in-new {
--- a/src/components/Registration/BusinessTypeConfirm.vue
+++ b/src/components/Registration/BusinessTypeConfirm.vue
@@ -120,7 +120,7 @@ export default class BusinessTypeConfirm extends Vue {
 }
 
 .invalid-section ::v-deep .v-input--checkbox .v-input__slot .v-label {
-  color: #d3272c !important;
+  color: $app-red !important;
 }
 
 .v-icon.mdi-open-in-new {
--- a/src/components/Registration/BusinessTypeConfirm.vue
+++ b/src/components/Registration/BusinessTypeConfirm.vue
@@ -15,16 +15,13 @@
       </v-col>
 
       <v-col cols="12" sm="9">
-        <p class="mb-0 checkbox-gp-text">General Partnership</p>
+        <p class="mb-0 checkbox-gp-text">{{ label }}</p>
         <div id="business-check-div">
           <v-checkbox
             class="mt-0"
             v-model="checked"
             hide-details
-            label="I acknowledge that a General Partnership cannot be changed into a
-            Sole Proprietorship (including DBA). If this is necessary, a new
-            Name Request Number and Statement of Registration (along with
-            associated fees) will be required."
+            :label="text"
           >
           </v-checkbox>
         </div>
@@ -46,13 +43,29 @@ export default class BusinessTypeConfirm extends Vue {
   @Prop({ default: false })
   readonly hasBusinessTypeChecked!: boolean;
 
+  /** Whether the business typ is SP or GP */
+  @Prop({ default: false })
+  readonly isTypePartnership!: boolean;
+
   // Local variables
-  protected checked = false;
+  protected checked = false
+  protected label = 'General Partnership'
+  protected labelSP = 'BC Sole Proprietorship / Doing Business As (DBA) Registration'
+  protected text = `I acknowledge that a General Partnership cannot be \
+    changed into a Sole Proprietorship (including DBA). If this is \
+    necessary, a new Name Request Number and Statement of Registration \
+    (along with associated fees) will be required.`
+  protected textSP = `I acknowledge that a Sole Proprietorship (including DBA) \
+    cannot be changed into a General Partnership. If this is necessary, a new \
+    Name Request Number and Statement of Registration (along with associated fees) \
+    will be required.`
 
   /** Called when component is mounted. */
   protected mounted (): void {
     // init model variable + validate
     this.checked = this.businessTypeConfirm
+    this.label = this.isTypePartnership ? this.label : this.labelSP
+    this.text = this.isTypePartnership ? this.text : this.textSP
     this.emitValid()
   }
 
--- a/src/components/Registration/BusinessTypeConfirm.vue
+++ b/src/components/Registration/BusinessTypeConfirm.vue
@@ -1,0 +1,103 @@
+<template>
+  <div id="business-type">
+    <v-row no-gutters>
+      <v-col cols="12" sm="3" class="pr-4">
+        <label class="d-block title-label">Business Type</label>
+        <v-chip
+          v-if="hasBusinessTypeChecked"
+          id="checked-chip"
+          x-small label
+          color="primary"
+          text-color="white"
+        >
+          Checked
+        </v-chip>
+      </v-col>
+
+      <v-col cols="12" sm="9">
+        <p class="mb-0">General Partnership</p>
+        <div id="business-type-div">
+          <v-checkbox
+            class="business-type-check mt-0"
+            hide-details
+            v-model="checked"
+          />
+          <p class="mb-0" id="business-type-text">
+            I acknowledge that a General Partnership cannot be changed into a
+            Sole Proprietorship (including DBA). If this is necessary, a new
+            Name Request Number and Statement of Registration (along with
+            associated fees) will be required.
+          </p>
+        </div>
+      </v-col>
+    </v-row>
+  </div>
+</template>
+
+<script lang="ts">
+import { Component, Emit, Prop, Vue, Watch } from 'vue-property-decorator'
+
+@Component
+export default class BusinessTypeConfirm extends Vue {
+  /** Whether to display Change features. */
+  @Prop({ default: false })
+  readonly hasBusinessTypeChecked!: boolean;
+
+  // Local variables
+  protected checked = false;
+
+  /** Watches for changes to the checkbox value. */
+  @Watch('checked')
+  private onValueChanged (): void {
+    // update business type + validate
+    this.emitBusinessType()
+    this.emitValid()
+  }
+
+  /** Emits updated business type event. */
+  @Emit('update:BusinessTypeConfirm')
+  private emitBusinessType (): boolean {
+    return this.checked
+  }
+
+  // /** Emits component validity event. */
+  @Emit('valid')
+  private emitValid (): boolean {
+    return this.checked
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import "@/assets/styles/theme.scss";
+
+#business-type {
+  .col-sm-3 {
+    color: $gray9;
+    font-weight: bold;
+  }
+
+  .col-sm-9 {
+    color: $gray7;
+  }
+}
+
+#business-type-div {
+  display: flex;
+  margin-top: 20px;
+}
+
+#business-type-div .business-type-check {
+  margin-left: -3px;
+  padding-top: 1px;
+}
+
+#business-type-text {
+  font-size: 15px;
+}
+
+.v-icon.mdi-open-in-new {
+  margin-top: -2px;
+  padding-left: 2px;
+}
+</style>
--- a/src/components/Registration/BusinessTypeConfirm.vue
+++ b/src/components/Registration/BusinessTypeConfirm.vue
@@ -39,12 +39,23 @@ import { Component, Emit, Prop, Vue, Watch } from 'vue-property-decorator'
 
 @Component
 export default class BusinessTypeConfirm extends Vue {
+  /** The registration business number. */
+  @Prop({ required: true })
+  readonly businessTypeConfirm!: boolean
+
   /** Whether to display Change features. */
   @Prop({ default: false })
   readonly hasBusinessTypeChecked!: boolean;
 
   // Local variables
   protected checked = false;
+
+  /** Called when component is mounted. */
+  protected mounted (): void {
+    // init model variable + validate
+    this.checked = this.businessTypeConfirm
+    this.emitValid()
+  }
 
   /** Watches for changes to the checkbox value. */
   @Watch('checked')
--- a/src/components/Registration/BusinessTypeConfirm.vue
+++ b/src/components/Registration/BusinessTypeConfirm.vue
@@ -53,7 +53,7 @@ export default class BusinessTypeConfirm extends Vue {
   /** Called when component is mounted. */
   protected mounted (): void {
     // init model variable + validate
-    this.checked = this.businessTypeConfirm || false
+    this.checked = this.businessTypeConfirm
     this.emitValid()
   }
 
--- a/src/components/Registration/BusinessTypeConfirm.vue
+++ b/src/components/Registration/BusinessTypeConfirm.vue
@@ -2,7 +2,7 @@
   <div id="business-type-confirm">
     <v-row no-gutters>
       <v-col cols="12" sm="3" class="pr-4">
-        <label class="d-block title-label">Business Type</label>
+        <label class="d-block">Business Type</label>
         <v-chip
           v-if="hasBusinessTypeChecked"
           id="checked-chip"
@@ -15,7 +15,7 @@
       </v-col>
 
       <v-col cols="12" sm="9">
-        <p class="mb-0">General Partnership</p>
+        <p class="mb-0 checkbox-gp-text">General Partnership</p>
         <div id="business-check-div">
           <v-checkbox
             class="mt-0"
@@ -104,6 +104,10 @@ export default class BusinessTypeConfirm extends Vue {
 
 ::v-deep .v-input--checkbox .v-input__slot {
   align-items: flex-start;
+}
+
+.invalid-section ::v-deep .v-input--checkbox .v-input__slot .v-label {
+  color: #d3272c !important;
 }
 
 .v-icon.mdi-open-in-new {
--- a/src/components/Registration/DefineRegistrationSummary.vue
+++ b/src/components/Registration/DefineRegistrationSummary.vue
@@ -24,6 +24,22 @@
         </v-row>
       </article>
 
+      <!-- Business Type Confirm -->
+      <template v-if="isTypePartnership">
+        <v-divider class="mx-6" />
+
+        <article class="section-container">
+          <v-row no-gutters>
+            <v-col cols="12" sm="3" class="pr-4">
+              <label>Business Type</label>
+            </v-col>
+            <v-col cols="12" sm="9">
+              <span>{{ businessTypeConfirm || false}}</span>
+            </v-col>
+          </v-row>
+        </article>
+      </template>
+
       <v-divider class="mx-6" />
 
       <!-- Nature of Business -->
@@ -153,6 +169,11 @@ export default class DefineRegistrationSummary extends Mixins(DateMixin, EnumMix
   /** The business number. */
   get businessNumber (): string {
     return this.getRegistration.businessNumber
+  }
+
+  /** The business type confirm. */
+  get businessTypeConfirm (): boolean {
+    return this.getRegistration.businessTypeConfirm
   }
 
   /** The business start date. */
--- a/src/components/Registration/DefineRegistrationSummary.vue
+++ b/src/components/Registration/DefineRegistrationSummary.vue
@@ -24,22 +24,6 @@
         </v-row>
       </article>
 
-      <!-- Business Type Confirm -->
-      <template v-if="isTypePartnership">
-        <v-divider class="mx-6" />
-
-        <article class="section-container">
-          <v-row no-gutters>
-            <v-col cols="12" sm="3" class="pr-4">
-              <label>Business Type</label>
-            </v-col>
-            <v-col cols="12" sm="9">
-              <span>{{ businessTypeConfirm || false}}</span>
-            </v-col>
-          </v-row>
-        </article>
-      </template>
-
       <v-divider class="mx-6" />
 
       <!-- Nature of Business -->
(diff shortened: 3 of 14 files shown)


severinbeauvais:
1. please update the link in the description; look at other PRs for syntax
2. please update the app version in the package.json file, run `npm i`, and then commit both package.json and package-lock.json
3. you may have to rebase in case someone else updated app version before you

src/views/Registration/RegistrationDefineBusiness.vue
```diff
@@ -9,7 +9,19 @@
...
+        <template v-if="isTypePartnership">
+          <v-divider class="mx-6" />
+
+          <!-- Business Type -->
+          <header id="business-type-header" />
+          <BusinessTypeGP
+            class="py-8 px-6"
+            :class="{ 'invalid-section': getShowErrors && !businessTypeGPValid }"
+            :showErrors="getShowErrors"
+            @update:businessTypeGP="setRegistrationBusinessTypeGP($event)"
```
severinbeauvais, line 21: Do you need to restore the value from a draft filing (and thus pass it in to this component)?

src/views/Registration/RegistrationDefineBusiness.vue
```diff
@@ -167,8 +181,10 @@ export default class RegistrationDefineBusiness extends Mixins(CommonMixin) {
   @Action setRegistrationDefineBusinessValid!: ActionBindingIF
   @Action setFolioNumber!: ActionBindingIF
   @Action setRegistrationBusinessNumber!: ActionBindingIF
+  @Action setRegistrationBusinessTypeGP!: ActionBindingIF
```
severinbeauvais, line 184: This is a new action, right? You'll need to add it, and its mutation, and check in those file.

src/components/Incorporation/UploadRules.vue
```diff
@@ -111,7 +111,7 @@
                   <v-col cols="1"><v-icon>mdi-circle-small</v-icon></v-col>
                   <v-col cols="11">
                     <p class="mb-0">
-                      The Cooperative name is identified <b>exactly</b> as follows throughout the Memorandum:
+                      The Cooperative name is identified <b>exactly</b> as follows throughout the Rules of the Association:
```
severinbeauvais, line 114: Please run lint locally. There is a warning here that you need to fix.

image

(Sometimes our GH CI action catches these and sometimes not. But it's good practise to fix it if you broke it :)  )

src/components/Registration/BusinessTypeGP.vue
```diff
@@ -0,0 +1,103 @@
...
+      <v-col cols="12" sm="9">
+        <p class="mb-0">General Partnership</p>
+        <div id="business-type-div">
+          <v-checkbox
+            class="business-type-check mt-0"
+            hide-details
+            v-model="checked"
+          />
+          <p class="mb-0" id="business-type-text">
+            I acknowledged that a General Partnership cannot be changed into a
```
severinbeauvais, line 26: Typo: acknowledged -> acknowledge

src/store/mutations/mutations-model.ts
```diff
@@ -349,3 +349,8 @@ export const mutateRegistrationBusinessType = (state: StateIF, businessType: Bus
   state.stateModel.registration.businessType = businessType
   if (!state.stateModel.ignoreChanges) mutateHaveChanges(state, true)
 }
+
+export const mutateRegistrationBusinessTypeGP = (state: StateIF, businessTypeGP: boolean) => {
+  state.stateModel.registration.businessTypeGP = businessTypeGP
+  if (!state.stateModel.ignoreChanges) mutateHaveChanges(state, true)
```
severinbeauvais, line 355: @Lcyc29 In case you're wondering, this is to set a "have changes" flag (as long as we're not ignoring changes, like during initial load) when the user changes this property. This flag is used to warn users if they try to leave this app with unsaved changes.

src/views/Registration/RegistrationDefineBusiness.vue
```diff
@@ -9,7 +9,19 @@
...
-        <!-- FUTURE: Business Type goes here -->
+        <template v-if="isTypePartnership">
+          <v-divider class="mx-6" />
+
+          <!-- Business Type -->
+          <header id="business-type-header" />
+          <BusinessTypeGP
+            class="py-8 px-6"
+            :class="{ 'invalid-section': getShowErrors && !businessTypeGPValid }"
+            :showErrors="getShowErrors"
```
severinbeauvais, line 20: You don't need to pass this prop since your component doesn't use it.

severinbeauvais:
4. Please add some unit tests for your new component.
    Unfortunately we don't yet have a test suite for RegistrationDefineBusiness (in which you could test that an event emitted by your component changes the page's validation state).

src/views/Registration/RegistrationDefineBusiness.vue
Lcyc29: I need to restore the value from the draft filing, and I've updated the code. Please review it.

src/views/Registration/RegistrationDefineBusiness.vue
Lcyc29: I've added the mutation and its subsequent contents. Please review the code.

src/views/Registration/RegistrationDefineBusiness.vue
severinbeauvais: Did you know that you could make your component support a `v-model` prop?

Ref: https://v2.vuejs.org/v2/guide/components.html?redirect=true#Using-v-model-on-Components

TLDR: if you component accepts a `value` prop and emits an `input` event, then the parent (THIS page) can simply use `v-model="yourLocalVariableName"`.

The reason we generally don't do this is that we need a getter/action to read/write from/to the store. So then we would need to separate the v-model's input and output behaviour, maybe like this:
```
get yourLocalVariableName (): boolean {
  // put store getter call here
}
set yourLocalVariableName (val: boolean): void {
  // put store action call here
}
```
As you can see, this doesn't save any code at all. In fact I think it's more complicated to understand, and I'm all for simple-to-understand code :)

src/views/Registration/RegistrationDefineBusiness.vue
Lcyc29: Okay. Thanks. I will remove it.

src/store/mutations/mutations-model.ts
Lcyc29: Thanks for the clarification.

src/components/Registration/BusinessTypeGP.vue
Lcyc29: Got it.

src/components/Incorporation/UploadRules.vue
Lcyc29: I think it's gone now. It was a warning from eslint that line 114 was over the limit of 120 characters.

src/mixins/filing-template-mixin.ts
```diff
@@ -400,7 +400,7 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
     this.setRegistrationBusinessAddress(draftFiling.registration.offices?.businessOffice)
 
     // restore Business Type
-    this.setRegistrationBusinessType(draftFiling.registration.businessType || BusinessTypes.SP)
+    this.setRegistrationBusinessType(draftFiling.registration.businessType)
```
severinbeauvais, line 403: For the reviewers: I asked for this change. It is so that Business Type is not defaulted to SP for General Partnership. I confirmed with @vysakh-menon-aot that Business Type does not apply to GPs.

src/components/Incorporation/UploadRules.vue
severinbeauvais: This line still appears to be 124 columns wide in this PR. Did you change it and commit the change?

src/views/Registration/RegistrationDefineBusiness.vue
severinbeauvais: Looks good.

src/components/Registration/BusinessTypeGP.vue
```diff
@@ -0,0 +1,103 @@
...
+      </v-col>
+    </v-row>
+  </div>
+</template>
+
+<script lang="ts">
+import { Component, Emit, Prop, Vue, Watch } from 'vue-property-decorator'
+
+@Component
+export default class BusinessTypeGP extends Vue {
```
severinbeauvais, line 41: I just realized that Business Type is an existing property; it is a SP sub-type. (Which explains why you named this BusinessTypeGP.)

I recomment that you rename this component to avoid confusion. What do you plan to call your store value? Perhaps that's a good name for this component too. Maybe something with "acknowledge" or "confirm" in it?

src/components/Incorporation/UploadRules.vue
Lcyc29: I am using a private variable for the long string and use disable eslint from there. This line of code is outdated. However, I am not sure if that's going to work, because I was unable to access the Incorporation Application.

src/components/Registration/BusinessTypeGP.vue
Lcyc29: I think BusinessConfirm is short and good. I will change it.

src/interfaces/store-interfaces/state-interfaces/registration-state-interface.ts
```diff
@@ -15,4 +15,5 @@ export interface RegistrationStateIF {
   naics: NaicsIF
   businessType: BusinessTypes
   businessNumber?: string // GPs only
+  BusinessTypeConfirm?: boolean // GPs only
```
severinbeauvais, line 18: Our code style for variables is camelCase.

We use PascalCase for Types, Enums an const classes.

src/components/Registration/BusinessTypeConfirm.vue
```diff
@@ -0,0 +1,103 @@
...
+  /** Watches for changes to the checkbox value. */
+  @Watch('checked')
+  private onValueChanged (): void {
+    // update business type + validate
+    this.emitBusinessType()
+    this.emitValid()
+  }
+
+  /** Emits updated business type event. */
+  @Emit('update:BusinessTypeConfirm')
```
severinbeauvais, line 58: This needs to be camelCase too.

I'm a bit surprised that the parent event spec (`@update:businessTypeConfirm=...`) works. I know it auto-converts from camelCase to kebab-case but this is new to me 🤷‍♂️

src/components/Registration/BusinessTypeGP.vue
severinbeauvais: Yeah, sorry about requesting that change. I see it caused a lot of code changes. But I think the new name is short and good and more clear :)

src/components/Incorporation/UploadRules.vue
severinbeauvais: I don't see this changed file any more. Maybe you clobbered it during a rebase or squash.

I saw your changes temporarily. They should work. You'll need to create the right kind of IA to test it.

PS Use eslint disable sparingly. I'd prefer to see a long line broken up into 2 lines.

severinbeauvais: Here's a link to your commit: https://github.com/bcgov/business-create-ui/commit/f783a75c1cf34096a94c6f1497d159e1442cd616

You should be able to cherry-pick it and then push it to GH.

src/views/Registration/RegistrationDefineBusiness.vue
severinbeauvais: Closing this conversation because some of the code has been updated.

src/views/Registration/RegistrationDefineBusiness.vue
```diff
@@ -9,7 +9,18 @@
...
+          <v-divider class="mx-6" />
+
+          <!-- Business Type -->
+          <header id="business-type-header" />
+          <BusinessTypeConfirm
+            class="py-8 px-6"
+            :class="{ 'invalid-section': getShowErrors && !businessTypeConfirmValid }"
+            @update:businessTypeConfirm="setRegistrationBusinessTypeConfirm($event)"
+            @valid="onBusinessTypeConfirmValidEvent($event)"
+          />
```
severinbeauvais, line 22: Don't forget to pass in the initial value (as restored from a draft filing).

src/components/Incorporation/UploadRules.vue
Lcyc29: I lost it somewhere, and it should be there now. Thanks a lot.

src/views/Registration/RegistrationDefineBusiness.vue
Lcyc29: Thanks. I added the initial value to both files: BusinessNumber.vue and RegistrationDefineBusiness.vue.

severinbeauvais: Yup, looks good 👍

severinbeauvais (commented):
Looks good. Please check open comments and checkboxes above for remaining items.

src/components/Incorporation/UploadRules.vue
severinbeauvais: The file is back! :)

Would you mind breaking up your long line up into 2 lines, so you don't have to use eslint-disable? If it looks ugly, consider just having the text on 2 lines in the template...?

Lcyc29: Sure. I broke the this long line into 2 lines in my latest commit.

src/components/Incorporation/UploadRules.vue
```diff
@@ -111,7 +111,8 @@
                   <v-col cols="1"><v-icon>mdi-circle-small</v-icon></v-col>
                   <v-col cols="11">
                     <p class="mb-0">
-                      The Cooperative name is identified <b>exactly</b> as follows throughout the Memorandum:
+                      The Cooperative name is identified <b>exactly</b> as follows throughout&nbsp;
```
severinbeauvais, line 114: You only need a non-blanking space in an inline element (like a span). This is what a paragraph with a carriage return looks like in the DOM:

image

In a nutshell, the above displays as you'd expect: the CR disappears and there is no extra whitespace at beginning, middle or end.

Long version: https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace

src/components/Incorporation/UploadRules.vue
```diff
@@ -218,6 +219,7 @@ export default class UploadRules extends Mixins(CommonMixin, DocumentMixin) {
   private uploadRulesDoc: File = null
   private uploadRulesDocKey: string = null
   private helpToggle = false
+  // eslint-disable-next-line max-len
```
severinbeauvais, line 222: Please delete this.

src/components/Incorporation/UploadRules.vue
Lcyc29: Thanks Severin. I modified the code in my latest commit. Please have a look.

src/components/Incorporation/UploadRules.vue
Lcyc29: Okay. It is now deleted. My bad.

severinbeauvais:
1. Sorry about the comment above that was not linked to a line number. This happens when I comment on an individual commit instead of the files page. Sometimes I forget to reload the files page and comment on the commit instead... oops.

2. Sorry if I seem really picky about changes. You're actually doing very well! I'm trying to avoid even little things that can accumulate and make things harder to manage over time in a large project such as this.

3. This PR only needs the app version (and both package files) committed, and ideally a unit test to verify that your new component works as intended. (See open comments above.) Then we can merge it :)

Lcyc29:
> 1. Sorry about the comment above that was not linked to a line number. This happens when I comment on an individual commit instead of the files page. Sometimes I forget to reload the files page and comment on the commit instead... oops.
> 2. Sorry if I seem really picky about changes. You're actually doing very well! I'm trying to avoid even little things that can accumulate and make things harder to manage over time in a large project such as this.
> 3. This PR only needs the app version (and both package files) committed, and ideally a unit test to verify that your new component works as intended. (See open comments above.) Then we can merge it :)

Thanks Severin for the peer review and all the help for my first ticket. Sorry for my sloppy codes and you have to go through them. I understand that in order to maintain good project scalability and code management, I should avoid little things that could become blockers later on and need improve my code quality. I will run a unit test with Jest Runner and change the app version to 3.15.17 in both package files in my next commit. Thanks Severin :)

vysakh-menon-aot (approved):

src/mixins/filing-template-mixin.ts
```diff
@@ -414,6 +415,9 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
     // restore Business Number
     this.setRegistrationBusinessNumber(draftFiling.registration.business.taxId || null)
 
+    // restore Business Type Confirm
+    this.setRegistrationBusinessTypeConfirm(draftFiling.registration.businessTypeConfirm || false)
```
Lcyc29, line 419: I changed null to false because otherwise it's initial value being null is giving me an error in the unfinished step/scrolling down mechanics.

src/store/state/state-model.ts
```diff
@@ -167,6 +167,7 @@ export const stateModel: StateModelIF = {
     feeAcknowledgement: false,
     naics: { ...EmptyNaics },
     businessType: null,
-    businessNumber: null
+    businessNumber: null,
+    businessTypeConfirm: false
```
Lcyc29, line 171: I changed null to false for the initial value to make sure the unfinished step/scroll down mechanics is correct.

src/components/Registration/DefineRegistrationSummary.vue
```diff
@@ -24,6 +24,22 @@
...
+            <v-col cols="12" sm="3" class="pr-4">
+              <label>Business Type</label>
+            </v-col>
+            <v-col cols="12" sm="9">
+              <span>{{ businessTypeConfirm || false}}</span>
+            </v-col>
+          </v-row>
+        </article>
+      </template>
+
```
Lcyc29, line 42: I added the above template block in this file in my latest commit, because the unfinished steps/scroll down mechanics didn't work properly. I am not sure whether I should add this block though.

src/components/Registration/DefineRegistrationSummary.vue
```diff
@@ -155,6 +171,11 @@ export default class DefineRegistrationSummary extends Mixins(DateMixin, EnumMix
     return this.getRegistration.businessNumber
   }
 
+  /** The business type confirm. */
+  get businessTypeConfirm (): boolean {
+    return this.getRegistration.businessTypeConfirm
+  }
```
Lcyc29, line 177: I added this method in my latest commit, for the same reason above. I am not sure I should add this.

src/mixins/filing-template-mixin.ts
severinbeauvais: Does the error/scrolling mechanism still work for a Sole Prop? Maybe something special needs to be done in there so it only applies to GPs.
Here, I still prefer null since it's neither false nor true, so we consider it correct for a SP. Another option is to wrap this code in an if statement so it's only set for GPs.

src/store/state/state-model.ts
severinbeauvais: It would be more correct to fix those mechanics so they return valid for businessTypeConfirm for a SP. There's already something like that for businessNumberValid that I think you can reuse.

src/components/Registration/DefineRegistrationSummary.vue
severinbeauvais: Looks fine to me.

src/components/Registration/DefineRegistrationSummary.vue
severinbeauvais: Lines 28-41?

The template + v-if seems correct to me since this feature only applies to GPs.

PS I think this component (a subcomponent of RegistrationReviewConfirm page/view) doesn't have the error scrolling mechanism.

severinbeauvais: ACTUALLY, I think this checkbox doesn't display on the review page.

src/mixins/filing-template-mixin.ts
severinbeauvais: Fixed

src/views/Registration/RegistrationDefineBusiness.vue
```diff
@@ -9,7 +9,19 @@
...
+          <v-divider class="mx-6" />
+
+          <!-- Business Type -->
+          <header id="business-type-confirm-header" />
+          <BusinessTypeConfirm
+            class="py-8 px-6"
+            :class="{ 'invalid-section': getShowErrors && !businessTypeConfirmValid }"
+            :businessTypeConfirm="getRegistration.businessTypeConfirm"
+            @update:businessTypeConfirm="setRegistrationBusinessTypeConfirm($event)"
+            @valid="onBusinessTypeConfirmValidEvent($event)"
```
severinbeauvais, line 22: You may need to pass the global error flag into your component to change is label red on error. Check with Yui what he wants.

src/components/Registration/BusinessTypeConfirm.vue
```diff
@@ -0,0 +1,113 @@
...
+  display: flex;
+  margin-top: 20px;
+}
+
+#business-check-div .mt-0 {
+  margin-left: -3px;
+  padding-top: 1px;
+}
+
+::v-deep .v-input--checkbox .v-input__slot {
```
Lcyc29, line 105: This was added in the commit: ac78c6c... to align the checkbox and the label.

src/components/Registration/BusinessTypeConfirm.vue
```diff
@@ -0,0 +1,113 @@
...
+        <p class="mb-0">General Partnership</p>
+        <div id="business-check-div">
+          <v-checkbox
+            class="mt-0"
+            v-model="checked"
+            hide-details
+            label="I acknowledge that a General Partnership cannot be changed into a
+            Sole Proprietorship (including DBA). If this is necessary, a new
+            Name Request Number and Statement of Registration (along with
+            associated fees) will be required."
```
Lcyc29, line 27: In commit ac78c6c, I removed the p tag and used label instead. The checkbox is now aligned with the label top.

src/store/state/state-model.ts
severinbeauvais: Fixed

src/views/Registration/RegistrationDefineBusiness.vue
Lcyc29: Okay. I'm on it.

src/mixins/filing-template-mixin.ts
```diff
@@ -400,7 +401,7 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
     this.setRegistrationBusinessAddress(draftFiling.registration.offices?.businessOffice)
 
     // restore Business Type
-    this.setRegistrationBusinessType(draftFiling.registration.businessType || BusinessTypes.SP)
+    this.setRegistrationBusinessType(draftFiling.registration.businessType)
```
severinbeauvais, line 404: I thought there was a `|| null` here before? But it looks like this will also work (ie, if businessType is not in the filing object then undefined will be assigned).

For consistency, would leaving off the fallback work for Business Type Confirm below?

src/views/Registration/RegistrationDefineBusiness.vue
severinbeauvais: I see that you're looking at the conditional 'invalid-section' classname on the object to set the label text colour. Sneaky but if it works then 👍

src/components/Registration/BusinessTypeConfirm.vue
```diff
@@ -0,0 +1,130 @@
...
+    will be required.`
+
+  /** Called when component is mounted. */
+  protected mounted (): void {
+    // init model variable + validate
+    this.checked = this.businessTypeConfirm
+    this.label = this.isTypePartnership ? this.label : this.labelSP
+    this.text = this.isTypePartnership ? this.text : this.textSP
+    this.emitValid()
+  }
```
Lcyc29, line 70: According to the latest [design](https://projects.invisionapp.com/share/3511UPXPMT2P#/screens/460287155_Step_1-2_SP_-_DBA_Errors), I included added SP for this component and used conditional statements to render the label and text message.

src/components/Registration/BusinessTypeConfirm.vue
```diff
@@ -0,0 +1,130 @@
...
+          x-small label
+          color="primary"
+          text-color="white"
+        >
+          Checked
+        </v-chip>
+      </v-col>
+
+      <v-col cols="12" sm="9">
+        <p class="mb-0 checkbox-gp-text">{{ label }}</p>
```
Lcyc29, line 18: label is now a local variable.

src/views/Registration/RegistrationDefineBusiness.vue
```diff
@@ -9,7 +9,18 @@
...
+        <!-- Business Type -->
+        <header id="business-type-confirm-header" />
+        <BusinessTypeConfirm
+          class="py-8 px-6"
+          :class="{ 'invalid-section': getShowErrors && !businessTypeConfirmValid }"
+          :businessTypeConfirm="getRegistration.businessTypeConfirm"
+          :isTypePartnership="isTypePartnership"
+          @update:businessTypeConfirm="setRegistrationBusinessTypeConfirm($event)"
+          @valid="onBusinessTypeConfirmValidEvent($event)"
+        />
```
Lcyc29, line 23: In the latest commit 17a9bb1, I removed the template tag and v-if condition, because we will the checkbox for SP too. This is the [design link](https://projects.invisionapp.com/share/3511UPXPMT2P#/screens/460287155_Step_1-2_SP_-_DBA_Errors).

src/components/Registration/BusinessTypeConfirm.vue
```diff
@@ -0,0 +1,130 @@
...
+  @Prop({ required: true })
+  readonly businessTypeConfirm!: boolean
+
+  /** Whether to display Change features. */
+  @Prop({ default: false })
+  readonly hasBusinessTypeChecked!: boolean;
+
+  /** Whether the business typ is SP or GP */
+  @Prop({ default: false })
+  readonly isTypePartnership!: boolean;
```
severinbeauvais, line 48: Are you sure this component displays for SP? My understanding was that this component is for GP only.

(I know the designs show both, but I think those are old designs 🤦‍♂️ )

severinbeauvais: OK, it looks like Yui left some comments in the ticket fairly recently so I guess those are the "latest".

Please update the ticket description to call out the SP/DBA version. (Delete "old description" and just clean it up.)

src/views/Registration/RegistrationDefineBusiness.vue
```diff
@@ -179,6 +194,7 @@ export default class RegistrationDefineBusiness extends Mixins(CommonMixin) {
   /** Object of valid flags. Must match validComponents. */
   get validFlags (): object {
     return {
+      businessTypeConfirmValid: this.isTypePartnership ? this.businessTypeConfirmValid : true,
```
severinbeauvais, line 197: You need to update this for validation to work for SP businesses.

Also, if there are any comments that say Business Type Confirm is for GP only, please remove them (and possibly remove the optional designator) (and possibly update the initialization and fallback in filing-template-mixin to False since that should be the initial value for both SP and GP).

src/components/Registration/BusinessTypeConfirm.vue
Lcyc29: Yes. I've updated the description in Zenhub. Here is the [link](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/9828) fyi.

src/components/Registration/BusinessTypeConfirm.vue
```diff
@@ -0,0 +1,117 @@
...
+  padding-top: 1px;
+}
+
+::v-deep .v-input--checkbox .v-input__slot {
+  align-items: flex-start;
+}
+
+.invalid-section ::v-deep .v-input--checkbox .v-input__slot .v-label {
+  color: #d3272c !important;
+}
```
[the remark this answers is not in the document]
Lcyc29: This is corrected, and the line has been moved to 123.

src/components/Registration/BusinessTypeConfirm.vue
severinbeauvais: It's nice to have a code comment here to say what you're doing.

src/mixins/filing-template-mixin.ts
Lcyc29: We will use false for Business Type Confirm below, so this will be okay.

src/mixins/filing-template-mixin.ts
```diff
@@ -414,6 +415,9 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
     // restore Business Number
     this.setRegistrationBusinessNumber(draftFiling.registration.business.taxId || null)
 
+    // restore Business Type Confirm
+    this.setRegistrationBusinessTypeConfirm(draftFiling.registration.businessTypeConfirm)
```
Lcyc29, line 419: In the latest commit 5511d94, businessTypeConfirm uses the initial value from state-model, which is false, for both SP and GP.

src/store/state/state-model.ts
Lcyc29, line 171: In the latest commit 5511d94, the initial value for businessTypeConfirm is now false for both SP and GP.

src/mixins/filing-template-mixin.ts
severinbeauvais: 1. Did you remove `|| null` from line 404, and if so, why?

2. What about the case where `businessTypeConfirm` is not a field in the draft filing (eg, a brand new filing)? That's where the fallback makes sense.

src/views/Registration/RegistrationDefineBusiness.vue
severinbeauvais: Fixed

src/mixins/filing-template-mixin.ts
severinbeauvais: Yes, businessTypeConfirm in the store is initialized correctly.

However, here a draft filing is loaded. For a brand-new draft registration filing, this is not set so you will get an `undefined` value.

When a user has created a NR, is on the My Business Registry page, and clicks "use this NR" for a registration filing, a draft filing is created here: https://github.com/bcgov/sbc-auth/blob/64a61f807537de467af53b5617b8aab030cc0602/auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue#L393

Lcyc29: I changed it back to draftFiling.registration.businessTypeConfirm || null in my latest commit 674dc35.

src/interfaces/store-interfaces/state-interfaces/registration-state-interface.ts
```diff
@@ -15,4 +15,5 @@ export interface RegistrationStateIF {
   naics: NaicsIF
   businessType: BusinessTypes
   businessNumber?: string // GPs only
+  businessTypeConfirm?: boolean
```
severinbeauvais, line 18: As this is no longer an optional field, please remove the `?`.

src/mixins/filing-template-mixin.ts
severinbeauvais: I unresolved this because my 2 points above have not been addressed.

src/mixins/filing-template-mixin.ts
```diff
@@ -414,6 +415,9 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
     // restore Business Number
     this.setRegistrationBusinessNumber(draftFiling.registration.business.taxId || null)
 
+    // restore Business Type Confirm
+    this.setRegistrationBusinessTypeConfirm(draftFiling.registration.businessTypeConfirm || null)
+
```
severinbeauvais, line 420: This code assigns null to the store value if the draft filing doesn't have the property (for example, in a new draft). I'm thinking assigning false would make more sense. What do you think?

src/store/state/state-model.ts
severinbeauvais: Looks good

src/views/Registration/RegistrationDefineBusiness.vue
```diff
@@ -167,8 +180,10 @@ export default class RegistrationDefineBusiness extends Mixins(CommonMixin) {
   @Action setRegistrationDefineBusinessValid!: ActionBindingIF
   @Action setFolioNumber!: ActionBindingIF
   @Action setRegistrationBusinessNumber!: ActionBindingIF
+  @Action setRegistrationBusinessTypeConfirm!: ActionBindingIF
 
   // local variables
+  businessTypeConfirmValid = false
```
severinbeauvais, line 186: This is good.

I just wanted to mention: do you remember when we looked at this initial value together and it didn't seem to be used? You definitely need to initialize local variables so that Vue makes them reactive. You cannot assign undefined here, but null would be OK. But false is even better as it sets the type of this local variable. An alternative would be `businessTypeConfirmValid: boolean = null` so that it has "no value" until something else sets it, but honestly the current code looks cleaner to me.

tests/unit/filing-template-mixin.spec.ts
```diff
@@ -96,6 +96,7 @@ describe('Registration Filing', () => {
     store.state.stateModel.registration.businessType = 'SP' // not DBA
 
     store.state.stateModel.registration.businessNumber = '111222333'
+    store.state.stateModel.registration.businessTypeConfirm = null
```
severinbeauvais, line 99: So what value do we want to initialize this to...
- null, meaning it has no value (is neither true nor false)?
- false, since that's the default for the checkbox (ie, not checked)? (and is also the initial value in the store state object)

Another test case could check that businessTypeConfirm in the store gets a legitimate value (true or false) when loading a sparse registration filing.

src/mixins/filing-template-mixin.ts
Lcyc29: I think setting it to false makes sense, so that when draftFiling.registration.businessTypeConfirm is null (or false), then it is false. It is kind of consistent with state-model initial value and readonly businessTypeConfirm!: boolean (line 40) in BusinessTypeConfirm.vue.

tests/unit/filing-template-mixin.spec.ts
Lcyc29: This has to be null if we use draftFiling.registration.businessTypeConfirm || null in the filing mixin, because the draft registration doesn't have the key businessTypeConfirm and the received value in the test run is always null. If we use draftFiling.registration.businessTypeConfirm (|| false), then we can change this to false.

src/views/Registration/RegistrationDefineBusiness.vue
Lcyc29: I remember the method findIndex(f => !f), which is quite powerful to handle an array in which there is null and turn it into false.

severinbeauvais: That too but I'm referring to the time when we were looking at code together and wondering why businessTypeConfirmValid was null (in the validation array) when it's clearly initialized to false here.

Anyway, I just wanted to mention proper init here is good, even though the valid is almost immediately set by the sub-component.

src/interfaces/store-interfaces/state-interfaces/registration-state-interface.ts
Lcyc29: I removed ? in my latest commit: c77ff41.

src/mixins/filing-template-mixin.ts
Lcyc29: I changed it to || false in my latest commit: c77ff41.

tests/unit/filing-template-mixin.spec.ts
Lcyc29: I changed it to false in my latest commit: c77ff41.

src/mixins/filing-template-mixin.ts
Lcyc29: I changed it to || false in my latest commit: c77ff41.

severinbeauvais (approved):
Looks good.

Unit tests will follow in a future PR.

PR #384, opened 2022-04-22
9014/11316 HTML tags show up in the warning popup window/Create UI - CP Dissolution - Update text

Lcyc29:
Issue #: [/bcgov/entity#9014](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/9014)

Description of changes:
When a payment error occurred, a popup window for "Unable to process payment" show up.
tags are included in the text in the popup window, and they need to be removed.
- [x] Create UI 

Issue #: [/bcgov/entity#11316](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/11316)

Description of changes:
Update help text for uploading mineo docs to advise people to leave a 1-inch margin at the top of their documents.
- [x] Voluntary dissolution, affidavit
- [x] Co-op incorp, memorandum
- [x] Co-op incorp, rules

Business Type Confirm is now saved to a draft filing in the client side

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "3.5.21",
+  "version": "3.5.22",
   "private": true,
   "appName": "Create UI",
   "sbcName": "SBC Common Components",
--- a/src/components/Dissolution/CompleteAffidavit.vue
+++ b/src/components/Dissolution/CompleteAffidavit.vue
@@ -125,6 +125,11 @@
           </li>
           <li class="mt-1">
             <v-icon>mdi-circle-small</v-icon>
+            <span class="ml-2">Allow a minimum 1.5" margin at the top of the first page
+              (for BC Registries certified stamp)</span>
+          </li>
+          <li class="mt-1">
+            <v-icon>mdi-circle-small</v-icon>
             <span class="ml-2">Use a white background and a legible font with contrasting
               font colour</span>
           </li>
--- a/src/components/Dissolution/CompleteAffidavit.vue
+++ b/src/components/Dissolution/CompleteAffidavit.vue
@@ -120,8 +120,7 @@
         <ul class="mt-5">
           <li class="mt-1">
             <v-icon>mdi-circle-small</v-icon>
-            <span class="ml-2">Must be set to fit onto 8.5" x 11" letter-size paper
-            </span>
+            <span class="ml-2">Must be set to fit onto 8.5" x 11" letter-size paper</span>
           </li>
           <li class="mt-1">
             <v-icon>mdi-circle-small</v-icon>
--- a/src/dialogs/PaymentErrorDialog.vue
+++ b/src/dialogs/PaymentErrorDialog.vue
@@ -7,7 +7,7 @@
         <!-- display common message -->
         <div class="font-15" v-if="!isRoleStaff">
           <p>We are unable to process your payment at this time. This {{filingName}} has been saved
-            as a DRAFT and you can retry your payment from the Business Dashboard at a later time.</p>
+            as a DRAFT, you can retry your payment from your Business Registry dashboard at a later time.</p>
         </div>
 
         <!-- display generic message (no errors or warnings) -->
--- a/src/dialogs/PaymentErrorDialog.vue
+++ b/src/dialogs/PaymentErrorDialog.vue
@@ -28,11 +28,10 @@
         <!-- display errors -->
         <div class="font-15 mb-4" v-if="numErrors > 0">
           <p>We were unable to process your payment due to the following errors:</p>
-          <ul>
-            <li v-for="(error, index) in formatedError" :key="index">
-              <span v-html="error" id="formated-error-msg"></span>
-            </li>
-          </ul>
+          <div v-for="(error, index) in errors" :key="index" class="d-flex">
+            <span class="flex-shrink-0 bullet"><v-icon>mdi-circle-medium</v-icon></span>
+            <span class="flex-grow-1" v-html="error.message" />
+          </div>
         </div>
 
         <!-- display warnings-->
@@ -40,7 +39,7 @@
           <p>Please note the following warnings:</p>
           <ul>
             <li v-for="(warning, index) in warnings" :key="index">
-              {{warning.message}}
+              {{ warning.warning }}
             </li>
           </ul>
         </div>
@@ -103,21 +102,16 @@ export default class PaymentErrorDialog extends Vue {
   get numWarnings (): number {
     return this.warnings?.length || 0
   }
-
-  /** The formated error message */
-  protected formatedError = []
-  protected mounted (): void {
-    this.formatedError = this.errors.map((item:any) => (
-      item.message.split('<br/>').map((item2:string, index2:number) => {
-        return index2 === 0 ? item2 + '<br/>' : '<span>' + item2 + '</span><br/>'
-      }).join('')
-    ))
-  }
 }
 </script>
 
 <style scoped>
-  #formated-error-msg >>> span {
-    margin-left: 22px
+  .flex-shrink-0.bullet {
+    padding-left: 1.4px;
+    padding-right: 5px;
+  }
+
+  ::v-deep .v-icon {
+    font-size: 18px;
   }
 </style>
--- a/src/dialogs/PaymentErrorDialog.vue
+++ b/src/dialogs/PaymentErrorDialog.vue
@@ -62,7 +62,7 @@
 </template>
 
 <script lang="ts">
-import { Component, Vue, Prop, Emit, Watch } from 'vue-property-decorator'
+import { Component, Vue, Prop, Emit } from 'vue-property-decorator'
 import { Getter } from 'vuex-class'
 import RegistriesContactInfo from '@/components/common/RegistriesContactInfo.vue'
 
--- a/src/dialogs/PaymentErrorDialog.vue
+++ b/src/dialogs/PaymentErrorDialog.vue
@@ -29,8 +29,8 @@
         <div class="font-15 mb-4" v-if="numErrors > 0">
           <p>We were unable to process your payment due to the following errors:</p>
           <div v-for="(error, index) in errors" :key="index" class="d-flex">
-            <span id="error-bullet" class="flex-shrink-0">
-              <v-icon id="error-circle">mdi-circle-medium</v-icon>
+            <span class="flex-shrink-0 error-bullet">
+              <v-icon class="error-circle">mdi-circle-medium</v-icon>
             </span>
             <span class="flex-grow-1" v-html="error.message" />
           </div>
@@ -108,12 +108,12 @@ export default class PaymentErrorDialog extends Vue {
 </script>
 
 <style scoped>
-  #error-bullet {
+  .error-bullet {
     padding-left: 1.4px;
     padding-right: 5px;
   }
 
-  #error-circle {
+  .error-circle {
     font-size: 18px;
   }
 </style>
--- a/src/dialogs/PaymentErrorDialog.vue
+++ b/src/dialogs/PaymentErrorDialog.vue
@@ -29,9 +29,7 @@
         <div class="font-15 mb-4" v-if="numErrors > 0">
           <p>We were unable to process your payment due to the following errors:</p>
           <div v-for="(error, index) in errors" :key="index" class="d-flex">
-            <span class="flex-shrink-0 error-bullet">
-              <v-icon class="error-circle">mdi-circle-medium</v-icon>
-            </span>
+            <span class="flex-shrink-0"><v-icon>mdi-circle-medium</v-icon></span>
             <span class="flex-grow-1" v-html="error.message" />
           </div>
         </div>
@@ -40,9 +38,7 @@
         <div class="font-15 mb-4" v-if="numWarnings > 0">
           <p>Please note the following warnings:</p>
           <div v-for="(warning, index) in warnings" :key="index" class="d-flex">
-            <span class="flex-shrink-0 error-bullet">
-              <v-icon class="error-circle">mdi-circle-medium</v-icon>
-            </span>
+            <span class="flex-shrink-0"><v-icon>mdi-circle-medium</v-icon></span>
             <span class="flex-grow-1" v-html="warning.message" />
           </div>
         </div>
@@ -107,14 +103,3 @@ export default class PaymentErrorDialog extends Vue {
   }
 }
 </script>
-
-<style scoped>
-  .error-bullet {
-    padding-left: 1.4px;
-    padding-right: 5px;
-  }
-
-  .error-circle {
-    font-size: 18px;
-  }
-</style>
--- a/src/dialogs/PaymentErrorDialog.vue
+++ b/src/dialogs/PaymentErrorDialog.vue
@@ -29,7 +29,9 @@
         <div class="font-15 mb-4" v-if="numErrors > 0">
           <p>We were unable to process your payment due to the following errors:</p>
           <div v-for="(error, index) in errors" :key="index" class="d-flex">
-            <span class="flex-shrink-0 bullet"><v-icon>mdi-circle-medium</v-icon></span>
+            <span id="error-bullet" class="flex-shrink-0">
+              <v-icon id="error-circle">mdi-circle-medium</v-icon>
+            </span>
             <span class="flex-grow-1" v-html="error.message" />
           </div>
         </div>
@@ -39,7 +41,7 @@
           <p>Please note the following warnings:</p>
           <ul>
             <li v-for="(warning, index) in warnings" :key="index">
-              {{ warning.warning }}
+              {{ warning.message }}
             </li>
           </ul>
         </div>
@@ -106,12 +108,12 @@ export default class PaymentErrorDialog extends Vue {
 </script>
 
 <style scoped>
-  .flex-shrink-0.bullet {
+  #error-bullet {
     padding-left: 1.4px;
     padding-right: 5px;
   }
 
-  ::v-deep .v-icon {
+  #error-circle {
     font-size: 18px;
   }
 </style>
--- a/src/dialogs/PaymentErrorDialog.vue
+++ b/src/dialogs/PaymentErrorDialog.vue
@@ -29,8 +29,8 @@
         <div class="font-15 mb-4" v-if="numErrors > 0">
           <p>We were unable to process your payment due to the following errors:</p>
           <ul>
-            <li v-for="(error, index) in errors[0].message.split('<br/>')" :key="index">
-              {{ error }}
+            <li v-for="(error, index) in formatedError" :key="index">
+              <span v-html="error" id="formated-error-msg"></span>
             </li>
           </ul>
         </div>
@@ -63,7 +63,7 @@
 </template>
 
 <script lang="ts">
-import { Component, Vue, Prop, Emit } from 'vue-property-decorator'
+import { Component, Vue, Prop, Emit, Watch } from 'vue-property-decorator'
 import { Getter } from 'vuex-class'
 import RegistriesContactInfo from '@/components/common/RegistriesContactInfo.vue'
 
@@ -103,5 +103,21 @@ export default class PaymentErrorDialog extends Vue {
   get numWarnings (): number {
     return this.warnings?.length || 0
   }
+
+  /** The formated error message */
+  protected formatedError = []
+  protected mounted (): void {
+    this.formatedError = this.errors.map((item:any) => (
+      item.message.split('<br/>').map((item2:string, index2:number) => {
+        return index2 === 0 ? item2 + '<br/>' : '<span>' + item2 + '</span><br/>'
+      }).join('')
+    ))
+  }
 }
 </script>
+
+<style scoped>
+  #formated-error-msg >>> span {
+    margin-left: 22px
+  }
+</style>
--- a/src/dialogs/PaymentErrorDialog.vue
+++ b/src/dialogs/PaymentErrorDialog.vue
@@ -29,8 +29,8 @@
         <div class="font-15 mb-4" v-if="numErrors > 0">
           <p>We were unable to process your payment due to the following errors:</p>
           <ul>
-            <li v-for="(error, index) in errors" :key="index">
-              {{error.message}}
+            <li v-for="(error, index) in errors[0].message.split('<br/>')" :key="index">
+              {{ error }}
             </li>
           </ul>
         </div>
--- a/src/dialogs/PaymentErrorDialog.vue
+++ b/src/dialogs/PaymentErrorDialog.vue
@@ -39,11 +39,12 @@
         <!-- display warnings-->
         <div class="font-15 mb-4" v-if="numWarnings > 0">
           <p>Please note the following warnings:</p>
-          <ul>
-            <li v-for="(warning, index) in warnings" :key="index">
-              {{ warning.message }}
-            </li>
-          </ul>
+          <div v-for="(warning, index) in warnings" :key="index" class="d-flex">
+            <span class="flex-shrink-0 error-bullet">
+              <v-icon class="error-circle">mdi-circle-medium</v-icon>
+            </span>
+            <span class="flex-grow-1" v-html="warning.message" />
+          </div>
         </div>
 
         <template v-if="!isRoleStaff">
--- a/tests/unit/PaymentErrorDialog.spec.ts
+++ b/tests/unit/PaymentErrorDialog.spec.ts
@@ -107,6 +107,38 @@ describe('Payment Error Dialog', () => {
     )
     expect(wrapper.findAll('p').at(2).text()).toContain('If this error persists')
     expect(wrapper.findAll('li').length).toBe(0)
+    expect(wrapper.findAll('span').length).toBe(2)
+    expect(wrapper.findAll('span').at(1).text()).toContain('Your account is in the 3 day PAD confirmation period.')
+
+    expect(wrapper.find(RegistriesContactInfo).exists()).toBe(true)
+    expect(wrapper.find('#dialog-exit-button').exists()).toBe(true)
+
+    wrapper.destroy()
+  })
+
+  it('renders PAD warning messages correctly when they are present', () => {
+    store.state.stateModel.tombstone.authRoles = ['edit', 'view']
+    const wrapper = shallowMount(PaymentErrorDialog,
+      {
+        vuetify,
+        store,
+        propsData: { dialog: true, warnings: [
+          { warning: 'Test Warning 1' },
+          { warning: 'Test Warning 2' }
+        ]}
+      })
+
+    expect(wrapper.attributes('contentclass')).toBe('payment-error-dialog')
+    expect(wrapper.isVisible()).toBe(true)
+    expect(wrapper.find('#dialog-title').text()).toBe('Unable to Process Payment')
+    expect(wrapper.findAll('p').length).toBe(3)
+    expect(wrapper.findAll('p').at(0).text()).toContain('We are unable to process your payment')
+    expect(wrapper.findAll('p').at(1).text()).toContain('Please note the following warnings')
+    expect(wrapper.findAll('li').length).toBe(2)
+    expect(wrapper.findAll('li').at(0).text()).toBe('Test Warning 1')
+    expect(wrapper.findAll('li').at(1).text()).toBe('Test Warning 2')
+    expect(wrapper.find('#dialog-exit-button').exists()).toBe(true)
+    expect(wrapper.find('#dialog-okay-button').exists()).toBe(false)
 
     expect(wrapper.find(RegistriesContactInfo).exists()).toBe(true)
     expect(wrapper.find('#dialog-exit-button').exists()).toBe(true)
--- a/tests/unit/PaymentErrorDialog.spec.ts
+++ b/tests/unit/PaymentErrorDialog.spec.ts
@@ -123,8 +123,8 @@ describe('Payment Error Dialog', () => {
         vuetify,
         store,
         propsData: { dialog: true, warnings: [
-          { warning: 'Test Warning 1' },
-          { warning: 'Test Warning 2' }
+          { message: 'Test Warning 1' },
+          { message: 'Test Warning 2' }
         ]}
       })
 
--- a/tests/unit/PaymentErrorDialog.spec.ts
+++ b/tests/unit/PaymentErrorDialog.spec.ts
@@ -134,9 +134,12 @@ describe('Payment Error Dialog', () => {
     expect(wrapper.findAll('p').length).toBe(3)
     expect(wrapper.findAll('p').at(0).text()).toContain('We are unable to process your payment')
     expect(wrapper.findAll('p').at(1).text()).toContain('Please note the following warnings')
-    expect(wrapper.findAll('li').length).toBe(2)
-    expect(wrapper.findAll('li').at(0).text()).toBe('Test Warning 1')
-    expect(wrapper.findAll('li').at(1).text()).toBe('Test Warning 2')
+    
+    expect(wrapper.findAll('li').length).toBe(0)
+    expect(wrapper.findAll('span').length).toBe(4)
+    expect(wrapper.findAll('span').at(1).text()).toContain('Test Warning 1')
+    expect(wrapper.findAll('span').at(3).text()).toContain('Test Warning 2')
+
     expect(wrapper.find('#dialog-exit-button').exists()).toBe(true)
     expect(wrapper.find('#dialog-okay-button').exists()).toBe(false)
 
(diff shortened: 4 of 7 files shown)


src/dialogs/PaymentErrorDialog.vue
```diff
@@ -29,8 +29,8 @@
         <div class="font-15 mb-4" v-if="numErrors > 0">
           <p>We were unable to process your payment due to the following errors:</p>
           <ul>
-            <li v-for="(error, index) in errors" :key="index">
-              {{error.message}}
+            <li v-for="(error, index) in errors[0].message.split('<br/>')" :key="index">
+              {{ error }}
```
Lcyc29, line 33: This part the fix for [9014](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/9014): HTML tags show up in the warning popup window.

src/components/Dissolution/CompleteAffidavit.vue
```diff
@@ -123,6 +123,11 @@
             <span class="ml-2">Must be set to fit onto 8.5" x 11" letter-size paper
             </span>
           </li>
+          <li class="mt-1">
+            <v-icon>mdi-circle-small</v-icon>
+            <span class="ml-2">Allow a minimum 1.5" margin at the top of the first page
+              (for BC Registries certified stamp)</span>
+          </li>
```
Lcyc29, line 130: This li tag and the same ones below are for the ticket [11316](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/11316): Create UI - CP Dissolution - Update text.

src/components/Dissolution/CompleteAffidavit.vue
```diff
@@ -123,6 +123,11 @@
             <span class="ml-2">Must be set to fit onto 8.5" x 11" letter-size paper
```
severinbeauvais, line 123: Could you fix this while you're here? Span tags should be one the same line since spans are inline elements.

Rationale: With the closing tag on a separate line, there's actually a newline at the end of this text (even if it doesn't display).

src/dialogs/PaymentErrorDialog.vue
severinbeauvais: This change will show each line as a separate bullet.

I think the intent is to show a formatted line... you could use the `v-html` directive for that. I'm not sure if the lines will all align correctly so this needs to be tested.

severinbeauvais: ^^ Check with UI/IX on both options.

vysakh-menon-aot (approved):

src/dialogs/PaymentErrorDialog.vue
argush3: yeah, I think we should be using `v-html` for this.  There could be potentially other types of html tags coming back from pay api too.

Lcyc29: image

Lcyc29: I couldn't find the proper UI/UX for this page, and I wasn't able to test this on my local environment.

argush3: I think `v-html` looks ok.  The only thing is I would probably add styling bullet text indents after the first line.

i.e.
image

Lcyc29: image

If we were to use v-html, then the backend needs to send back the error string in this way specifically for the frontend to render the v-html with .error-span class (or some other name).

Lcyc29: The current message is the following:
"An error occurred during the BC Online transaction.<br/>Please contact the help desk to resolve this issue.<br/>SERVCE BC HELP DESK:<br/>Toll-free:1800-663-6102 (Canada and USA only)<br/>Email:bcolhelp@gov.bc.ca."

The customized message to render the v-html above is like this:
'An error occurred during the BC Online transaction.<br/><span class="error-span">Please contact the help desk to resolve this issue.</span><br/><span class="error-span">SERVCE BC HELP DESK:</span><br/><span class="error-span">Toll-free:1800-663-6102 (Canada and USA only)</span><br/><span class="error-span">Email:bcolhelp@gov.bc.ca.</span>'

severinbeauvais: We do not want to change the BE to wrap each line in a span with a class.

You're trying to use the span to get the alignment correct, right? There's probably another way that works with your latest example message.

severinbeauvais: Worse case, you could change each list item to a flex box and use shrink/grow so the bullet is in one "column" and the text is in another.

Example (left side of this diff): https://github.com/bcgov/business-filings-ui/pull/315/commits/afa6033cdde3e7e5e0ed0ca0c3a48167278487bf

severinbeauvais: @Lcyc29 As there is no design doc for this, ask Yui or Tracey how they would like to see this.

Lcyc29: @argush3 Thanks Argus for the feedback. Please keep giving me feedback :)

Lcyc29: image

I used a local variable to transform the error message into tags that can use css styles. It looks okay, but I am unable to test this in my local environment and the unit test returns an error showing that there is no <li> tag element.

src/dialogs/PaymentErrorDialog.vue
```diff
@@ -29,8 +29,8 @@
         <div class="font-15 mb-4" v-if="numErrors > 0">
           <p>We were unable to process your payment due to the following errors:</p>
           <ul>
-            <li v-for="(error, index) in errors" :key="index">
-              {{error.message}}
+            <li v-for="(error, index) in formatedError" :key="index">
+              <span v-html="error" id="formated-error-msg"></span>
             </li>
```
Lcyc29, line 34: Local variable after transformation.

src/dialogs/PaymentErrorDialog.vue
```diff
@@ -103,5 +103,21 @@ export default class PaymentErrorDialog extends Vue {
...
+
+  /** The formated error message */
+  protected formatedError = []
+  protected mounted (): void {
+    this.formatedError = this.errors.map((item:any) => (
+      item.message.split('<br/>').map((item2:string, index2:number) => {
+        return index2 === 0 ? item2 + '<br/>' : '<span>' + item2 + '</span><br/>'
+      }).join('')
+    ))
+  }
```
Lcyc29, line 115: Transform the local variable into the array of error that can use the css style.

severinbeauvais: Don't do this. The error message should be a single block of text and should be displayed as such.

Did you look at my suggestion to use flexbox and 2 "columns" inside it, the first for the bullet and the second for the text?

src/dialogs/PaymentErrorDialog.vue
```diff
@@ -103,5 +103,21 @@ export default class PaymentErrorDialog extends Vue {
...
 }
 </script>
+
+<style scoped>
+  #formated-error-msg >>> span {
```
severinbeauvais, line 120: The deep CSS selector should not be needed as the span is not rendered inside a Vue component.

src/dialogs/PaymentErrorDialog.vue
Lcyc29: The error message is a single text. I cannot turn columns in just the text with {{ error.messasge }}.

src/dialogs/PaymentErrorDialog.vue
severinbeauvais: Why do you need the span? The v-html can be right in the List Item element.

This is probably simpler if you change the Unordered List / List Items to something else.

src/dialogs/PaymentErrorDialog.vue
severinbeauvais: Use v-html.

Use https://vuetifyjs.com/en/styles/flex/.

src/dialogs/PaymentErrorDialog.vue
```diff
@@ -103,5 +103,21 @@ export default class PaymentErrorDialog extends Vue {
   get numWarnings (): number {
     return this.warnings?.length || 0
   }
+
+  /** The formated error message */
+  protected formatedError = []
+  protected mounted (): void {
+    this.formatedError = this.errors.map((item:any) => (
```
severinbeauvais, line 110: The local variable and code in mounted() could be simply replaced with a local getter.

src/dialogs/PaymentErrorDialog.vue
Lcyc29: image

It cannot align.

severinbeauvais: Doesn't it align if you use flexbox and create a flex item for the bullet and another flex item for the message?

argush3: i've done something like this in create ui in the past.  You would just need to update the `<v-col cols="11">` to use `v-html`.  I think it indented the bullet text properly.  But I'm assuming it can be simplified to use fewer tags.  Something like what Severin is suggesting of just using flex box.  

https://github.com/bcgov/business-create-ui/blob/81b41f9ce19ab16af1a1d3764c57adc1e318e6bb/src/components/Dissolution/CompleteResolution.vue#L274-L282

Lcyc29: The problem is that there is no html element to wrap around each text for the v-html span tag.

Lcyc29: image

This is what v-html gives. I don't know how I can do css style to plain text for which there is no html element.

severinbeauvais: @Lcyc29 Your code in Vue SFC Playground doesn't have Vuetify available to it, so you won't have the Vuetify flex helper classes either.

Lcyc29: @argush3 Thanks Argus. I ended up using something similar to yours, but your code `<v-icon>mdi-circle-small</v-icon>` helped me a lot with the icon.

src/dialogs/PaymentErrorDialog.vue
Lcyc29: This is code is removed.

src/dialogs/PaymentErrorDialog.vue
Lcyc29: This is code is removed.

src/dialogs/PaymentErrorDialog.vue
```diff
@@ -28,19 +28,18 @@
...
           <p>We were unable to process your payment due to the following errors:</p>
-          <ul>
-            <li v-for="(error, index) in errors" :key="index">
-              {{error.message}}
-            </li>
-          </ul>
+          <div v-for="(error, index) in errors" :key="index" class="d-flex">
+            <span class="flex-shrink-0 bullet"><v-icon>mdi-circle-medium</v-icon></span>
+            <span class="flex-grow-1" v-html="error.message" />
+          </div>
```
Lcyc29, line 34: Thanks Severin for helping out.

src/dialogs/PaymentErrorDialog.vue
```diff
@@ -28,19 +28,18 @@
...
         </div>
 
         <!-- display warnings-->
         <div class="font-15 mb-4" v-if="numWarnings > 0">
           <p>Please note the following warnings:</p>
           <ul>
             <li v-for="(warning, index) in warnings" :key="index">
-              {{warning.message}}
+              {{ warning.warning }}
```
Lcyc29, line 42: I checked with the test unit, and the key is warning instead of message, so I changed it to warning.warning.

src/dialogs/PaymentErrorDialog.vue
```diff
@@ -105,3 +104,14 @@ export default class PaymentErrorDialog extends Vue {
...
+<style scoped>
+  .flex-shrink-0.bullet {
+    padding-left: 1.4px;
+    padding-right: 5px;
+  }
+
+  ::v-deep .v-icon {
+    font-size: 18px;
+  }
+</style>
```
Lcyc29, line 117: Comparing with the `<li>` tag bullets, the bullet is a bit too large and the alignment is off by a little, so I used the style to make sure our custom v-icon look exactly like the `<li>` bullets.

tests/unit/PaymentErrorDialog.spec.ts
```diff
@@ -106,8 +106,7 @@ describe('Payment Error Dialog', () => {
       'We were unable to process your payment due to the following errors:'
     )
     expect(wrapper.findAll('p').at(2).text()).toContain('If this error persists')
-    expect(wrapper.findAll('li').length).toBe(1)
-    expect(wrapper.findAll('li').at(0).text()).toContain(padError[0].message)
+    expect(wrapper.findAll('li').length).toBe(0)
```
Lcyc29, line 109: There is no longer a `<li>` tag since we are replacing it with our custom v-icon.

src/dialogs/PaymentErrorDialog.vue
severinbeauvais: Are you sure the unit test isn't wrong?

tests/unit/PaymentErrorDialog.spec.ts
severinbeauvais: Could you add a test that looks for a structure or text that should be there instead?

src/dialogs/PaymentErrorDialog.vue
Lcyc29: Yes. The unit test is fine.

Lcyc29: Hmmm. It’s fine but it could be wrong. I’ll check again.

tests/unit/PaymentErrorDialog.spec.ts
Lcyc29: Sure. I’ll test the component with the flex class.

src/dialogs/PaymentErrorDialog.vue
severinbeauvais: Yes, that's what I meant 😁

tests/unit/PaymentErrorDialog.spec.ts
severinbeauvais: Could you please clone this unit test to verify that PAD warning messagess are correctly displayed?

src/dialogs/PaymentErrorDialog.vue
severinbeauvais: 1. Although you have matched the previous look exactly, these styles are now different from all other bulleted lists in this UI, which I discourage since we're trying for consistency.

2. Please have a conversation with UI/UX (Tracey or Yui) on whether these styles are OK. For one thing, I don't think we've ever used decimal or odd padding values before.

3. Please don't use the Vuetify layout classes for selecting elements. I consider these layout-only classes. One way to look at it is that you wouldn't want all "flex-shrink-0" classes to have this padding. But the classname you have assigned (bullet) is good.

4. Don't override the icon font size. That's the point of specifying a small, regular or large icon name. Or you can use small/large/etc in the [v-icon](https://vuetifyjs.com/en/api/v-icon/#props).

tests/unit/PaymentErrorDialog.spec.ts
```diff
@@ -106,8 +106,39 @@ describe('Payment Error Dialog', () => {
...
+  })
+
+  it('renders PAD warning messages correctly when they are present', () => {
+    store.state.stateModel.tombstone.authRoles = ['edit', 'view']
+    const wrapper = shallowMount(PaymentErrorDialog,
+      {
+        vuetify,
+        store,
+        propsData: { dialog: true, warnings: [
+          { warning: 'Test Warning 1' },
```
severinbeauvais, line 126: Please check the BE code (or ask a BE developer like Argus/Vysakh/Lekshmi) what the message field name is. I'm not convinced that it's "warning".

tests/unit/PaymentErrorDialog.spec.ts
```diff
@@ -106,8 +106,39 @@ describe('Payment Error Dialog', () => {
...
+    expect(wrapper.isVisible()).toBe(true)
+    expect(wrapper.find('#dialog-title').text()).toBe('Unable to Process Payment')
+    expect(wrapper.findAll('p').length).toBe(3)
+    expect(wrapper.findAll('p').at(0).text()).toContain('We are unable to process your payment')
+    expect(wrapper.findAll('p').at(1).text()).toContain('Please note the following warnings')
+    expect(wrapper.findAll('li').length).toBe(2)
+    expect(wrapper.findAll('li').at(0).text()).toBe('Test Warning 1')
+    expect(wrapper.findAll('li').at(1).text()).toBe('Test Warning 2')
+    expect(wrapper.find('#dialog-exit-button').exists()).toBe(true)
+    expect(wrapper.find('#dialog-okay-button').exists()).toBe(false)
```
Lcyc29, line 141: Added unit test for warning message. However, I am not sure it is warning.warning or warning.message. In create-ui, SaveErrorDialog.vue uses `warning.warning`, but PaymentErrorDialog.vue uses `warning.message`. They both use `saveWarnings` from the parent component App.vue. In edit-ui, both PaymentErrorDialog.vue and StaffPaymentErrorDialog.vue use`warning.message`, but SaveErrorDialog.vue uses `warning.warning`. They both use `saveWarnings` from the parent component. In all cases, there is no unit test for the warning message in the test file `PaymentErrorDialog.spec.ts`. Hence, I am not sure whether I should use `warning.message` or `warning.warning` for PaymentErrorDialog.

tests/unit/PaymentErrorDialog.spec.ts
Lcyc29: I added a unit test for warning messages in my last commit.

tests/unit/PaymentErrorDialog.spec.ts
Lcyc29: @argush3 Can you help me take a look in BE whether the PaymentErrorDialog returns the warning message with the key `warning` or `message`? It is `warning.warning` while others `warning.message`.

src/dialogs/PaymentErrorDialog.vue
Lcyc29: Screenshot 2022-04-25 103855

Screenshot 2022-04-25 103929

For displaying purposes, I put the error message together with the warning message. Since the warning message is using the `<li>` tags, I want to use the scoped style to make sure everything in our custom flex box `div` tag are the same as the `<li>` tag. Overriding the classes and tags are bad, and I will use `id` for these tags.

tests/unit/PaymentErrorDialog.spec.ts
severinbeauvais: If you find that some UIs are incorrect, please create a new bug ticket to fix them.

In the meantime, if you don't get an answer right away for this UI, please revert the name change and update the unit test. (Ie, if it's broken now then it stays broken, but it's not blocking you.)

src/dialogs/PaymentErrorDialog.vue
severinbeauvais: I'm not sure what you're doing but I'll review your changes.

Don't overthink this. Do follow existing code styles/practises. Do check with UI/UX team on this.

src/dialogs/PaymentErrorDialog.vue
```diff
@@ -28,19 +28,20 @@
...
         <div class="font-15 mb-4" v-if="numErrors > 0">
           <p>We were unable to process your payment due to the following errors:</p>
-          <ul>
-            <li v-for="(error, index) in errors" :key="index">
-              {{error.message}}
-            </li>
-          </ul>
+          <div v-for="(error, index) in errors" :key="index" class="d-flex">
+            <span id="error-bullet" class="flex-shrink-0">
+              <v-icon id="error-circle">mdi-circle-medium</v-icon>
```
severinbeauvais, line 33: As this template is in a for loop, it may result in duplicate IDs, which is invalid on a HTML page.

tests/unit/PaymentErrorDialog.spec.ts
severinbeauvais: Code has been reverted. If this is found to be incorrect then a future ticket will fix this.

src/dialogs/PaymentErrorDialog.vue
severinbeauvais: I unresolved this conversation because my 4 points above have not been addressed yet.

tests/unit/PaymentErrorDialog.spec.ts
severinbeauvais: I unresolved this conversation so we don't forget to answer the questions above and create bug tickets if needed.

Another way to handle this is to go ahead a create a new ticket now to ask the questions and to look at the affected UIs.

src/dialogs/PaymentErrorDialog.vue
Lcyc29: Do you want to remove the scoped style?

src/dialogs/PaymentErrorDialog.vue
Lcyc29: I will use class instead.

src/dialogs/PaymentErrorDialog.vue
Lcyc29: Here are my answers to the [questions](https://github.com/bcgov/business-create-ui/pull/384#discussion_r857845181):

1. I added the scoped style so that the bullets in both the error messages and the warning messages are the same size. Removing any of those styles will have different looks. If we want to be consistent with all other `v-icon` and `.flex-shrink-0` components, then we should remove the scoped style. By doing so, the resulted image is the second image.

2. I am still waiting for responses from the UI team. I will update this after they confirm with me.

3. I am using classes "error-bullet" and "error-circle" for the styles in my latest commit 6067d62.

4. Using small/large/etc in the [v-icon](https://vuetifyjs.com/en/api/v-icon/#props) doesn't look the same as the bullet for `<li>`. If we don't use style, then the current implementation `mdi-circle-medium` is the best size.

src/dialogs/PaymentErrorDialog.vue
severinbeauvais: Fixed.

tests/unit/PaymentErrorDialog.spec.ts
Lcyc29: Argus is checking the BE now and determine what's the correct response data.

src/dialogs/PaymentErrorDialog.vue
severinbeauvais: 1. Oh, I understand why you're trying to get both lists the same size: it's because the first list uses v-html and a MDI icon, and the second list still uses UL and LI elements. I think both lists should use v-html.

severinbeauvais: 2. If you've waited more than a day then bring this up in Standup.

Lcyc29: image

@severinbeauvais While Tracy is waiting for Scott to review the above image I sent to her, she spotted an interesting observation: "One question - I am wondering why we have contact information listed twice with 2 different sets of contact info? I know for sure we wouldn’t want that. And what is the “Please note the following warnings” information all about? It just seems like a lot of information that isn’t super helpful to the user.". That is, we are displaying the phone number, fax number and email twice in this error dialog.

tests/unit/PaymentErrorDialog.spec.ts
argush3: From looking at the code where the legal api calls the payment api to create the invoice when the filing is saved, I can see two main error paths.  I don't see anything with `warnings` in the legal api code from a quick search.

1. invoice is unable to be created due to timeout or connection issue when hitting pay api invoice endpoint.
legal api returns following response payload in this case
```
{
 ...
  "errors": [
    { 
      "message": "unable to create invoice for payment."
    }
  ]
}
```
2. pay api returns bad request response when pay api invoice endpoint invoked.
legal api returns following response payload in this case
```
{
  "errors": [
    {
      "payment_error_type": "error type…",
      "message": "error details…"
    }
  ]
}
```

Lcyc29: Thanks @argush3. We will use "message" as the key for the errors.

src/dialogs/PaymentErrorDialog.vue
Lcyc29: Comment on [1](https://github.com/bcgov/business-create-ui/pull/384#discussion_r857939481):

I agree with both should use `v-html`, but the `warnings` and 'errors' have different structures. 'warnings` is an array of dictionaries `[{message: '...'}, {message: '...'}]`, whereas 'errors' is an array of dictionaries with `[{message: '...', payment_error_type: '...' }]. It is the errors `message` key that contains stringified html, so v-html only applies to the errors, not the warnings.

severinbeauvais: The content of the messages that we receive from Pay API (in errors and warnings arrays) is out of scope for this ticket.

The presence of "following errrors" vs "following warnings" sections is out of scope for this ticket.

Please let Tracey know that if she wants to re-design this dialog and the information displayed, to please create another ticket.

severinbeauvais: > I agree with both should use `v-html`, but the `warnings` and `errors` have different structures. `warnings` is an array of dictionaries like `[{message: '...'}, {message: '...'}]`, whereas `errors` is an array of dictionaries like `[{message: '...', payment_error_type: '...' }]`. It is the errors `message` key that contains stringified html, so v-html only applies to the errors, not the warnings.

In both cases you can iterate (for loop) over the arrays and display the "message" using v-html, right?

tests/unit/PaymentErrorDialog.spec.ts
Lcyc29: @severinbeauvais Can you give feedback on the backend warnings issue?

src/dialogs/PaymentErrorDialog.vue
Lcyc29: Sure. I will use `v-html` for warnings too, in the same way as for the errors.

tests/unit/PaymentErrorDialog.spec.ts
severinbeauvais: This dialog may be using generic error handling -- not only for payment errors.

Doesn't the BE sometimes return warnings when a filing is filed?

src/dialogs/PaymentErrorDialog.vue
Lcyc29: I've changed to `v-html` for the warnings message in my last commit: 4191574.

tests/unit/PaymentErrorDialog.spec.ts
severinbeauvais: > Thanks @argush3. We will use `message` as the key for the errors. Wait a second. If there is no `warnings` from BE response, then we can use either `warning` or `message` in the warnings array. FE now uses both `warning` and `message` in create-ui, edit-ui and filing-ui for warning messages. It will probably create inconsistent bugs in the future if BE includes warnings with `warning` as the key in the response.

I appreciate this discussion as we are trying to sort out whether there is a bug here or not, but this discussion has now completely de-railed this PR and ticket.

Let's take this discussion elsewhere -> create a ticket to investigate this and resolve it if needed. In THIS PR, only fix what was requested. In other words, leave the code as `warning.message`.

Lcyc29: Thanks Severin and Argus. We will leave the code as `warning.message`.

Lcyc29:
I am ready to merge this branch.

argush3 (approved):

src/dialogs/PaymentErrorDialog.vue
severinbeauvais: This conversation should not have been closed as my items 1, 2 and 4 have not been resolved to my satisfaction.

tests/unit/PaymentErrorDialog.spec.ts
severinbeauvais: What are we doing to keep track of the possible warnings bug? Is there a ticket number?

src/dialogs/PaymentErrorDialog.vue
```diff
@@ -7,7 +7,7 @@
         <!-- display common message -->
         <div class="font-15" v-if="!isRoleStaff">
           <p>We are unable to process your payment at this time. This {{filingName}} has been saved
-            as a DRAFT and you can retry your payment from the Business Dashboard at a later time.</p>
+            as a DRAFT, you can retry your payment from your Business Registry dashboard at a later time.</p>
```
Lcyc29, line 10: This text change is from 
tlebedovich's [comment](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/9014).

tests/unit/PaymentErrorDialog.spec.ts
Lcyc29: I'm not sure. I don't know how to create a ticket. Can you give me a walkthrough for creating a ticket this issue?

src/dialogs/PaymentErrorDialog.vue
Lcyc29: Here are my answers to the items 1, 2, 3, 4:

1. I added to scoped style to make sure the bullets and texts are the same as the `<li>` tags. If we don't want them to look like the `<li>` tags, then we can remove the scoped style.

2. The UXA team has provided feedback on the error messages, but not on the bullets or the text alignment. If we don't use decimal pixels in css, there is a small pixel difference. If we don't want them to look like `<li>` tags, then we can remove the scoped style.

4. If we don't use custom style for the `v-icon` size, the bullets don't look the same as the `<li>` tags. 

Sorry. I am not sure what is the best option here. Please give advice when you are available. Thanks.

severinbeauvais: We have 2 code styles for bulleted lists in this UI (and in the others):
- lists that actually use the `li` bullet
- lists that use v-icon for the bullet

In this case, for all 3 comments, unless you have specific direction from UI/UX team, make it the same as the other bulleted lists in this UI. Hint: search globally for `mdi-circle`.

src/dialogs/PaymentErrorDialog.vue
```diff
@@ -28,21 +28,19 @@
...
           <p>We were unable to process your payment due to the following errors:</p>
-          <ul>
-            <li v-for="(error, index) in errors" :key="index">
-              {{error.message}}
-            </li>
-          </ul>
+          <div v-for="(error, index) in errors" :key="index" class="d-flex">
+            <span class="flex-shrink-0"><v-icon>mdi-circle-medium</v-icon></span>
+            <span class="flex-grow-1" v-html="error.message" />
+          </div>
```
Lcyc29, line 34: Removed class and used default css.

src/dialogs/PaymentErrorDialog.vue
```diff
@@ -28,21 +28,19 @@
...
           <p>Please note the following warnings:</p>
-          <ul>
-            <li v-for="(warning, index) in warnings" :key="index">
-              {{warning.message}}
-            </li>
-          </ul>
+          <div v-for="(warning, index) in warnings" :key="index" class="d-flex">
+            <span class="flex-shrink-0"><v-icon>mdi-circle-medium</v-icon></span>
+            <span class="flex-grow-1" v-html="warning.message" />
+          </div>
```
Lcyc29, line 43: Removed class and used default css.

src/dialogs/PaymentErrorDialog.vue
severinbeauvais: Looks good.

src/dialogs/PaymentErrorDialog.vue
severinbeauvais: Thanks for the comments. You don't have to comment all code changes; only the ones that need an explanation :)

tests/unit/PaymentErrorDialog.spec.ts
severinbeauvais: Please add new ticket number to 9014. No need to respond to this conversation. I'm closing it now.

severinbeauvais (approved):
Thanks for all the little changes. Looks great.

Lcyc29:
> Thanks for all the little changes. Looks great.

Thanks for the feedback.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "3.5.20",
+  "version": "3.5.21",
```
severinbeauvais, line 3: I'm not sure why GH isn't showing a merge conflict here, since another PR has already updated this version. (The left side of the diff shows main branch, which isn't 20 any more.)

I recommend updating the version number to 22 so we can track your change through Dev/Test/Prod environments.

PR #385, opened 2022-04-22
11877_npm_upgrade_version

ozamani9gh:
*Issue #:* /bcgov/entity#11877

*Description of changes:*
update note version from 14.17.5 to 16.14.2


severinbeauvais (approved):
Are there any READMEs to update?

Please announce to all devs of this project that they will need to use the new node version locally. (They can use [nvm](https://github.com/nvm-sh/nvm) to switch between different versions.)

severinbeauvais:
> Are there any READMEs to update?

Maybe an update to this?
https://github.com/bcgov/entity/blob/master/docs/development.md

package.json
```diff
@@ -78,7 +78,7 @@
     "vue-plugin-helper-decorator": "^0.0.12",
     "vue-property-decorator": "^8.5.1",
     "vue-template-compiler": "2.6.12",
-    "vue-test-utils-helpers": "https://github.com/bcgov/vue-test-utils-helpers.git",
+    "vue-test-utils-helpers": "github:bcgov/vue-test-utils-helpers",
```
severinbeauvais, line 81: We will have to see if the CI/CD actions are OK with this. I assume you've run the tests locally, ie, this package was retrieved correctly using this spec?

severinbeauvais (approved):

package.json
ozamani9gh: i applied a git fetch upstream, git rebase and push back to origin.

should be ready to build

severinbeauvais: Did you run `npm audit fix` here also?

severinbeauvais: Oh, I see a commit that says you did. ✔️

severinbeauvais, line 81: I'm still curious/worried about why this declaration looks like this but the one in business-filings-ui looks like this:
`"vue-test-utils-helpers": "git+https://github.com/bcgov/vue-test-utils-helpers.git",`

What do you think? Merge it and see if this works?

severinbeauvais (approved):
One last comment. But we could try merging and seeing what happens.

package.json
severinbeauvais: OK, this now looks the same as in Filings UI.

severinbeauvais (approved):
LGTM

PR #387, opened 2022-05-04
9735 SP/GP Create UI - Staff payment and update certify component

lewischenstudio:
Issue #: [/bcgov/entity#9735](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/9735)

*Description of changes:*
- ~insert staff payment component~ _already done_
- ~editable completing party in documents delivery~ _already done_
- ~update certify section to be third party, not the "I, ___ certify"~ _already done_
- [x] Added "My Business Registry" in the breadcrumb (with reference to [Invision](https://projects.invisionapp.com/share/3511UPXPMT2P#/screens/464799355))
- [x] Display Business Number for Sole Proprietors

Comment: Additional requirement to redirect "My Business Registry" to the list of businesses page with staff login.
Suggestion: new SBC AUTH ticket to work on the redirect url.

### DIFF
--- a/src/components/Registration/DefineRegistrationSummary.vue
+++ b/src/components/Registration/DefineRegistrationSummary.vue
@@ -85,7 +85,7 @@
       </template>
 
       <!-- Business Number -->
-      <template v-if="isTypePartnership">
+      <template>
         <v-divider class="mx-6" />
 
         <article class="section-container">
@@ -152,6 +152,7 @@ export default class DefineRegistrationSummary extends Mixins(DateMixin, EnumMix
 
   /** The business number. */
   get businessNumber (): string {
+    console.info('this.getRegistration: ', this.getRegistration)
     return this.getRegistration.businessNumber
   }
 
--- a/src/components/Registration/DefineRegistrationSummary.vue
+++ b/src/components/Registration/DefineRegistrationSummary.vue
@@ -85,20 +85,18 @@
       </template>
 
       <!-- Business Number -->
-      <template>
-        <v-divider class="mx-6" />
+      <v-divider class="mx-6" />
 
-        <article class="section-container">
-          <v-row no-gutters>
-            <v-col cols="12" sm="3" class="pr-4">
-              <label>Business Number</label>
-            </v-col>
-            <v-col cols="12" sm="9">
-              <span>{{ businessNumber || '(Not entered)' }}</span>
-            </v-col>
-          </v-row>
-        </article>
-      </template>
+      <article class="section-container">
+        <v-row no-gutters>
+          <v-col cols="12" sm="3" class="pr-4">
+            <label>Business Number</label>
+          </v-col>
+          <v-col cols="12" sm="9">
+            <span>{{ businessNumber || '(Not entered)' }}</span>
+          </v-col>
+        </v-row>
+      </article>
     </section>
   </div>
 </template>
--- a/src/components/Registration/DefineRegistrationSummary.vue
+++ b/src/components/Registration/DefineRegistrationSummary.vue
@@ -152,7 +152,6 @@ export default class DefineRegistrationSummary extends Mixins(DateMixin, EnumMix
 
   /** The business number. */
   get businessNumber (): string {
-    console.info('this.getRegistration: ', this.getRegistration)
     return this.getRegistration.businessNumber
   }
 
--- a/src/resources/BreadCrumbResource.ts
+++ b/src/resources/BreadCrumbResource.ts
@@ -66,3 +66,11 @@ export function getStaffDashboardBreadcrumb (): BreadcrumbIF {
     href: `${sessionStorage.getItem('BUSINESSES_URL')}staff/${getParams()}`
   }
 }
+
+export function getStaffRegistryBreadcrumb (): BreadcrumbIF {
+  const accountId = sessionStorage.getItem('ACCOUNT_ID')
+  return {
+    text: 'My Business Registry',
+    href: `${sessionStorage.getItem('BUSINESSES_URL')}account/${accountId}/business`
+  }
+}
--- a/src/resources/BreadCrumbResource.ts
+++ b/src/resources/BreadCrumbResource.ts
@@ -67,7 +67,7 @@ export function getStaffDashboardBreadcrumb (): BreadcrumbIF {
   }
 }
 
-export function getStaffMyRegistryBreadcrumb (): BreadcrumbIF {
+export function getStaffMyBusinessRegistryBreadcrumb (): BreadcrumbIF {
   const accountId = sessionStorage.getItem('ACCOUNT_ID')
   return {
     text: 'My Business Registry',
--- a/src/resources/BreadCrumbResource.ts
+++ b/src/resources/BreadCrumbResource.ts
@@ -67,7 +67,7 @@ export function getStaffDashboardBreadcrumb (): BreadcrumbIF {
   }
 }
 
-export function getStaffRegistryBreadcrumb (): BreadcrumbIF {
+export function getStaffMyRegistryBreadcrumb (): BreadcrumbIF {
   const accountId = sessionStorage.getItem('ACCOUNT_ID')
   return {
     text: 'My Business Registry',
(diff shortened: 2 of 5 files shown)


src/components/Registration/DefineRegistrationSummary.vue
```diff
@@ -85,7 +85,7 @@
       </template>
 
       <!-- Business Number -->
-      <template v-if="isTypePartnership">
+      <template>
```
severinbeauvais, line 88: Just delete this pseudo-element then. Vue uses it for conditional logic only -- it doesn't render.

src/resources/BreadCrumbResource.ts
```diff
@@ -66,3 +66,11 @@ export function getStaffDashboardBreadcrumb (): BreadcrumbIF {
     href: `${sessionStorage.getItem('BUSINESSES_URL')}staff/${getParams()}`
   }
 }
+
+export function getStaffRegistryBreadcrumb (): BreadcrumbIF {
+  const accountId = sessionStorage.getItem('ACCOUNT_ID')
```
severinbeauvais, line 71: This should be handled by `navigate()` instead.

severinbeauvais: Oh wait, this is different.

Looks fine 👍

severinbeauvais: Actually, the ideal solution might be for Auth Web to redirect from, say:
`https://dev.bcregistry.ca/business/auth/account/business?accountid=2288`
to:
`https://dev.bcregistry.ca/business/auth/account/668/business`

(It partially does this already for non-staff/IDIR logins.)

Do you want to try looking into this?

src/resources/BreadCrumbResource.ts
```diff
@@ -66,3 +66,11 @@ export function getStaffDashboardBreadcrumb (): BreadcrumbIF {
     href: `${sessionStorage.getItem('BUSINESSES_URL')}staff/${getParams()}`
   }
 }
+
+export function getStaffRegistryBreadcrumb (): BreadcrumbIF {
```
severinbeauvais, line 70: It's actually not the "Registry" page (which is https://www.bcregistry.gov.bc.ca/) but the "(My) Business Registry" page

severinbeauvais:
Please un-check the items in the description that are incomplete or not part of this Pull Request.

Also, please update the app version so that you can track your change through dev/test/prod environments.

src/resources/BreadCrumbResource.ts
lewischenstudio: Yes. I need to try the Auth Web. Because it didn't redirect properly with my way.

severinbeauvais: The reason I recommend a fix to Auth Web is because many UIs/web apps have this same breadcrumb feature, and all would need to be updated now and any time the URL changes. (The URL is otherwise provided by external config, so no rebuilding needed.)

If the problem is that Auth Web isn't going to the right business page when the user is staff/IDIR, then that should be fixed directly.

OTOH, if the PO decides to accept this temporary fix now, and we have a ticket to implement a proper fix and delete this workaround later, then I'm OK with this.

lewischenstudio: This is a good question. Yui and Tracy did suggest to work on this fix in another ticket. I can now see how this has an impact on other UIs, such as filing-ui that doesn't have "My Business Registry" either (on staff). Meanwhile, I am not sure what I need to do with Auth Web. If "My Business Registry" should appear in all UIs, then it's better to work on it in another ticket and discard any change in this ticket.

src/components/Registration/DefineRegistrationSummary.vue
```diff
@@ -85,20 +85,18 @@
       </template>
 
       <!-- Business Number -->
-      <template v-if="isTypePartnership">
-        <v-divider class="mx-6" />
+      <v-divider class="mx-6" />
```
thorwolpert, line 88: It doesn't look like any change here, other than changing indenting.
please don't make changes to formatting, for the sake of changing formatting as it detracts from understanding what the PR is attempting to do.

thorwolpert (commented):
Please don't make changes to formatting for the sake of changing formatting.

src/components/Registration/DefineRegistrationSummary.vue
severinbeauvais: Looks like a conditional was removed, so I'm OK with fixing the indentation of the block.

src/components/Registration/DefineRegistrationSummary.vue
```diff
@@ -85,20 +85,18 @@
       </template>
 
       <!-- Business Number -->
-      <template v-if="isTypePartnership">
```
severinbeauvais, line 88: Wait - why did you delete this?

src/resources/BreadCrumbResource.ts
severinbeauvais: What's the other ticket? Perhaps this needs a larger discussion... please raise it at Standup. Thx.

If we take the breadcrumbs out of this ticket, does THIS ticket need to change any code?

lewischenstudio: If we take the breadcrumbs out of this ticket, then this ticket won't have any change.

src/components/Registration/DefineRegistrationSummary.vue
lewischenstudio: According to the [Invision design](https://projects.invisionapp.com/share/3511UPXPMT2P#/screens/464799355), SP also displays BN, so I removed this condition.

src/resources/BreadCrumbResource.ts
severinbeauvais: When I am logged in as staff (IDIR), the following URL doesn't redirect to My Business Registry. Do I have the right URL?
https://dev.bcregistry.ca/business/auth/account/668/business

severinbeauvais: Ohhhhh, I think I know!

If staff start from the [staff dashboard[(https://dev.bcregistry.ca/business/auth/staff/dashboard/active) and go the Entity Dashboard or Manage Account pages then a bunch of account-related fields are set in session storage and those pages display correctly.

However, if staff directly go to a business' dashboard (eg, https://dev.bcregistry.ca/business/FM1000007) then those session variables are not set the and going to My Business Registry (eg, https://dev.bcregistry.ca/business/auth/account/668/business) doesn't work.

All this to say this is more complicated than initially believed.

And also this change has been removed from this ticket.

lewischenstudio: Yes, you are right. "My Business Registry" on the breadcrumbs can redirect to either of the two pages: staff's registry page or the user's registry page. Suppose we have it in the breadcrumbs for both cases. 
1. If a staff is accessing a user's NR and wants to return to the user's registry page, then this button should do that. 
2. If a staff is accessing his/her own NR and wants to return his/her own registry page, then this button also needs to do that. 
This button is very complicated and I feel sorry to bring it up. 

Since BN isn't required for SP, there isn't any change at all in this branch. 

I think it's safe to close this pull request.

severinbeauvais: Agreed. I'm resolving this conversation.

src/components/Registration/DefineRegistrationSummary.vue
severinbeauvais: This change is obsolete as we want to continue displaying BN only for GPs.
I'm resolving this conversation.

src/resources/BreadCrumbResource.ts
severinbeauvais: Obsolete. Resolving.

src/components/Registration/DefineRegistrationSummary.vue
severinbeauvais: This change is obsolete as we want to continue displaying BN only for GPs.
Resolving.

lewischenstudio:
- Items in the Todo list are already implemented. 
- "My Business Registry" are not required for this ticket. 
- BN is not required for SP. 

I'm closing this branch since there is no change.

PR #394, opened 2022-05-17
12263 Fixed empty tax id bug

severinbeauvais:
*Issue #:* /bcgov/entity#12263

*Description of changes:*
- deleted empty tax id so it doesn't fail schema pattern
- added unit tests
- fixed lint error
- app version = 4.0.7

### DIFF
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -396,6 +396,10 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
         }
       }
       delete party.officer.businessNumber
+
+      // if tax id is empty (which doesn't match the schema pattern), delete it
+      if (!party.officer.taxId) delete party.officer.taxId
+
       return party as PartyIF
     })
   }
--- a/tests/unit/PaymentErrorDialog.spec.ts
+++ b/tests/unit/PaymentErrorDialog.spec.ts
@@ -122,10 +122,11 @@ describe('Payment Error Dialog', () => {
       {
         vuetify,
         store,
-        propsData: { dialog: true, warnings: [
-          { message: 'Test Warning 1' },
-          { message: 'Test Warning 2' }
-        ]}
+        propsData: { dialog: true,
+          warnings: [
+            { message: 'Test Warning 1' },
+            { message: 'Test Warning 2' }
+          ] }
       })
 
     expect(wrapper.attributes('contentclass')).toBe('payment-error-dialog')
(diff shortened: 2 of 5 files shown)


vysakh-menon-aot (approved):

src/mixins/filing-template-mixin.ts
```diff
@@ -396,6 +396,10 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
         }
       }
       delete party.officer.businessNumber
+
+      // if tax id is empty (which doesn't match the schema pattern), delete it
+      if (!party.officer.taxId) delete party.officer.taxId
```
severinbeauvais, line 401: If the user did not touch the input control for this value (which is labelled Business Number) then this v-model value is null or undefined, which is fine.

But if the user did touch the input control then this v-model value is "" (empty string), which fails schema validation due to the [required pattern].(https://github.com/bcgov/business-schemas/blob/e32e6dc561abebdc3034af48993da12f882dfb80/src/registry_schemas/schemas/party.json#L46).

Therefore, the solution is to delete this property if it's empty (null, undefined or empty string).

tests/unit/PaymentErrorDialog.spec.ts
```diff
@@ -122,10 +122,11 @@ describe('Payment Error Dialog', () => {
...
         store,
-        propsData: { dialog: true, warnings: [
-          { message: 'Test Warning 1' },
-          { message: 'Test Warning 2' }
-        ]}
+        propsData: { dialog: true,
+          warnings: [
+            { message: 'Test Warning 1' },
+            { message: 'Test Warning 2' }
+          ] }
```
severinbeauvais, line 129: Lint fixes only in this file.

NB: these are auto-fixed if you do `npm run lint`.

tests/unit/filing-template-mixin.spec.ts
```diff
@@ -98,6 +97,69 @@ describe('Registration Filing', () => {
...
+      {
+        officer: {
+          id: '4321',
+          email: 'kitty@example.com',
+          lastName: '',
+          firstName: '',
+          partyType: 'organization',
+          middleName: '',
+          organizationName: 'Crazy Cat Consulting Company',
+          businessNumber: '123456789'
```
severinbeauvais, line 136: Valid business number. It should be in the filing object.

tests/unit/filing-template-mixin.spec.ts
```diff
@@ -144,7 +206,68 @@ describe('Registration Filing', () => {
...
+            {
+              officer: {
+                id: '1234',
+                email: 'lucille@bcregtest.gov.bc.ca',
+                lastName: 'TWENTY',
+                firstName: 'BCREGTEST Lucille',
+                partyType: 'person',
+                middleName: '',
+                organizationName: ''
+              },
```
severinbeauvais, line 220: No tax id 👍

tests/unit/filing-template-mixin.spec.ts
```diff
@@ -144,7 +206,68 @@ describe('Registration Filing', () => {
...
+            {
+              officer: {
+                id: '4321',
+                email: 'kitty@example.com',
+                lastName: '',
+                firstName: '',
+                partyType: 'organization',
+                middleName: '',
+                organizationName: 'Crazy Cat Consulting Company',
+                taxId: '123456789'
```
severinbeauvais, line 245: Tax id 👍

lewischenstudio (approved):

PR #397, opened 2022-05-24
Added business check before loading parties

severinbeauvais:
*Issue #:* None

*Description of changes:*
- added business check before loading parties
- app version = 4.0.9

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -630,8 +630,8 @@ export default class App extends Mixins(
         throw error // go to catch()
       }
 
-      // load parties only for SP/GP
-      this.isTypeFirm && this.loadPartiesInformation()
+      // load parties only for SP/GP businesses
+      if (this.isTypeFirm && this.getBusinessId) this.loadPartiesInformation()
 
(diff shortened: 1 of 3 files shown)


src/App.vue
```diff
@@ -630,8 +630,8 @@ export default class App extends Mixins(
         throw error // go to catch()
       }
 
-      // load parties only for SP/GP
-      this.isTypeFirm && this.loadPartiesInformation()
+      // load parties only for SP/GP businesses
+      if (this.isTypeFirm && this.getBusinessId) this.loadPartiesInformation()
```
severinbeauvais, line 634: Only existing businesses can load parties, and only existing businesses have a business ID.

(IAs and Registrations have a temp reg number instead, and cannot load parties.)

severinbeauvais:
I tested this locally before and after the fix using:
http://localhost:8080/basePath/registration-people-roles?id=TTEdGdiXGP&accountid=2443

vysakh-menon-aot (approved):

PR #398, opened 2022-05-25
Bug/12321 new

ritvick:
*Issue #:* /bcgov/entity###
https://github.com/bcgov/entity/issues/12321

*Description of changes:*
Added validation on click on "file and save"

### DIFF
--- a/src/components/common/Actions.vue
+++ b/src/components/common/Actions.vue
@@ -118,6 +118,7 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
   @Getter isSavingResuming!: boolean
   @Getter isFilingPaying!: boolean
   @Getter getDissolutionDate!: string
+  @Getter isTypeFirm!: boolean
 
   @Action setIsSaving!: ActionBindingIF
   @Action setIsSavingResuming!: ActionBindingIF
@@ -210,8 +211,10 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
   private async onClickFilePay (): Promise<void> {
     // Prompt Step validations
     this.setValidateSteps(true)
-    const date = this.getDissolutionDate
-    if (!date) this.setDissolutionDateError('Dissolution date is required')
+    if (this.isTypeFirm) {
+      const date = this.getDissolutionDate
+      if (!date) this.setDissolutionDateError('Dissolution date is required')
+    }
 
     if (this.isApplicationValid) {
       // prevent double saving
--- a/src/components/common/Actions.vue
+++ b/src/components/common/Actions.vue
@@ -126,7 +126,6 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
   @Action setHaveChanges!: ActionBindingIF
   @Action setEffectiveDateTimeValid!: ActionBindingIF
   @Action setValidateSteps!: ActionBindingIF
-  @Action setDissolutionDateError!: ActionBindingIF
 
   /** Is True if Jest is running the code. */
   get isJestRunning (): boolean {
--- a/src/components/common/Actions.vue
+++ b/src/components/common/Actions.vue
@@ -211,10 +211,6 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
   private async onClickFilePay (): Promise<void> {
     // Prompt Step validations
     this.setValidateSteps(true)
-    if (this.isTypeFirm) {
-      const date = this.getDissolutionDate
-      if (!date) this.setDissolutionDateError('Business Dissolution date is required')
-    }
 
     if (this.isApplicationValid) {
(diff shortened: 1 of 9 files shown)


src/components/common/Actions.vue
```diff
@@ -208,6 +211,10 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
   private async onClickFilePay (): Promise<void> {
     // Prompt Step validations
     this.setValidateSteps(true)
+    if (this.isTypeFirm) {
+      const date = this.getDissolutionDate
```
shabeeb-aot, line 215: I think we can make all conditions in one line and also check for blank string

src/views/DissolutionFirm/DissolutionFirm.vue
```diff
@@ -360,7 +368,7 @@ export default class DissolutionFirm extends Mixins(DateMixin) {
   /** Validations rules for start date field. */
   get startDateRules (): Array<Function> {
     return [
-      (v: string) => !!v || 'Dissolution date is required',
+      (v: string) => !!v || 'Business Dissolution date is required',
```
shabeeb-aot, line 371: Is this change for all corptype?

src/interfaces/store-interfaces/state-interfaces/dissolution-state-interface.ts
```diff
@@ -5,6 +5,7 @@ export interface DissolutionStateIF {
   dissolutionType: DissolutionTypes
   dissolutionDate: string
   dissolutionStatementStep: DissolutionStatementIF
+  dissolutionDateFieldError: string
```
severinbeauvais, line 8: It would be nicer if the property name matched the getter and action/mutation.

Re:
getDissolutionDateError
setDissolutionDateError
mutateDissolutionDateError

shabeeb-aot:
update npm version and commit package-lock

src/components/common/Actions.vue
```diff
@@ -208,6 +211,10 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
   private async onClickFilePay (): Promise<void> {
     // Prompt Step validations
     this.setValidateSteps(true)
+    if (this.isTypeFirm) {
```
severinbeauvais, line 214: This check should be in src/views/DissolutionFirm/DissolutionFirm.vue, not here.

You can see that a global flah (validateSteps) is set to True above. You can watch that flag (or just use reactivity) to display the error there.

This change will probably simply a lot of things in this PR.

src/components/common/Actions.vue
```diff
@@ -117,6 +117,8 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
   @Getter isSaving!: boolean
   @Getter isSavingResuming!: boolean
   @Getter isFilingPaying!: boolean
+  @Getter getDissolutionDate!: string
+  @Getter isTypeFirm!: boolean
```
severinbeauvais, line 121: If these are no longer needed, please delete them.

src/views/DissolutionFirm/DissolutionFirm.vue
```diff
@@ -342,6 +343,12 @@ export default class DissolutionFirm extends Mixins(DateMixin) {
     return new Date(this.getBusinessFoundingDate)
   }
 
+  /** Dissolution Error */
+  private dissolutionError (): String {
+    return this.isTypeFirm && this.getValidateSteps &&
+    !this.getDissolutionDate && 'Dissolution date is required'
+  }
```
severinbeauvais, line 350: 1. Return type should be `string` (native type) not `String` (object).
2. What does this return if any of the conditions is False? Is `errorMsg` still OK with it?

src/interfaces/store-interfaces/state-interfaces/dissolution-state-interface.ts
severinbeauvais: Obsolete comment.

src/components/common/Actions.vue
severinbeauvais: You found an even better solution 👍

src/views/DissolutionFirm/DissolutionFirm.vue
severinbeauvais: OK, that works but it's complicated to understand/review. How about a simple if-then-else instead?

severinbeauvais (approved):
Approved. (I trust you with simplifying `dissolutionError()`.)

src/views/DissolutionFirm/DissolutionFirm.vue
severinbeauvais: Your fix will probably fix the lint errors, too :)

severinbeauvais: Nice!

PR #399, opened 2022-05-27
pass filingLabelText prop to feeSummary Component

meganawong:
*Issue #:* /bcgov/entity#11660

*Description of changes:*
* Pass new filingLabelText to feeSumary component for SP/GP

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -131,6 +131,7 @@
                   <SbcFeeSummary
                     :filingData="feeFilingData"
                     :payURL="payApiUrl"
+                    :filingLabel="filingLabelText"
                   />
                 </affix>
               </aside>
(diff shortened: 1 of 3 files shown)


src/App.vue
```diff
@@ -384,6 +386,10 @@ export default class App extends Mixins(
       this.fileAndPayInvalidNameRequestDialog
     )
   }
+  /** The Fee Summary filing text for firm businesses. */
+  get filingLabelText (): string {
+    return this.isTypeFirm && 'Dissolution'
```
severinbeauvais, line 391: It would be better to use a ternary expression or if-then-else because this may return `false`.

Since this getter is supposed to return a string, acceptable "empty" return values are `''` or `null`.

severinbeauvais, line 391: What's the default filing fee label here?

Why not create a new fee code with the desired label?

meganawong: Only a label change for SP/GP, fee code should stay the same and be treated as a voluntary dissolution as per Jyoti's instruction.

meganawong: fixed

severinbeauvais (approved):
Looks good.

PR #402, opened 2022-05-31
Bug: 12384

ritvick:
*Issue #:* https://github.com/bcgov/entity/issues/12384

*Description of changes:*

Fixed the Date Mixin method that returned an invalid value if the user is not in Pacific timezone.

### DIFF
--- a/src/mixins/date-mixin.ts
+++ b/src/mixins/date-mixin.ts
@@ -46,10 +46,12 @@ export default class DateMixin extends Mixins(CommonMixin) {
   createUtcDate (year: number, month: number, day: number, hours: number = 0, minutes: number = 0): Date {
     // use date from server to create a new date in Pacific timezone
     // (this sets the correct tz offset in the new date)
-    const date = new Date(this.getCurrentJsDate.toLocaleString('en-US', { timeZone: 'America/Vancouver' }))
 
-    // update all date and time fields
-    date.setFullYear(year, month, day)
+    let date = new Date(Date.UTC(year, month, day, hours, minutes))
+    let utcDate = new Date(date.toLocaleString('en-US', { timeZone: 'UTC' }))
+    let tzDate = new Date(date.toLocaleString('en-US', { timeZone: 'America/vancouver' }))
+    let offset = utcDate.getTime() - tzDate.getTime()
+    date.setTime(date.getTime() + offset)
 
     return date
   }
--- a/src/mixins/date-mixin.ts
+++ b/src/mixins/date-mixin.ts
@@ -50,7 +50,6 @@ export default class DateMixin extends Mixins(CommonMixin) {
 
     // update all date and time fields
     date.setFullYear(year, month, day)
-    date.setHours(hours, minutes, 0, 0) // zero out seconds and milliseconds
 
     return date
   }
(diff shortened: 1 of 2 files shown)


severinbeauvais (commented):
Whatever solution is found here will also need to be committed against the other UIs.

src/mixins/date-mixin.ts
```diff
@@ -46,12 +46,15 @@ export default class DateMixin extends Mixins(CommonMixin) {
   createUtcDate (year: number, month: number, day: number, hours: number = 0, minutes: number = 0): Date {
     // use date from server to create a new date in Pacific timezone
     // (this sets the correct tz offset in the new date)
```
severinbeauvais, line 48: Please delete these obsolete comments.

If you can add another helpful comment explaining what this does (or why it works), that would be great. Or, paste a reference to a SO response.

tests/unit/date-mixin.spec.ts
```diff
@@ -46,7 +46,7 @@ describe('Date Mixin', () => {
 
     expect(vm.yyyyMmDdToDate(null)).toBeNull()
     expect(vm.yyyyMmDdToDate('12345678901')).toBeNull()
-    expect(vm.yyyyMmDdToDate('2021-01-01')).toEqual(new Date(2021, 0, 1)) // PST
-    expect(vm.yyyyMmDdToDate('2021-07-01')).toEqual(new Date(2021, 6, 1)) // PDT
+    expect(vm.yyyyMmDdToDate('2021-01-01').toISOString()).toEqual('2021-01-01T08:00:00.000Z') // PST
+    expect(vm.yyyyMmDdToDate('2021-07-01').toISOString()).toEqual('2021-07-01T07:00:00.000Z') // PDT
```
severinbeauvais, line 50: I was originally thinking this should check explicitly for PST/PDT, but this is better because ISO string is a standard format while "toString()" is locale-dependent. 👍

src/mixins/date-mixin.ts
```diff
@@ -46,12 +46,15 @@ export default class DateMixin extends Mixins(CommonMixin) {
...
-    const date = new Date(this.getCurrentJsDate.toLocaleString('en-US', { timeZone: 'America/Vancouver' }))
```
severinbeauvais, line 49: Since `getCurrentDate` is no longer needed, could you please delete line 15 and 2?

src/mixins/date-mixin.ts
```diff
@@ -46,12 +46,15 @@ export default class DateMixin extends Mixins(CommonMixin) {
...
-    date.setFullYear(year, month, day)
-    date.setHours(hours, minutes, 0, 0) // zero out seconds and milliseconds
+    let date = new Date(Date.UTC(year, month, day, hours, minutes))
+    let utcDate = new Date(date.toLocaleString('en-US', { timeZone: 'UTC' }))
+    let tzDate = new Date(date.toLocaleString('en-US', { timeZone: 'America/vancouver' }))
+    let offset = utcDate.getTime() - tzDate.getTime()
+    date.setTime(date.getTime() + offset)
 
+    console.log(date.toISOString());
+    console.log(new Date(2021, 0, 1));
```
severinbeauvais, line 57: Delete these :)

src/mixins/date-mixin.ts
severinbeauvais: Maybe this?
```
    // 1. create the new date in UTC
    // 2. compute the offset between UTC and Vancouver
    // 3. add the offset to convert the date to Vancouver
```

severinbeauvais (approved):
OK to merge when test errors are fixed.

PR #410, opened 2022-06-15
12570 Add transactional folio number to SP/GP registration filings

lambert-alex:
*Issue #:* /bcgov/entity#12570

*Description of changes:*
SP/GP Registration filings:
1. Added Transactional Folio Number component in the Review and Confirm view for premium users.
2. Added unit tests for Registration Review Confirm view as there was no tests for it.
3. Remove business folio number for SP/GP registrations
4. Version number = 4.0.20


vysakh-menon-aot (approved):

severinbeauvais (commented):
The code looks great!

Is anything needed to save the TFN in the filing and to possibly parse it from a resumed filing? See https://github.com/bcgov/business-create-ui/blob/cf94573b2ae02982b643acd10123380019000832/src/mixins/filing-template-mixin.ts#L604.

lambert-alex:
> The code looks great!
> 
> Is anything needed to save the TFN in the filing and to possibly parse it from a resumed filing? See
> 
> https://github.com/bcgov/business-create-ui/blob/cf94573b2ae02982b643acd10123380019000832/src/mixins/filing-template-mixin.ts#L604
> 
> .

I will check I don't know.

lewischenstudio (approved):
It's looking good and clean!

src/views/Registration/RegistrationDefineBusiness.vue
```diff
@@ -119,26 +119,6 @@
         />
       </v-card>
     </section>
-
```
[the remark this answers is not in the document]
severinbeauvais: The folio number on this page is the "account" folio number that the Legal API will use in filings from now on (same as the folio number on the Edit Profile page for a BCOL account)...
Unless the user enters a Transactional Registration Number, which would be used only for THIS filing.

So I think you need both. Well actually, what do the other filings do?

src/interfaces/filing-interfaces/filing-interfaces.ts
```diff
@@ -78,6 +78,7 @@ export interface RegistrationFilingIF {
     effectiveDate?: string // not saved by UI but may be provided by API
     filingId?: number // not saved by UI but may be provided by API
     folioNumber?: string // optional and only displayed for certain account types
+    isTransactionalFolioNumber?: boolean // Optional - applies to registration Premium accounts only
```
lambert-alex, line 81: I am not sure if the legal api supports this, from my local tests it seems to. I looked through the legal api and I couldn't tell if it did or not.

severinbeauvais: The Legal API will save the filing JSON (ie, this object) verbatim in the filings table.

Whether the Filer looks at this property and does something special with it, I don't know. It could be tested by setting a TFN, filing a registration, and then looking at the receipt. (And subsequently filing a change with no TFN and confirming this original TFN is not used either, ie, it really was for the registration only.)

severinbeauvais, line 81: According to filing-template-mixin below, this property is used for dissolution premium accounts also.

lambert-alex: So the legal API only acts to save the JSON and doesn't actually care what is the contents of the object are?

lambert-alex: This property is already implemented for dissolution premium accounts. 

> https://github.com/bcgov/business-create-ui/blob/cf94573b2ae02982b643acd10123380019000832/src/interfaces/filing-interfaces.ts#L102

severinbeauvais: The Legal API only cares about the objects only when the filing is filed (draft = false), at which point it does schema and some custom validation.

severinbeauvais: Oh, right 😊 

Then it's redundant to say it only applies to one type of filing when these objects are already specific to that filing type 😉

src/interfaces/filing-interfaces/filing-interfaces.ts
```diff
@@ -78,6 +78,7 @@ export interface RegistrationFilingIF {
     effectiveDate?: string // not saved by UI but may be provided by API
     filingId?: number // not saved by UI but may be provided by API
     folioNumber?: string // optional and only displayed for certain account types
+    isTransactionalFolioNumber?: boolean // Optional - applies dissolution and registration. Premium accounts only
```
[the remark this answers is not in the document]
lambert-alex: Okay ya that makes sense otherwise it'll be confusing.

PR #407, opened 2022-06-13
Fix dissolution validations 11736

lambert-alex:
*Issue #:* /bcgov/entity[11736](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/11736)

*Description of changes:*
For both:
1. Removed shake animation from confirm sections as it was the only place with a shake animation.

For BEN dissolutions:
1. Complete affidavit step does not show red styling/label until the review and confirm step has been visited.
3. The complete resolution step do not show red styling/label until review and confirm step has been visited.

For CO dissolutions:
1. Upload affidavit step does not show red styling/label until the review and confirm step has been visited .
2. The complete special resolution step do not show red styling/label until review and confirm step has been visited.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the bcrs-entities-create-ui license (Apache 2.0).

PR #415, opened 2022-06-23
Change verification to show step icons as incomplete 11736

lambert-alex:
*Issue #:* /bcgov/entity#11736

*Description of changes:*

Fix for incomplete pull request #407
 
- Step icons are now marked incomplete after visiting the Review and Confirm step
-  App version = 4.0.21

### DIFF
--- a/src/components/common/Stepper.vue
+++ b/src/components/common/Stepper.vue
@@ -25,7 +25,7 @@
           <v-icon class="step__btn2" size="30" color="success darken-1" v-show="isValid(step.to)">
             mdi-check-circle
           </v-icon>
-          <v-icon class="step__btn2" size="30" color="error darken-1" v-show="!isValid(step.to) && getValidateSteps">
+          <v-icon class="step__btn2" size="30" color="error darken-1" v-show="!isValid(step.to) && getShowErrors">
             mdi-close-circle
           </v-icon>
         </div>
@@ -55,8 +55,8 @@ import { RegistrationStateIF } from '@/interfaces'
 @Component({})
 export default class Stepper extends Vue {
   @Getter getRegistration!: RegistrationStateIF
+  @Getter getShowErrors!: boolean
   @Getter getSteps!: Array<any>
-  @Getter getValidateSteps!: boolean
   @Getter isAddPeopleAndRolesValid!: boolean
   @Getter isAffidavitValid!: boolean
   @Getter isBusySaving!: boolean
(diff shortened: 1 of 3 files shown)


vysakh-menon-aot (approved):

src/components/common/Stepper.vue
```diff
@@ -25,7 +25,7 @@
           <v-icon class="step__btn2" size="30" color="success darken-1" v-show="isValid(step.to)">
             mdi-check-circle
           </v-icon>
-          <v-icon class="step__btn2" size="30" color="error darken-1" v-show="!isValid(step.to) && getValidateSteps">
+          <v-icon class="step__btn2" size="30" color="error darken-1" v-show="!isValid(step.to) && getShowErrors">
```
severinbeauvais, line 28: `getValidateSteps` becomes True when the user has gone to the review page.

`getShowErrors` becomes True when the user tries to file (I think).

These don't mean the same thing.

Instead of this, I'm thinking something is missing to make the subject steps invalid.

severinbeauvais (changes requested):
~I think this isn't the right solution.~

You can test this by leaving some pages incomplete and then going to the review page -- all invalid pages should show the X.

src/components/common/Stepper.vue
lambert-alex: `getValidateSteps` becomes True after the user tries to file,
 `getShowErrors` becomes True after the user visits the review page.

The step styling uses the same validation, for isInvalid.
https://github.com/bcgov/business-create-ui/blob/a1660d9ed66215ff35b2a19152d144cd3684b781/src/components/Dissolution/CompleteResolution.vue#L393

severinbeauvais: OK, I was wrong. Thanks for the references.

So, why is this broken only now and for only certain steps? It should have been broken for all steps up to now, right?

severinbeauvais: Resolving this comment. I need to understand what the problem is.

severinbeauvais (approved):
So it looks like I got my wires crossed...

I'm not sure why the existing logic was to show the red X only after trying to file and pay, but I agree with Yui's comment in the ticket, and your solution solves that.

Kudos, sorry for the confusion, good to merge!

lewischenstudio (approved):

src/components/common/Stepper.vue
lambert-alex: I think it was broken for all the steps, at least it was when I checked. I just think the ticket didn't specifically mention all the steps. Not really sure why.

PR #422, opened 2022-07-06
Dissolution ux fixes/12629

meganawong:
*Issue #:* /bcgov/entity#12629

*Description of changes:*
Implement UX fixes and update validation errors for dissolution dashboard.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the bcrs-entities-create-ui license (Apache 2.0).

PR #424, opened 2022-07-08
12307 Update business-create-ui to use legal API to V2

lewischenstudio:
*Issue #:* [/bcgov/entity#12307](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/12307)

*Description of changes:*
- renamed url in `legal.services.ts`
- renamed `legalApiUrl` in `config-helper.ts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "4.1.2",
+  "version": "4.1.3",
   "private": true,
   "appName": "Create UI",
   "sbcName": "SBC Common Components",
--- a/src/store/getters/state-getters.ts
+++ b/src/store/getters/state-getters.ts
@@ -468,8 +468,8 @@ export const isDissolutionValid = (state: StateIF): boolean => {
   const isCourtOrderValid = isRoleStaff(state) ? getCourtOrderStep(state).valid : true
   const isStaffPaymentValid = isRoleStaff(state) ? getStaffPaymentStep(state).valid : true
 
-  // eslint-disable-next-line
-  const isEffectiveDateTimeValid = (isBaseCompany(state) || isTypeFirm(state)) ? getEffectiveDateTime(state).valid : true
+  const isEffectiveDateTimeValid = (isBaseCompany(state) || isTypeFirm(state))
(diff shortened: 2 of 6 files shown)


vysakh-menon-aot (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "4.1.1",
+  "version": "4.1.2",
```
severinbeauvais, line 3: Another PR was just merged with this version. Please update.

severinbeauvais (commented):
Did you test that all existing Legal API calls (ie, any call that uses Base URL) still function correctly?

src/store/getters/state-getters.ts
```diff
@@ -468,6 +468,7 @@ export const isDissolutionValid = (state: StateIF): boolean => {
   const isCourtOrderValid = isRoleStaff(state) ? getCourtOrderStep(state).valid : true
   const isStaffPaymentValid = isRoleStaff(state) ? getStaffPaymentStep(state).valid : true
 
+  // eslint-disable-next-line
```
severinbeauvais, line 471: Why not split the line instead?

Also, you should disable all lint for a line; only the error you want to disable.

Also, Megan is updating this anyway: https://github.com/bcgov/business-create-ui/pull/422#discussion_r917089556

severinbeauvais: ^^ What I would do is make this multi-line in this PR, and then if Megan's PR goes first then you can remove it. If Megan's PR goes second, she can remove it.

severinbeauvais:
> Did you test that all existing Legal API calls (ie, any call that uses Base URL) still function correctly?

Specifically, all Legal API calls except the subject one will now use V2. Do they still work as before?

lewischenstudio:
> > Did you test that all existing Legal API calls (ie, any call that uses Base URL) still function correctly?
> 
> Specifically, all Legal API calls except the subject one will now use V2. Do they still work as before?

Yes. Endpoints using legal-api V2 are working as they before.

severinbeauvais (approved):

PR #428, opened 2022-07-27
Firm errors/13049

meganawong:
*Issue #:* /bcgov/entity#13049

*Description of changes:*
* Fix fee summary label for FIRM companies (previously always showing dissolution)
* Fix form validation for dissolution flow 
* Update sbc-common-components version

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "lodash": "^4.17.20",
     "pdfjs-dist": "2.0.943",
     "regenerator-runtime": "^0.13.7",
-    "sbc-common-components": "2.4.54",
+    "sbc-common-components": "3.0.1",
     "uuid": "^8.3.1",
     "vue": "2.6.12",
     "vue-affix": "^0.5.2",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "4.1.6",
+  "version": "4.1.7",
   "private": true,
   "appName": "Create UI",
   "sbcName": "SBC Common Components",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "4.1.5",
+  "version": "4.1.6",
   "private": true,
   "appName": "Create UI",
   "sbcName": "SBC Common Components",
--- a/src/App.vue
+++ b/src/App.vue
@@ -387,7 +387,7 @@ export default class App extends Mixins(
   }
   /** The Fee Summary filing text for firm businesses. */
   get filingLabelText (): string {
-    return this.isTypeFirm && this.isDissolutionFiling ? 'Dissolution' : ''
+    return this.isDissolutionFiling ? 'Dissolution' : ''
   }
 
   /** The About text. */
--- a/src/App.vue
+++ b/src/App.vue
@@ -387,7 +387,7 @@ export default class App extends Mixins(
   }
   /** The Fee Summary filing text for businesses. */
   get filingLabelText (): string {
-    return this.isDissolutionFiling ? 'Dissolution' : ''
+    return (this.isTypeFirm && this.isDissolutionFiling) ? 'Dissolution' : null
   }
 
   /** The About text. */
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -42,7 +42,7 @@
     "lodash": "^4.17.20",
     "pdfjs-dist": "2.0.943",
     "regenerator-runtime": "^0.13.7",
-    "sbc-common-components": "2.4.54",
+    "sbc-common-components": "3.0.1",
```
shabeeb-aot, line 45: 👍

shabeeb-aot (approved):

src/App.vue
```diff
@@ -387,7 +387,7 @@ export default class App extends Mixins(
   }
   /** The Fee Summary filing text for firm businesses. */
   get filingLabelText (): string {
-    return this.isTypeFirm ? 'Dissolution' : ''
+    return this.isTypeFirm && this.isDissolutionFiling ? 'Dissolution' : ''
```
severinbeauvais, line 390: I wonder if you should delete `isTypeFirm` from this?

Currently it looks like it might display "Dissolution" while registering a new firm, which would be incorrect.

src/store/getters/state-getters.ts
```diff
@@ -467,8 +467,10 @@ export const isDissolutionValid = (state: StateIF): boolean => {
   // only for Staff role
   const isCourtOrderValid = isRoleStaff(state) ? getCourtOrderStep(state).valid : true
   const isStaffPaymentValid = isRoleStaff(state) ? getStaffPaymentStep(state).valid : true
+  const isCompletingPartyValid = isRoleStaff(state) ? getCompletingParty(state).valid : true
+  const isDissolutionDateValid = !!getDissolutionDate(state)
```
severinbeauvais, line 471: Please add a blank line above this since it's not part of the staff block.

src/views/DissolutionFirm/DissolutionFirm.vue
```diff
@@ -313,7 +313,7 @@ export default class DissolutionFirm extends Mixins(DateMixin, EnumMixin) {
 
   /** Is true when the completing party conditions are not met. */
   get isCompletingPartyInvalid ():boolean {
-    return this.getValidateSteps && (!this.getCompletingParty.valid)
+    return this.getValidateSteps && (!this.completingPartyValid) && this.isRoleStaff
```
severinbeauvais, line 316: You don't need the internal braces, but I'd like to see outer braces so this looks like a "single return value".

src/App.vue
severinbeauvais: Please update the comment too. Thanks.

severinbeauvais: I don't understand how this worked correctly before... It looks like this getter value was always used by SbcFeeSummary, so did the fee summary component always show "Dissolution" previously???

meganawong: Yes, it was a bug. This change will resolve it

severinbeauvais: Would "null" be a better fallback value since we don't want the fee summary to use this prop (it's "not a value" rather than "empty string")?

severinbeauvais (approved):
Looks good.

Up to you whether to change '' to null, and then this can be merged.

src/App.vue
meganawong: > I wonder if you should delete `isTypeFirm` from this?
> 
> Currently it looks like it might display "Dissolution" while registering a new firm, which would be incorrect.

I just remembered, the check for the FIRM type is required as if it is a dissolution for a non-firm company it should use the original filing text "Voluntary Dissolution"

severinbeauvais: OK. Please add a comment saying this label text is specifically for firm dissolutions so we know next time ;)

src/App.vue
```diff
@@ -385,9 +385,9 @@ export default class App extends Mixins(
       this.fileAndPayInvalidNameRequestDialog
     )
   }
-  /** The Fee Summary filing text for firm businesses. */
+  /** The Fee Summary filing text for firm dissolutions. */
   get filingLabelText (): string {
-    return this.isTypeFirm ? 'Dissolution' : ''
+    return (this.isTypeFirm && this.isDissolutionFiling) ? 'Dissolution' : null
```
severinbeauvais, line 390: No, this getter is used for ALL entity types (not just firms or dissolutions).

I was thinking you should add a comment above the return statement describing that the text override is only for firm dissolutions.

PR #432, opened 2022-08-10
Business number displaying incorrectly/13140

meganawong:
*Issue #:* /bcgov/entity#13140

*Description of changes:*
* add call to legal-api/businesses and stores relevant values 
* displays correct business number in entity info header (taxId)


src/components/common/EntityInfo.vue
```diff
@@ -92,6 +92,7 @@ export default class EntityInfo extends Mixins(EnumMixin, DateMixin) {
   @Getter isIncorporationFiling!: boolean
   @Getter isRegistrationFiling!: boolean
   @Getter getBusinessId!: string
```
severinbeauvais, line 94: Is this now obsolete?

src/interfaces/store-interfaces/state-interfaces/business-id-interface.ts
```diff
@@ -0,0 +1,37 @@
...
+/** The Business Warning object. */
+export interface BusinessWarningIF {
+  code: string // FUTURE: use an enum
+  message: string
+  warningType: string
+  filing: string
+}
+
+/** The Business object from the API. */
+export interface BusinessIdIF {
```
severinbeauvais, line 13: I think we should have a single Business Interface. Could this object easily be merged with the existing BusinessIF?

src/App.vue
```diff
@@ -607,6 +619,15 @@ export default class App extends Mixins(
       // fetch the draft filing and resources
       try {
         if (this.getBusinessId) {
+          try {
+            await this.fetchBusinessData() // throws on error
+          } catch (error) {
```
severinbeauvais, line 624: Why not let the outer catch block handle this (line 649)?

src/App.vue
```diff
@@ -978,6 +1006,31 @@ export default class App extends Mixins(
     await updateLdUser(key, email, firstName, lastName, custom)
   }
 
+  /** Stores business info from Legal API. */
+  async storeBusinessInfo (response: any): Promise<void> {
```
severinbeauvais, line 1010: Doesn't need to be async, or awaited (line 726).

src/App.vue
```diff
@@ -978,6 +1006,31 @@ export default class App extends Mixins(
...
+    const business = response?.data?.business as BusinessIdIF
+
+    // if (!business) {
+    //   throw new Error('Invalid business info')
+    // }
```
severinbeauvais, line 1015: This is the only check that the API returned `response.data.business`. I think you should uncomment this.

src/components/common/EntityInfo.vue
meganawong: it is still used in other places

src/App.vue
meganawong: it appears to break some of the tests if I do this

severinbeauvais: Maybe the tests should be updated as well?

src/interfaces/store-interfaces/state-interfaces/business-id-interface.ts
severinbeauvais: Thanks. Looks good.

tests/unit/App.spec.ts
```diff
@@ -16,6 +16,7 @@ import Stepper from '@/components/common/Stepper.vue'
 import { ConfirmDialog } from '@bcrs-shared-components/confirm-dialog'
 import mockRouter from './MockRouter'
 import Vuelidate from 'vuelidate'
+import { Getter } from 'vuex-class'
```
severinbeauvais, line 19: I think you don't need this.

src/App.vue
meganawong: updated the tests

severinbeauvais: Awesome, thanks!

severinbeauvais (approved):
LGTM but it would nice to get @shabeeb-aot 's review before merging.

src/App.vue
```diff
@@ -978,6 +998,31 @@ export default class App extends Mixins(
...
+    }
+
+    // if (this.getBusinessId !== business.identifier) {
+    //   throw new Error('Business identifier mismatch')
+    // }
+
+    // FUTURE: change this to a single setter/object?
+    this.setAdminFreeze(business.adminFreeze)
+    this.setEntityName(business.legalName)
+    this.setEntityState(business.state)
```
shabeeb-aot, line 1016: can we use one value to set all these business objects? 
So we can have one mutation for business object

severinbeauvais: I looked into that and it may be difficult/messy since there are other places in the code that set just one of these values.

Maybe we could "set" an array of key-val tuples, and the action would only set the keys (properties) in the list?

shabeeb-aot: I think we can leave it like this , may more easy to read code

shabeeb-aot (approved):

PR #435, opened 2022-08-23
13372 Create UI: Review and Confirm step icon should not show error until there is error

lewischenstudio:
*Issue #:* /bcgov/entity#13372
*Issue #:* /bcgov/entity#13371

*Description of changes:*
- [x] The extension field should keep its dotted line until the user enters a value in that field.
- [x] The Review and Confirm icon should show neutral icon until the user presses File and Pay without having satisfied all required fields on that step.
- [x] Test NR number: NR 1030445
- [x] unit tests added

### DIFF
--- a/src/components/common/Actions.vue
+++ b/src/components/common/Actions.vue
@@ -117,6 +117,7 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
   @Getter isSaving!: boolean
   @Getter isSavingResuming!: boolean
   @Getter isFilingPaying!: boolean
+  @Getter isFinalRegistrationValid!: boolean
 
   @Action setIsSaving!: ActionBindingIF
   @Action setIsSavingResuming!: ActionBindingIF
@@ -124,6 +125,7 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
   @Action setHaveChanges!: ActionBindingIF
   @Action setEffectiveDateTimeValid!: ActionBindingIF
   @Action setValidateSteps!: ActionBindingIF
+  @Action setClickFileAndPay!: ActionBindingIF
 
   /** Is True if Jest is running the code. */
   get isJestRunning (): boolean {
@@ -206,6 +208,12 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
    * @returns a promise (ie, this is an async method)
    */
   private async onClickFilePay (): Promise<void> {
+    // Set the status of clickFileAndPay for registration
+    if (this.isTypeFirm) {
+      this.setClickFileAndPay(true)
+      if (!this.isFinalRegistrationValid) return
+    }
+
     // Prompt Step validations
     this.setValidateSteps(true)
 
--- a/src/components/common/Actions.vue
+++ b/src/components/common/Actions.vue
@@ -209,10 +209,7 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
    */
   private async onClickFilePay (): Promise<void> {
     // Set the status of clickFileAndPay for registration
-    if (this.isTypeFirm) {
-      this.setClickFileAndPay(true)
-      if (!this.isFinalRegistrationValid) return
-    }
+    if (this.isTypeFirm) this.setClickFileAndPay(true)
 
     // Prompt Step validations
     this.setValidateSteps(true)
--- a/src/components/common/Stepper.vue
+++ b/src/components/common/Stepper.vue
@@ -69,7 +69,10 @@ export default class Stepper extends Vue {
   @Getter isIncorporationAgreementValid!: boolean
   @Getter isIncorporationApplicationValid!: boolean
   @Getter isMemorandumValid!: boolean
-  @Getter isRegistrationValid!: boolean
+  @Getter isPartialRegistrationValid!: boolean
+  @Getter isFinalRegistrationValid!: boolean
+  @Getter getClickFileAndPay!: boolean
+  @Getter isTypeFirm!: boolean
   @Getter isResolutionValid!: boolean
   @Getter isRulesValid!: boolean
 
@@ -91,9 +94,18 @@ export default class Stepper extends Vue {
 
       case RouteNames.REGISTRATION_DEFINE_BUSINESS: return this.getRegistration.defineBusinessValid
       case RouteNames.REGISTRATION_PEOPLE_ROLES: return this.isAddPeopleAndRolesValid
-      case RouteNames.REGISTRATION_REVIEW_CONFIRM: return this.isRegistrationValid
+      case RouteNames.REGISTRATION_REVIEW_CONFIRM: return this.isRegistrationValid()
     }
     return false
+  }
+
+  private isRegistrationValid (): boolean {
+    if (this.isTypeFirm) {
+      if (!this.isPartialRegistrationValid) return false
+      if (!this.getClickFileAndPay) return this.isPartialRegistrationValid
+      return this.isFinalRegistrationValid
+    }
+    return this.isFinalRegistrationValid
   }
 
   /** show error styling for reveiw-confirm step after file and pay is selected */
--- a/src/components/common/Stepper.vue
+++ b/src/components/common/Stepper.vue
@@ -22,7 +22,8 @@
           >
             <v-icon class="step__icon" :class="{ 'selected-icon': isCurrentStep(step) }">{{ step.icon }}</v-icon>
           </v-btn>
-          <v-icon class="step__btn2" size="30" color="success darken-1" v-show="isValid(step.to)">
+          <v-icon class="step__btn2" size="30" color="success darken-1" v-show="isValid(step.to)
+            && isFirmReviewValid(step.to)">
             mdi-check-circle
           </v-icon>
           <v-icon class="step__btn2" size="30" color="error darken-1"
@@ -99,6 +100,8 @@ export default class Stepper extends Vue {
     return false
   }
 
+  /** Check for define and people components are valid first
+  * Check for review component when the FileAndPay button is clicked */
   private isRegistrationValid (): boolean {
     if (this.isTypeFirm) {
       if (!this.isPartialRegistrationValid) return false
@@ -106,6 +109,12 @@ export default class Stepper extends Vue {
       return this.isFinalRegistrationValid
     }
     return this.isFinalRegistrationValid
+  }
+
+  /** show error styling for registration review-confirm step after file and pay is selected */
+  private isFirmReviewValid (route: RouteNames): boolean {
+    const thisRoute = route === RouteNames.REGISTRATION_REVIEW_CONFIRM
+    return thisRoute ? this.getClickFileAndPay : !thisRoute
   }
 
   /** show error styling for reveiw-confirm step after file and pay is selected */
--- a/src/interfaces/resource-interfaces/resource-interface.ts
+++ b/src/interfaces/resource-interfaces/resource-interface.ts
@@ -51,7 +51,8 @@ export interface RegistrationResourceIF {
       certifyClause: string
       entityDisplay: string
     }
-  }
+  },
+  clickFileAndPay: boolean
 }
 
 // Interface to define the resource model
--- a/src/store/actions/actions-model.ts
+++ b/src/store/actions/actions-model.ts
@@ -375,3 +375,7 @@ export const setLastDirectorChangeDate: ActionIF = ({ commit }, date: string): v
 export const setWarnings : ActionIF = ({ commit }, businessWarnings: Array<BusinessWarningIF>): void => {
   commit('mutateBusinessWarnings', businessWarnings)
 }
+
+export const setClickFileAndPay: ActionIF = ({ commit }, clickFileAndPay: boolean): void => {
+  commit('mutateClickFileAndPay', clickFileAndPay)
+}
--- a/src/store/getters/resource-getters.ts
+++ b/src/store/getters/resource-getters.ts
@@ -107,3 +107,8 @@ export const getCustodialRecordsResources = (state: StateIF): CustodianResourceI
 export const getAffidavitResources = (state: StateIF): AffidavitResourceIF => {
   return state.resourceModel.affidavit
 }
+
+/** The state of the File and Pay button is clicked */
+export const getClickFileAndPay = (state: StateIF): boolean => {
+  return state.resourceModel.clickFileAndPay
+}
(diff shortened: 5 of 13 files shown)


src/components/common/BusinessContactInfo.vue
```diff
@@ -187,4 +188,19 @@ export default class BusinessContactInfo extends Mixins(CommonMixin) {
...
+    rgba(0, 0, 0, 0.38) 0px,
+    rgba(0, 0, 0, 0.38) 2px,
+    transparent 2px,
+    transparent 4px
+  ) 1 repeat;
+}
+
+::v-deep .phone-ext.phone-ext-filled .v-text-field > .v-input__control > .v-input__slot::before {
+  border-image: none;
+}
```
lewischenstudio, line 205: This part is for ticket [13371](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/13371). Two classes are added to tell the two different states of phone extension field: empty and filled.

src/components/common/Stepper.vue
```diff
@@ -91,11 +94,22 @@ export default class Stepper extends Vue {
...
+  * Check for review component when the FileAndPay button is clicked */
+  private isRegistrationValid (): boolean {
+    if (this.isTypeFirm) {
+      if (!this.isPartialRegistrationValid) return false
+      if (!this.getClickFileAndPay) return this.isPartialRegistrationValid
+      return this.isFinalRegistrationValid
+    }
+    return this.isFinalRegistrationValid
+  }
+
```
lewischenstudio, line 112: The logic in this method is the following:
1. Checks if it is a firm type or not
2. If two previous steppers are not valid, then return false; otherwise, move on to the next step.
3. If it's the first time clicking the button "File and Pay", then check "isPartialRegistrationValid".
4. If it is a firm type, then check "isFinalRegistrationValid".

src/interfaces/resource-interfaces/resource-interface.ts
```diff
@@ -51,7 +51,8 @@ export interface RegistrationResourceIF {
       certifyClause: string
       entityDisplay: string
     }
-  }
+  },
+  clickFileAndPay: boolean
```
lewischenstudio, line 55: This field is added to toggle the status whether the button "File and Pay" is clicked.

src/store/getters/state-getters.ts
```diff
@@ -533,8 +534,16 @@ export const isIncorporationApplicationValid = (state: StateIF): boolean => {
...
 }
 
+/** Whether the partial registration steps are valid. */
+export const isPartialRegistrationValid = (state: StateIF): boolean => {
+  return (
+    getRegistration(state).defineBusinessValid &&
+    isAddPeopleAndRolesValid(state)
+  )
+}
+
```
lewischenstudio, line 544: "isPartialRegistrationValid" only checks for the validities of the first two steppers.

src/components/common/Stepper.vue
```diff
@@ -22,7 +22,8 @@
           >
             <v-icon class="step__icon" :class="{ 'selected-icon': isCurrentStep(step) }">{{ step.icon }}</v-icon>
           </v-btn>
-          <v-icon class="step__btn2" size="30" color="success darken-1" v-show="isValid(step.to)">
+          <v-icon class="step__btn2" size="30" color="success darken-1" v-show="isValid(step.to)
+            && isFirmReviewValid(step.to)">
```
lewischenstudio, line 26: Added "isFirmReviewValid(step.to)" to make sure "The Review and Confirm icon should show neutral icon" (from ticket [13372](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/13372)).

src/components/common/Stepper.vue
lambert-alex: I think there are getters in create UI which already handle the validation before and after file and pay has been clicked https://github.com/bcgov/business-create-ui/blob/c0ab60baa549c77d87d5a29482c8d6a7f6b120b7/src/store/getters/state-getters.ts#L556 and https://github.com/bcgov/business-create-ui/blob/c0ab60baa549c77d87d5a29482c8d6a7f6b120b7/src/store/getters/state-getters.ts#L564

lambert-alex: Okay I see what you mean.
Have you checkout that this Component works in other places?
I know I have had to fix this exact error in this component for BEN dissolutions previously.
I think it would be worth it to check that this stepper function works the same across dissolutions, registrations and incorporations. 

If you have already checked it then I am happy with it.

src/store/getters/state-getters.ts
```diff
@@ -533,8 +534,16 @@ export const isIncorporationApplicationValid = (state: StateIF): boolean => {
...
 /** Whether all the registration steps are valid. */
-export const isRegistrationValid = (state: StateIF): boolean => {
+export const isFinalRegistrationValid = (state: StateIF): boolean => {
```
cameron-freshworks, line 546: I'm not sure about the name change here. 
This computed value is still the registration validity as a whole?

src/store/getters/state-getters.ts
cameron-freshworks: Curious about this, it appears we already have a getter for each respective step. What requirement is there that requests steps 1 and 2 validations in isolation?

src/store/getters/state-getters.ts
```diff
@@ -40,6 +40,7 @@ import {
   CompletingPartyIF,
   PartyIF
 } from '@/interfaces'
+import { stateModel } from '../state'
```
cameron-freshworks, line 43: Is this import in use?

src/store/actions/actions-model.ts
```diff
@@ -375,3 +375,7 @@ export const setLastDirectorChangeDate: ActionIF = ({ commit }, date: string): v
 export const setWarnings : ActionIF = ({ commit }, businessWarnings: Array<BusinessWarningIF>): void => {
   commit('mutateBusinessWarnings', businessWarnings)
 }
+
+export const setClickFileAndPay: ActionIF = ({ commit }, clickFileAndPay: boolean): void => {
+  commit('mutateClickFileAndPay', clickFileAndPay)
+}
```
cameron-freshworks, line 381: Was there no other means already in place that is prompted when clicking file and file? 
A validation flag perhaps?

I'm surprised that this is something that isn't already handled?

src/store/getters/resource-getters.ts
```diff
@@ -107,3 +107,8 @@ export const getCustodialRecordsResources = (state: StateIF): CustodianResourceI
 export const getAffidavitResources = (state: StateIF): AffidavitResourceIF => {
   return state.resourceModel.affidavit
 }
+
+/** The state of the File and Pay button is clicked */
+export const getClickFileAndPay = (state: StateIF): boolean => {
+  return state.resourceModel.clickFileAndPay
+}
```
cameron-freshworks, line 114: Same as above, isn't there already a validation flag/prompt when clicking file and pay?

src/store/getters/state-getters.ts
cameron-freshworks: Sorry, reading further into the ticket/context, a firm registration is 2 steps?

src/store/actions/actions-model.ts
cameron-freshworks: After another look, isn't `setValidateSteps` already a trigger for validations?

src/components/common/Stepper.vue
cameron-freshworks: I'm a little wary of this very firm explicit code in whats is a fairly common component. 

Maybe i'm off but shouldn't `step.to` (ie `isRegistrationValid` in this case) be a computed value that already encapsulates these conditions. 

If you only want to show the invalid icons on the stepper once a user has tried to submit, isn't that just a matter of adding the `validateSteps` flag or whichever flag is triggered, to the condition outlined below in the `isValid()` method?

lambert-alex (commented):
13371 LGTM, I think this should be split off into a separate PR to merge. For 13372 since the Stepper component is providing styling for three separate filings, I think we should look at making changes to ```getValidateSteps``` and ```getShowErrors``` so they work across all three filing types.

src/components/common/Stepper.vue
lewischenstudio: Because the ticket only shows the image of registration, I'm not sure whether this method is used for all filing types. If it is used for, then there will be a small tweak in the implementation. We will have to confirm with Yui later.

lewischenstudio:
> Styling looks fine!
> 
> I'm not sure about the additional actions/mutations/getters however with regards to the Stepper icon, i believe that could be achieved without additional state properties no?
> 
> Seems like we are adding in a bit of complexity for a rather straightforward scenario, i might have to defer this too Severin for when he is back.

I totally agree. This implementation is a much more complicated addition to the original straightforward one. The problem was that the Review and Confirm stepper validation is triggered whether the previous steppers are valid or not. If they are not valid, then the original implementation works perfectly as it's triggered automatically. If the previous steppers are valid, then the original implementation also triggers the validation, which becomes a problem. The validation for the registration "isRegistrationValid" includes checks for all components, including the Certify and StaffPayment components. That is, upon visiting the Review and Confirm stepper, this validation is triggered automatically whether or not the previous steppers and the Certify and StaffPayment components are valid or not. This ticket wants the Review and Confirm stepper to remain "neutral icon" (aka having neither the check nor the cross icon). Then the original validation "isRegistrationValid" which includes checks for everything must be split into two parts: one for previous steppers and one for its current stepper. If the previous steppers are valid, then it displays the neutral icon. "isFirmReviewValid" is to ensure that while the previous two steppers are valid, if it is a registration type, then return false. Then the checker icon will not display, while the cross icon doesn't display either. The validation for the Review and Confirm stepper is only triggered when the user clicks "File and Pay". This triggers some state changes, and all new getters and setters are created for this purpose. The entire implementation becomes complicated, mainly because the final validation becomes two parts.

Sorry if I come in a bit straight here, as I want to make sure the logic and reason is clear.

lewischenstudio:
> 13371 LGTM, I think this should be split off into a separate PR to merge. For 13372 since the Stepper component is providing styling for three separate filings, I think we should look at making changes to `getValidateSteps` and `getShowErrors` so they work across all three filing types.

Sure. Let me work on that in a bit. I should've thought of it earlier. Thanks.

src/store/getters/state-getters.ts
lewischenstudio: That is what I understand from this ticket. I'm not sure if this will be applied to all other filings.

src/components/common/Stepper.vue
lewischenstudio: The problem is that `isRegistrationValid` validates all five components. The cross icon will be displayed when the user visits the Review and Confirm step for the first time, because the Certify or StaffPayment component is not checked yet.

src/store/getters/state-getters.ts
lewischenstudio: I'll remove this line.

src/store/getters/state-getters.ts
lewischenstudio: The name change is to separate the partial registration validation from the final registration validation. I can change it back. Yes, this computed value is still the registration validity as a whole. If it is not a whole, then something bad might happen. Like if the user visits the final stepper and checks everything, returns to the previous step to remove something, then comes back the final stepper, it will still be valid.

src/store/actions/actions-model.ts
lewischenstudio: I'll find out what happens if I change "setClickFileAndPay" to "setValidateSteps".

src/components/common/BusinessContactInfo.vue
lewischenstudio: This part has been merged as of PR [13371 Create UI: Phone extension visual should remain "optional" until value entered](https://github.com/bcgov/business-create-ui/pull/436).

severinbeauvais:
Does this PR still resolve ticket 13371? If not, please update the description block.

src/components/common/Actions.vue
```diff
@@ -117,13 +117,15 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
   @Getter isSaving!: boolean
   @Getter isSavingResuming!: boolean
   @Getter isFilingPaying!: boolean
+  @Getter isFinalRegistrationValid!: boolean
```
severinbeauvais, line 120: This getter doesn't appeatr to be used in this component.

src/components/common/Actions.vue
```diff
@@ -206,6 +208,9 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
    * @returns a promise (ie, this is an async method)
    */
   private async onClickFilePay (): Promise<void> {
+    // Set the status of clickFileAndPay for registration
+    if (this.isTypeFirm) this.setClickFileAndPay(true)
```
severinbeauvais, line 212: I'm surprised this ticket/fix applies to firms only. Please confirm with UI/UX.

On the other hand, if this bug is only seen with firms then there's something wrong with the firms code that should be fixed directly instead of changing the app's architecture like this.

src/components/common/Stepper.vue
severinbeauvais: I agree with Cam. We want to stay away from special cases, and I'm not aware that firms are a special case in this project.

Also, this app's architecture (global functionality) should apply to all entity types. If this bug applies to all entity types then this code doesn't fix it. If this bug applies to firms only then there's something broken in the firms code that should be fixed instead.

src/components/common/Stepper.vue
```diff
@@ -91,11 +95,28 @@ export default class Stepper extends Vue {
...
+  /** Check for define and people components are valid first
+  * Check for review component when the FileAndPay button is clicked */
+  private isRegistrationValid (): boolean {
+    if (this.isTypeFirm) {
+      if (!this.isPartialRegistrationValid) return false
+      if (!this.getClickFileAndPay) return this.isPartialRegistrationValid
+      return this.isFinalRegistrationValid
+    }
+    return this.isFinalRegistrationValid
+  }
```
severinbeauvais, line 112: This is too much of a special case:
- firms only
- hard-coded previous step validity

severinbeauvais (changes requested):
See my comments above. I do not approve this solution.

lewischenstudio:
> See my comments above. I do not approve this solution.

Sorry Severin. I am a bit confused here. Do you not approve the solution to solve the problem "The Review and Confirm icon should show neutral icon until the user presses File and Pay without having satisfied all required fields on that step", the solution to firms only, or the proposition as "The Review and Confirm icon should show neutral icon until the user presses File and Pay without having satisfied all required fields on that step"?

src/interfaces/resource-interfaces/resource-interface.ts
lewischenstudio: This variable is added here because this ticket is about firm registration only. If this approach applies to all filings, then this variable should be removed and I will use the original variable.

lewischenstudio:
After the discussion, we finalized the decision and solution to the original ticket. Since the solution and the approach does not meet the final requirement, I am closing this PR. Thank you guys for the reviews. The new PR is [here](https://github.com/bcgov/business-create-ui/pull/437).

PR #437, opened 2022-08-30
13372 Create UI: Review and Confirm step icon should not show error until there is error

lewischenstudio:
*Issue #:* /bcgov/entity#13372

*Description of changes:*
- added "getValidateSteps" to the red cross icon validator.
- added unit test

### DIFF
--- a/src/components/common/Stepper.vue
+++ b/src/components/common/Stepper.vue
@@ -26,7 +26,7 @@
             mdi-check-circle
           </v-icon>
           <v-icon class="step__btn2" size="30" color="error darken-1"
-            v-show="!isValid(step.to) && getShowErrors && isReviewStepValid(step.to)">
+            v-show="!isValid(step.to) && getShowErrors && isReviewStepValid(step.to) && getValidateSteps">
             mdi-close-circle
           </v-icon>
         </div>
--- a/src/components/common/Stepper.vue
+++ b/src/components/common/Stepper.vue
@@ -26,7 +26,7 @@
             mdi-check-circle
           </v-icon>
           <v-icon class="step__btn2" size="30" color="error darken-1"
-            v-show="!isValid(step.to) && getShowErrors && isReviewStepValid(step.to) && getValidateSteps">
+            v-show="!isValid(step.to) && getShowErrors && getValidateSteps">
             mdi-close-circle
           </v-icon>
         </div>
@@ -94,14 +94,6 @@ export default class Stepper extends Vue {
       case RouteNames.REGISTRATION_REVIEW_CONFIRM: return this.isRegistrationValid
     }
     return false
-  }
-
-  /** show error styling for reveiw-confirm step after file and pay is selected */
-  protected isReviewStepValid (route: RouteNames): boolean {
-    return (
-      route !== RouteNames.DISSOLUTION_REVIEW_CONFIRM ||
-      this.getValidateSteps
-    )
   }
 
   private goTo (step) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-create-ui",
-  "version": "4.1.12",
+  "version": "4.1.13",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "business-create-ui",
-      "version": "4.1.12",
+      "version": "4.1.13",
       "dependencies": {
         "@babel/compat-data": "^7.11.0",
         "@bcrs-shared-components/bread-crumb": "1.0.2",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "4.1.12",
+  "version": "4.1.13",
   "private": true,
   "appName": "Create UI",
   "sbcName": "SBC Common Components",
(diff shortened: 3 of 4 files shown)


src/components/common/Stepper.vue
```diff
@@ -26,7 +26,7 @@
             mdi-check-circle
           </v-icon>
           <v-icon class="step__btn2" size="30" color="error darken-1"
-            v-show="!isValid(step.to) && getShowErrors && isReviewStepValid(step.to)">
+            v-show="!isValid(step.to) && getShowErrors && isReviewStepValid(step.to) && getValidateSteps">
```
severinbeauvais, line 29: I know I'm picking on existing code but I'm trying to understand the logic here (validate it in my head) and I'm still confused by it.

You check `getValidateSteps` here and it's also checked in `isReviewStepValid()`...

I also notice that this code displays the Close icon on error, but one of the conditions is that `isReviewStepValid` is True (ie, valid)!?!?

Can you fix this as it's really hard to understand ... I suggest renaming isReviewStepValid -> isReviewStepInvalid, or adding a ! in front of it and refactoring the method to return the opposite of what it does now.

It may turn out that `getValidateSteps` doesn't need to be used twice 🤷

lewischenstudio: I don't quite understand `isReviewStepValid()` either. It returns `route !== RouteNames.DISSOLUTION_REVIEW_CONFIRM || this.getValidateSteps`. This `v-icon` will be not displayed if `route === RouteNames.DISSOLUTION_REVIEW_CONFIRM` and `this.getValidateSteps == false`? It's really confusing.

The one that matters the most is `getValidateSteps`. The icon only displays when `getValidateSteps == true`, along with `!isValid(step.to) && getShowErrors`. I am thinking to remove `isReviewStepValid()`.

severinbeauvais: > I am thinking to remove `isReviewStepValid()`.

The team that implemented dissolutions must have done this. It's not exactly clear to me why, but I do know that dissolutions are a single step, so maybe this is to handle that case. Best bet: check with @seeker25 or @meganawong on this.

I'd like to see whole statement simplified if possible (or explained if not 😃 ).

lewischenstudio: This is done in this [pull request](https://github.com/bcgov/business-create-ui/pull/418).

lewischenstudio: I've removed the method in my latest commit. I can add the method back if necessary. I've tested all scenarios list in the ticket and they all worked well.

severinbeauvais: Oh, Alex did that, as you mentioned previously.

Please figure this out and either simplify it, or add comments explaining it.

src/components/common/Stepper.vue
```diff
@@ -96,14 +96,6 @@ export default class Stepper extends Vue {
...
   }
 
-  /** show error styling for reveiw-confirm step after file and pay is selected */
-  protected isReviewStepValid (route: RouteNames): boolean {
-    return (
-      route !== RouteNames.DISSOLUTION_REVIEW_CONFIRM ||
-      this.getValidateSteps
-    )
-  }
-
```
severinbeauvais, line 106: Please document and run a test case for this (in the ticket) to verify that you didn't break existing functionality.

lewischenstudio: Yeah about that. I tried to add unit tests to test for different scenarios, but the icons with class `step__btn2` always shows `display: true`, meaning that there was no way I could test whether the red cross icon or the green check icon is being displayed. A little help is needed here.

src/components/common/Stepper.vue
lewischenstudio: Alex is not here until next week. I can ask him why he implemented this when he is back.

src/components/common/Stepper.vue
severinbeauvais: A manual test case is all I'm asking for.
1. find out what the previous code was for
2. verify that it still works

Add the test case and result to your ticket for reviewers and QA to see. Thx!

src/components/common/Stepper.vue
severinbeauvais: Can you tell from Alex's ticket description, or ask Alex's ticket author?

It actually seems that your ticket pre-empts Alex's ticket, though they were both written by Yui.

src/components/common/Stepper.vue
lewischenstudio: The previous code was trying to do the same job: show error styling for reveiw-confirm step after file and pay is selected. However, it didn't and it's confusing for me. Alex will have to explain it. I left a comment in that [pull request](https://github.com/bcgov/business-create-ui/pull/418) after it was merged. There was no unit test when he added the code in his PR. For this pull request, I wanted to add unit test case, but there is some coding difficulty as mentioned above. If you are talking about test case that I run on local environment using filings, then I can confirm that the scenarios commented at the bottom of the ticket were tested in my local environment.

severinbeauvais: I am talking about test steps that you perform manually (locally), and document in the ticket comments so that we can see you did it.

lewischenstudio: Thanks for the clarification. I've tested various scenarios locally and added the validation scenarios in the ticket. I updated the comment just now for QA testing.

severinbeauvais (approved):
I approve this PR.

It would be nice to get some more reviews before merging this.

seeker25 (approved):

lewischenstudio:
Thanks for the merge. I agree. It would be nice to have more reviewers before merging this PR.

PR #440, opened 2022-09-06
11783 Get rid of "new Date()"

severinbeauvais:
*Issue #:* bcgov/entity#11783

*Description of changes:*
- changed "new Date()" to use internal values and better mixin methods
- misc cleanup
- fixed some unit tests
- app version = 4.1.15

### DIFF
--- a/src/components/Dissolution/CompleteResolution.vue
+++ b/src/components/Dissolution/CompleteResolution.vue
@@ -5,20 +5,19 @@
       <header>
         <h2>{{getCreateResolutionResource.introSection.header}}</h2>
         <p class="section-description mt-2">
-            <span v-for="(partialItem, index) in getCreateResolutionResource.introSection.items" :key="index">
-              <span v-if="partialItem.type === ItemTypes.TEXT" v-html="partialItem.value"></span>
-              <v-tooltip v-if="partialItem.type === ItemTypes.TOOLTIP"
-                 top max-width="15rem"
-                 content-class="top-tooltip"
-                 transition="fade-transition">
-                  <template v-slot:activator="{ on }">
-                    <span v-on="on" class="tool-tip dotted-underline"> {{partialItem.value.label}} </span>
-                  </template>
-                  <span>
-                    {{partialItem.value.text}}
-                  </span>
-              </v-tooltip>
-            </span>
+          <span v-for="(partialItem, index) in getCreateResolutionResource.introSection.items" :key="index">
+            <span v-if="partialItem.type === ItemTypes.TEXT" v-html="partialItem.value"></span>
+            <v-tooltip v-if="partialItem.type === ItemTypes.TOOLTIP"
+              top max-width="15rem"
+              content-class="top-tooltip"
+              transition="fade-transition"
+            >
+              <template v-slot:activator="{ on }">
+                <span v-on="on" class="tool-tip dotted-underline"> {{partialItem.value.label}} </span>
+              </template>
+              <span>{{partialItem.value.text}}</span>
+            </v-tooltip>
+          </span>
         </p>
       </header>
     </section>
@@ -134,7 +133,7 @@
                 :minDate="resolutionDateMinStr"
                 :maxDate="resolutionDateMaxStr"
                 :inputRules="resolutionDateRules"
-                @emitDateSync="onResolutionDateSync"
+                @emitDateSync="onResolutionDateSync($event)"
               />
             </v-col>
           </v-row>
@@ -168,7 +167,7 @@
                             :counter="MAX_RESOLUTION_TEXT_LENGTH"
                             v-model="resolutionText"
                             :rules="resolutionTextRules"
-                            @change="onResolutionTextChanged"
+                            @change="onResolutionTextChanged($event)"
                             v-observe-visibility="{ callback: onResolutionVisibilityChanged, once: true }"
                 />
               </v-form>
@@ -243,7 +242,7 @@
                 :minDate="signatureDateMinStr"
                 :maxDate="signatureDateMaxStr"
                 :inputRules="signatureDateRules"
-                @emitDateSync="onSigningDateSync"
+                @emitDateSync="onSigningDateSync($event)"
               />
             </v-col>
           </v-row>
@@ -350,18 +349,18 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin) {
     confirmResolutionChkRef: FormIF
   }
 
-  private MAX_RESOLUTION_TEXT_LENGTH: number = 1000
-  private resolutionText: string = ''
+  readonly MAX_RESOLUTION_TEXT_LENGTH = 1000
+  private resolutionText = ''
   private resolutionConfirmed = false
   private helpToggle = false
-  private PLACEHOLDER_LEGAL_NAME = 'legal_name'
-  private PLACEHOLDER_LEGAL_NAME_INLINE = 'legal_name_inline'
-  private signingPerson: PersonIF = null
+  readonly PLACEHOLDER_LEGAL_NAME = 'legal_name'
+  readonly PLACEHOLDER_LEGAL_NAME_INLINE = 'legal_name_inline'
+  private signingPerson = { ...EmptyPerson } as PersonIF
 
   // Date properties
-  private resolutionDateText: string = ''
-  private signatureDateText: string = ''
-  private foundingDate: Date = null
+  private resolutionDateText = ''
+  private signatureDateText = ''
+  private foundingDate = null as Date
 
   @Getter getShowErrors!: boolean
   @Getter isTypeCoop!: boolean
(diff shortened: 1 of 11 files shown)


src/components/Dissolution/CompleteResolution.vue
```diff
@@ -350,18 +349,18 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin) {
...
+  readonly MAX_RESOLUTION_TEXT_LENGTH = 1000
+  private resolutionText = ''
   private resolutionConfirmed = false
   private helpToggle = false
-  private PLACEHOLDER_LEGAL_NAME = 'legal_name'
-  private PLACEHOLDER_LEGAL_NAME_INLINE = 'legal_name_inline'
-  private signingPerson: PersonIF = null
+  readonly PLACEHOLDER_LEGAL_NAME = 'legal_name'
+  readonly PLACEHOLDER_LEGAL_NAME_INLINE = 'legal_name_inline'
+  private signingPerson = { ...EmptyPerson } as PersonIF
```
severinbeauvais, line 358: Needed for unit tests so they can de-reference `signingPerson` initially.

src/components/Dissolution/CompleteResolution.vue
```diff
@@ -443,96 +442,87 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin) {
     ]
   }
 
-  private validateIsBetweenDates (minDate: Date, maxDate: Date, dateToValidate: Date): boolean {
-    if (!dateToValidate) { return true }
-    if (dateToValidate >= minDate && dateToValidate <= maxDate) { return true }
-    return false
-  }
```
severinbeauvais, line 450: Never used, so I deleted it.

src/components/Dissolution/CompleteResolution.vue
```diff
@@ -614,7 +604,7 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin) {
 
   /** Called when component is created. */
   created (): void {
-    const foundingDate = new Date(this.getBusinessFoundingDate)
+    const foundingDate = this.apiToDate(this.getBusinessFoundingDate)
```
severinbeauvais, line 607: Sample business Founding Date: "2020-06-03T18:13:16.520032+00:00"

src/components/Dissolution/CompleteResolution.vue
```diff
@@ -443,96 +442,87 @@ export default class CompleteResolution extends Mixins(CommonMixin, DateMixin) {
...
+  get resolutionDateMin (): Date {
     return this.foundingDate
   }
 
   /** The maximum date that can be entered (today). */
-  private get resolutionDateMaxStr (): string {
-    const date = new Date()
-    return this.dateToYyyyMmDd(date)
+  get resolutionDateMaxStr (): string {
+    return this.dateToYyyyMmDd(this.getCurrentJsDate)
```
severinbeauvais, line 457: Alternately, this could use `this.getCurrentDate`, but the code as written is consistent with similar getters here so I think this is better.

src/components/Incorporation/IncorporationDateTime.vue
```diff
@@ -191,8 +191,7 @@ export default class IncorporationDateTime extends Mixins(DateMixin) {
   private constructDate (): void {
     if (this.isFutureEffective) {
       // Format the selected date string and create Date
-      const dateString = this.dateText.replace(/-/g, ', ')
-      const dateToValidate = new Date(dateString)
+      const dateToValidate = this.yyyyMmDdToDate(this.dateText)
```
severinbeauvais, line 194: Sample date text: "2022-09-05"

src/components/Incorporation/IncorporationDateTime.vue
```diff
@@ -244,28 +243,28 @@ export default class IncorporationDateTime extends Mixins(DateMixin) {
...
-    return this.dateToYyyyMmDd(date)
+    return this.dateToYyyyMmDd(this.getCurrentJsDate)
   }
 
   /** The maximum date that can be entered (today + 10 days). */
   private get maxDate (): string {
-    const date = new Date()
+    const date = new Date(this.getCurrentJsDate) // make a copy
     date.setDate(date.getDate() + 10)
     return this.dateToYyyyMmDd(date)
```
severinbeauvais, line 253: Need to make a copy so we don't change the original date object.

src/rules/date-rules.ts
```diff
@@ -9,10 +9,11 @@ export const DateRules = {
   }
 }
 
-// validation helper methods
+// validation helper method(s)
 export const DateRuleHelpers = {
   isBetweenDates (minDate: Date, maxDate: Date, dateStrToValidate: string): boolean {
     if (!dateStrToValidate) { return true }
-    return (new Date(dateStrToValidate) >= minDate) && (new Date(dateStrToValidate) <= maxDate)
+    const date = new Date(dateStrToValidate) // eg, September 5, 2022
```
severinbeauvais, line 16: I don't like this one but there's nothing I can do about it (other than refactoring the shared date picker).

The problem is that the shared date picker returns this date like this: "September 5, 2022". But this statement converts it to a Date object according to the browser's local timezone, which may be incorrect or spoofed, so the date could be Sept 4 or Sept 6.

One possibility to solve this is to convert the string version of the date to a UTC date (using known tz offset). However, we can't just compare against an array of month names, since the locale could be a different language :(

A better solution would be for the shared date picker to return a date as YYYY-MM-DD. Even then, this code would be ugly as we can't use the Date Mixin to convert dates so we'd have to copy the needed date mixin methods locally.

So I think we just leave this. Worst case, the date picker rejects a user's date selection, which doesn't break anything on our end :/

src/views/DissolutionFirm/DissolutionFirm.vue
```diff
@@ -372,33 +372,33 @@ export default class DissolutionFirm extends Mixins(DateMixin, EnumMixin) {
     )
   }
 
-  /** The minimum start date that can be entered (greater than registration date). */
-  private get startDateMin (): Date {
-    const date = this.apiToDate(this.getBusinessFoundingDate)
-    date.setDate(date.getDate() + 1)
-    return date
-  }
```
severinbeauvais, line 380: Moved below `dissolutionError()` to keep it with its siblings.

tests/unit/App.spec.ts
```diff
@@ -799,7 +799,8 @@ describe('Dissolution - Define Dissolution page for a BEN', () => {
                 business: {
                   identifier: 'BC0870803',
                   legalName: '0870803 B.C. LTD.',
-                  legalType: 'BEN'
+                  legalType: 'BEN',
+                  foundingDate: '2021-10-07T20:37:41+00:00'
```
severinbeauvais, line 803: Without this, the value is assigned `undefined` and causes breakage.

tests/unit/CompleteResolution.spec.ts
```diff
@@ -63,57 +63,42 @@ for (const test of completeResolutionTestCases) {
...
     })
 
-    it('renders intro section', async () => {
-      wrapper = wrapperFactory(
-        CompleteResolution,
-        null,
-        { entityType: test.entityType },
-        null,
-        DissolutionResources
-      )
```
severinbeauvais, line 85: These were all the same so I made them common.

Also, most tests here were'e async so I updated that.

tests/unit/CompleteResolution.spec.ts
```diff
@@ -63,57 +63,42 @@ for (const test of completeResolutionTestCases) {
...
         null,
-        { entityType: test.entityType },
+        {
+          entityType: test.entityType,
+          business: { foundingDate: '2021-10-07T20:37:41+00:00' },
+          createResolutionStep: {
+            givenName: 'Given',
+            additionalName: 'Additional',
+            familyName: 'Family'
+          }
```
severinbeauvais, line 77: This state data is needed for this component to display correctly.

tests/unit/IncorporationDateTime.spec.ts
```diff
@@ -22,6 +22,7 @@ document.body.setAttribute('data-app', 'true')
 describe('Incorporation Date Time component', () => {
   let wrapperFactory: any
   const today = new Date()
+  store.state.stateModel.currentJsDate = today
```
severinbeauvais, line 25: In the future, we may need to add this to other test suites.

src/rules/date-rules.ts
seeker25: I ran into this same problem

seeker25: Probably a good idea, unless you think the number of people using the site from another TZ is very small - I wonder what the stats are?

severinbeauvais: Well, some of BC is Mountain Time, and we probably have some users from across Canada (if not the world). I think it should be fixed, or at least kept track of.

seeker25 (approved):

src/rules/date-rules.ts
severinbeauvais: I changed my mind slightly. What we should do here is compare dates only (not date-times). I'll still create a new ticket for this.

severinbeauvais: New ticket for this: https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/13531

PR #443, opened 2022-09-13
13577 Cleanup before upgrade

severinbeauvais:
*Issue #:* bcgov/entity#13577

*Description of changes:*
- pre-upgrade (lint and other misc cleanup)
- app version = 4.2.1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-create-ui",
-  "version": "4.2.0",
+  "version": "4.2.1",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "business-create-ui",
-      "version": "4.2.0",
+      "version": "4.2.1",
       "dependencies": {
         "@babel/compat-data": "^7.11.0",
         "@bcrs-shared-components/bread-crumb": "1.0.2",
@@ -34,7 +34,7 @@
         "array.prototype.move": "0.0.4",
         "axios": "^0.21.1",
         "core-js": "^3.6.5",
-        "http-status-codes": "^1.4.0",
+        "http-status-codes": "^2.2.0",
         "launchdarkly-js-client-sdk": "^2.18.0",
         "lodash": "^4.17.20",
         "pdfjs-dist": "2.0.943",
@@ -11616,9 +11616,9 @@
       }
     },
     "node_modules/http-status-codes": {
-      "version": "1.4.0",
-      "resolved": "https://registry.npmjs.org/http-status-codes/-/http-status-codes-1.4.0.tgz",
-      "integrity": "sha512-JrT3ua+WgH8zBD3HEJYbeEgnuQaAnUeRRko/YojPAJjGmIfGD3KPU/asLdsLwKjfxOmQe5nXMQ0pt/7MyapVbQ=="
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/http-status-codes/-/http-status-codes-2.2.0.tgz",
+      "integrity": "sha512-feERVo9iWxvnejp3SEfm/+oNG517npqL2/PIA8ORjyOZjGC7TwCRQsZylciLS64i6pJ0wRYz3rkXLRwbtFa8Ng=="
     },
     "node_modules/https-browserify": {
       "version": "1.0.0",
@@ -21447,9 +21447,9 @@
       "integrity": "sha512-60TedWOHHuqTpaZsTni9yyc+FwH4qP/Kjt6QegK7Hv1G+tP/CJ7+K1mTRsrEJkjUq55Q0a7hjHTblv+Y+u4o0Q=="
     },
     "node_modules/vue-class-component": {
-      "version": "7.2.5",
-      "resolved": "https://registry.npmjs.org/vue-class-component/-/vue-class-component-7.2.5.tgz",
-      "integrity": "sha512-0CSftHY0bDTD+4FbYkuFf6+iKDjZ4h2in2YYJDRMk5daZIjrgT9LjFHvP7Rzqy9/s1pij3zDtTSLRUjsPWMwqg==",
+      "version": "7.2.6",
+      "resolved": "https://registry.npmjs.org/vue-class-component/-/vue-class-component-7.2.6.tgz",
+      "integrity": "sha512-+eaQXVrAm/LldalI272PpDe3+i4mPis0ORiMYxF6Ae4hyuCh15W8Idet7wPUEs4N4YptgFHGys4UrgNQOMyO6w==",
       "peerDependencies": {
         "vue": "^2.0.0"
       }
@@ -32299,9 +32299,9 @@
       }
     },
     "http-status-codes": {
-      "version": "1.4.0",
-      "resolved": "https://registry.npmjs.org/http-status-codes/-/http-status-codes-1.4.0.tgz",
-      "integrity": "sha512-JrT3ua+WgH8zBD3HEJYbeEgnuQaAnUeRRko/YojPAJjGmIfGD3KPU/asLdsLwKjfxOmQe5nXMQ0pt/7MyapVbQ=="
+      "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/http-status-codes/-/http-status-codes-2.2.0.tgz",
+      "integrity": "sha512-feERVo9iWxvnejp3SEfm/+oNG517npqL2/PIA8ORjyOZjGC7TwCRQsZylciLS64i6pJ0wRYz3rkXLRwbtFa8Ng=="
     },
     "https-browserify": {
       "version": "1.0.0",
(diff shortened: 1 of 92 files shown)


package-lock.json
```diff
@@ -21447,9 +21447,9 @@
       "integrity": "sha512-60TedWOHHuqTpaZsTni9yyc+FwH4qP/Kjt6QegK7Hv1G+tP/CJ7+K1mTRsrEJkjUq55Q0a7hjHTblv+Y+u4o0Q=="
     },
     "node_modules/vue-class-component": {
-      "version": "7.2.5",
-      "resolved": "https://registry.npmjs.org/vue-class-component/-/vue-class-component-7.2.5.tgz",
-      "integrity": "sha512-0CSftHY0bDTD+4FbYkuFf6+iKDjZ4h2in2YYJDRMk5daZIjrgT9LjFHvP7Rzqy9/s1pij3zDtTSLRUjsPWMwqg==",
+      "version": "7.2.6",
```
severinbeauvais, line 21450: This is a normal patch upgrade, seen here because I totally rebuild this package lock file. It should have no effect.

src/App.vue
```diff
@@ -444,7 +446,7 @@ export default class App extends Mixins(
   }
 
   /** Called when component is created. */
-  async created (): Promise<void> {
+  protected async created (): Promise<void> {
```
severinbeauvais, line 449: `created()` should not be public.

src/App.vue
```diff
@@ -498,7 +500,7 @@ export default class App extends Mixins(
   }
 
   /** Called when component is destroyed. */
-  private destroyed (): void {
+  protected beforeDestroy (): void {
```
severinbeauvais, line 503: This is the new lifecycle event name in Vue3 but it's back-compatible with Vue2.

src/components/Dissolution/AssociationDetails.vue
```diff
@@ -161,12 +161,12 @@ export default class AssociationDetails extends Mixins(CommonMixin, EnumMixin, D
   }
 
   /** Whether the address object is empty. */
-  private isEmptyAddress (address: AddressIF): boolean {
+  protected isEmptyAddress (address: AddressIF): boolean {
```
severinbeauvais, line 164: `isEmptyAddress()` needs to be protected so the template (HTML) can call it. (This is not yet enforced by the framework but this is correct and the previous code was incorrect.)

src/components/Dissolution/CooperativeType.vue
```diff
@@ -34,13 +34,14 @@ export default class CooperativeType extends Mixins(EnumMixin) {
...
 
+  // Prop
   @Prop({ default: false }) readonly showErrors!: boolean
 
   // Global getter
   @Getter getCooperativeType!: CoopTypes
 
   // Local properties
-  private readonly items: Array<any> = [
+  readonly items: Array<any> = [
```
severinbeauvais, line 44: This is safe without being private because external code still can't change it.

src/components/Incorporation/IncorporationDateTime.vue
```diff
@@ -120,28 +120,30 @@ export default class IncorporationDateTime extends Mixins(DateMixin) {
...
-  private selectMinute: string[] = []
-  private selectPeriod: string = 'AM'
+  protected selectHour = [] as string[]
+  protected selectMinute = [] as string[]
+  protected selectPeriod = 'AM'
 
   // Date Time Selectors
-  private hours: Array<string> = [...Array(12).keys()]
+  readonly hours: Array<string> = [...Array(12).keys()]
     .map(num => (++num).toLocaleString('en-US'))
```
severinbeauvais, line 138: This is a "constant" (but needs to be declared as readonly because it's a class member).

These selectors don't change. One more way of saying is they are static.

src/components/Incorporation/IncorporationDateTime.vue
```diff
@@ -152,21 +154,21 @@ export default class IncorporationDateTime extends Mixins(DateMixin) {
   }
 
   /** The array of validations rules for effective date hours field. */
-  private get hourRules (): Array<Function> {
+  get hourRules (): Array<Function> {
     return [
       v => this.dateText !== '' && (/^([1-9]|1[012])$/.test(v) || '')
```
severinbeauvais, line 159: This _cannot_ be a readonly array because it uses a variable in its code. So it's a getter instead.

src/components/Incorporation/IncorporationDateTime.vue
```diff
@@ -241,33 +243,33 @@ export default class IncorporationDateTime extends Mixins(DateMixin) {
...
     return this.dateToYyyyMmDd(date)
   }
 
   /** The minimum time that can be entered (now + 3 minutes). */
-  private minTime (): string {
+  protected minTime (): string {
     const date = this.getCurrentJsDate
     return new Date(date.getTime() + 180000)
       .toLocaleTimeString('en-US', { hour: '2-digit', minute: '2-digit', hour12: true })
   }
```
severinbeauvais, line 262: This could probably be a getter instead of a function but that's _really_ out of scope for this cleanup.

src/components/Registration/BusinessTypeConfirm.vue
```diff
@@ -53,14 +53,14 @@ export default class BusinessTypeConfirm extends Vue {
   protected checked = false
   protected label = 'General Partnership'
   protected labelSP = 'BC Sole Proprietorship / Doing Business As (DBA) Registration'
-  protected text = `I acknowledge that a General Partnership cannot be \
-    changed into a Sole Proprietorship (including DBA). If this is \
```
severinbeauvais, line 57: The latest ES Lint tried to reformat this in a way that's incompatible with our current framework so I manually reformatted this into something ES Lint likes.

src/components/common/BusinessContactInfo.vue
```diff
@@ -113,8 +113,8 @@ export default class BusinessContactInfo extends Mixins(CommonMixin) {
   // Rules for template
   readonly Rules = Rules
 
-  private contact: ContactPointIF = this.initialValue
-  private formValid: boolean = false
+  protected contact = this.initialValue
```
severinbeauvais, line 116: `contact` gets its type from `initialValue`.

src/components/common/NameRequestInfo.vue
```diff
@@ -104,7 +104,7 @@
               class="pt-0 mt-0"
               v-model="hasNameTranslation"
               id="name-translation-checkbox"
-              @click.native="confirmNameTranslation()"
```
severinbeauvais, line 107: This is not supported in Vue3 but I tested this and this code works correctly without it.

src/components/common/OfficeAddresses.vue
```diff
@@ -243,34 +243,35 @@ export default class OfficeAddresses extends Mixins(CommonMixin) {
...
+  /**
+   * State of the checkbox for determining whether or not the mailing address is the same as the delivery address.
+   * For Records Office.
+   */
+  private inheritRecMailingAddress = true
 
-  // State of the checkbox for determining whether the Record address is the same as the Registered address
-  private inheritRegisteredAddress: boolean = true
+  /** State of the checkbox for determining whether the Record address is the same as the Registered address. */
+  private inheritRegisteredAddress = true
```
severinbeauvais, line 267: The comment in JS Doc format allows VS Code to display the comment when hovering over the variable.

src/main.ts
```diff
@@ -43,17 +44,17 @@ async function start () {
   // must come first as inits below depend on config
   await fetchConfig()
 
-  if (window['sentryEnable'] === 'true') {
+  if ((window as any).sentryEnable === 'true') {
```
severinbeauvais, line 47: ES Lint wanted to change this to `window.sentryEnable` but the window object doesn't have that property and this caused an error.

The correct to fix this is to extend the window object to declare any additional properties ... I'm calling that out of scope for this cleanup.

tests/unit/DissolutionStatement.spec.ts
```diff
@@ -5,10 +5,9 @@ import { CorpTypeCd, DissolutionStatementTypes } from '@/enums'
 import { DissolutionResources } from '@/resources'
 
 // Input field selectors to test changes to the DOM elements.
-const typeSelector: string = '#dissolution-statement-'
-const summaryErrorSelector: string = '.invalid-section'
```
severinbeauvais, line 9: Wasn't used.

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
```diff
@@ -28,16 +28,16 @@ export interface NameRequestApplicantIF {
...
 }
 
 // NB: use cloneDeep when assigning EmptyOrgPerson
 export const EmptyNameRequest: NameRequestIF = {
   nrNumber: '',
   entityType: null,
-  details: {},
-  applicant: {},
+  details: {} as NameRequestDetailsIF,
+  applicant: {} as NameRequestApplicantIF,
```
severinbeauvais, line 41: These should probably be null and handled accordingly in code but that's out of scope for this cleanup.

What I did here is remove the `{}` union type from `details` to pre-empt an error where `details.approvedName` was not recognized as valid.

As far as I can tell, the code runs normally atm.

src/mixins/filing-template-mixin.ts
```diff
@@ -368,7 +369,7 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
             : {}
         },
         nameRequest: {
-          legalName: this.getNameRequest.details['approvedName'],
```
severinbeauvais, line 371: This was a sloppy way to work around a type issue.

severinbeauvais: Using `(myObject as any).myProperty` is also not ideal, but lint can flag variables used as any so we could find and fix these in the future. Plus there aren't a lot of places where "as any" is used 😅

tests/unit/FileUploadPreview.spec.ts
```diff
@@ -35,17 +35,6 @@ function myArrayBuffer () {
...
- * Utility method to get around with the timing issues
- */
-async function waitForUpdate (numTimesToFlushPromises) {
-  await Vue.nextTick()
-  for (let i = 0; i < numTimesToFlushPromises; i++) {
-    await flushPromises()
-  }
-  await Vue.nextTick()
-}
-
```
severinbeauvais, line 48: I moved this to its own TS file so it can be imported in a couple of test suites (instead of duplicated).

severinbeauvais:
@seeker25 @vysakh-menon-aot @cameron-freshworks @lekshmimallika-aot 

Please note the updates made by the next version of ES Lint, which I fixed here pre-emptively. Please follow this code style from now on (to prevent future cleanup).

lewischenstudio (approved):
This is a lot of cleanups. Thanks.

PR #444, opened 2022-09-14
13577 Upgrade to Vue3 (WIP)

severinbeauvais:
*Issue #:* bcgov/entity#13577

*Description of changes:*
- updated packages (WIP) for Vue3
- app version = 5.0.0
- DO NOT MERGE


.eslintrc.js
```diff
@@ -5,8 +5,9 @@ module.exports = {
   },
   extends: [
     'plugin:vue/essential',
+    // 'plugin:vue/recommended',
     '@vue/standard',
-    '@vue/typescript'
+    '@vue/typescript/recommended'
```
severinbeauvais, line 10: I'm thinking lines 8 and 10 are worthy changes but I'm trying to limit the impact of the upgrade -- we can use Vue Recommended rules, and fix what it finds, later.

.eslintrc.js
```diff
@@ -5,8 +5,10 @@ module.exports = {
   },
   extends: [
     'plugin:vue/essential',
+    // 'plugin:vue/recommended',
     '@vue/standard',
     '@vue/typescript'
+    // '@vue/typescript/recommended'
```
severinbeauvais, line 11: I'm thinking lines 9 and 11 are worthy changes but I'm trying to limit the impact of the upgrade -- we can use recommended lint rules, and fix what it finds, later.

.browserslistrc
```diff
@@ -1,3 +1,3 @@
 > 1%
 last 2 versions
-not ie <= 10
+not dead
```
severinbeauvais, line 3: This _should_ remove polyfills for IE11 and maybe other old browsers.

I saw somewhere that Vue3 is not compatible with IE11 anyway, so I think this is a good change.

package.json
```diff
@@ -13,78 +13,60 @@
...
-    "@bcrs-shared-components/document-delivery": "1.0.0",
-    "@bcrs-shared-components/effective-date-time": "1.0.2",
-    "@bcrs-shared-components/enums": "1.0.19",
-    "@bcrs-shared-components/help-business-number": "1.0.0",
-    "@bcrs-shared-components/interfaces": "1.0.39",
-    "@bcrs-shared-components/mixins": "1.0.6",
-    "@bcrs-shared-components/nature-of-business": "1.0.3",
-    "@bcrs-shared-components/staff-comments": "1.2.0",
-    "@bcrs-shared-components/staff-payment": "2.0.1",
-    "@bcrs-shared-components/web-chat": "2.0.0",
```
severinbeauvais, line 34: I had to remove these dependencies (and sbc-common-components below) because they are not upgraded and they cause package conflicts.

package.json
```diff
@@ -13,78 +13,60 @@
...
-    "@mdi/font": "^5.5.55",
```
severinbeauvais, line 35: I moved this to Dev Dependencies as per some examples I saw. It will be pretty obvious if this change doesn't work (once the project builds and runs).

package.json
```diff
@@ -13,78 +13,60 @@
...
     "@sentry/browser": "^5.21.1",
     "@sentry/integrations": "^5.21.1",
+    "@vue/compat": "^3.2.39",
```
severinbeauvais, line 18: May not be needed if we don't need any Vue2 backward-compatibility features. See also `compatConfig` in vue.config.js.

If we go pure Vue3 then we probably don't need this package. To be determined later.

package.json
```diff
@@ -13,78 +13,60 @@
...
     "vue-affix": "^0.5.2",
     "vue-observe-visibility": "^1.0.0",
-    "vue-router": "^3.4.3",
+    "vue-router": "^4.1.5",
     "vue-the-mask": "^0.11.1",
     "vue2-filters": "^0.10.0",
     "vuelidate": "^0.7.5",
-    "vuetify": "^2.3.9",
-    "vuex": "^3.5.1"
+    "vuetify": "3.0.0-beta.4",
```
severinbeauvais, line 34: We really should postpone this upgrade until Vuetify is out of beta.

package.json
```diff
@@ -13,78 +13,60 @@
...
     "flush-promises": "^1.0.2",
     "intersection-observer": "^0.12.0",
+    "s-vuex-class": "^0.4.1",
     "sass": "^1.26.10",
     "sass-loader": "^8.0.2",
     "sinon": "^9.0.3",
-    "typescript": "^3.9.7",
-    "vue-cli-plugin-vuetify": "^2.0.7",
+    "typescript": "^4.8.3",
+    "vue-class-component": "8.0.0-rc.1",
```
severinbeauvais, line 63: We really should postpone this upgrade until this is released.

Possibly, if we refactor everything from Class API to the other option (not the same as Composition API) then we won't need this package or vue-property-decorator below.

package.json
```diff
@@ -13,78 +13,60 @@
...
     "vue-plugin-helper-decorator": "^0.0.12",
-    "vue-property-decorator": "^8.5.1",
-    "vue-template-compiler": "2.6.12",
+    "vue-property-decorator": "10.0.0-rc.3",
     "vue-test-utils-helpers": "git+https://github.com/bcgov/vue-test-utils-helpers.git",
-    "vuetify-loader": "^1.6.0",
-    "vuex-class": "^0.3.2",
-    "vuex-module-decorators": "^0.16.1"
+    "vuex-composition-helpers": "^1.1.2",
+    "vuex-module-decorators": "^2.0.0",
```
severinbeauvais, line 69: There's a lot of discussion on whether the store decorators work in Vue3. (I think they do but I didn't get far enough to test them.)

Since Vuex is deprecated in Vue3, there's also discussion about using [**pinia**](https://vuejs.org/guide/scaling-up/state-management.html#pinia) instead.

package.json
```diff
@@ -13,78 +13,60 @@
...
+    "@vue/cli-service": "^5.0.8",
+    "@vue/compiler-sfc": "^3.2.39",
+    "@vue/eslint-config-standard": "^8.0.1",
+    "@vue/eslint-config-typescript": "^11.0.1",
+    "@vue/test-utils": "^2.0.2",
+    "eslint": "^8.23.1",
+    "eslint-plugin-vue": "^9.4.0",
     "flush-promises": "^1.0.2",
     "intersection-observer": "^0.12.0",
+    "s-vuex-class": "^0.4.1",
```
severinbeauvais, line 58: This may be a replacement or shim for the Vuex class decorators. See also `vuex-module-decorators` below.

package.json
```diff
@@ -13,78 +13,60 @@
...
-    "vue-property-decorator": "^8.5.1",
-    "vue-template-compiler": "2.6.12",
+    "vue-property-decorator": "10.0.0-rc.3",
     "vue-test-utils-helpers": "git+https://github.com/bcgov/vue-test-utils-helpers.git",
-    "vuetify-loader": "^1.6.0",
-    "vuex-class": "^0.3.2",
-    "vuex-module-decorators": "^0.16.1"
+    "vuex-composition-helpers": "^1.1.2",
+    "vuex-module-decorators": "^2.0.0",
+    "webpack-plugin-vuetify": "2.0.0-alpha.11"
```
severinbeauvais, line 70: I believe this replaces `vuetify-loader` in the old code.

vue.config.js
```diff
@@ -9,6 +9,25 @@ const aboutText1 = (appName && appVersion) ? `${appName} v${appVersion}` : ''
...
+        return {
+          ...options,
+          compilerOptions: {
+            compatConfig: {
+              MODE: 2
+            }
+          }
+        }
+      })
+  },
```
severinbeauvais, line 29: This config was recommended by the Vue3 migration side ([**step 3**](https://v3-migration.vuejs.org/migration-build.html#upgrade-workflow)).

If we go pure Vue3 then we probably don't need this config. To be determined later.

package.json
severinbeauvais: Ideally, we shouldn't have any shims or compatibility packages or other temporary code in our project (since that's tech debt).

PR #445, opened 2022-09-14
Feature 13412 - Allow Changing Completing Party in Registrations

ritvick:
*Issue #:* /bcgov/entity/issues/13412

*Description of changes:*
We want to keep the orgperson detail emplty in case of registration by a staff. So that staff will manually have to fill the information on behalf of the customer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "4.2.0",
+  "version": "4.2.1",
   "private": true,
   "appName": "Create UI",
   "sbcName": "SBC Common Components",
--- a/src/components/Registration/RegPeopleAndRoles.vue
+++ b/src/components/Registration/RegPeopleAndRoles.vue
@@ -224,7 +224,7 @@ export default class RegPeopleAndRoles extends Mixins(PeopleRolesMixin) {
     this.currentOrgPerson.officer.partyType = partyType
 
     // pre-populate Completing Party's name, email address and mailing address only if the logged in user is not staff (registries or sbc)
-    if (roleType === RoleTypes.COMPLETING_PARTY && partyType === PartyTypes.PERSON && (!this.isRoleStaff || !this.isSbcStaff)) {
+    if (roleType === RoleTypes.COMPLETING_PARTY && partyType === PartyTypes.PERSON && !(this.isRoleStaff || this.isSbcStaff)) {
       this.currentOrgPerson.officer.firstName = this.getUserFirstName || ''
       this.currentOrgPerson.officer.lastName = this.getUserLastName || ''
       this.currentOrgPerson.officer.email = this.getTombstone.userEmail
--- a/src/components/Registration/RegPeopleAndRoles.vue
+++ b/src/components/Registration/RegPeopleAndRoles.vue
@@ -223,8 +223,8 @@ export default class RegPeopleAndRoles extends Mixins(PeopleRolesMixin) {
     // assign party type (org or person)
     this.currentOrgPerson.officer.partyType = partyType
 
-    // pre-populate Completing Party's name, email address and mailing address
-    if (roleType === RoleTypes.COMPLETING_PARTY && partyType === PartyTypes.PERSON) {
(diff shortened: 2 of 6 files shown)


severinbeauvais (commented):
This looks fine but does it fully satisfy the ticket requirements?

Also, please update the app version and commit the package files.

ritvick:
> This looks fine but does it fully satisfy the ticket requirements?
> 
> Also, please update the app version and commit the package files.

Yeah it fulfils all the requirement, unless I am missing something? Can you be more specific what equirements I may have missed here?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "4.2.0",
+  "version": "4.2.1",
```
severinbeauvais, line 3: This project is [already at v4.2.1](https://github.com/bcgov/business-create-ui/blob/c1afbed036ccd93f75b7da4be7994e3ff79c2647/package.json#L3). Please rebase and then update app version again.

severinbeauvais:
> Yeah it fulfils all the requirement, unless I am missing something? Can you be more specific what equirements I may have missed here?

When I refer back to the ticket, I see that Completing Party should be blank and editable for SBC Staff (which is not the same as BCROS staff). I think this isn't done.

severinbeauvais:
I have added some reviewers to this ticket. I cannot be the sole reviewer for Entity UIs. From now on, I will wait for other reviewers to have a look at PRs before I do.

cc: @pwei1018 @davemck513

src/components/Registration/RegPeopleAndRoles.vue
```diff
@@ -223,8 +223,8 @@ export default class RegPeopleAndRoles extends Mixins(PeopleRolesMixin) {
     // assign party type (org or person)
     this.currentOrgPerson.officer.partyType = partyType
 
-    // pre-populate Completing Party's name, email address and mailing address
-    if (roleType === RoleTypes.COMPLETING_PARTY && partyType === PartyTypes.PERSON) {
+    // pre-populate Completing Party's name, email address and mailing address only if the logged in user is not staff
+    if (roleType === RoleTypes.COMPLETING_PARTY && partyType === PartyTypes.PERSON && !this.isRoleStaff) {
```
lambert-alex, line 227: I think we should have a unit test checking this new functionality.

ritvick: Hi @lambert-alex ,
I am mostly backend side and finiding it difficult to how to put unit test for this. 
I checked the unit tests folder but was not able to locate tests which are related to logged in user behaviour. 
Can you point to some if they already exists ( Probably I would not have looked in detail) ?

severinbeauvais: In my opinion, the best way to handle this (and the lack of tests elsewhere) is to create a "staff" test suite where we can mount different components and verify that they display/hide the expected elements.

As this ^^ is beyond the scope of this ticket, Ritvick, please create a Tech Debt ticket (in Entities team space) to do this.

severinbeauvais (approved):
LGTM.

Please update your ticket with the current Sprint and Releases. Also please add a Test Notes comment identifying what to test (for QA purposes). Thanks!

lewischenstudio (approved):
LGTM 👍 It would be nice to add a test case for SbcStaff or add a test note for QA.

src/components/Registration/RegPeopleAndRoles.vue
ritvick: Created Tech debt ticket here: https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/13614

ritvick:
Added test Notes in the original ticket.

lambert-alex:
> Added test Notes in the original ticket.

Thank you!

src/components/Registration/RegPeopleAndRoles.vue
severinbeauvais: @lambert-alex OK to resolve this conversation?

PR #460, opened 2022-10-07
sp/gp diss - completing party editable for sbc staff

kialj876:
*Issue #:* /bcgov/entity#13822

*Description of changes:*

### DIFF
--- a/src/store/getters/state-getters.ts
+++ b/src/store/getters/state-getters.ts
@@ -190,8 +190,7 @@ export const isPremiumAccount = (state: StateIF): boolean => {
 
 /** Whether the user is SBC Staff (which is not the same as Staff). */
 export const isSbcStaff = (state: StateIF): boolean => {
-  const orgInfo = getOrgInformation(state)
-  return (orgInfo?.orgType === AccountTypes.SBC_STAFF)
+  return (getAccountInformation(state).accountType === AccountTypes.SBC_STAFF)
 }
 
 /** The Org Information object. */
--- a/src/store/getters/state-getters.ts
+++ b/src/store/getters/state-getters.ts
@@ -98,11 +98,6 @@ export const isGovAccountUser = (state: StateIF): boolean => {
   return getTombstone(state).authRoles.includes('gov_account_user')
 }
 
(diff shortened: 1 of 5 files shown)


src/views/DissolutionFirm/DissolutionFirm.vue
```diff
@@ -361,6 +362,20 @@ export default class DissolutionFirm extends Mixins(DateMixin) {
...
+    console.log(currentAccount)
+    if (!currentAccount) return false
+    try {
+      console.log(JSON.parse(currentAccount).accountType)
+      return JSON.parse(currentAccount).accountType === AccountTypes.SBC_STAFF
+    } catch (error) {
+      console.log('Error parsing current account =', error)
+      return false
+    }
+  }
```
severinbeauvais, line 377: As in Edit UI, this already exists in App.vue.

Just add a new getter to state-getters.ts.

src/store/getters/state-getters.ts
```diff
@@ -100,8 +100,7 @@ export const isGovAccountUser = (state: StateIF): boolean => {
 
 /** Whether the user is SBC Staff (which is not the same as Staff). */
 export const isSbcStaff = (state: StateIF): boolean => {
-  const orgInfo = getOrgInformation(state)
-  return (orgInfo?.orgType === 'SBC_STAFF')
+  return state.stateModel.accountInformation?.accountType === AccountTypes.SBC_STAFF
```
severinbeauvais, line 103: You might as well use `getAccountInformation` getter here instead of dereferencing the state.

severinbeauvais: And maybe move this getter near `isPremiumAccount` in this file to keep account getters together.

kialj876: sure

severinbeauvais (approved):
Ok to merge after cleanup.

severinbeauvais:
Use getAccountInformation instead of getOrgInformation ?

kialj876:
> Use getAccountInformation instead of getOrgInformation ?

sure. This is why shared composables are nice haha. One place with the logic for everywhere instead of all these different things that have a lot of the same data

PR #463, opened 2022-10-31
13896 Pages should gray out after clicking Save and Resume Later and File Now, to avoid multiple clicks

lambert-alex:
*Issue #:* /bcgov/entity#13896

*Description of changes:*

Note to consistently see this error you have to set throttling in dev tools

 - Upon clicking Save and Resume and File Now removed setSaving to false because if the page is slow to redirect then the buttons become reenabled.
 - App version = 3.8.13 -> 3.8.14
 - Related PR [Edit UI](https://github.com/bcgov/business-edit-ui/pull/408)

### DIFF
--- a/src/components/common/Actions.vue
+++ b/src/components/common/Actions.vue
@@ -197,7 +197,6 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
       return
     }
 
-    this.setIsSavingResuming(false)
     this.emitGoToDashboard()
   }
 
@@ -219,7 +218,6 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
         !this.isValidDateTime(this.getEffectiveDateTime.effectiveDate)
       ) {
         this.setEffectiveDateTimeValid(false)
-        this.setIsFilingPaying(false)
 
         // don't call window.scrollTo during Jest tests because jsdom doesn't implement it
         if (!this.isJestRunning) window.scrollTo({ top: 1250, behavior: 'smooth' })
--- a/src/components/common/Actions.vue
+++ b/src/components/common/Actions.vue
@@ -181,6 +181,8 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
   protected async onClickSaveResume (): Promise<void> {
     // prevent double saving
     if (this.isBusySaving) return
+    // If Save and Resume is successful setIsFilingPaying should't be reset to false,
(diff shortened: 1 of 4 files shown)


src/components/common/Actions.vue
```diff
@@ -219,7 +218,6 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
         !this.isValidDateTime(this.getEffectiveDateTime.effectiveDate)
       ) {
         this.setEffectiveDateTimeValid(false)
-        this.setIsFilingPaying(false)
```
lambert-alex, line 222: This is done again on line 224

severinbeauvais: Please add comment in code here and line 200 why we don't want to clear the flags.

lambert-alex: Will do!

ChrisGabel23 (approved):

src/components/common/Actions.vue
```diff
@@ -181,6 +181,8 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
   protected async onClickSaveResume (): Promise<void> {
     // prevent double saving
     if (this.isBusySaving) return
+    // If Save and Resume is successful setIsFilingPaying should't be reset to false,
+    // this prevent buttons from being re-enabled if the page is slow to redirect.
```
severinbeauvais, line 185: This should be on line 202, right?

src/components/common/Actions.vue
```diff
@@ -212,14 +213,15 @@ export default class Actions extends Mixins(DateMixin, FilingTemplateMixin, Name
     if (this.isApplicationValid) {
       // prevent double saving
       if (this.isBusySaving) return
+      // If File and Pay is successful setIsFilingPaying should't be reset to false,
+      // this prevent buttons from being re-enabled if the page is slow to redirect.
```
severinbeauvais, line 217: This should be on line 225, right?

severinbeauvais (approved):

src/components/common/Actions.vue
lambert-alex: I figured I should say it where isSavingResuming is set so anyone reading the code sees isSavingResuming is set to true vs if they see a comment about setSavingResume at the end of the function with no corresponding code.

severinbeauvais: OK, I agree with you. Resolving this conversation.

PR #429, opened 2022-07-28
13058 COLIN API failure - Front and Backend Change validation of the first name field in Coop IA UI and/or Corp Party table

lewischenstudio:
*Issue #:* [/bcgov/entity#13058](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/bcgov/entity/13058)

*Description of changes:*
- change the max length of "firstName" to 20
- change UI before the schema change

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the bcrs-entities-create-ui license (Apache 2.0).

PR #464, opened 2022-11-01
14119 Revert first name and middle name max lengths

severinbeauvais:
*Issue #:* bcgov/entity#14119

This PR reverts the following change: https://github.com/bcgov/business-create-ui/pull/429

*Description of changes:*

First commit:
- revert first name max length to 30
- revert middle name length to 30
- updated unit tests
- app version = 4.3.5

Second commit:
- bugfix: moved setCompletingParty after auth data is fetched (otherwise isRoleStaff is false)
- used direct getter for completing party's email

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -634,9 +634,6 @@ export default class App extends Mixins(
         console.log('Launch Darkly update error =', error) // eslint-disable-line no-console
       })
 
-      // set completing party before draft filing dissolution create
-      this.setCompletingParty(this.getCompletingParties())
-
       // fetch the draft filing and resources
       try {
         if (this.getBusinessId) {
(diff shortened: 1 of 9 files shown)


argush3 (approved):

src/App.vue
```diff
@@ -634,9 +634,6 @@ export default class App extends Mixins(
         console.log('Launch Darkly update error =', error) // eslint-disable-line no-console
       })
 
-      // set completing party before draft filing dissolution create
-      this.setCompletingParty(this.getCompletingParties())
-
```
severinbeauvais, line 639: At this point in the flow, `isRoleStaff` is false because the auth info has not been fetched. Because of this, `getCompletingParties()` won't return the correct info for staff.

I know the comment says "before draft dissolution create" but I could find nothing in draft dissolution create code that depends on completing party, so I simply moved this code to AFTER the call to `handleDraftDissolution()` on line 643.

src/App.vue
```diff
@@ -670,6 +667,9 @@ export default class App extends Mixins(
         throw error // go to catch()
       }
 
+      // set completing party
+      this.setCompletingParty(this.getCompletingParties())
```
severinbeauvais, line 671: Auth info (ie, whether user is staff) was fetched in `handleDraftDissolution()` or `handleDraftApplication()`, called above.

PR #466, opened 2022-11-07
14160

ketaki-deodhar:
*Issue #:* /bcgov/entity#14160

*Description of changes:*

Update first and middle name max length validation rules for BENs/COOPs to only allow for name length <= 20 and respective unit test

### DIFF
--- a/src/components/common/AddEditOrgPerson.vue
+++ b/src/components/common/AddEditOrgPerson.vue
@@ -73,7 +73,7 @@
                   label="First Name"
                   id="person__first-name"
                   v-model.trim="orgPerson.officer.firstName"
-                  :rules="enableRules ? Rules.FirstNameRules : []"
+                  :rules="enableRules ? Rules.FirstNameRules20Char : []"
                   :readonly="isCompletingParty && !isRoleStaff && isTypeCoop"
                 />
                 <v-text-field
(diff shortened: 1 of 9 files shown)


src/components/common/AddEditOrgPerson.vue
```diff
@@ -82,7 +82,7 @@
                   label="Middle Name (Optional)"
                   id="person__middle-name"
                   v-model.trim="orgPerson.officer.middleName"
-                  :rules="enableRules ? Rules.MiddleNameRules: []"
+                  :rules="enableRules ? Rules.MiddleNameRules20Char: []"
```
severinbeauvais, line 85: I think it would actually be better to make firms the special case, since FirstNameRules is used for other things (like Signing Party and Custodian of Records).

Otherwise the implementation looks OK to me.

tests/unit/AddEditOrgPerson.spec.ts
```diff
@@ -319,8 +319,8 @@ describe('Add/Edit Org/Person component', () => {
 
     const messages = wrapper.findAll('.v-messages__message')
     expect(messages.length).toBe(3)
-    expect(messages.at(0).text()).toBe('Cannot exceed 30 characters')
-    expect(messages.at(1).text()).toBe('Cannot exceed 30 characters')
+    expect(messages.at(0).text()).toBe('Cannot exceed 20 characters')
+    expect(messages.at(1).text()).toBe('Cannot exceed 20 characters')
     expect(messages.at(2).text()).toBe('Cannot exceed 30 characters')
```
severinbeauvais, line 324: Please add a set of unit tests to verify this for firms vs others.

lambert-alex (approved):
LGTM, could you just link the ticket in the description

tests/unit/AddEditOrgPerson.spec.ts
ketaki-deodhar: added unit test specific to firms

src/components/common/AddEditOrgPerson.vue
ketaki-deodhar: This makes sense as only firms are kind of a special case. Updated accordingly.

PR #468, opened 2022-11-08
14161: Create UI - file uploader pdf page size validation issue

lewischenstudio:
*Issue #:* /bcgov/entity#14161

*Description of changes:*
- page validation checks all pages of a pdf file
- added test case when the second page of a pdf file has invalid size 
- added a test file "invalidSize.pdf"


src/mixins/document-mixin.ts
```diff
@@ -63,12 +63,16 @@ export default class DocumentMixin extends Vue {
     const pdfBufferData = await file.arrayBuffer()
     const pdfData = new Uint8Array(pdfBufferData) // put it in a Uint8Array
     const pdf = await pdfjsLib.getDocument({ data: pdfData })
-    const p1 = await pdf.getPage(1)
-    const [x, y, w, h] = p1._pageInfo.view
-    const width = w - x
-    const height = h - y
-    const isvalidPageSize = (width / pageSizeInfo.pointsPerInch === pageSizeInfo.width) &&
-      (height / pageSizeInfo.pointsPerInch === pageSizeInfo.height)
+    let isvalidPageSize = true
```
ChrisGabel23, line 66: Not sure how picky we are about case, but shouldn't it be "isValidPageSize"? (capital V)

src/mixins/document-mixin.ts
```diff
@@ -63,12 +63,16 @@ export default class DocumentMixin extends Vue {
...
+    for (let pageNum = 1; pageNum <= pdf.numPages; pageNum++) {
+      const p1 = await pdf.getPage(pageNum)
+      const [x, y, w, h] = p1._pageInfo.view
+      const width = w - x
+      const height = h - y
+      isvalidPageSize = (width / pageSizeInfo.pointsPerInch === pageSizeInfo.width) &&
+        (height / pageSizeInfo.pointsPerInch === pageSizeInfo.height)
+      if (!isvalidPageSize) return isvalidPageSize
```
severinbeauvais, line 74: You might as well just return False here, since you've found an invalid page.

Also, you might as well return True on line 76.

severinbeauvais (commented):
Looks good.

Would it be possible to add a test suite for the PDF things?

argush3:
We'll need to add a test that tests the scenario that we are seeing where the first page was a valid size and a subsequent one wasn't.   There are already existing tests that test page size like the one below.  You can either update this pdf or add another test.

https://github.com/bcgov/business-create-ui/blob/6676d5a2271500a47fe35185d3b793fa0d3596a7/tests/unit/FileUploadPreview.spec.ts#L195-L216

src/mixins/document-mixin.ts
ChrisGabel23: Personally, I like coding things as false by default, then whatever's being tested must prove itself as true.

severinbeauvais: That works, too.

A couple of things I look for:
- Easy to understand. In this code, I had to look around and run the code in my head to verify what values are being returned.
- I like short-circuiting results (which you've done here) 👍 so I'm ok with returning False when you've found an exception.

src/mixins/document-mixin.ts
lewischenstudio: Yes, it should be a capital V, but I'll remove this line in the next commit.

src/mixins/document-mixin.ts
lewischenstudio: It's done. It explicitly returns false in the loop.

tests/unit/FileUploadPreview.spec.ts
```diff
@@ -203,6 +203,29 @@ describe('FileUploadPreview component', () => {
     wrapper.destroy()
   }, 30000)
 
+  it('rejects when the second page is not an accepted size', async () => {
```
severinbeauvais, line 206: Does the test on line 163 use a PDF with multiple pages? I think it should.

It's not clear to me from the new PDF filename that the first page is the correct size but the second page is invalid. Compare this filename with the one on line 185.

Thanks for the test!

lewischenstudio: Yeah, it does look for all pages. Since the second page has the invalid size, if it only verifies the first page, then it would not return the error.

argush3: Test looks good mostly I think.

Agree with Severin's comments:
1.  that we should add at least another letter sized page to pdf used on L163
2. more clear naming pdf file name in this test

lewischenstudio: It's done. The invalid pdf file is renamed to "invalidSecondPage" and another valid page is now added to the valid pdf file "letterSize".

argush3 (approved):

ChrisGabel23 (approved):

tests/unit/FileUploadPreview.spec.ts
severinbeauvais: Perfect, thanks!

severinbeauvais (approved):
Good job, thanks!

lewischenstudio:
Thanks for the approvals.

PR #486, opened 2022-12-05
14429 Enable Hotjar - special branch for testing only

severinbeauvais:
*Issue #:* bcgov/entity#14429

**This code change is for testing only, since it must deployed on a real URL (cannot use localhost). I am merging this into a special branch that I will deploy to Dev directly. So this code isn't going into main branch yet, and the next main branch deployment will overwrite Dev. I will work with Vysakh to ensure this doesn't accidentally get deployed to Test :)**

*Description of changes:*
- added vue-hotjar package
- added Hotjar ID to config (only temporary -- it needs to be '' by default and provided by 1Password/config map)
- initialized Hotjar
- app version = 4.4.0

### DIFF
--- a/public/config/configuration.json
+++ b/public/config/configuration.json
@@ -7,6 +7,7 @@
   "BUSINESSES_URL": "https://dev.bcregistry.ca/business/auth/",
   "BUSINESS_CREATE_LD_CLIENT_ID": "5e6142dd29b3030a850285b3",
   "DASHBOARD_URL": "https://dev.bcregistry.ca/business/",
+  "HOTJAR_ID": "3274073",
   "KEYCLOAK_CONFIG_PATH": "/local-keycloak-config-url/keycloak.json",
   "LEGAL_API_URL": "https://legal-api-dev.apps.silver.devops.gov.bc.ca",
   "LEGAL_API_VERSION_2": "/api/v2",
--- a/src/main.ts
+++ b/src/main.ts
@@ -13,6 +13,7 @@ import Vue2Filters from 'vue2-filters' // needed by SbcFeeSummary
 import * as Sentry from '@sentry/browser'
 import * as Integrations from '@sentry/integrations'
 import VueObserveVisibility from 'vue-observe-visibility' // added to help with rendering of text area heights properly
+import Hotjar from 'vue-hotjar'
 
 // Styles
 // NB: order matters - do not change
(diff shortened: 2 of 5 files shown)


public/config/configuration.json
```diff
@@ -7,6 +7,7 @@
   "BUSINESSES_URL": "https://dev.bcregistry.ca/business/auth/",
   "BUSINESS_CREATE_LD_CLIENT_ID": "5e6142dd29b3030a850285b3",
   "DASHBOARD_URL": "https://dev.bcregistry.ca/business/",
+  "HOTJAR_ID": "3274073",
```
severinbeauvais, line 10: When I figure out how Hotjar/surveys can work for our purposes, then I will store this key in 1Password (and probably leave it blank here for local development).

At the moment, I don't know what's possible or what it looks like, and the design team needs to weigh is as well. So this is a real experiment, haha.

I'm not even sure that we will kick off surveys from here. We might want to kick them off from Filings UI after returning from an IA filing/payment.

src/utils/FetchConfig.ts
```diff
@@ -94,6 +94,9 @@ export async function FetchConfig (): Promise<any> {
     console.info('Set Siteminder Logout URL to: ' + siteminderLogoutUrl)
   }
 
+  const hotjarId: string = response.data.HOTJAR_ID;
+  (<any>window).hotjarId = hotjarId
```
severinbeauvais, line 98: This is to load the ID from the local config file (when run locally) or from the fetched config file (when run in Dev/Test/Prod).

src/main.ts
```diff
@@ -52,8 +53,15 @@ async function start () {
     })
   }
 
+  // Initialize Hotjar
+  if (window['hotjarId']) {
+    console.info('Initializing Hotjar...') // eslint-disable-line no-console
+    Vue.use(Hotjar, { id: window['hotjarId'], isProduction: true })
```
ChrisGabel23, line 59: I get nervous when I see "isProduction: true" in Dev code, but I'm assuming that nothing in Prod will be touched here. Can you confirm?

severinbeauvais: Confirmed!

This is a Hotjar option to enable tracking. This is what we do with HJ in other places we use it. I don't know yet if we need tracking. See https://www.npmjs.com/package/vue-hotjar.

ChrisGabel23 (approved):

PR #488, opened 2022-12-07
14473-CCC, ULC BEN, BC Incorporation agreement step is not reflected correctly in Review and Confirm step

ketaki-deodhar:
*Issue #:* /bcgov/entity#14473

*Description of changes:*
Fix - Review and Confirm step incorporation agreement section now displays only the summary description of selected agreement choice.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "4.4.0",
+  "version": "4.4.1",
   "private": true,
   "appName": "Create UI",
   "sbcName": "SBC Common Components",
--- a/src/components/common/AgreementType.vue
+++ b/src/components/common/AgreementType.vue
@@ -21,8 +21,8 @@
       </div>
     </template>
 
+    <!-- EDIT SECTION -->
     <template v-if="!isSummary">
-      <!-- EDIT SECTION -->
       <div v-if="isTypeBcUlcCompany || isTypeBcCcc" :class="{ 'invalid-section': showErrorSummary && getShowErrors }">
         <v-checkbox
           v-for="(item, index) in getIncorporationAgreementDocuments"
@@ -39,7 +39,6 @@
         </v-checkbox>
       </div>
 
-      <!-- EDIT SECTION -->
       <div class="py-8 px-6" v-else :class="{ 'invalid-section': showErrorSummary && getShowErrors }">
         <v-radio-group
           class="mt-0 pt-0"
--- a/src/components/common/AgreementType.vue
+++ b/src/components/common/AgreementType.vue
@@ -21,43 +21,45 @@
       </div>
     </template>
 
-    <!-- EDIT SECTION -->
-    <div v-if="isTypeBcUlcCompany || isTypeBcCcc" :class="{ 'invalid-section': showErrorSummary && getShowErrors }">
-      <v-checkbox
-        v-for="(item, index) in getIncorporationAgreementDocuments"
-        v-model="agreementType"
-        class="ml-6 mt-0 agreement-option-list"
-        :id="`agreement-type-${item.code}`"
-        :key="index"
-        :value="item.code"
-        @change="changeAgreementType()"
-      >
-        <template v-slot:label>
-          <div v-html="item.description" class="ml-6 py-4 agreement-option" />
-        </template>
-      </v-checkbox>
-    </div>
-
-    <!-- EDIT SECTION -->
-    <div class="py-8 px-6" v-else :class="{ 'invalid-section': showErrorSummary && getShowErrors }">
-      <v-radio-group
-        class="mt-0 pt-0"
-        hide-details
-        v-model="agreementType"
-        @change="changeAgreementType()"
-      >
-        <v-radio
+    <template v-if="!isSummary">
+      <!-- EDIT SECTION -->
+      <div v-if="isTypeBcUlcCompany || isTypeBcCcc" :class="{ 'invalid-section': showErrorSummary && getShowErrors }">
+        <v-checkbox
           v-for="(item, index) in getIncorporationAgreementDocuments"
+          v-model="agreementType"
+          class="ml-6 mt-0 agreement-option-list"
+          :id="`agreement-type-${item.code}`"
           :key="index"
           :value="item.code"
-          :id="`agreement-type-${item.code}`"
+          @change="changeAgreementType()"
         >
           <template v-slot:label>
-            <div v-html="item.description" class="agreement-option ml-6" />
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "4.4.0",
+  "version": "4.4.1",
```
severinbeauvais, line 3: @vysakh-menon-aot FYI - will need to update GH release and release ticket.

src/components/common/AgreementType.vue
```diff
@@ -21,43 +21,45 @@
...
+            :value="item.code"
+            :id="`agreement-type-${item.code}`"
+          >
+            <template v-slot:label>
+              <div v-html="item.description" class="agreement-option ml-6" />
+            </template>
+          </v-radio>
+        </v-radio-group>
+      </div>
+    </template>
```
severinbeauvais, line 62: Can you please show some screenshots of the edit page and the summary page?

src/components/common/AgreementType.vue
```diff
@@ -21,43 +21,45 @@
...
-    <div class="py-8 px-6" v-else :class="{ 'invalid-section': showErrorSummary && getShowErrors }">
-      <v-radio-group
-        class="mt-0 pt-0"
-        hide-details
-        v-model="agreementType"
-        @change="changeAgreementType()"
-      >
-        <v-radio
+    <template v-if="!isSummary">
+      <!-- EDIT SECTION -->
```
severinbeauvais, line 25: I see what you're doing: this whole section is the Edit Section. Looks good.

Please move the comment on line 25 above the template element so it's still displayed when this section is code-folded.

lewischenstudio:
Sorry Ketaki. I think the solution is good in using the radio button in the the AgreementType.vue, but I don't see how making this change will have the checkbox checked in the Review and Confirm step. Another question is that do we want UCL to have the radio button or some other entity type? I'm quite confused here.

src/components/common/AgreementType.vue
```diff
@@ -21,43 +21,45 @@
...
-            <div v-html="item.description" class="agreement-option ml-6" />
+            <div v-html="item.description" class="ml-6 py-4 agreement-option" />
           </template>
-        </v-radio>
-      </v-radio-group>
-    </div>
+        </v-checkbox>
+      </div>
+
+      <!-- EDIT SECTION -->
```
severinbeauvais, line 42: This comment doesn't make sense any more.

By the way, I had to look at this a couple of times before I realized that the above section is checkboxes but the section below is radio buttons.

src/components/common/AgreementType.vue
```diff
@@ -21,43 +21,45 @@
...
-      <v-radio-group
-        class="mt-0 pt-0"
-        hide-details
-        v-model="agreementType"
-        @change="changeAgreementType()"
-      >
-        <v-radio
+    <template v-if="!isSummary">
+      <!-- EDIT SECTION -->
+      <div v-if="isTypeBcUlcCompany || isTypeBcCcc" :class="{ 'invalid-section': showErrorSummary && getShowErrors }">
```
severinbeauvais, line 26: Specifically, this section is for entity types that have a single incorporation agreement document, right? It would be make the code more generic if you checked that instead. Thoughts?

severinbeauvais (approved):
This looks good.

I picked on a few things but I think it's OK to merge/deploy this ASAP and then change the other things later (future release).

severinbeauvais:
> Sorry Ketaki. I think the solution is good in using the radio button in the the AgreementType.vue, but I don't see how making this change will have the checkbox checked in the Review and Confirm step. Another question is that do we want UCL to have the radio button or some other entity type? I'm quite confused here.

I think she just updated the edit section. The summary section is unchanged and should display whatever the user selected.

lewischenstudio:
> > Sorry Ketaki. I think the solution is good in using the radio button in the the AgreementType.vue, but I don't see how making this change will have the checkbox checked in the Review and Confirm step. Another question is that do we want UCL to have the radio button or some other entity type? I'm quite confused here.
> 
> I think she just updated the edit section. The summary section is unchanged and should display whatever the user selected.

But this ticket is about the Review and Confirm (summary) section not displaying the checkbox even though this has been checked in the Incorporation Agreement (edit) section. The change here doesn't reflect the summary section, so I am bit uncertain of this change.

src/components/common/AgreementType.vue
ketaki-deodhar: I think ULC and CC have only one template so check box but other s have option to choose from

severinbeauvais:
> > > Sorry Ketaki. I think the solution is good in using the radio button in the the AgreementType.vue, but I don't see how making this change will have the checkbox checked in the Review and Confirm step. Another question is that do we want UCL to have the radio button or some other entity type? I'm quite confused here.
> > 
> > 
> > I think she just updated the edit section. The summary section is unchanged and should display whatever the user selected.
> 
> But this ticket is about the Review and Confirm (summary) section not displaying the checkbox even though this has been checked in the Incorporation Agreement (edit) section. The change here doesn't reflect the summary section, so I am bit uncertain of this change.

Both the summary and edit sections were displayed in this bug. A before screenshot and after screenshot should clarify this.

src/components/common/AgreementType.vue
ketaki-deodhar: Sure can be refactored.

src/components/common/AgreementType.vue
ketaki-deodhar: Please check these before and after screenshots.

lewischenstudio (approved):

ketaki-deodhar:
> > > > Sorry Ketaki. I think the solution is good in using the radio button in the the AgreementType.vue, but I don't see how making this change will have the checkbox checked in the Review and Confirm step. Another question is that do we want UCL to have the radio button or some other entity type? I'm quite confused here.
> > > 
> > > 
> > > I think she just updated the edit section. The summary section is unchanged and should display whatever the user selected.
> > 
> > 
> > But this ticket is about the Review and Confirm (summary) section not displaying the checkbox even though this has been checked in the Incorporation Agreement (edit) section. The change here doesn't reflect the summary section, so I am bit uncertain of this change.
> 
> Both the summary and edit sections were displayed at the same time in this bug. A before screenshot and after screenshot should clarify this.

Edit and summary, both sections were on Review and confirm step. I think Review and confirm step should not have option to select. It should only show summary of the selection in previous step.

src/components/common/AgreementType.vue
ketaki-deodhar: Done!

PR #493, opened 2023-01-03
14635 Cleanup before future work

severinbeauvais:
*Issue #:* bcgov/entity#14635

*Description of changes:*

Commit 1:
- replaced Mixins decorator with component mixins property
- replaced incorrect use of constants with enum
- added Volar-Vetur package + config
- misc cleanup
- app version = 4.5.0

Commit 2:
- renamed some folders and file for cleanliness
- renamed resource objects to pre-empt conflicts
- renamed step objects to pre-empt conflicts
- fixed 0-dollar fees bug
- misc cleanup


src/App.vue
```diff
@@ -247,14 +248,15 @@ import { SessionStorageKeys } from 'sbc-common-components/src/util/constants'
...
+    NameRequestMixin
+  ]
 })
-export default class App extends Mixins(
-  CommonMixin,
-  DateMixin,
-  FilingTemplateMixin,
-  NameRequestMixin
-) {
+export default class App extends Vue {
```
severinbeauvais, line 259: The reasons for doing this are manifold:
- Vue 2.7 (and future Vue 3) don't like the special decorators, esp. `Mixins`
- the VS Code Volar (Vue Language Features) plugin only intermittently recognizes mixins used the old way; often it would show errors on mixin functions
- in some cases, you can't extend more than 1 mixin at a time; this bypasses that

In a nutshell, the new code is better for the future.

src/enums/index.ts
```diff
@@ -1,24 +1,24 @@
 export * from './accountTypes'
+export * from './bulletListTypes'
 export * from './dissolutionStatementTypes'
 export * from './dissolutionTypes'
 export * from './effectOfOrders'
 export * from './entityState'
 export * from './entitySubtypes'
-export * from './filingStatus'
```
severinbeauvais, line 7: Simply alphabetized.

src/mixins/filing-template-mixin.ts
```diff
@@ -142,7 +141,7 @@ export default class FilingTemplateMixin extends DateMixin {
     // Build the main filing.
     const filing: IncorporationFilingIF = {
       header: {
-        name: INCORPORATION_APPLICATION,
```
severinbeauvais, line 145: Changed constant to an enum.

src/store/getters/state-getters.ts
```diff
@@ -6,8 +6,7 @@ import {
   DissolutionTypes,
   FilingNames,
   FilingTypes,
-  FilingTypesSubTitle,
-  NameRequestStates
```
severinbeauvais, line 10: Deleted unused enum.

src/views/DissolutionFirm/DissolutionFirm.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="mt-10" id="dissolution-firm-form">
```
severinbeauvais, line 2: Our convention is that components should have an ID and it should match the filename (and component name).

vysakh-menon-aot (approved):

package.json
```diff
@@ -63,6 +63,7 @@
     "@types/vuelidate": "^0.7.15",
     "@typescript-eslint/eslint-plugin": "^5.38.0",
     "@typescript-eslint/parser": "^5.38.0",
+    "@volar-plugins/vetur": "latest",
```
jonathan-longe, line 66: What do you think about declaring a specific version number instead of "latest"?

jonathan-longe (approved):
Looks good. I've added one suggestion to the package.json file.

ChrisGabel23 (approved):

package.json
severinbeauvais: This is how Travis did it in [Auth Web](https://github.com/bcgov/sbc-auth/blob/main/auth-web/package.json) and how I did it in [Filings UI](https://github.com/bcgov/business-filings-ui/blob/main/package.json).

I think "latest" probably makes good sense since we should generally be using the latest VS Code extension versions anyway (and this dependency should match that).

severinbeauvais:
Hey guys - I'm doing a bit more pre-feature cleanup and will commit to this branch/PR soon.

src/resources/Dissolution/index.ts
```diff
@@ -0,0 +1,7 @@
+export * from '@/resources/Dissolution/BC'
+export * from '@/resources/Dissolution/BEN'
+export * from '@/resources/Dissolution/CC'
+export * from '@/resources/Dissolution/CP'
+export * from '@/resources/Dissolution/GP'
+export * from '@/resources/Dissolution/SP'
+export * from '@/resources/Dissolution/ULC'
```
severinbeauvais, line 7: This file and the others below are actually only slightly changed from before, but I renamed the folder they're in.

src/services/pay-services.ts
```diff
@@ -23,7 +23,7 @@ export default class PayServices {
     return axios.get(url)
       .then(response => {
         const fees = response?.data
-        if (!fees.filingFees) {
```
severinbeauvais, line 26: This was throwing an error when fees.filingFees=0, ie,

image

src/services/pay-services.ts
```diff
@@ -23,7 +23,7 @@ export default class PayServices {
...
+        if (!fees) {
```
severinbeauvais, line 26: This checks that the `response.data` object exists.

PR #499, opened 2023-01-17
14798 Added applicant information (incremental)

severinbeauvais:
*Issue #:* bcgov/entity#14798

*Description of changes:*
- cleaned up feeFilingData re: Coop dissolutions
- implemented multiple rolesSubtitle
- parsed restoration data
- updated resources for multiple filingData and rolesSubtitle
- removed unneeded addIncorporator logic
- added disable role getters
- added restoration actions/mutations
- added restoration getters
- added StaffPayment to restoration last step (to test things)
- fixed a unit test
- app version = 4.5.4

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -183,50 +183,17 @@ import * as Dialogs from '@/dialogs'
 import * as Views from '@/views'
 
 // Mixins, interfaces, etc
-import {
-  CommonMixin,
-  DateMixin,
-  FilingTemplateMixin,
-  NameRequestMixin
-} from '@/mixins'
-import {
-  AccountInformationIF,
-  ActionBindingIF,
-  AddressIF,
-  BreadcrumbIF,
-  BusinessIF,
-  CompletingPartyIF,
-  ConfirmDialogType,
-  EmptyFees,
-  FilingDataIF,
-  OrgInformationIF,
-  ResourceIF,
-  StepIF
-} from '@/interfaces'
-import {
-  DissolutionResources,
-  IncorporationResources,
-  RegistrationResources,
-  RestorationResources,
-  getEntityDashboardBreadcrumb,
-  getMyBusinessRegistryBreadcrumb,
-  getRegistryDashboardBreadcrumb,
-  getSbcStaffDashboardBreadcrumb,
-  getStaffDashboardBreadcrumb
-} from '@/resources'
+import { CommonMixin, DateMixin, FilingTemplateMixin, NameRequestMixin } from '@/mixins'
+import { AccountInformationIF, ActionBindingIF, AddressIF, BreadcrumbIF, BusinessIF, CompletingPartyIF,
+  ConfirmDialogType, EmptyFees, FilingDataIF, OrgInformationIF, ResourceIF, StepIF } from '@/interfaces'
+import { DissolutionResources, IncorporationResources, RegistrationResources, RestorationResources,
+  getEntityDashboardBreadcrumb, getMyBusinessRegistryBreadcrumb, getRegistryDashboardBreadcrumb,
+  getSbcStaffDashboardBreadcrumb, getStaffDashboardBreadcrumb } from '@/resources'
 import { AuthServices, LegalServices, PayServices } from '@/services/'
 
 // Enums and Constants
-import {
-  CorpTypeCd,
-  FilingCodes,
-  FilingNames,
-  FilingStatus,
-  FilingTypes,
-  NameRequestStates,
-  RouteNames,
-  StaffPaymentOptions
-} from '@/enums'
+import { CorpTypeCd, FilingCodes, FilingNames, FilingStatus, FilingTypes, NameRequestStates, RouteNames,
+  StaffPaymentOptions } from '@/enums'
 import { SessionStorageKeys } from 'sbc-common-components/src/util/constants'
 
 @Component({
@@ -364,23 +331,23 @@ export default class App extends Vue {
   /** Data for fee summary component. */
   get feeFilingData (): Array<FilingDataIF> {
     let filingData = [] as Array<FilingDataIF>
+
     if (this.getFilingData) {
       filingData = cloneDeep(this.getFilingData)
-      if (this.isTypeCoop && this.getFilingData.length > 0) {
-        // Only set Future Effective and Priority to Special Resolution Fee
-        const specialResolutionFilingData = filingData.find(x => x.filingTypeCode === FilingCodes.SPECIAL_RESOLUTION)
-        if (specialResolutionFilingData) {
-          if (this.getStaffPaymentStep.staffPayment.option === StaffPaymentOptions.NO_FEE) {
-            filingData.forEach(x => {
-              x.waiveFees = true
-            })
-          } else {
-            specialResolutionFilingData.futureEffective = this.getEffectiveDateTime.isFutureEffective
-            specialResolutionFilingData.priority = this.getStaffPaymentStep.staffPayment.isPriority
-          }
+
+      if (this.isTypeCoop && this.isDissolutionFiling) {
+        const specialResolutionFilingData =
+          filingData.find(x => x.filingTypeCode === FilingCodes.SPECIAL_RESOLUTION)
+
+        if (this.getStaffPaymentStep.staffPayment.option === StaffPaymentOptions.NO_FEE) {
+          filingData.forEach(x => { x.waiveFees = true })
+        } else if (specialResolutionFilingData) {
+          // update futureEffective/priority on Special Resolution data only
+          specialResolutionFilingData.futureEffective = this.getEffectiveDateTime.isFutureEffective
+          specialResolutionFilingData.priority = this.getStaffPaymentStep.staffPayment.isPriority
         }
-      } else if (this.getFilingData[0]) {
-        // Avoid waiveFee with priority or futureEffective in the same request
+      } else if (filingData[0]) {
+        // avoid waiveFee with futureEffective/priority in the same request
         if (this.getStaffPaymentStep.staffPayment.option === StaffPaymentOptions.NO_FEE) {
           filingData[0].waiveFees = true
         } else {
@@ -389,6 +356,7 @@ export default class App extends Vue {
         }
       }
     }
+
     return filingData
   }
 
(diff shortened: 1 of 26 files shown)


src/App.vue
```diff
@@ -183,50 +183,17 @@ import * as Dialogs from '@/dialogs'
...
-  getRegistryDashboardBreadcrumb,
-  getSbcStaffDashboardBreadcrumb,
-  getStaffDashboardBreadcrumb
-} from '@/resources'
+import { CommonMixin, DateMixin, FilingTemplateMixin, NameRequestMixin } from '@/mixins'
+import { AccountInformationIF, ActionBindingIF, AddressIF, BreadcrumbIF, BusinessIF, CompletingPartyIF,
+  ConfirmDialogType, EmptyFees, FilingDataIF, OrgInformationIF, ResourceIF, StepIF } from '@/interfaces'
+import { DissolutionResources, IncorporationResources, RegistrationResources, RestorationResources,
+  getEntityDashboardBreadcrumb, getMyBusinessRegistryBreadcrumb, getRegistryDashboardBreadcrumb,
+  getSbcStaffDashboardBreadcrumb, getStaffDashboardBreadcrumb } from '@/resources'
```
severinbeauvais, line 191: I prefer all-on-one-line (with line breaks) instead of the old, tall import.

What do you guys think?

src/App.vue
```diff
@@ -364,23 +331,23 @@ export default class App extends Vue {
   /** Data for fee summary component. */
   get feeFilingData (): Array<FilingDataIF> {
     let filingData = [] as Array<FilingDataIF>
+
     if (this.getFilingData) {
       filingData = cloneDeep(this.getFilingData)
-      if (this.isTypeCoop && this.getFilingData.length > 0) {
-        // Only set Future Effective and Priority to Special Resolution Fee
-        const specialResolutionFilingData = filingData.find(x => x.filingTypeCode === FilingCodes.SPECIAL_RESOLUTION)
-        if (specialResolutionFilingData) {
```
severinbeauvais, line 372: In this old code, this logic is executed for ALL coop filings (per line 369).

`specialResolutionFilingData` only exists for [dissolution filings](https://github.com/bcgov/business-create-ui/blob/a689f82a319d7c16134eb4a31002efacbae2803e/src/resources/Dissolution/CP.ts#L21).

This means that non-dissolution coop filings (IAs) didn't run this code block, so they didn't update the filing data flags... fortunately we did not (yet) implement staff payment for coop IAs, so this didn't break.

Anyway, the new code is only for coop dissolution filings, so coop IAs will use the fallback code on new line 349 (and therefore work when we implement staff payment).

Also, the new code checks for the `specialResolutionFilingData` object on new line 344 (so we don't get an exception), but otherwise it works correctly and safely.

Coop dissolution fees before any staff payment options selected:
image

Coop dissolution fees with No Fee option selected:
image

Coop dissolution fees with a payment option and Priority checked:
image

As for why I touched this, it's because I had to deal with Full vs Limited restorations, which implement a similar array for full vs limited fees (and a subtitle). So I cleaned up this method, and then I ended up implementing full vs limited a different way.

src/components/common/AddEditOrgPerson.vue
```diff
@@ -54,11 +54,12 @@
             </p>
           </v-card>
 
+          <div v-if="isCompletingParty && !isRoleStaff && isTypeCoop" class="mt-8" />
```
severinbeauvais, line 57: I prefer this explicit white-space block (which is needed only when the block above is rendered), rather than an un-obvious conditional class on the form below. (It took me a while to understand why the conditional class existed.)

src/components/common/AddEditOrgPerson.vue
```diff
@@ -160,15 +160,14 @@
...
-                    <!-- *** TODO: disable conditionally? -->
                     <v-col cols="4" v-if="showHeirLegalRepRole">
                       <v-checkbox
                         id="heir-legal-rep-checkbox"
                         class="mt-0"
                         v-model="selectedRoles"
                         :value="RoleTypes.HEIR_LEGAL_REP"
                         :label="RoleTypes.HEIR_LEGAL_REP"
-                        :disabled="false"
+                        :disabled="disableHeirLegalRepRole"
```
severinbeauvais, line 170: New local getters (computeds) in mixin.

src/components/common/AddEditOrgPerson.vue
```diff
@@ -204,19 +203,19 @@
...
-                    <v-col cols="4" v-if="showOther">
+                    <v-col cols="4" v-if="showOtherRole">
                       <v-checkbox
                         id="other-checkbox"
                         class="mt-0"
                         v-model="selectedRoles"
                         :value="RoleTypes.OTHER"
                         :label="RoleTypes.OTHER"
-                        :disabled="false"
+                        :disabled="disableOtherRole"
```
severinbeauvais, line 218: Screenshot:

image

src/components/common/PeopleAndRoles.vue
```diff
@@ -189,17 +188,8 @@ export default class PeopleAndRoles extends Vue {
     // first assign empty org/person object
     this.currentOrgPerson = cloneDeep(EmptyOrgPerson)
 
-    if (roleType) {
-      // a role was provided - pre-select it
-      this.currentOrgPerson.roles = [{ roleType }]
-    } else if (this.validNumCompletingParty && !this.getPeopleAndRolesResource.addIncorporator) {
-      // only Director role is possible - pre-select it
```
severinbeauvais, line 196: We can no longer assume that this must be a Director if we already have the completing party and we don't need to add incorporator(s). There are too many other things to check and it's too complicated to generalize this.

Actually, I think "feature" was broken early on (probably when we allowed the replacement of the completing party) because I can't find any examples of adding a person whose only role can be Director.

src/enums/restorationTypes.ts
```diff
@@ -1,4 +1,4 @@
 export enum RestorationTypes {
-  FULL = 'FULL',
-  LIMITED = 'LIMITED'
+  FULL = 'fullRestoration',
+  LIMITED = 'limitedRestoration'
```
severinbeauvais, line 3: Per [schema](https://github.com/bcgov/business-schemas/blob/fceafb95400c69c9ceb07c4aecbdcc1ffea93f6d/src/registry_schemas/schemas/restoration.json#L25).

src/mixins/add-edit-org-person-mixin.ts
```diff
@@ -205,11 +210,36 @@ export default class AddEditOrgPersonMixin extends Vue {
...
-    if (this.isRestorationFiling) return false
+    if (this.isFullRestorationFiling || this.isLimitedRestorationFiling) return false
     // disable this role if it's the only role displayed
     return (!this.showCompletingPartyRole && !this.showIncorporatorRole && this.showDirectorRole)
   }
 
+  /** Whether the Heir or Legal Representative role should be disabled. */
+  get disableHeirLegalRepRole (): boolean {
+    return false
+  }
```
severinbeauvais, line 221: I added these getters, even though their return value is static, to keep the template consistent and for future use.

src/mixins/filing-template-mixin.ts
```diff
@@ -519,10 +522,9 @@ export default class FilingTemplateMixin extends DateMixin {
     this.setFoundingDate(draftFiling.business.foundingDate)
 
     // restore Restoration data
-
-    // *** TODO: restore date
-    // *** TODO: restore type
-    // *** TODO: restore expiry
+    this.setRestorationDate(draftFiling.restoration.date)
+    this.setRestorationType(draftFiling.restoration.type)
+    this.setRestorationExpiry(draftFiling.restoration.expiry || null)
```
severinbeauvais, line 527: I had to decide whether to do this, or this:
```
this.setRestoration({
   date: draftFiling.restoration.date,
   type: draftFiling.restoration.type,
   expiry: draftFiling.restoration.expiry || null
})
```

I chose the former because we're going to need those setters anyway, since those 3 properties are set from different components.

src/mixins/people-roles-mixin.ts
```diff
@@ -317,10 +318,12 @@ export default class PeopleRolesMixin extends Vue {
...
     return (
       this.validNumCompletingParty &&
-      this.validMinimumIncorporators &&
-      this.validMinimumDirectors &&
+      this.validNumIncorporators &&
+      this.validNumDirectors &&
       this.validDirectorCountry &&
       this.validDirectorProvince &&
+      this.validApplicantPerson &&
+      this.validApplicantOrg &&
```
severinbeauvais, line 326: New checks for applicants. These local getters (computeds) return True if the rule isn't defined, so this works fine for non-restoration filings.

src/store/actions/actions-model.ts
```diff
@@ -280,10 +277,6 @@ export const setCustodianOfRecords: ActionIF = ({ commit }, custodian: OrgPerson
   commit('mutateCustodianOfRecords', custodian)
 }
 
-//
-// Registration actions
-//
```
severinbeauvais, line 285: This section comment is a good idea but there are actions below that aren't for registrations.

src/store/getters/resource-getters.ts
```diff
@@ -24,7 +15,14 @@ export const getCompanyDisplayName = (state: StateIF): string => {
...
 export const getPeopleAndRolesResource = (state: StateIF): any => {
-  return state.resourceModel.peopleAndRoles
+  const peopleAndRoles = state.resourceModel.peopleAndRoles
+  if (isFullRestorationFiling(state)) {
+    peopleAndRoles.rolesSubtitle = peopleAndRoles.rolesSubtitle[0]
+  }
+  if (isLimitedRestorationFiling(state)) {
+    peopleAndRoles.rolesSubtitle = peopleAndRoles.rolesSubtitle[1]
+  }
+  return peopleAndRoles
```
severinbeauvais, line 25: Yes, this isn't super-general (filing-agnostic), but at least it keeps the components clean.

Ditto line 65 below.

src/views/Restoration/RestorationReviewConfirm.vue
```diff
@@ -60,6 +60,18 @@
...
+    <section id="staff-payment-section" class="mt-10" v-if="isRoleStaff">
+      <header>
+        <h2>Staff Payment</h2>
+        <p class="mt-4"></p>
+      </header>
+
+      <v-card flat class="mt-6">
+        <StaffPayment class="py-8 px-6" />
+      </v-card>
+    </section>
```
severinbeauvais, line 74: I added this so I could test that the staff payment options work. They do!

image

image

src/App.vue
ChrisGabel23: I prefer all-on-one-line as well.

JazzarKarim (approved):

ChrisGabel23 (approved):

PR #503, opened 2023-01-24
14429 Filing Survey dialog and launch filing survey

severinbeauvais:
*Issue #:* bcgov/entity#14429

*Description of changes:*
- added js-cookie package
- added hotjar-shared config for IA_SURVEY_ID
- fetch IA_SURVEY_ID
- added code to launch filing survey
- added code to set/get/remove cookie
- app version = 4.5.8
- updated unit tests

### DIFF
--- a/devops/vaults.json
+++ b/devops/vaults.json
@@ -57,5 +57,11 @@
             "base",
             "business-create-ui"
         ]
+    },
+    {
+        "vault": "hotjar",
+        "application": [
+            "shared"
+        ]
     }
 ]
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "4.5.7",
+  "version": "4.5.8",
   "private": true,
   "appName": "Create UI",
   "sbcName": "SBC Common Components",
@@ -38,6 +38,7 @@
     "axios": "^0.27.2",
     "core-js": "^3.25.2",
     "http-status-codes": "^2.2.0",
+    "js-cookie": "^3.0.1",
     "launchdarkly-js-client-sdk": "^2.22.1",
     "lodash": "^4.17.21",
     "pdfjs-dist": "2.0.943",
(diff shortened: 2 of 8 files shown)


devops/vaults.json
```diff
@@ -57,5 +57,11 @@
             "base",
             "business-create-ui"
         ]
+    },
+    {
+        "vault": "hotjar",
+        "application": [
+            "shared"
+        ]
```
severinbeauvais, line 65: This is a new vault in 1Password. The "shared" application (section) contains keys that are shared between projects. There are other sections specific to each project.

Example (key ids are not yet updated):

image

package.json
```diff
@@ -38,6 +38,7 @@
     "axios": "^0.27.2",
     "core-js": "^3.25.2",
     "http-status-codes": "^2.2.0",
+    "js-cookie": "^3.0.1",
```
severinbeauvais, line 41: There's also `vue-cookies` but this one is more cross-platform and still super-easy to use.

public/config/configuration.json
```diff
@@ -8,6 +8,7 @@
   "BUSINESS_CREATE_LD_CLIENT_ID": "5e6142dd29b3030a850285b3",
   "DASHBOARD_URL": "https://dev.bcregistry.ca/business/",
   "HOTJAR_ID": "",
+  "IA_SURVEY_ID": "",
```
severinbeauvais, line 11: When configured, checks whether to display Filing Survey dialog. Not configured when running locally.

src/App.vue
```diff
@@ -527,6 +530,28 @@ export default class App extends Vue {
...
+  /** Called to save or delete "hide survey" cookie. */
+  protected updateSurveyCookie (doNotShow: boolean): void {
+    const domain = window.location.hostname
+    if (doNotShow) {
+      // hide survey for 30 days
+      Cookies.set(this.IA_SURVEY_KEY, 'do-not-show', { domain, expires: 30 })
+    } else {
+      Cookies.remove(this.IA_SURVEY_KEY, { domain })
+    }
+  }
```
severinbeauvais, line 553: Sample cookies:

image

I'm expecting/hoping the Domain will be "dev.bcregistry.ca" or whatever. This is currently running locally.

The Expires field is a cookie feature: the cookie is auto-removed then the date/time is reached.

The other fields are default values and should be OK as-is.

PR #507, opened 2023-01-26
14429 Change "hide survey" from cookie to local storage

severinbeauvais:
*Issue #:* bcgov/entity#14429

_Sorry about all these changes. The business keeps "refining" their requirements. 😳_

*Description of changes:*
- change from cookie (which expires) to local storage (which doesn't)
- uninstall js-cookie package
- app version = 4.5.12

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -46,7 +46,7 @@
       :dialog="filingSurveyDialog"
       @no="filingSurveyDialog = false"
       @yes="filingSurveyDialog = false; launchFilingSurvey()"
-      @doNotShow="updateSurveyCookie($event)"
+      @doNotShow="doNotShowSurvey($event)"
     />
 
     <PaymentErrorDialog
@@ -170,7 +170,6 @@
 // Libraries
 import Vue from 'vue'
 import axios from 'axios'
-import Cookies from 'js-cookie'
 import { Component, Watch } from 'vue-property-decorator'
 import { Action, Getter } from 'vuex-class'
 import { StatusCodes } from 'http-status-codes'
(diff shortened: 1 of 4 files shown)


src/App.vue
```diff
@@ -539,22 +536,23 @@ export default class App extends Vue {
 
   /** Opens Auth Web in a new tab with the survey query param. */
   protected launchFilingSurvey (): void {
-    const iaSurveyId = sessionStorage.getItem('IA_SURVEY_ID')
-
     // safety check
```
severinbeauvais, line 539: This method should not be called if IA Survey ID was 0 in the first place, but we still don't want to open a new window with an invalid id.

src/App.vue
```diff
@@ -539,22 +536,23 @@ export default class App extends Vue {
...
   }
 
-  /** Called to save or delete "hide survey" cookie. */
-  protected updateSurveyCookie (doNotShow: boolean): void {
+  /** Called to update "do not show again" state. */
+  protected doNotShowSurvey (doNotShow: boolean): void {
     if (doNotShow) {
-      // hide survey for 30 days
-      Cookies.set(this.IA_SURVEY_KEY, 'do-not-show', { domain: this.domain, expires: 30 })
+      // safety check
```
severinbeauvais, line 549: This method should not be called if IA Survey ID was 0 in the first place, but we still don't want to set a local storage variable with an invalid id.

src/App.vue
```diff
@@ -1106,11 +1104,11 @@ export default class App extends Vue {
...
         if (this.isIncorporationFiling) {
-          if (+sessionStorage.getItem('IA_SURVEY_ID')) {
-            const cookie = Cookies.get(this.IA_SURVEY_KEY, { domain: this.domain })
-            if (!cookie) {
+          if (this.iaSurveyId > 0) {
+            const savedId = +localStorage.getItem('HIDE_SURVEY')
+            if (savedId !== this.iaSurveyId) {
               this.filingSurveyDialog = true
             }
           }
```
severinbeauvais, line 1114: This updated mechanism also allows us to set (in 1Password/config map/configuration.json) a NEW survey id and have it display even if a previous survey (id) was hidden.

ChrisGabel23 (approved):

PR #508, opened 2023-01-27
14798 Remaining implementation of Applicant Information

severinbeauvais:
*Issue #:* bcgov/entity#14798

*Description of changes:*

Commit 1:
- don't render fee summary until data is loaded
- don't check FF if Jest is running
- added conditional columns to ListPeopleAndRoles
- added Email column to ListPeopleAndRoles
- disable add buttons when there is a valid applicant
- auto-assign Applicant role
- deleted obsolete roles
- used shared Contact Point interface
- added restoration properties to interface
- used sub-component for rule list items
- create RuleListItem component
- fixed applicant information validation
- misc cleanup (including some TODO comments)
- created/updated unit tests
- app version = 4.5.13

Commit 2:
- reordered v-cards
- updated labels
- updated Document Delivery props
- updated DocumentDelivery dependency

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -138,7 +138,8 @@
             </v-col>
 
             <v-col cols="12" lg="3">
-              <aside>
+              <!-- Render fee summary only after data is loaded. -->
+              <aside v-if="haveData">
                 <affix relative-element-selector=".col-lg-9" :offset="{ top: 100, bottom: -100 }">
                   <SbcFeeSummary
                     :filingData="feeFilingData"
@@ -545,14 +546,14 @@ export default class App extends Vue {
 
   /** Called to update "do not show again" state. */
   protected doNotShowSurvey (doNotShow: boolean): void {
-    if (doNotShow) {
-      // safety check
-      if (this.iaSurveyId > 0) {
+    // safety check
+    if (this.iaSurveyId > 0) {
+      if (doNotShow) {
         // save id of survey to hide
         localStorage.setItem('HIDE_SURVEY', this.iaSurveyId)
+      } else {
+        localStorage.removeItem('HIDE_SURVEY')
       }
-    } else {
-      localStorage.removeItem('HIDE_SURVEY')
     }
   }
 
@@ -592,12 +593,15 @@ export default class App extends Vue {
     // reset errors in case this method is invoked more than once (ie, retry)
     this.resetFlags()
 
-    // check that current route matches a supported filing type
-    const supportedFilings = await GetFeatureFlag('supported-filings')
-    if (!supportedFilings?.includes(this.$route.meta.filingType)) {
-      window.alert('This filing type is not available at the moment. Please check again later.')
-      this.goToDashboard(true)
-      return
+    // don't check FF during Jest tests
+    if (!this.isJestRunning) {
+      // check that current route matches a supported filing type
+      const supportedFilings = await GetFeatureFlag('supported-filings')
+      if (!supportedFilings?.includes(this.$route.meta.filingType)) {
+        window.alert('This filing type is not available at the moment. Please check again later.')
+        this.goToDashboard(true)
+        return
+      }
     }
 
     try {
@@ -637,6 +641,10 @@ export default class App extends Vue {
         this.fetchErrorDialog = !this.isErrorDialog
         throw error // go to catch()
       }
+
+      // FUTURE:
+      // by now, we know what type of filing this is
+      // add staff check for certain filings (ie, restorations)
 
       // get user info
       const userInfo = await this.loadUserInfo().catch(error => {
--- a/src/components/common/ListPeopleAndRoles.vue
+++ b/src/components/common/ListPeopleAndRoles.vue
@@ -14,6 +14,10 @@
             id="router-link"
             :to="{ path: `/${RouteNames.REGISTRATION_PEOPLE_ROLES}` }"
           >Return to this step to finish it</router-link>
+          <router-link v-if="isFullRestorationFiling || isLimitedRestorationFiling"
+            id="router-link"
+            :to="{ path: `/${RouteNames.RESTORATION_APPLICANT_INFORMATION}` }"
+          >Return to this step to finish it</router-link>
         </span>
       </div>
 
@@ -60,7 +64,7 @@
             <MailingAddress :address="orgPerson.mailingAddress" />
           </v-col>
 
-          <v-col class="delivery-address-column">
+          <v-col class="delivery-address-column" v-if="showDeliveryAddressColumn">
             <template v-if="isDirector(orgPerson) || isProprietor(orgPerson) || isPartner(orgPerson)">
               <p v-if="isSame(orgPerson.mailingAddress, orgPerson.deliveryAddress)">
                 Same as Mailing Address
@@ -69,10 +73,14 @@
             </template>
           </v-col>
 
-          <v-col class="roles-column">
+          <v-col class="roles-column" v-if="showRolesColumn">
             <p v-for="(role, index) in orgPerson.roles" :key="index">
               {{ role.roleType }}
             </p>
+          </v-col>
+
+          <v-col class="email-column" v-if="showEmailColumn">
+            <p>{{ orgPerson.officer.email }}</p>
           </v-col>
 
           <v-col v-if="!isSummary" class="actions-column" :class="{'disabled':disabled}">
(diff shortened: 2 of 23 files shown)


src/App.vue
```diff
@@ -545,14 +546,14 @@ export default class App extends Vue {
...
+    if (this.iaSurveyId > 0) {
+      if (doNotShow) {
         // save id of survey to hide
         localStorage.setItem('HIDE_SURVEY', this.iaSurveyId)
+      } else {
+        localStorage.removeItem('HIDE_SURVEY')
       }
-    } else {
-      localStorage.removeItem('HIDE_SURVEY')
     }
```
severinbeauvais, line 557: Cleanup -- should not remove item if we don't have a current survey ID.

src/mixins/filing-template-mixin.ts
```diff
@@ -402,7 +402,9 @@ export default class FilingTemplateMixin extends DateMixin {
           nrNumber: this.getNameRequestNumber
         },
         nameTranslations: this.getNameTranslations,
-        parties: this.orgPersonsToParties(this.getAddPeopleAndRoleStep.orgPeople)
+        parties: this.orgPersonsToParties(this.getAddPeopleAndRoleStep.orgPeople),
+        offices: undefined, // *** TODO: update
+        contactPoint: undefined // *** TODO: update
```
severinbeauvais, line 407: Needed due to required fields in interface.

@ketaki-deodhar please update this.

tests/unit/App.spec.ts
```diff
@@ -908,24 +908,6 @@ describe('Restoration - Restore Business Name for a BEN', () => {
...
-        priorityFees: 0,
-        processingFees: 0,
-        serviceFees: 2.0,
-        tax: {
-          gst: 0,
-          pst: 0
-        },
-        total: 125.0
-      }
-    }))
```
severinbeauvais, line 927: Moved (and reduced) inside GET call below.

src/App.vue
```diff
@@ -138,7 +138,8 @@
             </v-col>
 
             <v-col cols="12" lg="3">
-              <aside>
+              <!-- Render fee summary only after data is loaded. -->
+              <aside v-if="haveData">
```
severinbeauvais, line 142: Still displays (at same time as spinner is hidden):

image

src/components/common/RegPeopleAndRoles.vue
```diff
@@ -163,8 +163,8 @@
           outlined
           color="primary"
           class="btn-outlined-primary mt-6"
-          :disabled="showOrgPersonForm"
-          @click="addOrgPerson(null, PartyTypes.ORGANIZATION)"
+          :disabled="showOrgPersonForm || validApplicantPerson || validApplicantOrg"
```
severinbeauvais, line 166: With an individual:
image

With a business or corporation:
image

With neither:
image

src/components/common/RegPeopleAndRoles.vue
```diff
@@ -18,36 +18,41 @@
...
-            <v-icon v-else>mdi-circle-small</v-icon>
-            <span class="rule-item-txt">{{rule.text}}</span>
-          </li>
+          <RuleListItem class="completing-party-rule"
+            v-if="rule.id === RuleIds.NUM_COMPLETING_PARTY"
+            :key="index"
+            :valid="validNumCompletingParty"
+            :showErrors="getShowErrors"
+            :text="rule.text"
+          />
```
severinbeauvais, line 27: These sub-components might not save much code vertical space, but they're a lot simpler.

src/components/common/RegPeopleAndRoles.vue
```diff
@@ -18,36 +18,41 @@
...
-            <span class="rule-item-txt">{{rule.text}}</span>
-          </li>
+          <RuleListItem class="completing-party-rule pt-2"
+            v-if="rule.id === RuleIds.NUM_COMPLETING_PARTY"
+            :key="index"
+            :valid="validNumCompletingParty"
+            :showErrors="getShowErrors"
+            :text="rule.text"
+          />
+          <RuleListItem class="proprietors-rule pt-2"
```
severinbeauvais, line 28: Note that the spacing class is here in the parent, not in the sub-component. This allows the parent to control the spacing as it sees fit without "inheriting" whitespace from the component. Another way to say this is that the spacing in the component itself is "tight" -- which may be useful if the component is reused on a different page with different spacing requirements.

src/components/common/RuleListItem.vue
```diff
@@ -0,0 +1,26 @@
+<template>
+  <li class="rule-list-item">
```
severinbeauvais, line 2: In most components, we'd use an `id` here, but there may be several instances of this component on a page, and ids are supposed to be unique, so I'm using a classname here instead.

src/mixins/filing-template-mixin.ts
```diff
@@ -402,7 +400,9 @@ export default class FilingTemplateMixin extends DateMixin {
...
```
severinbeauvais, line 405: @ketaki-deodhar ^^^

src/views/Restoration/RestorationReviewConfirm.vue
```diff
@@ -10,15 +10,21 @@
           </p>
       </header>
 
-      <v-card id="company-summary-vcard" flat class="mt-6">
```
severinbeauvais, line 13: As per uxpin, the Applicant Information comes before the Business Information.

src/views/Restoration/RestorationReviewConfirm.vue
```diff
@@ -27,16 +33,18 @@
...
         <DocumentDelivery
           class="py-8 px-6"
-          :contactValue="getBusinessContact.email"
-          :completingPartyEmail="getUserEmail"
-          contactLabel="Registered Office"
+          contactLabel="Business Office"
+          :contactValue="businessOfficeEmail"
+          additionalLabel="Applicant"
+          :additionalValue="applicantEmail"
+          :showCompletingParty="false"
```
severinbeauvais, line 47: Note that "false" needs to be a bound prop. This doesn't work with `showCompletingParty="false"`.

src/views/Restoration/RestorationReviewConfirm.vue
```diff
@@ -116,6 +126,18 @@ export default class RestorationReviewConfirm extends Vue {
...
+  get businessOfficeEmail (): string {
+    return this.getBusinessContact.email
+  }
+
+  get applicantEmail (): string {
+    const orgPeople = this.getAddPeopleAndRoleStep.orgPeople
+    const applicants = orgPeople.filter(
+      orgPerson => orgPerson.roles.some(role => role.roleType === RoleTypes.APPLICANT)
+    )
+    return applicants[0]?.officer.email
```
severinbeauvais, line 138: If this returns undefined then the DocumentDelivery component displays "Not Entered".

src/components/common/RegPeopleAndRoles.vue
jonathan-longe: That's definitely cleaner!

src/components/common/ListPeopleAndRoles.vue
```diff
@@ -146,19 +154,31 @@ import { PartyTypes, RoleTypes, RouteNames } from '@/enums'
...
   protected activeIndex: number // is NaN for new org/person
 
+  get tableHeaders (): Array<string> {
+    const headers = ['Name', 'Mailing Address']
+    if (this.showDeliveryAddressColumn) headers.push('Delivery Address')
+    if (this.showRolesColumn) headers.push('Roles')
+    if (this.showEmailColumn) headers.push('Email')
+    return headers
+  }
+
```
jonathan-longe, line 181: What do you think about making a sub-component to display a row in the table?  Then, if I'm thinking about this correctly, these properties could be computed row attributes.

jonathan-longe (approved):
Thanks for the opportunity to review @severinbeauvais

src/components/common/ListPeopleAndRoles.vue
severinbeauvais: I think you're right, but I think this isn't complicated enough, or used elsewhere, that it needs to be its own component IMHO. It's a gray area.

src/mixins/filing-template-mixin.ts
ketaki-deodhar: Got it. I will keep an eye on this while merging :)

ChrisGabel23 (approved):

PR #509, opened 2023-01-31
14811- Confirm Business Information step

ketaki-deodhar:
*Issue #:* /bcgov/entity#14811

*Description of changes:*

Restoration - Confirm Business Information step
- 4 addresses (reuse existing component)
- save/resume this data
- unit tests


src/components/common/OfficeAddresses.vue
```diff
@@ -194,7 +194,7 @@ import { Getter } from 'vuex-class'
 import { isEmpty } from 'lodash'
 import { OfficeAddressSchema } from '@/schemas'
 import BaseAddress from 'sbc-common-components/src/components/BaseAddress.vue'
-import { AddressIF, DefineCompanyIF, IncorporationAddressIF } from '@/interfaces'
+import { AddressIF, DefineCompanyIF, RegisteredRecordsAddressesIF } from '@/interfaces'
```
ketaki-deodhar, line 197: Renamed to a more generic name

src/store/getters/state-getters.ts
```diff
@@ -526,7 +526,7 @@ export const isRestorationValid = (state: StateIF): boolean => {
 
   return (
     isAddPeopleAndRolesValid(state) &&
-    getRestoration(state).businessInfoValid &&
+    getDefineCompanyStep(state).valid &&
     getRestoration(state).businessNameValid &&
```
ketaki-deodhar, line 530: Validations

src/components/common/OfficeAddresses.vue
```diff
@@ -311,7 +311,7 @@ export default class OfficeAddresses extends Vue {
           )
         }
       }
-      if (this.isBaseCompany) {
+      if (this.isBaseCompany && this.addresses.recordsOffice) {
         this.recMailingAddress = this.addresses.recordsOffice?.mailingAddress
```
ketaki-deodhar, line 315: Added a check to resolve issue with loading address on first load

src/components/Incorporation/SummaryDefineCompany.vue
```diff
@@ -5,9 +5,13 @@
...
           <span class="error-text mx-1">This step is unfinished.</span>
-          <router-link
+          <router-link v-if="isIncorporationFiling"
             :to="{ path: `/${RouteNames.INCORPORATION_DEFINE_COMPANY}` }"
           >Return to this step to finish it</router-link>
+          <router-link v-if="isFullRestorationFiling || isLimitedRestorationFiling"
+            id="router-link"
+            :to="{ path: `/${RouteNames.RESTORATION_BUSINESS_INFORMATION}` }"
+          >Return to this step to finish it</router-link>
         </span>
```
ketaki-deodhar, line 15: Review and Confirm, Return to step with validations.

image

tests/unit/RestorationBusinessInformation.spec.ts
```diff
@@ -1,18 +1,49 @@
...
+  {
+    entityType: 'BC'
+  },
+  {
+    entityType: 'ULC'
+  },
+  {
+    entityType: 'CC'
+  }
+]
```
ketaki-deodhar, line 21: Added all types of entities in case functionality changes in later

src/components/common/OfficeAddresses.vue
jonathan-longe: Great idea!

src/mixins/filing-template-mixin.ts
```diff
@@ -959,4 +975,29 @@ export default class FilingTemplateMixin extends DateMixin {
...
   }
+
+  /**
+   * Ensure consistent object structure for a Restoration, whether it contains a Name
+   * Request or not, and whether it is an initial draft or it has been previously saved.
+   * Object merging does not work very well otherwise due to nested properties.
+   * @param filing the filing fetched from legal-api
+   * @returns the filing in safe-empty state if applicable
+   */
+  formatEmptyRestoration (filing: any): RestorationFilingIF {
```
jonathan-longe, line 986: Rather than `any`, can we declare a specific filing type?

jonathan-longe (approved):
Good work @ketaki-deodhar !

ChrisGabel23 (approved):

JazzarKarim (approved):

src/components/common/OfficeAddresses.vue
severinbeauvais: We should fix that first load issue.

src/interfaces/filing-interfaces/filing-interfaces.ts
```diff
@@ -185,7 +185,11 @@ export interface RestorationFilingIF {
...
     nameRequest?: RegistrationNameRequestIF
     parties: PartyIF[]
-    offices: IncorporationAddressIF
-    contactPoint: ContactPointIF
+    offices: RegisteredRecordsAddressesIF | object
+    contactPoint: {
+      email: string
+      phone: string
+      extension?: number
+    }
```
severinbeauvais, line 193: I think you should restore the previous type for contact point.

src/interfaces/filing-interfaces/filing-interfaces.ts
```diff
@@ -185,7 +185,11 @@ export interface RestorationFilingIF {
     nameTranslations?: NameTranslationIF[]
     nameRequest?: RegistrationNameRequestIF
     parties: PartyIF[]
-    offices: IncorporationAddressIF
-    contactPoint: ContactPointIF
+    offices: RegisteredRecordsAddressesIF | object
```
severinbeauvais, line 188: What are you trying to do here that requires a possibly untyped object?

src/mixins/filing-template-mixin.ts
```diff
@@ -536,6 +543,15 @@ export default class FilingTemplateMixin extends DateMixin {
     // restore Persons and Organizations
     this.setOrgPersonList(this.partiesToOrgPersons(draftFiling.restoration.parties || []))
 
+    // restore Office addresses
+    this.setOfficeAddresses(draftFiling.restoration.offices)
```
severinbeauvais, line 547: Are office addresses guaranteed to be in the draft filing? If so then this is OK. If not then you should provide a suitable fallback value.

src/mixins/filing-template-mixin.ts
severinbeauvais: I can think of 2 better solutions for this issue:
1. when Filings UI create a draft restoration, it pre-populates these empty objects, or
2. create an "empty restoration" object that you can use in App.vue to fill in missing values.

Example of 2:
- create something like this: https://github.com/bcgov/business-create-ui/blob/cf242543f0aee56e94832628545f56f482a37a55/src/interfaces/common/address-interface.ts#L14
- use it in App.vue like this:
```
        draftFiling = {
          ...emptyRestorationFiling, // pre-fill needed properties         
          ...this.buildRestorationFiling(), // fill properties from store data
          ...draftFiling // fill properties from draft
        }
```

severinbeauvais: The root problem is that we don't want the store to have invalid data, for example a null object when it should be an empty object. We should work to fix that instead of creating special cases.

src/store/state/resource-model.ts
```diff
@@ -69,8 +69,34 @@ const dissolutionResourceModel: DissolutionResourceIF = {
...
+    header: null,
+    blurb: null,
+    helpSection: null,
+    blurb2: null,
+    addIncorporator: null,
+    addOrganization: null,
+    addBusiness: null,
+    rules: []
+  }
+}
```
severinbeauvais, line 95: To reduce duplicate code and keep things consistent, we should have an "empty people and roles" object and an empty "review and confirm" object that we can just spread or clone here.

You don't have to do this in this ticket -- we should have a refactor / tech debt ticket to clean up some of this kind of stuff.

tests/unit/RestorationApplicantInformation.spec.ts
```diff
@@ -3,7 +3,7 @@ import { RestorationApplicantInformation } from '@/views'
 import { RestorationResources } from '@/resources/'
 
 describe('Restoration Review Confirm view', () => {
-  it('renders the component properly', () => {
+  xit('renders the component properly', () => {
```
severinbeauvais, line 6: Does this test not work?

tests/unit/RestorationBusinessInformation.spec.ts
severinbeauvais: Nice! 👍

severinbeauvais (commented):
It looks really good. Great job! Your code fits in nicely with the other code.

Please see my comments. We might want to improve some architectural issue (eg, empty initial data) here, or we could do it as part of a future ticket. What do you think?

src/interfaces/filing-interfaces/filing-interfaces.ts
ketaki-deodhar: okay

tests/unit/RestorationApplicantInformation.spec.ts
ketaki-deodhar: It was failing. So uncommented it but I can fix it and commit. Just didn't want to make changes if someone else was parallelly working on it

src/views/Restoration/RestorationBusinessName.vue
```diff
@@ -1,18 +1,5 @@
...
-      <header id="name-header">
-        <h2>Name</h2>
-      </header>
-
-      <v-card flat class="mt-5">
-        <NameRequestInfo
-          @hasNameTranslation="onNameTranslation($event)"
-        />
-      </v-card>
-    </section>
```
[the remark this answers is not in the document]
ketaki-deodhar: I can retain the code but I will still have to comment it out as this will be in another page/view file

src/interfaces/filing-interfaces/filing-interfaces.ts
ketaki-deodhar: Frankly, just followed the example in Incorporation application case. I can take a look and fix this.

tests/unit/RestorationApplicantInformation.spec.ts
severinbeauvais: Yes, please put it back. Any idea why it's failing? LMK if I can help.

src/views/Restoration/RestorationBusinessName.vue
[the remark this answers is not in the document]
severinbeauvais: According to the uxpin, the "Add Business Name" step -- this page -- displays this section.

Some of these components may also be on the Review and Confirm step.

What other duplication do you see?

src/interfaces/filing-interfaces/filing-interfaces.ts
severinbeauvais: Yeah, I know. That was a work-around for the "missing prop in typed object" issue -- kind of thing you've been bumping into and I mentioned.

If your code doesn't work with the object type then just leave it -- we will have a future ticket to fix such things.

src/store/state/state-model.ts
```diff
@@ -113,7 +113,16 @@ export const stateModel: StateModelIF = {
...
-    officeAddresses: {}
+    officeAddresses: {
+      registeredOffice: {
+        mailingAddress: { ...EmptyAddress },
+        deliveryAddress: { ...EmptyAddress }
+      },
+      recordsOffice: {
+        mailingAddress: { ...EmptyAddress },
+        deliveryAddress: { ...EmptyAddress }
+      }
```
ketaki-deodhar, line 124: Adding this and conditional logic change to set default address in 'RestorationBusinessInformation.vue' file resolved the issue with null address object issue

src/views/Restoration/RestorationBusinessInformation.vue
```diff
@@ -202,6 +108,22 @@ export default class RestorationBusinessInformation extends Vue {
     })
   }
 
+  private isEmptyRecordsAddress () : boolean {
+    const recordsAddress = this.addresses.recordsOffice
+    if (!recordsAddress.mailingAddress ||
```
severinbeauvais, line 113: If `recordsAddress` itself is null/undefined, this will throw an exception.

Ditto for `registeredAddress` below.

src/mixins/filing-template-mixin.ts
severinbeauvais: I think you should have a check here, otherwise you could set office addresses to null or undefined. (And undefined is bad because the property is no longer reactive if it has that value.)

src/views/Restoration/RestorationBusinessInformation.vue
ketaki-deodhar: Oh yes! Had the check there not sure why I removed it :(
Adding it back

severinbeauvais: You could use [optional chaining](https://www.google.com/search?q=optional+chaining) to simplify the code. Up to you.

src/mixins/filing-template-mixin.ts
ketaki-deodhar: Just thinking I missed something from our discussion. Adding it.

ketaki-deodhar: Done

src/views/Restoration/RestorationBusinessInformation.vue
ketaki-deodhar: Done

src/store/state/resource-model.ts
ketaki-deodhar: I can add a ticket for this.

severinbeauvais: Yes, please and thanks!

src/mixins/filing-template-mixin.ts
```diff
@@ -959,4 +977,29 @@ export default class FilingTemplateMixin extends DateMixin {
...
+   * Object merging does not work very well otherwise due to nested properties.
+   * @param filing the filing fetched from legal-api
+   * @returns the filing in safe-empty state if applicable
+   */
+  formatEmptyRestoration (filing: any): RestorationFilingIF {
+    const toReturn = filing
+    if (toReturn.restoration) {
+      // set offices
+      if (!toReturn.restoration?.offices) {
+        toReturn.restoration.offices = []
```
severinbeauvais, line 993: Is this still needed? If so then you should assign an object here, not an array.

src/mixins/filing-template-mixin.ts
```diff
@@ -959,4 +977,29 @@ export default class FilingTemplateMixin extends DateMixin {
...
+      }
+      // set contact point
+      if (!toReturn.restoration?.contactPoint) {
+        toReturn.restoration.contactPoint = {
+          email: '',
+          phone: ''
+        }
```
severinbeauvais, line 1000: You could use `EmptyContactPoint` here :)

src/views/Restoration/RestorationBusinessInformation.vue
```diff
@@ -202,6 +108,18 @@ export default class RestorationBusinessInformation extends Vue {
     })
   }
 
+  private isEmptyRecordsAddress () : boolean {
+    if (!this.addresses.recordsOffice?.mailingAddress || !this.addresses.recordsOffice?.deliveryAddress) {
+      return true
+    }
```
severinbeauvais, line 114: 1. this method doesn't return a value if the statement is false
2. this could be a getter instead of a method
3. should you check the sub-objects?

ketaki-deodhar: Can make it a get and return if false. Need to check sub-objects as they are undefined.

src/mixins/filing-template-mixin.ts
severinbeauvais: I'm happy with this conversation. @jonathan-longe Please resolve if OK with you, or follow up as needed.

src/mixins/filing-template-mixin.ts
```diff
@@ -988,16 +988,9 @@ export default class FilingTemplateMixin extends DateMixin {
...
-      if (!toReturn.restoration?.offices) {
-        toReturn.restoration.offices = []
-      }
       // set contact point
       if (!toReturn.restoration?.contactPoint) {
-        toReturn.restoration.contactPoint = {
-          email: '',
-          phone: ''
-        }
+        toReturn.restoration.contactPoint = EmptyContactPoint
```
severinbeauvais, line 993: This is an object assignment. I believe that if you did
```
toReturn.restoration.contactPoint.email = 'severin@example.com`
```
that it would update the Empty Contact Point object itself.

Instead, you can spread the empty object into a new object (or use cloneDeep if the object is more than 1 layer deep).

severinbeauvais: Do you need to init Contact Point here, or is its initialization in the store OK?

ketaki-deodhar: Yikes! I am thinking this again. Let me try not keeping it here and test again. It is already in state-model. 
image

severinbeauvais: It's called "iterating", and the code keeps getting better 😆

ketaki-deodhar: initialization in the store works well with a null/empty check for first load. Do not need formatEmptyRestoration call.

severinbeauvais: Excellent - one less special case!

severinbeauvais (approved):
Approved! Thanks for the clean code.

ketaki-deodhar:
Updated package version. Merging now

PR #510, opened 2023-02-01
14798 UI updates

severinbeauvais:
*Issue #:* bcgov/entity#14798

*Description of changes:*
- updated people and roles list header
- added fallback "not registered in BC" text
- added fallback confirm checkbox text
- added fallback "registered in BC" text
- added subheader resource for Add People And Roles component
- removed `restoration.date` property (see https://github.com/bcgov/business-schemas/pull/125)
- app version = 4.5.14

### DIFF
--- a/src/interfaces/filing-interfaces/filing-interfaces.ts
+++ b/src/interfaces/filing-interfaces/filing-interfaces.ts
@@ -176,7 +176,6 @@ export interface RestorationFilingIF {
     foundingDate: string
   }
   restoration: {
-    date: string // today, as YYYY-MM-DD
     type: RestorationTypes
     relationships?: any // *** TODO: udpate type
     expiry?: string // YYYY-MM-DD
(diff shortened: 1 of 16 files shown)


JazzarKarim (approved):

jonathan-longe (approved):
Thanks for the opportunity to review

src/interfaces/filing-interfaces/filing-interfaces.ts
```diff
@@ -176,7 +176,6 @@ export interface RestorationFilingIF {
     foundingDate: string
   }
   restoration: {
-    date: string // today, as YYYY-MM-DD
```
ChrisGabel23, line 179: Aren't restorations date-specific? I'm confused why the date has been removed.

severinbeauvais: This "restoration date" field was not needed and was removed from the schema by Vysakh this morning.

This filing still has both a "date" and "effectiveDate", which are set by the API when the filing is filed-and-paid.

If you're wondering about "when does the restoration start", it's when it's filed. If you're wondering about the time period of a limited restoration, that also starts when it's filed -- there is a separate date property, "expiry" for that.

PR #511, opened 2023-02-07
Restoration type component

JazzarKarim:

*Description of changes:*
- Created Restoration Type Component:

https://user-images.githubusercontent.com/122301442/217377618-0deb8343-d068-477e-965f-5642a8db2fe6.mp4

- Created Restoration Type Summary Component:

  - Full Restoration but no relationship is selected:
full error

  - Full Restoration with relationships selected:
full good

  - Limited Restoration selected:
Limited good


src/interfaces/store-interfaces/state-interface.ts
```diff
@@ -69,6 +69,7 @@ export interface StateModelIF {
   completingParty?: CompletingPartyIF
   parties?: Array<PartyIF>
   restoration: RestorationStateIF
+  isRestorationTypeValid: boolean
```
severinbeauvais, line 72: Can this be moved to the restoration state object?

JazzarKarim: I thought about it. Thing is, restoration schema doesn't have an isValid field.

severinbeauvais: There is an interface for the restoration state, which is different from the restoration JSON object.

src/enums/index.ts
```diff
@@ -27,5 +27,6 @@ export {
   EffectiveDateTypes,
   EntityStates,
   FilingCodes,
-  StaffPaymentOptions
+  StaffPaymentOptions,
+  RelationshipTypes
```
severinbeauvais, line 31: Please sort this list. It's not a big deal for a small list like this, but for larger lists it saves times/errors when looking for something.

src/constants/constants.ts
```diff
@@ -1,3 +1,5 @@
 export const ADDRESSCHANGED = 'addressChanged'
 export const ISFUTUREEFFECTIVE = 'isFutureEffective'
 export const ISIMMEDIATE = 'isImmediate'
+export const ISLIMITEDRESTORATION = 'isLimited'
+export const ISFULLRESTORATION = 'isFull'
```
severinbeauvais, line 5: Are these used anywhere?

severinbeauvais: Oh, found it.

src/enums/index.ts
JazzarKarim: Noted.

src/interfaces/store-interfaces/state-interface.ts
JazzarKarim: Oh wait, I think I have things mixed up a little. In that case, I think I can move it to the state object.

src/constants/constants.ts
severinbeauvais: This use of constants is a bad pattern. In your case, please create a new enum (type) and use that. This will also make `selectRestorationType` nicely typed. (You can assign null as an initial "unknown" value if needed.)

src/views/Restoration/RestorationBusinessName.vue
```diff
@@ -6,105 +6,20 @@
         <h2>Name</h2>
       </header>
 
-      <v-card flat class="mt-5">
-        <NameRequestInfo
-          @hasNameTranslation="onNameTranslation($event)"
-        />
-      </v-card>
```
severinbeauvais, line 13: Please restore this. I will be updating it in ticket 14783.

src/views/Restoration/RestorationBusinessName.vue
```diff
@@ -6,105 +6,20 @@
...
-            @hasCooperativeType="onCooperativeType($event)"
-          />
-        </v-card>
-      </div>
-    </section>
-
-    <!-- Registered Office Addresses -->
-    <section class="mt-10" v-show="isEntityType">
+        <!--Restoration Type is here-->
+        <section class="mt-10" v-show="isEntityType">
```
severinbeauvais, line 12: This v-show is suspicious.
1. I'm not sure why you need to check this here. Is entity type not set sometimes?
2. Why it is a v-show instead of a v-if? Do the component within need to be rendered when we don't know entity type?

src/views/Restoration/RestorationReviewConfirm.vue
```diff
@@ -97,6 +97,7 @@ import ListPeopleAndRoles from '@/components/common/ListPeopleAndRoles.vue'
 import { GetCorpFullDescription } from '@bcrs-shared-components/corp-type-module'
 import StaffPayment from '@/components/common/StaffPayment.vue'
 import SummaryDefineCompany from '@/components/Incorporation/SummaryDefineCompany.vue'
+import RestorationTypeSummary from '@/components/Restoration/RestorationTypeSummary.vue'
```
severinbeauvais, line 100: Is this component supposed to be rendered on this page? I don't see it.

JazzarKarim: I just imported it but I didn't render it since it's going to be a part of a bigger component. The component is going to have other things like Approval type. So I left there to be used later.

tests/unit/RestorationType.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+    const wrapper = shallowWrapperFactory(
+      RestorationType
+    )
+
+    // verify the component is rendered
+    expect(wrapper.find('#restoration-type').exists()).toBe(true)
+
+    // verify sub-components on page are rendered
+    // expect(wrapper.findComponent(RelationshipsPanel).exists()).toBe(true)
+    // expect(wrapper.findComponent(LimitedRestorationPanel).exists()).toBe(true)
```
severinbeauvais, line 17: Fix or delete please.

tests/unit/RestorationType.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+  })
+})
```
severinbeauvais, line 19: If this sub-component has appropriate tests in the shared repo, then you don't need to test it again here.

However, the component that hosts this sub-component should make sure the sub-component is displayed when needed, and also verify that the right thing is done with any events (or getters/computed variables).

tests/unit/RestorationTypeSummary.spec.ts
```diff
@@ -0,0 +1,13 @@
...
+describe(`Initialize Restoration Type Component`, () => {
+  it('renders the component properly', () => {
+    const wrapper = shallowWrapperFactory(
+        RestorationTypeSummary
+    )
+
+    // verify the component is rendered
+    expect(wrapper.find('#restoration-type-summary').exists()).toBe(true)
+  })
+})
```
severinbeauvais, line 13: This test suite should also verify that store properties display correctly, plus any other things that this component does.

tests/unit/RestorationType.spec.ts
JazzarKarim: Yes, I was confused what should I exactly test since I've done a lot of the testing in the sub-components (bcrs shared repo).

src/components/Restoration/RestorationType.vue
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+label {
+  font-weight: bold;
```
severinbeauvais, line 136: Labels typically already have a base style that makes them bold.

If you're only styling a single label then I recommend using the Vuetify helper classes to set font weight: https://vuetifyjs.com/en/styles/text-and-typography/

src/components/Restoration/RestorationTypeSummary.vue
```diff
@@ -0,0 +1,67 @@
...
+          <v-icon color="error">mdi-information-outline</v-icon>
+          <span class="error-text mx-1">This step is unfinished.</span>
+          <router-link
+            :to="{ path: `/${RouteNames.RESTORATION_BUSINESS_NAME}` }"
+          >Return to this step to finish it</router-link>
+        </span>
+      </div>
+
+      <v-row>
+        <v-col class="px-4">
```
severinbeauvais, line 15: How wide is this column? Can you show screenshots on a wide screen and narrow screen?

src/components/Restoration/RestorationTypeSummary.vue
```diff
@@ -0,0 +1,67 @@
...
+          <label>Restoration Type</label>
+        </v-col>
+
+        <v-col cols="12" sm="9">
+          <template v-if="getRestoration.type === 'limitedRestoration'">
```
severinbeauvais, line 20: Please use enum for value.

src/views/Restoration/RestorationReviewConfirm.vue
severinbeauvais: Isn't it part of your ticket to show restoration/approval type on the review page? If not, do we need another ticket?

tests/unit/RestorationType.spec.ts
severinbeauvais: Then you don't need to test the sub-component again.

But you should test the props you would send to it and how you handle the events you would get back from it -- we don't do this nearly as well as we should, but we should try :)

Think "tiers" -- THIS component should be tested here, but not its parent or children components.

src/views/Restoration/RestorationReviewConfirm.vue
JazzarKarim: Yes it is! Okay, I'll add it to the review page. 

summary

This is what's going to look like in the end. My summary component is the one in the middle. I'll add and keep it in the review page.

tests/unit/RestorationType.spec.ts
severinbeauvais: If extra unit tests will take more than 1 day, I'm OK with creating a new Tech Debt ticket to expand them later.

src/components/Restoration/RestorationTypeSummary.vue
JazzarKarim: Wide:
wide

Narrow:
narrow

severinbeauvais: Can you show a little more on the narrow screen, please? I want to see how the other sections display.

src/components/Restoration/RestorationType.vue
```diff
@@ -0,0 +1,136 @@
...
+  @Getter getCurrentDate!: string
+  @Getter isRestorationTypeValid!: boolean
+
+  @Action setRestorationType!: ActionBindingIF
+  @Action setRestorationExpiry!: ActionBindingIF
+  @Action setRestorationRelationships!: ActionBindingIF
+  @Action setRestorationTypeValid!: ActionBindingIF
+
+  // Local properties
+  protected isFull = false
```
severinbeauvais, line 78: Instead of managing this property, why not use a computed value (getter) instead?
```
get isFull (): boolean {
   return (this.getRestoration.type === RestorationTypes.FULL)
}
```
And ditto for `isLimited`.

Let the framework do the reactivity for you.

src/components/Restoration/RestorationTypeSummary.vue
JazzarKarim: localhost_8080_businesses_create_restoration-review-confirm_id=BC0871154 accountid=3040 (1)

src/components/Restoration/RestorationType.vue
```diff
@@ -0,0 +1,136 @@
...
+   * Called when component is mounted.
+   * Automatically check limited or full restoration when user continues a draft as per last selection.
+   */
+  mounted (): void {
+    if (this.getRestoration.type === RestorationTypes.LIMITED) {
+      this.selectRestorationType = RestorationTypes.LIMITED
+      this.isFull = !this.isLimited
+    } else {
+      this.selectRestorationType = RestorationTypes.FULL
+    }
```
severinbeauvais, line 94: Instead of this code, did you try initializing `selectRestorationType` from the store value? I think the property initialization is done when this component is created (not mounted), but it may work.
```
protected selectRestorationType = this.getRestoration.type
```

src/components/Restoration/RestorationType.vue
```diff
@@ -0,0 +1,136 @@
...
+      )
+    }
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+</style>
```
severinbeauvais, line 136: You don't have to, but you can, delete this style block, since it's not being used.

src/components/Restoration/RestorationTypeSummary.vue
```diff
@@ -0,0 +1,64 @@
...
+  })
+
+export default class RestorationTypeSummary extends Vue {
+  @Getter getRestoration!: RestorationStateIF
+  @Getter isRestorationTypeValid!: boolean
+
+  readonly RouteNames = RouteNames
+  readonly RestorationTypes = RestorationTypes
+}
+
```
severinbeauvais, line 55: We don't have an official code style (other than what our linter enforces), but typically we don't have blank lines at the top or bottom of a block.

src/store/mutations/mutations-model.ts
```diff
@@ -413,3 +413,11 @@ export const mutateRestorationType = (state: StateIF, type: RestorationTypes) =>
 export const mutateRestorationExpiry = (state: StateIF, expiry: string) => {
   state.stateModel.restoration.expiry = expiry
 }
+
+export const mutateRestorationRelationships = (state: StateIF, relationships: RelationshipTypes[]) => {
+  state.stateModel.restoration.relationships = relationships
+}
+
+export const mutateRestorationIsValid = (state: StateIF, isRestorationTypeValid: boolean) => {
```
severinbeauvais, line 421: There's a mismatch between the mutation name and the thing you're modifying. Please fix.

src/interfaces/store-interfaces/state-interfaces/restoration-state-interface.ts
```diff
@@ -11,6 +11,8 @@ export interface RestorationStateIF {
   businessNameValid: boolean
   type: RestorationTypes
   expiry?: string // YYYY-MM-DD
+  relationships?: RelationshipTypes[]
+  isRestorationTypeValid: boolean
```
severinbeauvais, line 15: Sorry, I just noticed this now, but how about renaming this flag to be consistent with the other validity flags in this object?

src/views/Restoration/RestorationBusinessName.vue
```diff
@@ -165,43 +177,36 @@ export default class RestorationBusinessName extends Vue {
...
   @Getter getFolioNumber!: string
-
+  @Getter isRestorationTypeValid!: boolean
   @Action setBusinessContact!: ActionBindingIF
   @Action setCooperativeType!: ActionBindingIF
   @Action setFolioNumber!: ActionBindingIF
   @Action setOfficeAddresses!: ActionBindingIF
   @Action setDefineCompanyStepValidity!: ActionBindingIF
   @Action setIgnoreChanges!: ActionBindingIF
-
```
severinbeauvais, line 175: Again, our non-official code style is to leave whitespace between separate blocks. Please restore it.

src/components/Restoration/RestorationTypeSummary.vue
severinbeauvais: Even narrower!

What I want to check is that keys and values stack correctly on mobile, ie:
image

JazzarKarim: localhost_8080_businesses_create_restoration-review-confirm_id=BC0871154 accountid=3040 (2)

severinbeauvais: OK, nice!

src/components/Restoration/RestorationTypeSummary.vue
```diff
@@ -0,0 +1,64 @@
...
+          <v-icon color="error">mdi-information-outline</v-icon>
+          <span class="error-text mx-1">This step is unfinished.</span>
+          <router-link
+            :to="{ path: `/${RouteNames.RESTORATION_BUSINESS_NAME}` }"
+          >Return to this step to finish it</router-link>
+        </span>
+      </div>
+
+      <v-row>
+        <v-col class="px-4">
```
severinbeauvais, line 15: There's some unwanted leading whitespace here.

Compare this with https://github.com/bcgov/business-create-ui/blob/bf73ca09f6f81fe527689ac2b1c43ed3e1df8626/src/components/Incorporation/SummaryDefineCompany.vue#L21

src/components/Restoration/RestorationType.vue
JazzarKarim: This won't work! I tried it out, `this.getRestoration.type` is returning undefined.

severinbeauvais: It might return null since that's what the value in the store is initialized to.

Anyway, this means this component is created before the restoration is fetched -- not surprising since the fetch is async (network call). I was hoping maybe this component wasn't rendered right away.

Side note: what would be the the harm in initializing `selectRestorationType` as null?

src/components/Restoration/RestorationType.vue
JazzarKarim: These two properties are used to determine which panel to expand based on radio button selection. Are you sure having computed value getters would help?

src/components/Restoration/RestorationType.vue
JazzarKarim: Yes makes sense. There's no harm, I could do that. This is because I'm setting it again based on what was previously selected in the `mounted` block.

src/components/Restoration/RestorationType.vue
severinbeauvais: I'm 99% sure. Getters are reactive -- they update depending on the properties they look at. And the template is also reactive, so I think your v-cards will conditionally display just fine.

src/components/Restoration/RestorationType.vue
severinbeauvais: Actually, it should be initialized to null. Currently you are initializing it to an empty string, but it's not a string.

What I mean is: what if you remove this code from `mounted()` ? Is there any issue with that?

JazzarKarim: Thing is, in the mounted block, I am auto-selecting either the Full Restoration or Limited Restoration radio buttons when a user loads their previous draft. Without this code, when a user loads a draft, both radio buttons are not selected.

mounted

src/components/Restoration/RestorationType.vue
```diff
@@ -0,0 +1,135 @@
...
+  @Getter getCurrentDate!: string
+  @Getter getRestorationTypeValid!: boolean
+
+  @Action setRestorationType!: ActionBindingIF
+  @Action setRestorationExpiry!: ActionBindingIF
+  @Action setRestorationRelationships!: ActionBindingIF
+  @Action setRestorationTypeValid!: ActionBindingIF
+
+  // Local properties
+  protected selectRestorationType = null
```
severinbeauvais, line 78: Please add type.

src/store/getters/state-getters.ts
```diff
@@ -608,6 +608,11 @@ export const getRestoration = (state: StateIF): RestorationStateIF => {
   return state.stateModel.restoration
 }
 
+/** The restoration object. */
```
severinbeauvais, line 611: Incorrect comment.

src/components/Restoration/RestorationType.vue
severinbeauvais: Doesn't the first radio group get set as soon as the draft restoration filing is loaded?

severinbeauvais: Does it work correctly atm or does something need to be figured out?

JazzarKarim: > Doesn't the first radio group get set as soon as the draft restoration filing is loaded?

Without the mounted block, the expansion panel would work, but the radio buttons wouldn't be selected. They would both be empty as in the screenshot above.

src/components/Restoration/RestorationType.vue
```diff
@@ -0,0 +1,135 @@
...
+  @Getter getCurrentDate!: string
+  @Getter getRestorationTypeValid!: boolean
+
+  @Action setRestorationType!: ActionBindingIF
+  @Action setRestorationExpiry!: ActionBindingIF
+  @Action setRestorationRelationships!: ActionBindingIF
+  @Action setRestorationTypeValid!: ActionBindingIF
+
+  // Local properties
+  protected selectRestorationType:string = null
```
severinbeauvais, line 78: Isn't it of type RestorationTypes?

severinbeauvais (approved):

severinbeauvais:
Please get 1 or 2 more approved reviews before merging this.

src/components/Restoration/RestorationType.vue
JazzarKarim: You're right. So sorry about that 😢

JazzarKarim:
> Please get 1 or 2 more approved reviews before merging this.

@jonathan-longe @ketaki-deodhar @ChrisGabel23 Please whenever you can. Thanks a lot!

ChrisGabel23 (approved):

ketaki-deodhar (approved):

PR #512, opened 2023-02-08
14783 Prepared NR components for reuse

severinbeauvais:
*Issue #:* bcgov/entity#14783

*Description of changes:*

- separated Name Translation from Name Request Info
- ported Business Name from Edit UI component
- copied CorrectXXX from Edit UI
- removed obsolete code from Restoration Business Name
- deleted some local enums
- imported some shared enums
- cleaned up dissolution types
- created/fixed/expanded unit tests
- misc cleanup
- app version = 4.5.17


src/components/Restoration/BusinessName.vue
```diff
@@ -0,0 +1,288 @@
+<template>
```
severinbeauvais, line 1: This is a port of "YourCompany.vue" from Edit UI.

In Edit UI, this handles the editing of Business Name, Business Type, Name Translations, etc.

Here, it only handles Business Name (which includes NR stuff).

src/components/Restoration/BusinessName.vue
```diff
@@ -0,0 +1,288 @@
...
+  @Getter getNameRequest!: NameRequestIF
+  @Getter getBusinessNumber!: string
+  @Getter getNameRequest!: NameRequestIF
+  @Getter getEntityType!: CorpTypeCd
+
+  // Global actions
+  // @Action setEditingCompanyName!: ActionBindingIF
+  // @Action setValidComponent!: ActionBindingIF
+
+  // *** TODO
```
severinbeauvais, line 133: In progress, but I want to commit incrementally.

Right now this doesn't break anything (since it's only used by restoration filings).

src/components/Restoration/CorrectCompanyName.vue
```diff
@@ -0,0 +1,85 @@
+<template>
```
severinbeauvais, line 1: A straight copy from Edit UI. (Ditto the other CorrectXXX components.)

I intentionally didn't change the name, so we can maybe extract this into the shared components repo later.

src/components/Restoration/CorrectCompanyName.vue
```diff
@@ -0,0 +1,85 @@
...
+  ]
+})
+export default class CorrectCompanyName extends Vue {
+  /** Form Submission Prop */
+  @Prop({ default: null }) readonly formType!: NameChangeOptions
+
+  @Action setNameRequest!: ActionBindingIF
+
+  @Getter getNameRequestLegalName!: string
+  @Getter getNameRequest!: NameRequestIF
```
severinbeauvais, line 37: Ideally, this low-level component wouldn't talk to the store, but would use props and events to communicate with its parent.

Or we could use only props for data in, and for functions in that get called to update something. In the parent, these functions would be references to the store methods or whatever.

src/components/common/NameRequestInfo.vue
```diff
@@ -1,14 +1,9 @@
 <template>
-  <div id="name-request-summary">
-    <ConfirmDialog
-      ref="confirmTranslationRemovalDialog"
-      attach="#name-request-summary"
-    />
```
severinbeauvais, line 6: I took all the name translation stuff out of here and moved it to a new component, where I can reuse it independently from THIS component..

src/components/common/NameTranslation.vue
```diff
@@ -0,0 +1,225 @@
+<template>
```
severinbeauvais, line 1: This is what I took out of NameRequestInfo.vue.

src/enums/index.ts
```diff
@@ -11,6 +11,7 @@ export * from './filingTypes'
 export * from './itemTypes'
 export * from './partyTypes'
 export * from './modes'
+export * from './name-change-options'
```
severinbeauvais, line 14: In the shared repo we use this kebab-case. We also use kebab-case for interfaces. So I'm not starting a new standard but slowly cleaning up something.

src/enums/name-change-options.ts
```diff
@@ -0,0 +1,5 @@
+export enum NameChangeOptions {
+  CORRECT_NAME = 'correct-name',
+  CORRECT_NAME_TO_NUMBER = 'correct-name-to-number',
+  CORRECT_NEW_NR = 'correct-new-nr'
+}
```
severinbeauvais, line 5: This is right from Edit UI. Ultimately this would go in the shared components repo along with the components that use it... future.

src/interfaces/index.ts
```diff
@@ -66,12 +73,6 @@ export * from './utils-interfaces/pdf-info-interface'
 export * from './utils-interfaces/validation-detail-interface'
 export * from './utils-interfaces/validation-item-detail-interface'
 
-// Common interfaces
-export * from './common/address-interface'
-
-// Registration interfaces
-export * from './Registration/business-address-interface'
-
```
severinbeauvais, line 74: I moved these up to keep this list alphabetical.

src/views/Restoration/RestorationBusinessName.vue
```diff
@@ -4,123 +4,23 @@
...
-      <header id="folio-number-header">
-        <h2>Folio / Reference Number (Optional)</h2>
-        <p>Add an optional Folio or Reference Number about this business for your own tracking purposes.
-           This information is not used by the BC Business Registry.
-        </p>
+    <!-- Restoration Type -->
+    <section class="mt-10">
+      <header id="restoration-type-header">
+        <h2>Restoration Type</h2>
+        <p>Determine the restoration type and approval type.</p>
```
severinbeauvais, line 22: This is to be merged with Karim's changes.

src/components/Incorporation/UploadMemorandum.vue
```diff
@@ -414,10 +414,8 @@ export default class UploadMemorandum extends Vue {
 <style lang="scss" scoped>
 @import '@/assets/styles/theme.scss';
 
-header {
-  p {
-    padding-top: 0.5rem;
-  }
+header p {
```
severinbeauvais, line 417: This is shallower so it's a tiny bit more efficient.

src/interfaces/store-interfaces/state-interfaces/name-request-interface.ts
```diff
@@ -33,7 +33,7 @@ export interface NameRequestNameIF {
  * Includes only the properties we care about.
  */
 export interface NameRequestIF {
-  applicants: NameRequestApplicantIF
+  applicants: NameRequestApplicantIF // object, not array
```
severinbeauvais, line 36: Breaks our (unofficial) naming convention that only arrays should be plural.

src/components/Restoration/BusinessName.vue
jonathan-longe: When you write, "port", is this an example we could use to illustrate code duplication between the front-end UIs?  If so, it would be valuable to flag it for reference later.

src/components/Restoration/CorrectCompanyName.vue
jonathan-longe: As we've discussed, I'm concerned about code duplication.  The more we copy code from one repro to another, the more we create a mess we'll have to clean up later.  I think it might be wise to raise this with the business to make sure they're aware of the issue and get approval before proceeding.

src/components/Restoration/BusinessName.vue
severinbeauvais: Yes, this is probably 95% duplicated code with Edit UI.

How do you propose to flag this?

src/components/Restoration/CorrectCompanyName.vue
severinbeauvais: Yep, will do.

To move these sub-components to the shared repo, I first have to refactor them slightly to use props/events instead of talking to the store directly. Then I'll have to go back to Edit UI and update their usage there. It's 1-2 days of work.

src/components/Dissolution/CareAndCustodySelect.vue
```diff
@@ -45,12 +45,14 @@ import { RoleTypes } from '@/enums'
...
 
+  // for template
   readonly RoleTypes = RoleTypes
 
+  // local variable
   protected liquidatorOrCustodian = null as RoleTypes
 
   protected changeCareAndCustodyType (): void {
-    // Todo: Apply option to store when required
+    // FUTURE: save `liquidatorOrCustodian` to store when required
```
severinbeauvais, line 55: This component is not used at the moment, but I hesitate to delete it altogether.

src/components/Restoration/BusinessName.vue
severinbeauvais: Looks like we're not discussing this with Dave until Monday, but I'd like to get this ticket wrapped up well before then.

Flag ideas -- a code comment? A tech debt ticket?

src/components/Restoration/CorrectCompanyName.vue
severinbeauvais: As discussed at Standup this morning, I will make the duplicate components shared.

src/enums/filingTypes.ts
```diff
@@ -1,6 +0,0 @@
-export enum FilingTypes {
-  INCORPORATION_APPLICATION = 'incorporationApplication',
-  REGISTRATION = 'registration',
-  RESTORATION = 'restoration',
-  VOLUNTARY_DISSOLUTION = 'dissolution',
```
severinbeauvais, line 5: The is the source of the dissolution evil -- other UIs had a value of `voluntaryDissolution` here, which is invalid as there is no such filing type (it's called just `dissolution` and has a subtype).

src/App.vue
```diff
@@ -115,8 +115,9 @@
               <header>
                 <h1>{{ getFilingName }}</h1>
               </header>
-              <p class="mt-4" v-if="getFilingSubtitle">
```
severinbeauvais, line 118: This was overly complicated as it used an enum with a single key (FilingTypesSubTitle) and a getter with a single case.

We can make this complicated again in the future if we need to :)

src/App.vue
```diff
@@ -671,7 +676,7 @@ export default class App extends Vue {
       // then try to re-route them
       if (this.$route.meta.filingType !== this.getFilingType) {
         switch (this.getFilingType) {
-          case FilingTypes.VOLUNTARY_DISSOLUTION:
```
severinbeauvais, line 674: As mentioned in another comment, there is no such filing type. However, the old code worked because VOLUNTARY_DISSOLUTION = `dissolution`.

Two wrongs make a right?

src/store/getters/state-getters.ts
```diff
@@ -57,7 +57,7 @@ export const getFilingName = (state: StateIF): FilingNames => {
     case FilingTypes.INCORPORATION_APPLICATION: return FilingNames.INCORPORATION_APPLICATION
     case FilingTypes.REGISTRATION: return FilingNames.REGISTRATION
     case FilingTypes.RESTORATION: return FilingNames.RESTORATION_APPLICATION
-    case FilingTypes.VOLUNTARY_DISSOLUTION:
+    case FilingTypes.DISSOLUTION:
       return isTypeFirm(state) ? FilingNames.DISSOLUTION_FIRM : FilingNames.VOLUNTARY_DISSOLUTION
```
severinbeauvais, line 61: As you see here, this UI assumes that a dissolution filing is a voluntary dissolution. That works today and we can fix this up in the future if needed.

(We might have to check the dissolution filing subtype if we need to know what kind of dissolution it is. Today, this UI only handles voluntary dissolutions.)

tests/unit/App.spec.ts
```diff
@@ -636,7 +636,7 @@ describe('Incorporation - Define Company page for a BEN (named)', () => {
   })
 })
 
-describe('Dissolution - Define Dissolution page for a BEN', () => {
+describe('Voluntary Dissolution - Define Dissolution page for a BEN', () => {
```
severinbeauvais, line 639: See type (name) on line 845 below, and subtype on line 831 below.

src/components/Restoration/CorrectCompanyName.vue
ketaki-deodhar: Makes sense!

src/enums/index.ts
ketaki-deodhar: Eventually start using kebab-case everywhere OR we need to decode on standardization yet?

ketaki-deodhar (approved):

src/components/Restoration/BusinessName.vue
severinbeauvais: As discussed at Standup yesterday, I will make the duplicate components shared.

src/enums/index.ts
severinbeauvais: I think we should gradually change enum filenames to kebab-case.

I think we should leave component names in TitleCase.

Thoughts?

ketaki-deodhar: Yes. This approach makes sense

jonathan-longe (approved):

PR #513, opened 2023-02-10
14783 WIP

severinbeauvais:
*Issue #:* bcgov/entity#14783

*Description of changes:*


src/components/Incorporation/IncorporationDateTime.vue
```diff
@@ -423,7 +423,7 @@ label {
     top: -2rem;
     padding: 0 12px;
     font-size: $px-12;
-    font-weight: 500;
+    font-weight: normal;
```
severinbeauvais, line 426: We only use "normal" and "bold", as declared in the font section of the base CSS.

src/components/Incorporation/SummaryDefineCompany.vue
```diff
@@ -22,7 +22,7 @@
             <label id="company-label">Name</label>
           </v-col>
           <v-col cols="12" sm="9" class="pt-4 pt-sm-0">
-            <div id="company-name">{{ getNameRequestApprovedName || '[Incorporation Number] B.C. Ltd.' }}</div>
+            <div id="company-name">{{ getNameRequestApprovedName || '[Incorporation Number] B.C. LTD.' }}</div>
```
severinbeauvais, line 25: Everywhere else it's "LTD" so this is a fix. (It's also now consistent with the name components I'm porting over from Edit UI.)

src/components/Restoration/BusinessName.vue
```diff
@@ -104,15 +110,12 @@
         </v-col>
       </v-row>
     </div>
-
-    <!-- Name Translation -->
-    <NameTranslation @hasNameTranslation="emitHasNameTranslation($event)" />
```
severinbeauvais, line 109: Moved to parent component.

src/components/Restoration/BusinessName.vue
```diff
@@ -223,9 +212,10 @@ export default class BusinessName extends Vue {
     return this.toDisplayPhone(this.nrApplicant.phoneNumber)
   }
 
-  /** The current options for change of name correction or edit. */
-  get nameChangeOptions (): Array<NameChangeOptions> {
+  /** The current options for name changes. */
+  get correctionNameChoices (): Array<NameChangeOptions> {
     return [
+      // NameChangeOptions.CORRECT_NAME,
```
severinbeauvais, line 218: I think I'll leave this option here and commented out -- if we ever reuse this component in Create UI again then we can revisit this.

src/components/Restoration/CorrectCompanyName.vue
```diff
@@ -16,32 +16,18 @@
...
-})
+@Component({})
 export default class CorrectCompanyName extends Vue {
-  /** Form Submission Prop */
   @Prop({ default: null }) readonly formType!: NameChangeOptions
-
-  @Action setNameRequest!: ActionBindingIF
-
-  @Getter getNameRequestLegalName!: string
-  @Getter getNameRequest!: NameRequestIF
```
severinbeauvais, line 37: No longer using Actions or Getters -- now using Props and Emits!

src/components/Restoration/CorrectNameOptions.vue
```diff
@@ -205,7 +226,7 @@ export default class CorrectNameOptions extends Vue {
 }
 
 .action-btns {
-  margin: 30px 0;
+  margin-top: 30px;
```
severinbeauvais, line 229: Example:
image

src/components/Restoration/CorrectNameRequest.vue
```diff
@@ -1,59 +1,57 @@
...
+  <v-form id="correct-name-request-form" ref="form" v-model="formValid" lazy-validation>
     <ConfirmDialogShared
       ref="confirm"
       attach="#app"
     />
 
     <v-row no-gutters>
       <v-col cols="1" class="mt-1">
-        <v-btn x-small fab outlined :ripple="false" color="gray7" class="step-icon" tabindex="-1">1</v-btn>
+        <v-chip class="step-icon">1</v-chip>
```
severinbeauvais, line 10: Old:
image

New (and I'll update the style to match):
image

src/components/Restoration/CorrectNameRequest.vue
```diff
@@ -63,12 +61,23 @@
...
+
+interface NrCorrectionIF {
+  legalType: CorpTypeCd,
+  nrNumber: string,
+  legalName: string,
+  expiry: string,
+  requestType: string,
+  status: NameRequestStates,
+  applicant: NameRequestApplicantIF | object
+}
```
severinbeauvais, line 78: This may go away if I can reuse one of the other two (!) NR interfaces... Otherwise this is used here only.

src/components/Restoration/CorrectNameRequest.vue
```diff
@@ -63,12 +61,23 @@
...
+
+type FetchNameRequest = (nrNumber: string) => Promise<NrResponseIF>
```
severinbeauvais, line 80: This is so I can type a function that I pass in as a prop.

src/components/Restoration/RestorationType.vue
```diff
@@ -1,56 +1,63 @@
 <template>
-  <v-card flat id="restoration-type" class="pa-6">
-    <v-row>
+  <div id="restoration-type">
+    <div class="section-container" :class="{ 'invalid-section': invalidSection }">
+      <v-row no-getters>
```
severinbeauvais, line 4: "no-getters" gets rid of the weird negative margins, and means I don't need the "pa-6" you see in the old code on line 2.

src/components/Restoration/RestorationType.vue
```diff
@@ -1,56 +1,63 @@
 <template>
-  <v-card flat id="restoration-type" class="pa-6">
```
severinbeauvais, line 2: To be consistent with other components hanging off the parent page, I moved the v-card to the parent and I moved the invalid-section styling in here.

Example of WORKING validation error:

image

src/components/Restoration/RestorationType.vue
```diff
@@ -115,21 +130,5 @@ export default class RestorationType extends Vue {
       this.setRestorationExpiry(null)
     }
   }
-
-  @Watch('$route')
-  private async scrollToInvalidComponent (): Promise<void> {
-    if (this.$route.name === RouteNames.RESTORATION_BUSINESS_NAME) {
```
severinbeauvais, line 121: This is done in the parent, consistent with all other step components.

src/components/common/AddNameTranslation.vue
```diff
@@ -78,3 +78,10 @@ export default class AddNameTranslation extends Vue {
   protected removeTranslation (): void {}
 }
 </script>
+
+<style lang="scss" scoped>
+// un-bold the text field label
+.v-text-field :deep(.v-label) {
+  font-weight: normal;
+}
```
severinbeauvais, line 86: Before:
image
and
image

After:
image
and
image

This is now consistent with all other textfields in our UIs.

src/mixins/filing-template-mixin.ts
```diff
@@ -281,9 +285,13 @@ export default class FilingTemplateMixin extends DateMixin {
         this.setIncorporationAgreementStepData({
           agreementType: draftFiling.incorporationApplication.incorporationAgreement?.agreementType
         })
-        // set courtOrder attribute
-        this.setCourtOrderFileNumber(draftFiling.incorporationApplication.courtOrder?.fileNumber || '')
-        this.setHasPlanOfArrangement(draftFiling.incorporationApplication.courtOrder?.hasPlanOfArrangement || false)
+        // set court order fields
+        if (draftFiling.incorporationApplication.courtOrder?.fileNumber) {
+          this.setCourtOrderFileNumber(draftFiling.incorporationApplication.courtOrder.fileNumber)
```
severinbeauvais, line 290: I think I prefer this style, where we don't overwrite the store initial values (and so we don't set the initial value a second time).

src/components/Incorporation/SummaryDefineCompany.vue
```diff
@@ -15,39 +15,41 @@
...
-              v-for="(nameTranslation, index) in getNameTranslations"
-              class="text-uppercase"
-              :key="`name_translation_${index}`"
-            >
-              {{ nameTranslation.name }}
-            </div>
-          </v-col>
-        </v-row>
-      </article>
+      <template v-if="showCompanyName">
```
severinbeauvais, line 18: Hide whitespace changes to see what's updated here (not very much).

src/components/Incorporation/SummaryDefineCompany.vue
```diff
@@ -15,39 +15,41 @@
...
+        <!-- Name -->
+        <article class="section-container">
+          <v-row no-gutters>
+            <v-col cols="12" sm="3" class="pr-4">
+              <label id="company-label">Name</label>
+            </v-col>
+            <v-col cols="12" sm="9" class="pt-4 pt-sm-0">
+              <div id="company-name">{{ getNameRequestApprovedName || '[Incorporation Number] B.C. LTD.' }}</div>
```
severinbeauvais, line 26: To be consistent with everywhere else.

src/components/Restoration/ApprovalType.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+  <div id="approval-type">
+    <div class="section-container" :class="{ 'invalid-section': invalidSection }">
+      <v-row no-getters>
```
severinbeauvais, line 4: This is stub code and is to be replaced by @JazzarKarim 's recent work.

src/components/Restoration/BusinessName.vue
```diff
@@ -1,132 +1,66 @@
 <template>
   <div id="business-name">
-    <div class="section-container" :class="{'invalid-section': invalidSection}">
-      <!-- Business Name -->
+    <pre>getNameRequestApprovedName={{ getNameRequestApprovedName }}</pre>
+    <pre>isCorrectingName={{ isCorrectingName }}</pre>
```
severinbeauvais, line 4: For debugging only. I'll remove this before flight.

src/components/Restoration/BusinessName.vue
```diff
@@ -1,132 +1,66 @@
...
-        </template>
-      </v-row>
-
-      <!-- Name Request Applicant -->
-      <v-row v-if="hasNewNr" no-gutters class="sub-section">
-        <v-col cols="12" sm="3" class="pr-4">
-          <label class="pr-4">
-            <strong>Name Request Applicant</strong>
+            <div>ULC => NR 4766680</div>
+            <div>BEN => NR 3335087</div>
```
severinbeauvais, line 22: For my debugging only.

src/components/Restoration/BusinessName.vue
```diff
@@ -1,132 +1,66 @@
 <template>
   <div id="business-name">
-    <div class="section-container" :class="{'invalid-section': invalidSection}">
-      <!-- Business Name -->
+    <pre>getNameRequestApprovedName={{ getNameRequestApprovedName }}</pre>
+    <pre>isCorrectingName={{ isCorrectingName }}</pre>
+    <!-- Display Mode -->
+    <template v-if="isNewName">
+      <NameRequestInfo />
```
severinbeauvais, line 7: Using existing readonly component for displaying.

Below, using new component for editing.

src/components/Restoration/BusinessName.vue
```diff
@@ -135,183 +69,124 @@ import NameTranslation from '@/components/common/NameTranslation.vue'
...
-    return (originalName !== currentName)
+  /** The request action code for this filing type. */
+  get requestActionCode (): NrRequestActionCodes {
+    if (this.isRestorationFiling) return NrRequestActionCodes.RESTORE
+    return null
+  }
+
+  /** Whether a new business legal name was entered. */
+  get isNewName (): boolean {
+    console.log('*** nr approved name =', this.getNameRequestApprovedName)
```
severinbeauvais, line 122: WIP.

src/components/Restoration/BusinessType.vue
```diff
@@ -0,0 +1,34 @@
+<template>
```
severinbeauvais, line 1: New component that we didn't notice in the uxpin. Fortunately it's a simple one.

src/components/Restoration/CorrectCompanyName.vue
```diff
@@ -1,10 +1,10 @@
 <template>
```
severinbeauvais, line 1: This is the component copied from Edit UI, then refactored here to be drop-in ready into the shared repo.

src/components/Restoration/CorrectCompanyName.vue
```diff
@@ -50,36 +36,43 @@ export default class CorrectCompanyName extends Vue {
 
   /** Called when component is mounted. */
   mounted (): void {
+    // *** TODO: get name request approved name?
+    const legalName = this.nameRequest.legalName
```
severinbeauvais, line 40: WIP

src/components/Restoration/CorrectCompanyName.vue
```diff
@@ -50,36 +36,43 @@ export default class CorrectCompanyName extends Vue {
...
+    // process only when current form type matches
     if (this.formType === NameChangeOptions.CORRECT_NAME) {
       // set the new company name
-      this.setNameRequest({
-        ...this.getNameRequest,
-        legalName: this.companyName
+      // *** TODO: update name request approved name?
+      this.emitNameRequest({
+        ...this.nameRequest,
+        legalName: this.textfield
```
severinbeauvais, line 56: WIP -- the Name Request object actually doesn't have a `legalName` property. Instead, the approved name needs to be emitted to the parent to eventually be set in the store.

src/components/Incorporation/SummaryDefineCompany.vue
```diff
@@ -120,7 +122,8 @@ export default class SummaryDefineCompany extends Vue {
   // for template
   readonly RouteNames = RouteNames
 
-  // Getters
+  @Prop({ default: true }) readonly showCompanyName!: boolean
```
severinbeauvais, line 125: I made Company Name optional so I could reuse this component on the Restoration Review Confirm page.

src/components/Restoration/ApprovalType.vue
```diff
@@ -0,0 +1,51 @@
...
+  /** This section's validity state (when prompted by app). */
+  get invalidSection (): boolean {
+    return (this.getShowErrors && !this.getApprovalTypeValid)
+  }
+
+  get approvedByRegistrar (): boolean {
+    return (!!this.getRestoration.noticeDate && !!this.getRestoration.applicationDate)
+  }
+}
+</script>
```
severinbeauvais, line 51: @JazzarKarim This is a "shim" component between this app and your shared component. It's here that all store getters/actions are done, so that the parent component doesn't have to know about these things.

src/components/Restoration/BusinessName.vue
```diff
@@ -1,132 +1,70 @@
...
-      </v-row>
-
-      <!-- Name Request Applicant -->
-      <v-row v-if="hasNewNr" no-gutters class="sub-section">
-        <v-col cols="12" sm="3" class="pr-4">
-          <label class="pr-4">
-            <strong>Name Request Applicant</strong>
+            <!-- *** TODO: remove before flight -->
+            <div>ULC => NR 4766680</div>
+            <div>BC => NR 3335087</div>
```
severinbeauvais, line 25: For debugging only. I'll remove this before flight.

src/components/Restoration/BusinessName.vue
```diff
@@ -135,183 +73,132 @@ import NameTranslation from '@/components/common/NameTranslation.vue'
...
-  }
+  /** The current options for name changes. */
+  get correctionNameChoices (): Array<NameChangeOptions> {
+    // *** TODO: uncomment after testing
+    // if (this.isRestorationFiling) {
+    //   return [
+    //     NameChangeOptions.CORRECT_NAME_TO_NUMBER,
+    //     NameChangeOptions.CORRECT_NEW_NR
+    //   ]
+    // }
```
severinbeauvais, line 115: For debugging only. I'll remove this before flight.

src/components/Restoration/CorrectNameOptions.vue
```diff
@@ -1,8 +1,9 @@
 <template>
```
severinbeauvais, line 1: The component hierarchy is as follows:
```
RestorationBusinessName (step 1)
   |
   Business Name
      |
      Correct Name Options (this component)
         |
         CorrectCompanyName (not used for restorations -- Edit UI only)
         CorrectNameRequest
         CorrectNameToNumber
```

src/components/Restoration/CorrectNameOptions.vue
```diff
@@ -81,17 +98,22 @@ import CorrectNameRequest from '@/components/Restoration/CorrectNameRequest.vue'
...
 export default class CorrectNameOptions extends Vue {
-  /** The options to display */
-  @Prop() readonly correctionNameChoices!: Array<string>
+  @Prop({ default: null }) readonly actionTxt!: string
+  @Prop({ default: () => [] }) readonly correctionNameChoices!: Array<NameChangeOptions>
+  @Prop({ required: true }) readonly businessId!: string
+  @Prop({ required: true }) readonly entityType!: CorpTypeCd
+  @Prop({ required: true }) readonly nameRequest!: NameRequestIF
+  @Prop({ required: true }) readonly companyName!: string
+  @Prop({ required: true }) readonly fetchAndValidateNr!: () => Promise<NameRequestIF>
```
severinbeauvais, line 107: I decided to make these required instead of having defaults. An error is thrown if this component is called and the required props are not provided.

severinbeauvais, line 107: Look, I'm passing in a function that fetches and validates the NR (so that THIS component doesn't need to know about axios or urls, etc).

src/components/Restoration/CorrectNameRequest.vue
```diff
@@ -1,59 +1,58 @@
...
     <ConfirmDialogShared
       ref="confirm"
       attach="#app"
     />
 
     <v-row no-gutters>
-      <v-col cols="1" class="mt-1">
-        <v-btn x-small fab outlined :ripple="false" color="gray7" class="step-icon" tabindex="-1">1</v-btn>
+      <v-col cols="1" class="mt-3">
+        <v-chip outlined class="step-icon">1</v-chip>
```
severinbeauvais, line 10: Before:
image

After:
image

V-chip ftw!

src/components/Restoration/CorrectNameRequest.vue
```diff
@@ -1,59 +1,58 @@
...
+          id="nr-number"
           filled
+          persistent-hint
+          class="text-input-field"
           label="Enter the NR Number"
           hint="Example: NR 1234567"
-          persistent-hint
-          :rules="done && nrNumRules"
-          id="nr-number"
-          @keyup="uppercase('nameRequestNumber')"
```
severinbeauvais, line 23: This required a mixin, yuck.

src/components/Restoration/CorrectNameRequest.vue
```diff
@@ -177,71 +167,61 @@ export default class CorrectNameRequest extends Vue {
...
       }
     }
   }
 
-  /**
-   * Parse and Set the Name Request date to Store.
-   * @param nr The name request data
-   */
-  private parseNameRequest (nr: NrResponseIF): void {
-    const nrCorrection: NrCorrectionIF = {
```
severinbeauvais, line 196: This was some weird thing to convert from a proper NR object to some intermediate (local only) object, and then in filing-template-mixin only a few fields from this object are saved.

src/components/Restoration/RestorationType.vue
severinbeauvais: (I also did the same thing with the Approval Type component.)

src/components/Restoration/RestorationTypeSummary.vue
```diff
@@ -1,63 +0,0 @@
-<template>
```
severinbeauvais, line 1: This became the Summary Restore Business component.

src/components/Restoration/SummaryRestoreBusiness.vue
```diff
@@ -0,0 +1,173 @@
...
+}
+
+#company-description,
+.name-translation,
+#expires-on,
+#applicants-relationships,
+#court-order-number,
+#notice-date,
+#application-date {
+  color: $gray7;
```
severinbeauvais, line 171: I would have just used a class, but we already declared the IDs for unit testing.

src/components/common/NameRequestInfo.vue
```diff
@@ -80,42 +82,29 @@
         </v-row>
       </div>
     </template>
-
-    <!-- Name Translation -->
-    <NameTranslation
-      v-if="showNameTranslation"
-      @hasNameTranslation="emitHasNameTranslation($event)"
-    />
```
severinbeauvais, line 88: This was moved to the parent component so that THIS component could be NR stuff only.

src/components/common/NameRequestInfo.vue
```diff
@@ -176,58 +163,9 @@ export default class NameRequestInfo extends Vue {
   }
 
   /** The applicant. */
-  get applicant (): NameRequestApplicantIF {
-    return this.getNameRequest.applicants // not an array
-  }
-
-  /** The applicant's name. */
-  get applicantName (): string {
```
severinbeauvais, line 184: Now using a common (soon to be shared) mixin method for this.

Ditto for applicant address below.

src/components/common/NameTranslations.vue
```diff
@@ -185,41 +190,48 @@ export default class NameRequestInfo extends Vue {
...
-.translation-checkbox-label {
+.checkbox-label {
   margin-top: 1px;
   font-size: $px-14;
   color: $gray9;
 }
+
+p {
+  font-size: $px-14;
+}
```
severinbeauvais, line 236: Before:

image

After (now like all other text input field):

image

src/interfaces/filing-interfaces/filing-interfaces.ts
```diff
@@ -170,21 +170,23 @@ export interface RestorationFilingIF {
...
-    relationships?: RelationshipTypes[]
+    courtOrder: CourtOrderIF
+    expiry: string // YYYY-MM-DD
+    nameRequest: RegistrationNameRequestIF
+    nameTranslations: NameTranslationIF[]
+    noticeDate: string // YYYY-MM-DD
+    offices: RegisteredRecordsAddressesIF | object
+    parties: PartyIF[]
+    relationships: RelationshipTypes[]
+    type: RestorationTypes
```
severinbeauvais, line 190: ^^^ Sorted. (Ditto elsewhere.)

src/interfaces/resource-interfaces/component-resource-interfaces/certifyStatement-interface.ts
```diff
@@ -1,5 +0,0 @@
-export interface CertifyStatementIF {
```
severinbeauvais, line 1: Wasn't used.

src/mixins/common-mixin.ts
```diff
@@ -86,4 +86,71 @@ export default class CommonMixin extends Vue {
     const result = charSet.charAt(intVal)
     return result
   }
+
+  /**
```
severinbeauvais, line 90: This was copied over from Edit UI.

Ultimately (but please not in this PR), we could move this mixin to the shared repo.

src/mixins/filing-template-mixin.ts
```diff
@@ -209,7 +213,9 @@ export default class FilingTemplateMixin extends DateMixin {
     this.setOfficeAddresses(draftFiling.incorporationApplication.offices)
 
     // restore Name Translations
-    this.setNameTranslationState(draftFiling.incorporationApplication.nameTranslations || [])
+    if (draftFiling.incorporationApplication.nameTranslations) {
+      this.setNameTranslations(draftFiling.incorporationApplication.nameTranslations)
+    }
```
severinbeauvais, line 218: I think I prefer this style, where we don't overwrite the store initial values (and so we don't set the initial value a second time and in a second place).

src/resources/Dissolution/BC.ts
```diff
@@ -57,7 +57,6 @@ export const DissolutionResourceBc: DissolutionResourceIF = {
   },
   reviewAndConfirm: {
     completingPartyStatement: {
-      certifyStatementHeader: null,
```
severinbeauvais, line 60: Never used.

src/views/Restoration/RestorationBusinessName.vue
```diff
@@ -1,29 +1,32 @@
...
-    <!-- Name -->
+    <!-- Business Name and Name Translations -->
     <section id="name-section" class="mt-10">
       <header>
         <h2>Name</h2>
         <p>Add a Name Request that is reserved for this restoration application or restore
           as a numbered company.</p>
+          <pre>getBusinessNameValid={{ getBusinessNameValid }}</pre>
+          <pre>getNameTranslationsValid={{ getNameTranslationsValid }}</pre>
+          <pre>getRestorationTypeValid={{ getRestorationTypeValid }}</pre>
```
severinbeauvais, line 11: I'll remove these before flight.

src/components/Restoration/ApprovalType.vue
jonathan-longe: Rather than passing a prop and event for each form field, what do you think of passing the entire "ApprovalTypeShared" object?  If we're separating concerns, the parent shouldn't have to know the fields contained within a shared component and should have to know when we add or remove a field inside the shared component.

src/components/Restoration/CorrectCompanyName.vue
jonathan-longe: > So, no getters or actions. Props and events only.

You're trolling me!  :-)

src/components/Restoration/CorrectNameOptions.vue
jonathan-longe: I love the way that JavaScript makes functions first class citizens -- this is such a powerful technique!  Do we plan to pass in different functions, if so, can you provide some examples?

src/mixins/filing-template-mixin.ts
```diff
@@ -530,18 +546,39 @@ export default class FilingTemplateMixin extends DateMixin {
...
+    if (draftFiling.restoration.relationships) {
+      this.setRestorationRelationships(draftFiling.restoration.relationships)
+    }
+    if (draftFiling.restoration.noticeDate) {
+      this.setRestorationNoticeDate(draftFiling.restoration.noticeDate)
+    }
+    if (draftFiling.restoration.noticeDate) {
+      this.setRestorationApplicationDate(draftFiling.restoration.noticeDate)
+    }
```
jonathan-longe, line 568: Is there an opportunity to separate the code associated with each filingType?  I don't know if this would work, but I'd like to try diverting the logic by filingType early so we can simplify our code and reduce the number of times where we have to write a series of `if-then`, `if-then-else` or `switch` statements.

jonathan-longe:
Wow, @severinbeauvais there's lots of great work here.  You've had to modify a large number of files to make these changes.

severinbeauvais:
> Wow, @severinbeauvais there's lots of great work here. You've had to modify a large number of files to make these changes.

Yes, a lot of untangling. Some of these changes will also be moved to the shared repo.

severinbeauvais:
> You know the domain better than I do. My gut feel is that the following logic would be better placed in a getter. Having this logic {{ getNameRequestApprovedName || '[Incorporation Number] B.C. LTD.' }} inside a template means that we can't test without instantiating the component and all the associated requirements.

I don't mind putting "simple" logic like this in the template, otherwise I definitely use local getters (computed).

As for testing, in which cases would we test this component without instantiating (mounting) it? And I'm not sure how we could test anything without providing its requirements. Open to new ideas...

severinbeauvais:
> Rather than passing a prop and event for each form field, what do you think of passing the entire "ApprovalTypeShared" object? If we're separating concerns, the parent shouldn't have to know the fields contained within a shared component and should have to know when we add or remove a field inside the shared component.

What do you mean, "passing the entire shared component object"?

If you mean passing an object of properties instead of individual properties then I think sure, why not.

The components here are:
```
|
+ RestorationBusinessName.vue (step 1)
   |
   + ApprovalType.vue (this component, which is a shim to the shared component)
      |
      + ApprovalTypeShared.vue (actually ApprovalType but renamed in parent)
```

The top parent doesn't handling any props or events for its child.

The shim handled the store interactions and the props/events for its child.

The bottom component does the UI work and doesn't (and shouldn't) know about the store stuff.

How would you improve this?

severinbeauvais:
> I love the way that JavaScript makes functions first class citizens -- this is such a powerful technique! Do we plan to pass in different functions, if so, can you provide some examples?

Different functions for this scenario (CorrectNameOptions.vue) ? No,  just the one function. But I pass it in so the child can use it as a black-box service. Meanwhile the parent can deal with axios and whatever.

I didn't bother with a full type for this function, and almost use "any" for it. I only typed the return value for the child's benefit.

If I didn't answer your question, what are you thinking here?

severinbeauvais:
> Is there an opportunity to separate the code associated with each filingType? I don't know if this would work, but I'd like to try diverting the logic by filingType early so we can simplify our code and reduce the number of times where we have to write a series of if-then, if-then-else or switch statements.

Yes but not the code you attached your comment to. This particular code does a conditional restore of each field, so it doesn't overwrite the initial store value if the draft filing (JSON) doesn't contain it.

jonathan-longe:
> > @severinbeauvais what you do think about using a standard validation library like "VeeValidate" rather than exposing an event that the parent component must listen to? If I remember correctly, VeeValidate handles progressive form disclosure really well so that the parent doesn't need to know which form elements have been displayed and which are hidden. The parent shouldn't know which form elements have been renderd in order to know if the form is hidden.
> 
> We should look at this more deeply. I've played with VeeValidate before but didn't love it then. But we do have difficulty currently with progressive form disclosure. There's a ticket to look at form validation in more detail but it's not on anyone's radar. If you could champion this -- maybe after our current high-priority features are done -- that would be great.

Sure, I'd be happy to create a demo for your evaluation.  Let's revisit this after the high-priority features are complete.

jonathan-longe:
> > I love the way that JavaScript makes functions first class citizens -- this is such a powerful technique! Do we plan to pass in different functions, if so, can you provide some examples?
> 
> Different functions for this scenario (CorrectNameOptions.vue) ? No, just the one function. But I pass it in so the child can use it as a black-box service. Meanwhile the parent can deal with axios and whatever.
> 
> I didn't bother with a full type for this function, and almost use "any" for it. I only typed the return value for the child's benefit.
> 
> If I didn't answer your question, what are you thinking here?

If we have a plan to pass different functions into CorrectNameOptions to achieve different behaviors, then passing a function as an argument is the right approach.  In this case, I wonder if you would consider dispatching a store action instead?  Actions are specifically designed to handle API calls.  By moving the `fetchAndValidateNr()` method to the store it allows for other components to use it too and it removes business logic from components.

severinbeauvais:
> If we have a plan to pass different functions into CorrectNameOptions to achieve different behaviors, then passing a function as an argument is the right approach. In this case, I wonder if you would consider dispatching a store action instead? Actions are specifically designed to handle API calls. By moving the `fetchAndValidateNr()` method to the store it allows for other components to use it too and it removes business logic from components.

I don't have a plan to pass in a different function. I just wanted the child (which is a shared repo component) to not need to know how to fetch a name request nor have to deal with possible validation errors. I'm thinking those are all "parent" concerns, not the child's, even though the child is prompting for the NR number. These components are not a great architecture atm but I didn't want to reinvent them completely, especially since the plan is to slot them back into Edit UI.

Are you OK with this callback function as-is?

jonathan-longe:
> > Rather than passing a prop and event for each form field, what do you think of passing the entire "ApprovalTypeShared" object? If we're separating concerns, the parent shouldn't have to know the fields contained within a shared component and should have to know when we add or remove a field inside the shared component.
> 
> What do you mean, "passing the entire shared component object"?
> 
> If you mean passing an object of properties instead of individual properties then I think sure, why not.
> 
> The components here are:
> 
> ```
> |
> + RestorationBusinessName.vue (step 1)
>    |
>    + ApprovalType.vue (this component, which is a shim to the shared component)
>       |
>       + ApprovalTypeShared.vue (actually ApprovalType but renamed in parent)
> ```
> 
> The top parent doesn't handling any props or events for its child.
> 
> The shim handled the store interactions and the props/events for its child.
> 
> The bottom component does the UI work and doesn't (and shouldn't) know about the store stuff.
> 
> How would you improve this?

I would suggest that ApprovalTypeShared.vue *should* talk to the store.  We pass ApprovalTypeShared a store location (so it can be reused anywhere) and ApprovalTypeShared gets its status from the store and mutates the store as the user interacts with it.  ApprovalType.vue doesn't have to know anything about the inner workings of the ApprovalTypeShared.  We're free to add and remove fields from ApprovalTypeShared without having to make changes to ApprovalType.vue.  

If you're uncomfortable with the idea of using the store, an alternative approach is to use v-model to set / get the ApprovalTypeShared object.   Using v-model objects is a little tricky (as we've discussed previously), but it's another way of encapsulating the ApprovalTypeShared without having to set individual field propos and emit individual field updates.

jonathan-longe:
> > You know the domain better than I do. My gut feel is that the following logic would be better placed in a getter. Having this logic {{ getNameRequestApprovedName || '[Incorporation Number] B.C. LTD.' }} inside a template means that we can't test without instantiating the component and all the associated requirements.
> 
> I don't mind putting "simple" logic like this in the template, otherwise I definitely use local getters (computed).
> 
> As for testing, in which cases would we test this component without instantiating (mounting) it? And I'm not sure how we could test anything without providing its requirements. Open to new ideas...

I agree, it is simple logic, but it's business logic that I would suggest is better not in the template.  One reason is that many (all?) code editors can only show lines covered by tests when they're inside the <script>.  Logic in the template isn't shown as tested.

Also, my knowledge of the business is still a little shaky, but I think I remember reading that not all numbered incorporated entities end with "B.C. LTD.".  For example, wouldn't an incorporated CCC be written like this?  "0123456 B.C. 
Community Contribution Company Ltd."  If true, this suggests that the business logic is more complicated or that we have similar logic repeated elsewhere to cover different entity types.  If true, then I would suggest this logic belongs in a store getter so it can be reused in many components.

jonathan-longe:
I see what you mean about comments added as part of a PR can't be replied to!  That's super irritating!  I won't do that again. Thanks for the tip @severinbeauvais

severinbeauvais:
> > > Rather than passing a prop and event for each form field...

> > What do you mean, "passing the entire shared component object"?

> If you're uncomfortable with the idea of using the store, an alternative approach is to use v-model to set / get the ApprovalTypeShared object. Using v-model objects is a little tricky (as we've discussed previously), but it's another way of encapsulating the ApprovalTypeShared without having to set individual field propos and emit individual field updates.

I created ticket [15429](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/gh/bcgov/entity/15429) to look at this later. OK?

src/components/Incorporation/SummaryDefineCompany.vue
severinbeauvais: ^^ this and the related comment below are all very good points.

I have updated this code in a new commit. (Note that the "B.C. LTD." change for ULC and CCC is a separate ticket.)

src/components/Restoration/ApprovalType.vue
severinbeauvais: This thread continues in comments below.

src/components/Restoration/ApprovalType.vue
```diff
@@ -0,0 +1,52 @@
+<template>
+  <div id="approval-type">
+    <div class="section-container" :class="{ 'invalid-section': invalidSection }">
+      <ApprovalTypeShared
```
JazzarKarim, line 4: Does the label change to red when the approval type is not valid?

src/mixins/filing-template-mixin.ts
```diff
@@ -530,18 +546,39 @@ export default class FilingTemplateMixin extends DateMixin {
...
+    this.setRestorationApprovalType(draftFiling.restoration.approvalType)
+    if (draftFiling.restoration.courtOrder) {
+      this.setRestorationCourtOrder(draftFiling.restoration.courtOrder)
+    }
     this.setRestorationType(draftFiling.restoration.type)
-    this.setRestorationExpiry(draftFiling.restoration.expiry || null)
-    this.setRestorationRelationships(draftFiling.restoration.relationships || [])
+    if (draftFiling.restoration.expiry) {
+      this.setRestorationExpiry(draftFiling.restoration.expiry)
```
JazzarKarim, line 558: Is there a benefit for this rather than doing ```this.setRestorationExpiry(draftFiling.restoration.expiry || null)``` as before?

src/store/getters/state-getters.ts
```diff
@@ -400,10 +405,20 @@ export const isBusySaving = (state: StateIF): boolean => {
...
+}
 
-  return getDefineCompanyStep(state).valid && isValidCoopType
+/** Is true when the step is valid. */
+export const isRestoreBusinessNameValid = (state: StateIF): boolean => {
+  return (
+    getBusinessNameValid(state) &&
+    getNameTranslationsValid(state) &&
+    getRestorationTypeValid(state) &&
+    getApprovalTypeValid(state)
```
JazzarKarim, line 420: I really like this 😄. Instead of calling all the getters in the code when checking if valid, you created a new getter that calls all of them. Learning a lot here!

src/views/Restoration/RestorationBusinessName.vue
```diff
@@ -110,33 +123,6 @@ h2::before {
...
-    > label:first-child {
-      flex: 0 0 auto;
-      padding-right: 2rem;
-      width: 12rem;
-    }
-  }
-}
-
 header p {
   padding-top:0.5rem
```
JazzarKarim, line 127: It's not possible to do this with Vuetify spacing right? I searched last time everywhere for a way to do it but couldn't find any. I remember I asked Ethan about what to do with 15px, he told me just make it 16 which is pt-4.

src/components/Restoration/SummaryRestoreBusiness.vue
```diff
@@ -0,0 +1,179 @@
...
+
+  get relationships (): string {
+    return this.getRestoration.relationships.join(', ')
+  }
+
+  get noticeDate (): string {
+    return this.yyyyMmDdToPacificDate(this.getRestoration.noticeDate)
+  }
+
+  get applicationDate (): string {
```
JazzarKarim, line 135: Much cleaner than what I've previously done. Having all these computed value getters instead of inserting everything directly into the template. I'll be adapting this approach moving forward.

src/components/Restoration/ApprovalType.vue
severinbeauvais: This code changes the left border of the div to be red.

There is other code that changes the label to red.

image

src/mixins/filing-template-mixin.ts
severinbeauvais: Yes -- we don't need to set a default value here. It's already set when the store state is initialized. (DRY)

src/store/getters/state-getters.ts
severinbeauvais: Thanks :)

So there are 3 levels of getters:
- per section (eg, business name)
- per page/step (this one)
- for the overall filing (restoration)

src/views/Restoration/RestorationBusinessName.vue
severinbeauvais: Absolutely: it would be helper class "pt-2".

But this selects all paragraphs in all headers, which may be less code. (I didn't write this.)

src/components/Restoration/SummaryRestoreBusiness.vue
severinbeauvais: Jonathan also commented elsewhere that these are now testable in code (not looking at element text).

👍

src/components/Restoration/CorrectNameOptions.vue
severinbeauvais: This component is now in BCRS shared repo.

src/components/Incorporation/SummaryDefineCompany.vue
severinbeauvais: Updated and committed. This code is now on line 142.

src/components/Restoration/BusinessName.vue
severinbeauvais: I changed the code around a bit but my comment is still valid: using existig component + new code for editing.

src/components/Restoration/BusinessName.vue
```diff
@@ -1,132 +1,70 @@
 <template>
   <div id="business-name">
-    <div class="section-container" :class="{'invalid-section': invalidSection}">
-      <!-- Business Name -->
+    <!-- *** TODO: remove before flight -->
+    <!-- <pre>companyName={{ companyName }}</pre> -->
+    <!-- <pre>getNameRequestApprovedName={{ getNameRequestApprovedName }}</pre> -->
+    <!-- <pre>isCorrectingName={{ isCorrectingName }}</pre> -->
+
+    <!-- Editing Mode -->
```
severinbeauvais, line 8: Using new shared CorrectName component for editing.

Below, using existing NameRequestInfo component for displaying.

src/components/Restoration/CorrectCompanyName.vue
severinbeauvais: This component is now in BCRS shared repo.

src/components/Incorporation/SummaryDefineCompany.vue
severinbeauvais: This now uses a computed property + a store getter (since there are already getters in here). No more prop.

src/components/Restoration/CorrectNameOptions.vue
severinbeauvais: This component is now in BCRS shared repo.

src/components/Restoration/CorrectNameOptions.vue
severinbeauvais: This component is now in BCRS shared repo.

severinbeauvais: This component is now in BCRS shared repo.

src/components/Restoration/CorrectNameRequest.vue
severinbeauvais: This component is now in BCRS shared repo.

src/mixins/filing-template-mixin.ts
severinbeauvais: Yes but not the code you attached your comment to. This particular code does a conditional restore of each field, so it doesn't overwrite the initial store value if the draft filing (JSON) doesn't contain it.

src/views/Restoration/RestorationBusinessName.vue
severinbeauvais: I fixed the syntax on this line of code, so your comment no longer points to it ("Outdated").

Anyway, if you're OK with my answer then please resolve this conversation that you started,

src/components/Restoration/CorrectCompanyName.vue
severinbeauvais: This component is now in BCRS shared repo.

src/components/Restoration/CorrectCompanyName.vue
severinbeauvais: This component is now in BCRS shared repo.

src/components/Restoration/RestorationType.vue
severinbeauvais: To be consistent with other components hanging off the parent page, I moved the v-card to the parent and I moved the invalid-section styling in here.

(I also did the same thing with the Approval Type component. @JazzarKarim )

Example of restoration validation error:

image

Example of approval type validation error:

image

src/components/Restoration/RestorationType.vue
severinbeauvais: This is done in the parent, consistent with all other step components (@JazzarKarim ).

src/components/Restoration/CorrectNameRequest.vue
severinbeauvais: This component is now in BCRS shared repo.

src/components/Restoration/CorrectNameRequest.vue
severinbeauvais: This component is now in BCRS shared repo.

src/components/common/NameRequestInfo.vue
severinbeauvais: We can look at making the common mixin shared later.

src/components/Incorporation/SummaryDefineCompany.vue
severinbeauvais: @jonathan-longe OK to resolve this conversation?

src/components/Restoration/ApprovalType.vue
severinbeauvais: @jonathan-longe OK to resolve this conversation?

src/mixins/filing-template-mixin.ts
severinbeauvais: @jonathan-longe OK to resolve this conversation?

src/components/Restoration/ApprovalType.vue
severinbeauvais: @JazzarKarim OK to resolve this conversation?

src/mixins/filing-template-mixin.ts
severinbeauvais: @JazzarKarim OK to resolve this conversation?

src/store/getters/state-getters.ts
severinbeauvais: @JazzarKarim OK to resolve this conversation?

src/views/Restoration/RestorationBusinessName.vue
severinbeauvais: @JazzarKarim OK to resolve this conversation?

src/components/Restoration/SummaryRestoreBusiness.vue
severinbeauvais: @JazzarKarim OK to resolve this conversation?

src/components/Restoration/RestorationType.vue
```diff
@@ -1,56 +1,70 @@
 <template>
-  <v-card flat id="restoration-type" class="pa-6">
```
severinbeauvais, line 2: To be consistent with other components hanging off the parent page, I moved the v-card to the parent and I moved the invalid-section styling in here.

(I also did the same thing with the Approval Type component. @JazzarKarim )

Example of restoration validation error:

image

Example of approval type validation error:

image

src/components/Restoration/RestorationType.vue
```diff
@@ -65,71 +79,58 @@ import { LimitedRestorationPanel } from '@bcrs-shared-components/limited-restora
...
       this.setRestorationRelationships([])
-      this.setRestorationTypeValid(true)
     } else {
       this.setRestorationExpiry(null)
     }
   }
 
-  @Watch('$route')
-  private async scrollToInvalidComponent (): Promise<void> {
-    if (this.$route.name === RouteNames.RESTORATION_BUSINESS_NAME) {
```
severinbeauvais, line 121: This is done in the parent, consistent with all other step components (@JazzarKarim ).

severinbeauvais:
@jonathan-longe OK if I merge this? It will make it easier to finish the work in a subsequent PR.

JazzarKarim (approved):
Looks good to me 😃

jonathan-longe:
Amazing amount of work here @severinbeauvais .   Thanks for the opportunity to review.

PR #514, opened 2023-02-14
14797 Updated Restoration type to respond to UXA notes

JazzarKarim:

*Description of changes:*
 - New invalid design (default state):
new default invalid

 - New design with correct colors when relationships are selected:
new state

This default state might change. I'm just waiting on Ethan's reply: 
https://github.com/bcgov/entity/issues/14797#issuecomment-1428898641

### DIFF
--- a/src/components/Restoration/RestorationType.vue
+++ b/src/components/Restoration/RestorationType.vue
@@ -3,17 +3,22 @@
     <v-row>
 
       <v-col cols="12" sm="3" class="pr-4">
-        <label class="font-weight-bold">Restoration Type</label>
+        <label
+        v-bind:class="{'invalid': !getRestorationTypeValid, 'valid-title': getRestorationTypeValid}"
+        class="font-weight-bold">Restoration Type</label>
       </v-col>
 
       <v-col cols="12" sm="9">
         <v-radio-group column class="pt-0 mt-0" v-model="selectRestorationType">
-          <v-radio id="full-radio-button" label="Full Restoration" :value=RestorationTypes.FULL />
+          <v-radio id="full-radio-button" :value=RestorationTypes.FULL>
+            <span slot="label" class="valid-text">Full Restoration</span>
+          </v-radio>
 
           <!-- Relationship To Company Checkboxes -->
           <v-expand-transition>
-            <v-card flat v-if="isFull">
-              <div class="ml-8">
+            <v-card v-bind:class="{'invalid': !getRestorationTypeValid, 'valid-text': getRestorationTypeValid}"
+            flat v-if="isFull">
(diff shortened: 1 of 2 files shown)


src/components/Restoration/RestorationType.vue
```diff
@@ -3,38 +3,45 @@
...
+            <span slot="label" class="valid-text">Full Restoration</span>
+          </v-radio>
 
           <!-- Relationship To Company Checkboxes -->
           <v-expand-transition>
-            <v-card flat v-if="isFull">
-              <div class="ml-8">
+            <v-card v-bind:class="{'invalid': !getRestorationTypeValid, 'valid-text': getRestorationTypeValid}"
+            flat v-if="isFull">
+              <div class="ml-8 test">
```
severinbeauvais, line 21: Remove test code.

src/components/Restoration/RestorationType.vue
```diff
@@ -3,38 +3,45 @@
...
                     </span>
                   </template>
-                  Full restoration application must be related to the dissolved business.
+                  Full restoration applicant must be related to the dissolved business.
                 </v-tooltip>
                 to the company at the time the company was dissolved:
               </div>
               <RelationshipsPanel :draftRelationships="getRestoration.relationships"
-                @changed="setRestorationRelationships($event)" @valid="setRestorationTypeValid($event)" />
+                @changed="setRestorationRelationships($event)" @valid="setRestorationTypeValid($event)"/>
```
severinbeauvais, line 34: Actually, we like a space character before `/>`.

src/components/Restoration/RestorationType.vue
```diff
@@ -3,38 +3,45 @@
     <v-row>
 
       <v-col cols="12" sm="3" class="pr-4">
-        <label class="font-weight-bold">Restoration Type</label>
+        <label
+        v-bind:class="{'invalid': !getRestorationTypeValid, 'valid-title': getRestorationTypeValid}"
```
severinbeauvais, line 7: You can leave out `v-bind`.

You probably don't need exclusive conditional classes -- set one class here and then let CSS specificity overwrite it conditionally.

src/components/Restoration/RestorationType.vue
```diff
@@ -133,3 +140,18 @@ export default class RestorationType extends Vue {
...
+.valid-title {
+  color: $gray9;
+}
+.invalid {
+  color: $app-red;
+}
+
+.valid-text {
+  color: $gray7;
+}
```
severinbeauvais, line 156: Instead of adding spans with a class and then selecting that class here, I highly recommend you select the existing Vuetify classes (same as my comment in other PR).

PR #516, opened 2023-02-27
15183 Added Genesys Web Message to Create UI

JazzarKarim:
*Issue #:* /bcgov/entity#15183

*Description of changes:*
- Imported the new Genesys Web Chat Component.
  - Show only if filing type is SP/GP.
new chat


src/App.vue
```diff
@@ -84,6 +84,14 @@
       :webChatUrl="window['webChatUrl']"
     />
 
+    <!-- Display the Genesys WebMessage for SP/GP registrations only -->
+    <GenesysWebMessage
+      v-if="getFilingType === FilingTypes.REGISTRATION"
+      genesysURL="https://apps.cac1.pure.cloud/genesys-bootstrap/genesys.min.js"
+      environmentKey="cac1"
+      deploymentKey="42ed05f4-d545-436b-ba2c-94b66ed3396f"
+    />
```
severinbeauvais, line 93: Don't hard-code this -- get it from the configuration.json file.

jonathan-longe: Since this is a public forum, should we delete and create a new deployment key out of an abundance of caution?

JazzarKarim: My fault, I forgot about this! Should be okay now.

severinbeauvais: > Since this is a public forum, should we delete and create a new deployment key out of an abundance of caution?

This is a dev key -- I don't care if someone else sees it. If we do get spammed then we can change it. (I also don't want to destabilize Maximus by asking them for a new key.)

public/config/configuration.json
```diff
@@ -24,7 +27,7 @@
   "REGISTRY_HOME_URL": "https://dev.bcros.ca/",
   "REGISTRIES_SEARCH_API_URL": "https://bcregistry-dev.apigee.net/registry-search",
   "REGISTRIES_SEARCH_API_VERSION": "/api/v1",
-  "WEBCHAT_REASON": "",
+  "WEBCHAT_REASON": "zzz",
```
severinbeauvais: Remove debugging code.

src/utils/FetchConfig.ts
```diff
@@ -113,6 +113,18 @@ export async function FetchConfig (): Promise<any> {
   const sentryDsn: string = response.data.SENTRY_DSN;
   (<any>window).sentryDsn = sentryDsn
 
+  const genesysEnv: string = response.data.GENESYS_ENV;
+  (<any>window).genesysEnv = genesysEnv
+  if (genesysEnv) console.info('Set Genesys Environment to: ' + genesysEnv)
```
severinbeauvais, line 118: Remove these: we don't want to show them.

public/config/configuration.json
JazzarKarim: I removed it.

src/App.vue
```diff
@@ -84,6 +84,14 @@
       :webChatUrl="window['webChatUrl']"
     />
 
+    <!-- Display the Genesys WebMessage for SP/GP registrations only -->
+    <GenesysWebMessage
+      v-if="getFilingType === FilingTypes.REGISTRATION"
+      :genesysURL="window['genesysUrl']"
+      :environmentKey="window['genesysEnv']"
+      :deploymentKey="window['genesysId']"
+    />
```
jonathan-longe, line 93: Does the GenesysWebMessage component have to be initialized in App.vue?  I know it's not how App.vue is used at present, but typically App.vue would render the appropriate component from the routes and not much more.  Would it be possible to put this component in the appropriate view instead?

severinbeauvais: Actually, it's not even a Vue component. We made it a Vue component so we could easily drop it on a page like this. Where would such a component (or similarly, a dialog) be mounted?

JazzarKarim: Since we want to render the component for only registrations, I believe we can put this component in the Registration views. What is the real benefit for doing that though?

severinbeauvais: ^^ That's actually a decent idea 👍

src/App.vue
jonathan-longe: Hey Karim, I know this is an easy mistake to make.  I've come very close to committing secrets to public repos in Github before. The challenge is that Github history is difficult to delete.  When we squash and merge this PR into the main branch, there will be no trace of this key in the commit history, but it will remain in your personal Github branch and (I'm guessing) this PR too.  When I worked on a previous BCGov project, they took the disclosure of secrets -- even in DEV -- seriously and it had to be reported even when no personal information was disclosed.  If it's our decision whether or not to report it, then I'm happy to forget about it.  As Severin says the downside risk is small.

severinbeauvais: I don't know who we'd report this to. (And yes, I posted the keys first.)

Can we just forget about it? And if the dev webchat endpoint gets spammed, then we reset the keys?

jonathan-longe: Thinking about this further, the deployment key can't be a secret -- it's available to anyone who knows where to look in JavaScript.  This definitely doesn't need to be reported.  Sorry if I caused any alarm.

severinbeauvais: No, good question. Keep asking.

The key won't be in the JS but it will be in the configuration.json that is fetched.

JazzarKarim: @jonathan-longe That's actually good to know. Thanks for bringing that up Jonathan.

severinbeauvais: I'm satisfied in this conversation -- resolving it.

Jonathan, please open a new conversation if you want to continue chatting about this.

src/App.vue
JazzarKarim: @severinbeauvais Shall I proceed with this change or keep things as is? What do you think? If I do end up putting this component in the Registration views, I have to put in 3 different files. I don't think that's a good idea.

severinbeauvais: On the other hand, if we put the new component in the Registration view here, and in the App view in Namerequest UI, then it lacks consistency  :/

What's best practice here.

severinbeauvais (approved):
LGTM

src/App.vue
JazzarKarim: That's also a good point 😟

jonathan-longe: > Since we want to render the component for only registrations, I believe we can put this component in the Registration views. What is the real benefit for doing that though?

There's a couple of benefits.  First, the v-if can be removed and the component only rendered when needed. Second, it cleans up the code.

severinbeauvais: > @severinbeauvais Shall I proceed with this change or keep things as is? What do you think? If I do end up putting this component in the Registration views, I have to put in 3 different files. I don't think that's a good idea.

Oh yeah, good point -- we don't a single "Registration" view.

I'm OK with having in App.vue, but Jonathan started this conversation so I'll leave it for him to resolve.

JazzarKarim: I think I'm going to leave it as. This way, it is consistent with the Namerequest UI and also I don't have to import it over and over again.

jonathan-longe: I agree let's leave it as is.

PR #520, opened 2023-03-01
14783 Code cleanup

severinbeauvais:
*Issue #:* bcgov/entity#14783

*Description of changes:*
- replaced some local computed properties with existing store getters
- added handling for unknown restoration type
- added a unit test
- app version = 4.6.4

### DIFF
--- a/src/components/Restoration/SummaryRestoreBusiness.vue
+++ b/src/components/Restoration/SummaryRestoreBusiness.vue
@@ -51,13 +51,16 @@
           </v-col>
 
           <v-col cols="12" sm="9">
-            <template v-if="isLimitedRestoration">
+            <template v-if="isLimitedRestorationFiling">
               <p id="limited-restoration">Limited Restoration</p>
               <p id="expires-on">Expires on {{ expiry }}</p>
             </template>
-            <template v-else>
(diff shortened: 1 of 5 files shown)


src/components/Restoration/SummaryRestoreBusiness.vue
```diff
@@ -51,14 +51,17 @@
...
               <p id="expires-on">Expires on {{ expiry }}</p>
             </template>
-            <template v-else>
+            <template v-else-if="isFullRestorationFiling">
               <p id="full-restoration">Full Restoration</p>
               <p id="applicants-relationships">Applicant's relationship(s): {{ relationships }}</p>
             </template>
+            <template v-else>
+              <p id="unknown-restoration-type">[Unknown]</p>
+            </template>
```
severinbeauvais, line 64: This _shouldn't_ happen because we always initialize the restoration type, but just in case, it's incorrect to say "if it's not limited then it must be full", so now it's handled.

JazzarKarim: I'll be fixing that actually. The default state is going to change to nothing being selected. So, this is great.

JazzarKarim (approved):
LGTM

PR #522, opened 2023-03-03
Fix restoration type ux

JazzarKarim:

*Description of changes:*
- Final update to the restoration type design.
- New Default State:
new default state
- New error state of relationships:
new error relationships
- An overall view:
overal

@ethantspitt Please take a look and waiting for your feedback.

### DIFF
--- a/src/components/Restoration/RestorationType.vue
+++ b/src/components/Restoration/RestorationType.vue
@@ -23,7 +23,7 @@
             <!-- Relationship To Company Checkboxes -->
             <v-expand-transition>
               <div v-if="isFullRestorationFiling" :class="{ 'error-text': invalidSection }">
-                <div class="ml-8">
+                <div class="ml-8 tooltip-text">
                   Please select
                   <v-tooltip content-class="top-tooltip" transition="fade-transition" top>
                     <template v-slot:activator="{ on, attrs }">
@@ -117,7 +117,7 @@ export default class RestorationType extends Vue {
    */
   mounted (): void {
     if (this.isFullRestorationFiling && this.getRestoration.relationships.length === 0) {
-      this.setRestorationType('Full/Limited')
+      this.setRestorationType(null)
       this.setRestorationTypeValid(false)
     }
   }
@@ -139,16 +139,16 @@ export default class RestorationType extends Vue {
   @Watch('getRestoration.type', { immediate: true })
   private onRestorationType (val: RestorationTypes): void {
     this.selectRestorationType = val
-
-    // if Limited, set validity to true
-    // if Full then validity is set by RelationshipsPanel
-    if (this.isLimitedRestorationFiling) this.setRestorationTypeValid(true)
   }
 }
 </script>
 
 <style lang="scss" scoped>
 @import '@/assets/styles/theme.scss';
+.tooltip-text {
+  color: $gray7;
+}
+
 :deep() {
   // Fix font of all the radio buttons.
   .radio-button {
--- a/src/components/Restoration/RestorationType.vue
+++ b/src/components/Restoration/RestorationType.vue
@@ -22,8 +22,8 @@
 
             <!-- Relationship To Company Checkboxes -->
             <v-expand-transition>
-              <div v-if="isFullRestorationFiling" :class="{ 'error-text': invalidSection }">
-                <div class="ml-8 tooltip-text">
+              <div v-if="isFullRestorationFiling">
+                <div class="ml-8 tooltip-text" :class="{ 'error-text': invalidSection }">
                   Please select
                   <v-tooltip content-class="top-tooltip" transition="fade-transition" top>
                     <template v-slot:activator="{ on, attrs }">
--- a/src/components/Restoration/RestorationType.vue
+++ b/src/components/Restoration/RestorationType.vue
@@ -51,7 +51,7 @@
 
             <!-- Limited Restoration Radio Panel -->
             <v-expand-transition>
-              <div v-if="isLimitedRestorationFiling" class="mb-n12">
+              <div v-if="isLimitedRestorationFiling">
                 <LimitedRestorationPanel
                   :currentDate="getCurrentDate"
                   :expiryDate="getRestoration.expiry"
@@ -112,16 +112,6 @@ export default class RestorationType extends Vue {
   }
 
   /**
-   * Default State (if applicable) when we create new Restoration.
-   * No radio button is selected.
-   */
-  mounted (): void {
-    if (this.isFullRestorationFiling && this.getRestoration.relationships.length === 0) {
-      this.setRestorationType(null)
-    }
-  }
-
-  /**
    * Sets the selected Limited Restoration choice.
    * @param val the value of the selected radio button
    */
@@ -162,7 +152,7 @@ export default class RestorationType extends Vue {
     color: $gray7;
     font-weight: normal;
     align-items: stretch;
-    height: 1.3rem;
+    height: 1.25rem;
   }
 }
 </style>
--- a/src/components/Restoration/RestorationType.vue
+++ b/src/components/Restoration/RestorationType.vue
@@ -14,11 +14,15 @@
             v-model="selectRestorationType"
             @change="changeRestorationType()"
           >
-            <v-radio id="full-radio-button" label="Full Restoration" :value=RestorationTypes.FULL />
+            <v-radio
+              class="radio-button"
+              id="full-radio-button"
+              label="Full Restoration"
+              :value=RestorationTypes.FULL />
 
             <!-- Relationship To Company Checkboxes -->
             <v-expand-transition>
-              <div v-if="isFullRestorationFiling">
+              <div v-if="isFullRestorationFiling" :class="{ 'error-text': invalidSection }">
                 <div class="ml-8">
                   Please select
                   <v-tooltip content-class="top-tooltip" transition="fade-transition" top>
@@ -39,15 +43,20 @@
               </div>
             </v-expand-transition>
 
-            <v-radio id="limited-radio-button" label="Limited Restoration" :value=RestorationTypes.LIMITED />
+            <v-radio
+              class="pt-4 radio-button"
+              id="limited-radio-button"
+              label="Limited Restoration"
+              :value=RestorationTypes.LIMITED />
 
             <!-- Limited Restoration Radio Panel -->
             <v-expand-transition>
-              <div v-if="isLimitedRestorationFiling">
+              <div v-if="isLimitedRestorationFiling" class="mb-n12">
                 <LimitedRestorationPanel
                   :currentDate="getCurrentDate"
                   :expiryDate="getRestoration.expiry"
                   @expiry="setRestorationExpiry($event)"
+                  @valid="setRestorationTypeValid($event)"
                 />
               </div>
             </v-expand-transition>
@@ -103,6 +112,17 @@ export default class RestorationType extends Vue {
   }
 
   /**
+   * Default State (if applicable) when we create new Restoration.
+   * No radio button is selected.
+   */
+  mounted (): void {
+    if (this.isFullRestorationFiling && this.getRestoration.relationships.length === 0) {
+      this.setRestorationType('Full/Limited')
+      this.setRestorationTypeValid(false)
+    }
+  }
+
+  /**
    * Sets the selected Limited Restoration choice.
    * @param val the value of the selected radio button
    */
@@ -126,3 +146,24 @@ export default class RestorationType extends Vue {
   }
 }
 </script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+:deep() {
+  // Fix font of all the radio buttons.
+  .radio-button {
+    label {
+      font-weight: normal;
+      color: $gray7;
+    }
+  }
+
+  // Fix font of the relationships panel & center text with the checkboxes.
+  .v-input--checkbox .v-input__control .v-input__slot .v-label {
+    color: $gray7;
+    font-weight: normal;
+    align-items: stretch;
+    height: 1.3rem;
+  }
+}
+</style>
--- a/src/components/Restoration/RestorationType.vue
+++ b/src/components/Restoration/RestorationType.vue
@@ -118,7 +118,6 @@ export default class RestorationType extends Vue {
   mounted (): void {
     if (this.isFullRestorationFiling && this.getRestoration.relationships.length === 0) {
       this.setRestorationType(null)
-      this.setRestorationTypeValid(false)
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/components/Restoration/RestorationType.vue
```diff
@@ -14,11 +14,15 @@
...
+            <v-radio
+              class="radio-button"
+              id="full-radio-button"
+              label="Full Restoration"
+              :value=RestorationTypes.FULL />
 
             <!-- Relationship To Company Checkboxes -->
             <v-expand-transition>
-              <div v-if="isFullRestorationFiling">
+              <div v-if="isFullRestorationFiling" :class="{ 'error-text': invalidSection }">
```
JazzarKarim, line 25: Changing tooltip text to red (error) if invalid.

src/components/Restoration/RestorationType.vue
```diff
@@ -39,15 +43,20 @@
               </div>
             </v-expand-transition>
 
-            <v-radio id="limited-radio-button" label="Limited Restoration" :value=RestorationTypes.LIMITED />
+            <v-radio
+              class="pt-4 radio-button"
```
JazzarKarim, line 47: This extra padding is to keep spacing consistent between all the radio buttons.

src/components/Restoration/RestorationType.vue
```diff
@@ -39,15 +43,20 @@
...
+              id="limited-radio-button"
+              label="Limited Restoration"
+              :value=RestorationTypes.LIMITED />
 
             <!-- Limited Restoration Radio Panel -->
             <v-expand-transition>
-              <div v-if="isLimitedRestorationFiling">
+              <div v-if="isLimitedRestorationFiling" class="mb-n12">
```
JazzarKarim, line 54: Decrease the spacing between the restoration type and approval type.

src/components/Restoration/RestorationType.vue
```diff
@@ -102,6 +111,17 @@ export default class RestorationType extends Vue {
     return (this.getShowErrors && !this.getRestorationTypeValid)
   }
 
+  /**
+   * Default State (if applicable) when we create new Restoration.
+   * No radio button is selected.
+   */
+  mounted (): void {
```
JazzarKarim, line 118: When we're creating a new restoration, we create a new full restoration with no relationships. The UX team wants the default state to no radio button being selected. This deals with that.

src/components/Restoration/RestorationType.vue
```diff
@@ -126,3 +146,24 @@ export default class RestorationType extends Vue {
   }
 }
 </script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+:deep() {
+  // Fix font of all the radio buttons.
+  .radio-button {
```
JazzarKarim, line 154: Changing the color of all the radio button labels to our `$gray7`.

JazzarKarim:
> @JazzarKarim the screencap looks good, let me know when the ticket is moved to UXA, I'll take a look at the interactions. Thanks Karim!

Thank you Ethan!

src/components/Restoration/RestorationType.vue
JazzarKarim: Fee summary in default state:
fee summary in default state

src/mixins/filing-template-mixin.ts
```diff
@@ -427,7 +427,7 @@ export default class FilingTemplateMixin extends DateMixin {
         offices: this.getDefineCompanyStep.officeAddresses,
         parties: this.orgPersonsToParties(this.getAddPeopleAndRoleStep.orgPeople),
         relationships: this.getRestoration.relationships,
-        type: this.getRestoration.type
+        type: this.getRestoration.type || RestorationTypes.FULL
```
JazzarKarim, line 430: When saving with restoration type in default state, an error was being shown (unable to save). This is to fix that.

src/mixins/filing-template-mixin.ts
```diff
@@ -418,7 +418,7 @@ export default class FilingTemplateMixin extends DateMixin {
             ? { extension: +this.getBusinessContact.extension }
             : {}
         },
-        expiry: this.getRestoration.expiry || undefined, // can't be null
+        expiry: this.getRestoration.restorationTypeValid ? this.getRestoration.expiry : undefined, // can't be null
```
JazzarKarim, line 421: If user saves with invalid number of months and refreshes, handle it.

src/components/Restoration/RestorationType.vue
```diff
@@ -119,10 +139,31 @@ export default class RestorationType extends Vue {
   @Watch('getRestoration.type', { immediate: true })
   private onRestorationType (val: RestorationTypes): void {
     this.selectRestorationType = val
-
-    // if Limited, set validity to true
-    // if Full then validity is set by RelationshipsPanel
-    if (this.isLimitedRestorationFiling) this.setRestorationTypeValid(true)
```
JazzarKarim, line 125: Validation now is in the sub-component (Limited Restoration Panel).

src/components/Restoration/RestorationType.vue
severinbeauvais: In my opinion, you should only use negative bottom margin if this component is too tall... and there's usually a better way to fix that.

The spacing between component should be set in the parent. THIS component should not decrease the spacing to the next component.

JazzarKarim: I was actually debating whether I should fix this spacing in the parent or in this component. I got my answer. Thanks Sev.

src/components/Restoration/RestorationType.vue
severinbeauvais: I don't think clearing the restoration type (as below) is the right way to handle this. We should instead NOT set the initial restoration draft to Full. That will probably break other things, like the draft filing in the Todo List, but that's an easy fix. Can you identify anything else that will break?

src/components/Restoration/RestorationType.vue
```diff
@@ -119,10 +138,31 @@ export default class RestorationType extends Vue {
...
+      color: $gray7;
+    }
+  }
+
+  // Fix font of the relationships panel & center text with the checkboxes.
+  .v-input--checkbox .v-input__control .v-input__slot .v-label {
+    color: $gray7;
+    font-weight: normal;
+    align-items: stretch;
+    height: 1.3rem;
```
severinbeauvais, line 165: This is a very weird value. How did you decide on it? Why is it not a multiple of 0.25rem? Or if it's a specific value then maybe it should be in px instead.

src/mixins/filing-template-mixin.ts
severinbeauvais: So the user can store an invalid number of months (thus making expiry non-null but not valid)?

If we can't easily fix the underlying problem then this needs a comment (what you wrote above is nice) in the code to explain it.

src/components/Restoration/RestorationType.vue
severinbeauvais: Components should have "tight" spacing (no whitespace around them) so they can be reused more easily.

src/mixins/filing-template-mixin.ts
JazzarKarim: It would as follows:
1. The user saves with invalid number of months.
2. Since the resoration type was invalid, expiry will be undefined.
no expiry
3. When user refreshes, the selection will still be limited restoration, but the number of months will revert to the original which is 2 years.

src/components/Restoration/RestorationType.vue
JazzarKarim: Thing is, I didn't want to go into that rabbit hole of changing Filings UI for this specific case of default state. I'm trying now to fix it in Filings UI but to no avail. I will look into it more.

src/mixins/filing-template-mixin.ts
severinbeauvais: This is a code smell -- a (small) red flag that something is not right in the design or handling of the data, needing a special work-around to handle it.

What's the root problem we are trying to solve here -- that an invalid number of months was entered? Then one of 2 things should happen instead:
- preferably, we don't save (to store or draft) the invalid number of months; the component can still have a variable set to that value, but don't save it to the store
- or, we save invalid months, but the component indicates that it in an invalid value -- that way, upon resuming a draft, we can still indicate to the user that it is invalid and the filing cannot be submitted

I constantly ask myself, "is this the right place to handle something?" Is it the current code's job to do something with that data (logic), and more generally, should the current code know about this level of information?

In the current situation, it seems to me that this code should only save the data that is in the store, not apply rules to it. So, some other code needs to be responsible for ensuring that the data in the store is valid.

src/components/Restoration/RestorationType.vue
severinbeauvais: There are a couple of code smells here. First, the special-case logic (which additionally only happens at a particular point in time -- when the component is mounted). Second, the double fee codes (I think it should show no fee codes if a type hasn't been selected).

If the UX team wants the default to be "neither full nor limited resto" then let's handle that directly. But are we saying that if the user doesn't select any relationships, saves the draft, and resumes it, that we should clear their previous full resto selection? That doesn't sound right to me.

I think the thing to do here is separately handle the full/limited selection from the relationships selection.

In Filings UI, the Type is defaulted when the initial draft is created. You should be able to just remove that, but as I said, a few areas will need to be updated to handle "no" Type selected. Shouldn't be too bad though. LMK if I can help.

One last thing: if this is a significant updated requirement then we may need to review the impacts and push back a bit, maybe create a new ticket and prioritize it against all the other tickets. This feels a bit out of scope for "just some UX tweaks".

src/mixins/filing-template-mixin.ts
JazzarKarim: I 100% agree. Thing is, no matter how much how tried, I couldn't get the component to validate on load. So I opted to go this path which is not good.

I think on Friday my mind was so fried, I couldn't think straight. I just figured a small change to let my component emit a validation on load. This change is going to be in the sub-component so the current code doesn't need to worry about that.

The way it's going to work, I'll let the user save an invalid number of months. When the user saves and refreshes, the component is going to be invalid. Just like we do with the other components.

src/components/Restoration/RestorationType.vue
JazzarKarim: Agreed. I was thinking, shall merge this PR without the default state change and create another ticket that will handle that? This is because this involves changing Filings UI and moreover, it's not an "UX" change. 

Also, having not worked with Filings UI before, I just a little bit of time to get familiar with the codebase there. Shouldn't take too long though.

severinbeauvais: I'm in favour of leaving the initial restoration type out of this PR.

PS - in Filings UI, start here: https://github.com/bcgov/business-filings-ui/blob/0b49f6659e120bef4acc9f9bfc09f1f8855b8cc3/src/components/Dashboard/StaffNotation.vue#L121

JazzarKarim: Alright, sounds good. Thanks Sev 😄

src/mixins/filing-template-mixin.ts
severinbeauvais: This change was reverted. Closing this conversation.

src/views/Restoration/RestorationBusinessName.vue
```diff
@@ -22,7 +22,7 @@
       </header>
       <v-card flat class="mt-5">
         <RestorationType id="restoration-type" />
-        <ApprovalType id="approval-type" class="mt-n8" />
+        <ApprovalType id="approval-type" class="mt-n10" />
```
severinbeauvais, line 25: This can be fixed later, but it looks like ApprovalType has too much bottom whitespace in it. Again, components should have "tight" spacing (no whitespace around them), and then the parent can add however much whitespace it needs.

JazzarKarim: The problem is actually the error message of the number of month(s) text field. It's taking a lot of space and pushing things downward. Another fix would be to decrease the height of it in the shared component.

severinbeauvais: There may be a (Vuetify) component property to hide the error message when it's empty, thus removing that abundant bottom whitespace. However, that may cause screen jerkiness when the message is shown/hidden -- need to test and see which is best.

Yes, the shared component should not have excess whitespace. But that can be fixed later.

jonathan-longe:
Thanks for the opportunity to review @JazzarKarim.  Is the text shown in the image below created in RestorationType.vue?
image

JazzarKarim:
> Thanks for the opportunity to review @JazzarKarim. Is the text shown in the image below created in RestorationType.vue? image

Thanks for taking the time to review Jonathan!
It's this line: https://github.com/JazzarKarim/business-create-ui/blob/03f0eb7dcb5287f8d9130d7be68f65085c3f7e19/src/views/Restoration/RestorationBusinessName.vue#L20
It's actually in the RestorationBusinessName.vue.

jonathan-longe:
> Thanks for taking the time to review Jonathan! It's this line: https://github.com/JazzarKarim/business-create-ui/blob/03f0eb7dcb5287f8d9130d7be68f65085c3f7e19/src/views/Restoration/RestorationBusinessName.vue#L20 It's actually in the RestorationBusinessName.vue.

Yes, I saw that, but where's the number "2." coming from?
```
      <header>
        <h2>Restoration Type</h2>
        <p>Determine the restoration and approval type.</p>
      </header>
```

jonathan-longe (approved):
Thanks for the opportunity to review @JazzarKarim

JazzarKarim:
> > Thanks for taking the time to review Jonathan! It's this line: https://github.com/JazzarKarim/business-create-ui/blob/03f0eb7dcb5287f8d9130d7be68f65085c3f7e19/src/views/Restoration/RestorationBusinessName.vue#L20 It's actually in the RestorationBusinessName.vue.
> 
> Yes, I saw that, but where's the number "2." coming from?
> 
> ```
>       <header>
>         <h2>Restoration Type</h2>
>         <p>Determine the restoration and approval type.</p>
>       </header>
> ```

Hey Jonathan. So the 2. is coming from here:
https://github.com/JazzarKarim/business-create-ui/blob/03f0eb7dcb5287f8d9130d7be68f65085c3f7e19/src/views/Restoration/RestorationBusinessName.vue#L117-L121

src/components/Restoration/RestorationType.vue
```diff
@@ -14,12 +14,16 @@
             v-model="selectRestorationType"
             @change="changeRestorationType()"
           >
-            <v-radio id="full-radio-button" label="Full Restoration" :value=RestorationTypes.FULL />
+            <v-radio
+              class="radio-button"
+              id="full-radio-button"
+              label="Full Restoration"
+              :value=RestorationTypes.FULL />
```
severinbeauvais, line 21: Next time, I'd prefer to see the closing bracket on the next line (in line with the opening bracket).

severinbeauvais (approved):

PR #527, opened 2023-03-16
Fixed Business Lookup in SPs/GPs by readding apigee

JazzarKarim:
*Issue #:* /bcgov/entity#15183

*Description of changes:*
After updating security header when updating the Maximus webchat, `*apigee.net` was omitted which in turn made business lookup in SP/GP filings not work. This is to fix that issue.


severinbeauvais (approved):
Please update app version to `4.6.8a`.

argush3 (approved):

PR #531, opened 2023-03-30
Fixed the content path issue.

pwei1018:
*Issue #:* [15499](https://github.com/bcgov/entity/issues/15499)

*Description of changes:*
The current VUE_APP_PATH of GCP CD setting cause a null value exception after deploy to firebase. It must be changed to "/" value to avoid exception.

Also, open a ticket to investigate BASE_URL issue. https://github.com/bcgov/entity/issues/15817

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,5 @@
-# Base Path Openshift: businesses/create Firebase:
-VUE_APP_PATH=businesses/create
+# Base Path Openshift: /businesses/create Firebase: /
+VUE_APP_PATH=/businesses/create
 
 #vaults Shared
 VUE_APP_ADDRESS_COMPLETE_KEY=
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,5 @@
 # Base Path Openshift: /businesses/create Firebase: /
-VUE_APP_PATH=/businesses/create
+VUE_APP_PATH="/businesses/create"
 
 #vaults Shared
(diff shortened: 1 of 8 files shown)


.env.example
```diff
@@ -1,5 +1,5 @@
-# Base Path Openshift: businesses/create Firebase:
-VUE_APP_PATH=businesses/create
+# Base Path Openshift: /businesses/create Firebase: /
+VUE_APP_PATH=/businesses/create
```
severinbeauvais, line 2: Are quotes needed?

pwei1018: Doesn't matter, you can add or remove it.

severinbeauvais: Then let's keep it consistent with the other values in here.

Also in the fine below please.

pwei1018: Ok, I will add the quote for all places.

.env.example
```diff
@@ -1,5 +1,5 @@
...
 
 #vaults Shared
 VUE_APP_ADDRESS_COMPLETE_KEY=
```
severinbeauvais, line 5: Are you going to update lines 9 and 10 below?

severinbeauvais (approved):

.env.example
severinbeauvais: Yeah, sorry -- I was confused by the today's auto-redirect to the GCP environment, and the env file you sent me. They're 2 separate concerns. Nothing to change here.

PR #534, opened 2023-04-18
15840 added new Dev LD key to sample config file

severinbeauvais:
*Issue #:* bcgov/entity#15840

*Description of changes:*
- added common LD project key (Dev) to example env file
- app version = 5.0.5


.env.example
```diff
@@ -27,7 +27,7 @@ VUE_APP_REGISTRIES_SEARCH_API_VERSION="/api/v1"
 VUE_APP_REGISTRIES_SEARCH_API_KEY=
 
 #vaults launchdarkly
-VUE_APP_BUSINESS_CREATE_LD_CLIENT_ID=
+VUE_APP_BUSINESS_CREATE_LD_CLIENT_ID="642af8daefad6f134479c602"
```
severinbeauvais, line 30: This will make is easier to do local development. (Still need to copy this file to `.env` though.)

Let me know if you disagree with putting the Dev key in here.

pwei1018: Just not a good practices. It doesn't hurt. You can enable secret scanning in your github repository to see any complaints about it. https://docs.github.com/en/code-security/secret-scanning/about-secret-scanning

severinbeauvais: If it doesn't hurt, can I merge this change?

severinbeauvais: I enabled Secret Scanning for this repo and we'll see if I get an email with any secrets found.

JazzarKarim (approved):

.env.example
severinbeauvais: No secrets found yet, but maybe they'll show up after I merge this PR.

PR #540, opened 2023-04-28
16175 Vue-CLI and related packages upgrade + misc fixes

severinbeauvais:
**Note that this will be merged to a _feature branch_.**

*Issue #:* bcgov/entity#16175

*Description of changes:*

- updated .browserslistrc
- added eslint rule overrides
- replaced "array.prototype.move" by "array-move"
- upgraded pdfjs-dist to a specific (working) version
- added pinia packages (to check for conflicts)
- upgraded Vue-CLI to v5
- upgrade typescript to v4
- upgraded eslint to v7
- upgraded other misc packages
- misc cleanup
- changed `mixins: [...]` to `extends Mixins(...)` to retain typings
- changed some properties and methods to public to resolve template warnings
- fixed misc typing errors
- fixed misc lint errors (console in build, v-for and template, etc)
- fixed "this" in template
- fixed stepper styling
- fixed null address type
- fixed FileUploadPreview validation + cleanup
- fixed memory leak in document-mixin
- now import Vue from 'vue-property-decorator' instead of 'vue'
- renamed memorandumDoc -> memorandumFile
- renamed rulesDoc -> rulesFile
- renamed affidavitDoc -> affidavitFile
- deleted unneeded doc-interface.ts (docIF)
- imported App before styles to pre-empt global style issues
- imported lodash.omit to work around a lodash type issue
- refactored document-mixin to work with new library
- changed Vue.nextTick() to this.$nextTick (except in tests)
- fixed unit test naming issue
- fixed some unit tests (timeout issues, memory leaks, async optimization)


.eslintrc.js
```diff
@@ -9,6 +9,9 @@ module.exports = {
     '@vue/typescript'
   ],
   rules: {
+    'no-undef': 'off',
+    'no-unused-vars': 'off',
+    'prefer-const': 'off',
```
severinbeauvais, line 14: To shush new errors flagged by the updated eslint.

src/mixins/document-mixin.ts
```diff
@@ -4,8 +4,8 @@ import { AxiosResponse } from 'axios'
 import { AxiosInstance as axios } from '@/utils'
 import { DocumentUpload, PdfInfoIF } from '@/interfaces'
 import { PdfPageSize } from '@/enums'
-import pdfjsLib from 'pdfjs-dist/build/pdf'
-pdfjsLib.GlobalWorkerOptions.workerSrc = require('pdfjs-dist/build/pdf.worker.entry')
+// import pdfjsLib from 'pdfjs-dist/build/pdf'
+// pdfjsLib.GlobalWorkerOptions.workerSrc = require('pdfjs-dist/build/pdf.worker.entry')
```
severinbeauvais, line 8: This package has a webpack conflict so I deleted it temporarily.

seeker25: Just update it, it's what I did in `business-edit-ui`

`    "pdfjs-dist": "2.6.347",`

seeker25: Note: haven't tested

severinbeauvais: ^^ That doesn't work. Getting the following when trying to upload a PDF:

image

However, this works:
```
// import library and create worker
// NB: do not convert require to an import
const pdfjsLib = require('pdfjs-dist/build/pdf.js')
pdfjsLib.GlobalWorkerOptions.workerPort = new Worker(
  new URL('pdfjs-dist/build/pdf.worker.js', import.meta.url)
)
```
then
```
    ...
    const data = new Uint8Array(arrayBuffer) // put it in a Uint8Array
    const document = await pdfjsLib.getDocument({ data }).promise // note use of "promise"
    for (let pageNum = 1; pageNum <= document.numPages; pageNum++) 
    ...
```

severinbeauvais: However, the unit tests are unhappy with the above, working code:

[tests.log](https://github.com/bcgov/business-create-ui/files/11375465/tests.log)

src/components/Incorporation/ListShareClass.vue
```diff
@@ -204,17 +200,17 @@ export default class ListShareClass extends Vue {
    * @param seriesIndex The index of the series
    */
   protected moveIndex (indexFrom: number, direction: string, seriesIndex = -1): void {
-    let indexTo
+    let indexTo: number
     if (seriesIndex >= 0) {
       indexTo = direction === 'up' ? seriesIndex - 1 : seriesIndex + 1
       this.shareClasses[indexFrom].series[seriesIndex].priority = indexTo
       this.shareClasses[indexFrom].series[seriesIndex].priority = indexFrom
-      this.shareClasses[indexFrom].series.move(seriesIndex, indexTo)
```
severinbeauvais, line 212: `array.prototype.move`, imported above, caused build conflicts with new packages.

src/components/common/Actions.vue
```diff
@@ -314,7 +306,7 @@ export default class Actions extends Vue {
 
   // FUTURE: merge this with NameRequestMixin::validateNameRequest()
   /** Fetches NR and validates it. */
-  private async validateNameRequest (nrNumber: string): Promise<void> {
+  private async _validateNameRequest (nrNumber: string): Promise<void> {
```
severinbeauvais, line 309: I had to rename this to fix a name conflict.

src/components/common/Actions.vue
```diff
@@ -350,7 +342,7 @@ export default class Actions extends Vue {
   /** Label for the Next button. */
   get nextButtonLabel (): string {
     const nextStep = this.next()
-    return nextStep ? nextStep.text.replace('\n', ' ') : ''
+    return nextStep ? nextStep.text.replaceAll('\n', ' ') : ''
```
severinbeauvais, line 345: Another fix -- replace() only changes the first match, so we need to use replaceAll().

src/mixins/common-mixin.ts
```diff
@@ -1,6 +1,6 @@
-import Vue from 'vue'
-import { Component } from 'vue-property-decorator'
-import { omit, isEqual } from 'lodash'
+import { Component, Vue } from 'vue-property-decorator'
+import { isEqual } from 'lodash'
+import omit from 'lodash.omit'
```
severinbeauvais, line 3: `omit` couldn't be imported properly from lodash, but this works.

src/mixins/document-mixin.ts
```diff
@@ -1,10 +1,11 @@
...
-import { Component } from 'vue-property-decorator'
+import { Component, Vue } from 'vue-property-decorator'
 import { AxiosResponse } from 'axios'
 import { AxiosInstance as axios } from '@/utils'
 import { DocumentUpload, PdfInfoIF } from '@/interfaces'
 import { PdfPageSize } from '@/enums'
-import pdfjsLib from 'pdfjs-dist/build/pdf'
+
+// NB: must use require instead of import
+const pdfjsLib = require('pdfjs-dist/build/pdf.js')
```
severinbeauvais, line 8: If imported, the next line fails. I'm guessing `require` is async.

src/mixins/document-mixin.ts
```diff
@@ -60,15 +61,15 @@ export default class DocumentMixin extends Vue {
...
     const pageSizeInfo = this.pageSizeDict[pageSize]
-    const pdfBufferData = await file.arrayBuffer()
-    const pdfData = new Uint8Array(pdfBufferData) // put it in a Uint8Array
-    const pdf = await pdfjsLib.getDocument({ data: pdfData })
-    for (let pageNum = 1; pageNum <= pdf.numPages; pageNum++) {
-      const p1 = await pdf.getPage(pageNum)
-      const [x, y, w, h] = p1._pageInfo.view
+    const arrayBuffer = await file.arrayBuffer()
+    const data = new Uint8Array(arrayBuffer) // put it in a Uint8Array
+    const document = await pdfjsLib.getDocument({ data }).promise
```
severinbeauvais, line 66: This is another change: need to await the promise (or use `.then(...)`).

src/shims-tsx.d.ts
```diff
@@ -1,13 +0,0 @@
-import Vue, { VNode } from 'vue'
```
severinbeauvais, line 1: I don't think this was used by anything, as we don't have JSX or TSX files.

src/views/Incorporation/IncorporationShareStructure.vue
```diff
@@ -97,7 +97,7 @@
       />
     </v-card>
 
-    <v-card flat v-if="this.shareClasses.length > 0">
```
severinbeauvais, line 100: This previously worked but is not flagged as an error.

src/views/Restoration/RestorationApplicantInformation.vue
```diff
@@ -36,7 +32,7 @@ export default class RestorationApplicantInformation extends Vue {
   private async scrollToInvalidComponent (): Promise<void> {
     if (this.getShowErrors && this.$route.name === RouteNames.INCORPORATION_PEOPLE_ROLES) {
       // scroll to invalid components
-      await Vue.nextTick()
+      await this.$nextTick()
```
severinbeauvais, line 35: This is equivalent to the old code except I don't need to import Vue.

tests/unit/FileUploadPreview.spec.ts
```diff
@@ -35,7 +35,7 @@ function myArrayBuffer () {
   })
 }
 
-describe('FileUploadPreview component', () => {
+xdescribe('FileUploadPreview component', () => {
```
severinbeauvais, line 38: There's a webpack error in this test due to the updated pdfjs-dist package. I'll keep trying to figure this out.

tests/unit/ListShareClass.spec.ts
```diff
@@ -137,15 +137,15 @@ describe('List Shares and Series component', () => {
     const seriesListItem1 = wrapper.vm.$el.querySelectorAll('.v-data-table .series-row')[1]
 
     expect(seriesListItem1.querySelectorAll('td')[0].textContent).toContain('Share Series 2')
-    expect(seriesListItem1.querySelectorAll('td')[1].textContent).toContain(100)
+    expect(seriesListItem1.querySelectorAll('td')[1].textContent).toContain('100')
```
severinbeauvais, line 140: Yay, typechecking!

src/mixins/document-mixin.ts
severinbeauvais: FWIW, I tried async import and that didn't work either.

```
  private pdfjsLib: any

  async created (): Promise<void> {
    this.pdfjsLib = await import('pdfjs-dist')
    this.pdfjsLib.GlobalWorkerOptions.workerSrc = await import('pdfjs-dist/build/pdf.worker')
  }
  ...
  const document = await this.pdfjsLib.getDocument(...).promise
```

package.json
```diff
@@ -8,11 +8,12 @@
     "serve": "vue-cli-service serve",
```
severinbeauvais, line 8: PS The package-lock file has a lot of changes because of the package updates (and because I deleted and re-created it, so there's no left-over package dependencies in it, yay!).

src/App.vue
```diff
@@ -234,15 +232,9 @@ import { SessionStorageKeys } from 'sbc-common-components/src/util/constants'
...
-  mixins: [
-    CommonMixin,
-    DateMixin,
-    FilingTemplateMixin,
-    NameRequestMixin
-  ]
+  }
 })
-export default class App extends Vue {
+export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMixin, NameRequestMixin) {
```
JazzarKarim, line 237: Does this change for using Mixins have to happen for the other repos as well Sev?

severinbeauvais: Yes, but I was going to make them in a future ticket.

History: in originally thought that for better Vue3 compatibility, we should import mixins using the options syntax, but:
- that broke typings (and displayed build errors, even though it did work)
- there is a similar class decorator for Vue3 after all

JazzarKarim: Oh okay, makes sense. Thanks for the explanation!

I was thinking to make this change in Filings UI since I'm already in the process of changing all the files for Pinia. But I'm just going to focus on Pinia now and we'll fix that in a later ticket like you said.

src/App.vue
```diff
@@ -302,19 +293,19 @@ export default class App extends Vue {
   @Action setWindowWidth!: ActionBindingIF
 
   // Local properties
-  protected accountAuthorizationDialog = false
```
JazzarKarim, line 305: We won't be using access modifiers moving on Sev?

src/components/common/FileUploadPreview.vue
```diff
@@ -44,25 +39,28 @@ export default class FileUploadPreview extends Vue {
...
-      if (this.maxSize) {
-        const maxSizeMB = this.maxSize / 1024
-        const errorMsg = 'Exceeds maximum ' + maxSizeMB.toString() + ' MB file size'
-        return (file?.size <= (this.maxSize * 1024)) || errorMsg
-      }
-      return true
-    }
-  ]
+    ]
+  }
```
severinbeauvais, line 63: The old fileUploadRules array was static, so it didn't pick up changes due to different variable values.

The new fileUploadRules getter is reactive and now works properly.

src/components/common/FileUploadPreview.vue
```diff
@@ -73,18 +71,18 @@ export default class FileUploadPreview extends Vue {
     }
   }
 
-  protected fileChange (file) {
-    this.emitFileSelected(file)
+  async fileChange (file): Promise<void> {
+    await this.emitFileSelected(file)
```
severinbeauvais, line 75: This is a fix but there was no symptom of the bug.

Ditto line 131 below.

src/mixins/document-mixin.ts
```diff
@@ -1,11 +1,8 @@
-import Vue from 'vue'
-import { Component } from 'vue-property-decorator'
+import { Component, Vue } from 'vue-property-decorator'
 import { AxiosResponse } from 'axios'
 import { AxiosInstance as axios } from '@/utils'
 import { DocumentUpload, PdfInfoIF } from '@/interfaces'
 import { PdfPageSize } from '@/enums'
-import pdfjsLib from 'pdfjs-dist/build/pdf'
-pdfjsLib.GlobalWorkerOptions.workerSrc = require('pdfjs-dist/build/pdf.worker.entry')
```
severinbeauvais, line 8: With a `require` up here, Jest allocates the memory and does not free it between test suites.

So, I moved the requires into the class `created()` lifecycle hook. As far as I can tell, there's no need to free the memory explicitly when the class is unmounted/destroyed as that's done automatically.

tests/unit/DissolutionReviewConfirm.spec.ts
```diff
@@ -53,6 +53,8 @@ for (const test of reviewConfirmTestCases) {
 
       // verify page content
       expect(wrapper.find('h2').text()).toBe('Review and Confirm')
+
+      wrapper.destroy()
```
severinbeauvais, line 57: Doing this for every test in this file saved about 100 MB of heap memory when running the entire test suites.

To verify heap usage I changed the script to this (in package.json):
```
"test:unit": "vue-cli-service test:unit --testPathPattern --runInBand --logHeapUsage",
```

tests/unit/FileUploadPreview.spec.ts
```diff
@@ -36,6 +29,10 @@ function myArrayBuffer () {
 }
 
 describe('FileUploadPreview component', () => {
+  // mock some large PDF files
+  const oneMBFile = new File([new ArrayBuffer(1048576)], 'oneMBFile.pdf', { type: 'application/pdf' })
+  const elevenMBFile = new File([new ArrayBuffer(1048576 * 11)], 'elevenMBFile.pdf', { type: 'application/pdf' })
+
```
severinbeauvais, line 35: I moved these inside the describe block so Jest frees them between test suites.

tests/unit/FileUploadPreview.spec.ts
severinbeauvais: Fixed now.

src/App.vue
severinbeauvais: That depends on Typescript and the VS Code Vue extensions that we use (like volar and eslint).

At the moment, TS ignores access modifiers, so privates aren't really privates. Also, currently, volar and eslint generate errors when a Vue file's template uses a method/property that isn't public -- imho a protected should be fine since the template is sort-of part of the class. So, right now, methods/properties used in the template need to be public to avoid apparent errors.

Maybe in the future we can tighten up our access modifiers. We don't use classes as in C++ but good OO practises are generally a good idea.

severinbeauvais:
_Finally_ all the tests pass. Only took 1 week of frustration and outside help (thanks T!). Please review.

Ultimately we should make these changes to Edit UI and Filings UI. We can plan that next week.

src/App.vue
severinbeauvais: I got into this to help Ketaki focus on Pinia only (and that only worked so-so).

The catch is that you'll have to fix a bunch of files (as in this PR) to pre-empt mixin errors and the like.

On the other hand, do you really want me to make these changes to Filings UI and require you to rebase? I don't have a sense of whether auto-merge will work well or not 🤷‍♂️

JazzarKarim: Yes, I figured as much. No Sev, I'll try to fix the files and we'll go from there. I think rebasing would be a nightmare If I try to do it after the same changes end up being done in Filings UI.

src/App.vue
JazzarKarim: Noted! I'll keep that in mind. Thanks Sev 👍

src/components/common/ListPeopleAndRoles.vue
```diff
@@ -205,42 +201,42 @@ export default class ListPeopleAndRoles extends Vue {
   }
 
   /** Returns true if org-person is a person. */
-  protected isPerson (orgPerson: OrgPersonIF): boolean {
+  isPerson (orgPerson: OrgPersonIF): boolean {
     return (orgPerson.officer?.partyType === PartyTypes.PERSON)
   }
 
   /** Returns true if org-person is an organization (corporation/firm). */
-  protected isOrg (orgPerson: OrgPersonIF): boolean {
```
seeker25, line 213: Way better

seeker25 (approved):
These are great changes.. no more worrying about if it's private public or protected within a vue class component

JazzarKarim (approved):
Looks good to me! 😃 Thanks for taking care of this Sev! This helped me and Ketaki focus on the Pinia upgrade.

tests/unit/FileUploadPreview.spec.ts
```diff
@@ -211,20 +200,19 @@ describe('FileUploadPreview component', () => {
...
     })
+
     const fileInput = wrapper.find('input[type="file"]')
     setupFileInput(fileInput)
     inputValue = invalidSecondPagePdf.name
     inputFilesGet.mockReturnValue([invalidSecondPagePdf])
-    fileInput.trigger('change')
-
+    await fileInput.trigger('change')
     await waitForUpdate(5)
```
severinbeauvais, line 209: I tried with a count of 3 but it worked intermittently in the GH CI check, so 5 it is.

src/components/common/FileUploadPreview.vue
severinbeauvais: @ketaki-deodhar This is a code fix and it resolved the unit test errors for this component.

tests/unit/BusinessContactInfo.spec.ts
```diff
@@ -43,7 +43,7 @@ function createComponent (
   email = '',
   confirmEmail = '',
   phone = '',
-  extension: number = null,
+  extension = NaN,
```
severinbeauvais, line 46: This is a good initial value for a number since NaN is falsy.

src/components/Registration/BusinessTypeConfirm.vue
```diff
@@ -15,15 +15,17 @@
         </v-col>
 
         <v-col cols="12" sm="9">
-          <p class="mb-0">{{ label }}</p>
+          <p class="mb-0">{{ isTypePartnership ? labelGP : labelSP }}</p>
```
severinbeauvais, line 18: IMHO this is a better solution than overwriting some text strings as before.

src/components/Registration/BusinessTypeConfirm.vue
```diff
@@ -50,16 +51,16 @@ export default class BusinessTypeConfirm extends Vue {
...
+
+  readonly labelGP = 'General Partnership'
+  readonly labelSP = 'BC Sole Proprietorship / Doing Business As (DBA) Registration'
+  readonly textGP = 'I acknowledge that a General Partnership cannot be ' +
     'changed into a Sole Proprietorship (including DBA). If this is ' +
     'necessary, a new Name Request Number and Statement of Registration ' +
     '(along with associated fees) will be required.'
-
-  protected textSP = 'I acknowledge that a Sole Proprietorship (including DBA) ' +
+  readonly textSP = 'I acknowledge that a Sole Proprietorship (including DBA) ' +
```
severinbeauvais, line 63: Now that we don't overwrite some strings, they can all be readonly (static).

src/components/common/Stepper.vue
```diff
@@ -1,53 +1,51 @@
 <template>
   <v-card id="step-buttons-container">
-    <template v-for="(step, index) in getSteps">
-      <div
-        class="step"
-        :key="index"
```
severinbeauvais, line 6: eslint complained about the key not being in the template, but if I changed that then eslint complained about a key in a template.

So, I combined the template and the div tags, and eslint is happy.

severinbeauvais: PS If this diff looks weird in GH, click the ⚙️ icon at the top of this page and enable the Hide Whitespace option.

ketaki-deodhar (approved):
This is going to be so helpful in future. Looks good!

src/mixins/filing-template-mixin.ts
```diff
@@ -113,6 +112,14 @@ export default class FilingTemplateMixin extends DateMixin {
    * @returns the filing body to save
    */
   buildIncorporationFiling (): IncorporationFilingIF {
+    function fixNullAddressType (orgPeople: OrgPersonIF[]): OrgPersonIF[] {
+      return orgPeople.map(p => {
+        if (p.deliveryAddress?.addressType === null) delete p.deliveryAddress.addressType
+        if (p.mailingAddress?.addressType === null) delete p.mailingAddress.addressType
```
severinbeauvais, line 118: The conditional chaining is required for the unit tests :( but I not fixing that now.

severinbeauvais:
@pwei1018 Could you please create a temporary URL to review this PR? Thanks!

severinbeauvais:
> ~@pwei1018 Could you please create a temporary URL to review this PR? Thanks!~
> 
> This is no longer important -- I will deploy "pinia-upgrade" branch to Dev and we can look at it there.

Darn, looks like I cannot deploy a branch to GCP. @pwei1018 Can you help please?

PR #545, opened 2023-05-09
16175 Fixed undefined legal name in a dissolution (and also restoration)

severinbeauvais:
*Issue #:* bcgov/entity#16175

*Description of changes:*
- app version = 5.2.1
- deleted setEntityName and used setLegalName instead
- deleted getBusiness and used getBusinessOfficeAddress instead
- fixed undefined legal name -- added fallback for parsing draft without legal name (or legal type or founding date)

### DIFF
--- a/src/components/Dissolution/AssociationDetails.vue
+++ b/src/components/Dissolution/AssociationDetails.vue
@@ -27,8 +27,8 @@
         <v-col cols="12" sm="4" class="pr-4 pt-4 pt-sm-0">
           <label class="mailing-address-header">Mailing Address</label>
           <MailingAddress
-            v-if="!isEmptyAddress(getBusiness.officeAddress.mailingAddress)"
-            :address="getBusiness.officeAddress.mailingAddress"
+            v-if="!isEmptyAddress(getBusinessOfficeAddress.mailingAddress)"
+            :address="getBusinessOfficeAddress.mailingAddress"
             :editing="false"
           />
           <div v-else>(Not entered)</div>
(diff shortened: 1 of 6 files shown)


src/components/Dissolution/AssociationDetails.vue
```diff
@@ -27,8 +27,8 @@
         <v-col cols="12" sm="4" class="pr-4 pt-4 pt-sm-0">
           <label class="mailing-address-header">Mailing Address</label>
           <MailingAddress
-            v-if="!isEmptyAddress(getBusiness.officeAddress.mailingAddress)"
```
severinbeauvais, line 30: `getBusiness` was only ever used to get tge office address, so I replaced the general getter with a more specific one.

src/mixins/filing-template-mixin.ts
```diff
@@ -571,9 +572,9 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
     this.setFilingId(+draftFiling.header.filingId)
 
     // restore Business data
-    this.setEntityType(draftFiling.business.legalType)
-    this.setLegalName(draftFiling.business.legalName)
-    this.setFoundingDate(draftFiling.business.foundingDate)
+    this.setEntityType(draftFiling.business.legalType || this.getBusinessLegalType)
+    this.setLegalName(draftFiling.business.legalName || this.getBusinessLegalName)
+    this.setFoundingDate(draftFiling.business.foundingDate || this.getBusinessFoundingDate)
```
severinbeauvais, line 577: Because the draft restoration (or dissolution, below) didn't have legal name, it was being cleared out here. (It was actually loaded via a business fetch in App.vue, so this code here is probably not needed, but that cleanup is out of scope for this quick fix.)

JazzarKarim (approved):
LGTM!

PR #548, opened 2023-05-12
16375 fix "None is not a String" filing error

severinbeauvais:
*Issue #:* bcgov/entity#16375

*Description of changes:*
- app version = 5.2.4
- don't save legal name in restoration business object

### DIFF
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -413,7 +413,6 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
       business: {
         legalType: this.getEntityType,
         identifier: this.getBusinessId,
-        legalName: this.getBusinessLegalName,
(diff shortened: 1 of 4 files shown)


argush3 (approved):

vysakh-menon-aot (approved):

src/mixins/filing-template-mixin.ts
```diff
@@ -413,7 +413,6 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
       business: {
         legalType: this.getEntityType,
         identifier: this.getBusinessId,
-        legalName: this.getBusinessLegalName,
```
severinbeauvais, line 416: Some of the other filings don't save legal name here in the business object, and if a NR is used then legal name is saved below on line 457, so I _hope_ this is OK in the back end.

I'm testing all 4 possibilities (full vs resto, numbered vs named) and so far it looks OK.

PR #551, opened 2023-05-27
16548 Update SP/GP start date validation

severinbeauvais:
*Issue #:* bcgov/entity#16548

*Description of changes:*
- app version = 5.2.7
- changed min date to 10 years ago
- removed min start date for staff
- removed min validation for staff
- added some date rules
- updated start date message for staff
- updated start date message for reg users (10 vs 2 years in the past)
- updated a unit test

### DIFF
--- a/src/rules/date-rules.ts
+++ b/src/rules/date-rules.ts
@@ -9,11 +9,24 @@ export const DateRules = {
   }
 }
 
-// validation helper method(s)
+// validation helper methods
 export const DateRuleHelpers = {
+  /** Whether date string to validate is between min date and max date. */
   isBetweenDates (minDate: Date, maxDate: Date, dateStrToValidate: string): boolean {
-    if (!dateStrToValidate) { return true }
+    if (!dateStrToValidate) return true
     const date = new Date(dateStrToValidate) // eg, September 5, 2022
     return (date >= minDate) && (date <= maxDate)
+  },
+  /** Whether date string to validate is not before min date. */
+  isNotBeforeDate (minDate: Date, dateStrToValidate: string): boolean {
+    if (!dateStrToValidate) return true
+    const date = new Date(dateStrToValidate) // eg, September 5, 2022
+    return (date >= minDate)
+  },
+  /** Whether date string to validate is not after max date. */
+  isNotAfterDate (maxDate: Date, dateStrToValidate: string): boolean {
+    if (!dateStrToValidate) return true
+    const date = new Date(dateStrToValidate) // eg, September 5, 2022
+    return (date <= maxDate)
   }
 }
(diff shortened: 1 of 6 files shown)


src/rules/date-rules.ts
```diff
@@ -9,11 +9,24 @@ export const DateRules = {
   }
 }
 
-// validation helper method(s)
+// validation helper methods
 export const DateRuleHelpers = {
+  /** Whether date string to validate is between min date and max date. */
   isBetweenDates (minDate: Date, maxDate: Date, dateStrToValidate: string): boolean {
-    if (!dateStrToValidate) { return true }
+    if (!dateStrToValidate) return true
```
chenhongjing, line 16: A question. Why should it return true in this case?

src/views/Registration/RegistrationDefineBusiness.vue
```diff
@@ -99,16 +99,21 @@
...
+          <template v-else>
+            Enter the start date of the business. The start date can be
+            <v-tooltip top max-width="20rem" content-class="top-tooltip" transition="fade-transition">
+              <template v-slot:activator="{ on }">
+                <span v-on="on" class="tool-tip dotted-underline">no more than 10 years in the past</span>
+              </template>
+              <span>Choose the oldest date possible even if the actual start date is older than 10 years in the
+                past.</span>
+            </v-tooltip>
+            and 90 days in the future.
```
chenhongjing, line 114: Does it have rules to restrict start date no more than 90 days in the future?

src/rules/date-rules.ts
severinbeauvais: If there is no date string then we cannot compare it below.

Meanwhile, there is other code (on src/components/Registration/StartDate.vue line 88) that validates the date is required.

If you look at the rules array in src/components/Registration/StartDate.vue, the way Vuetify validation works is that it runs each function in the array until a function returns a string instead of True.

Example: the first function checks if the value exists (`!!v`); if False then it returns the error string, otherwise if True then it runs the next function.

So line 16 above should never return True. It is mostly a code safety check.

chenhongjing: yeah, it makes sense 👍

src/views/Registration/RegistrationDefineBusiness.vue
severinbeauvais: Yes -- it's the second rule pushed to the rules array (used by Vuetify for validating the component's value) in src/components/Registration/StartDate.vue on line 91 for staff and line 97 for regular users.

src/rules/date-rules.ts
severinbeauvais: I guess I didn't _really_ answer your question... I think it could just as well return False and the code would work the same way, since THIS function isn't called (via rules array) if the value doesn't exist (first rule would have caught it).

But maybe this validation function is called by other code, where an empty value is valid.

I guess the point is that validating whether the value exists is not THIS function's purpose.

chenhongjing (approved):
No more questions. LGTM 😃

JazzarKarim (approved):
LGTM 👍

ChrisGabel23 (approved):

PR #552, opened 2023-05-30
16444 Use Keycloak roles for LaunchDarkly update

severinbeauvais:
*Issue #:* bcgov/entity#16444

*Description of changes:*
- app version = 5.2.7
- cleaned up a couple of dialogs
- renamed some state variables and getters/setters
- get user's roles from Keycloak token
- used proper roles in LaunchDarkly update
- misc cleanup
- updated unit tests

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -51,6 +51,7 @@
 
     <PaymentErrorDialog
       attach="#app"
+      filingName="Application"
       :dialog="paymentErrorDialog"
       :errors="saveErrors"
       :warnings="saveWarnings"
--- a/src/App.vue
+++ b/src/App.vue
@@ -51,7 +51,6 @@
 
     <PaymentErrorDialog
       attach="#app"
-      filingName="Application"
       :dialog="paymentErrorDialog"
       :errors="saveErrors"
       :warnings="saveWarnings"
@@ -61,7 +60,6 @@
 
     <SaveErrorDialog
       attach="#app"
-      :filingName="saveErrorDialogName"
       :dialog="saveErrorDialog"
       :errors="saveErrors"
       :warnings="saveWarnings"
@@ -246,6 +244,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
   @Getter(useStore) getFilingName!: FilingNames
   @Getter(useStore) getFilingType!: FilingTypes
   @Getter(useStore) getHaveChanges!: boolean
+  @Getter(useStore) getKeycloakRoles!: Array<string>
   @Getter(useStore) getOrgInformation!: OrgInformationIF
   @Getter(useStore) getSteps!: Array<StepIF>
   @Getter(useStore) getUserFirstName!: string
@@ -260,7 +259,6 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
 
   @Action(useStore) setAccountInformation!: ActionBindingIF
   @Action(useStore) setAdminFreeze!: ActionBindingIF
-  @Action(useStore) setAuthRoles!: ActionBindingIF
   @Action(useStore) setBusinessId!: ActionBindingIF
   @Action(useStore) setBusinessNumber!: ActionBindingIF
   @Action(useStore) setCompletingParty!: ActionBindingIF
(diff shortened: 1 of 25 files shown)


src/App.vue
```diff
@@ -51,7 +51,6 @@
 
     <PaymentErrorDialog
       attach="#app"
-      filingName="Application"
```
severinbeauvais, line 54: Moved into dialog since it's static text.

src/App.vue
```diff
@@ -260,7 +259,6 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
 
   @Action(useStore) setAccountInformation!: ActionBindingIF
   @Action(useStore) setAdminFreeze!: ActionBindingIF
-  @Action(useStore) setAuthRoles!: ActionBindingIF
```
severinbeauvais, line 263: We don't do anything with the auth roles, so no need to store them.

src/store/store.ts
```diff
@@ -126,24 +125,9 @@ export const useStore = defineStore('store', {
       }
     },
 
-    /** Whether the user has "staff" auth role. */
+    /** Whether the user has "staff" Keycloak role. */
     isRoleStaff (): boolean {
-      return this.getTombstone.authRoles.includes('staff')
```
severinbeauvais, line 131: I got rid of `getTombstone` as it was only used in here and, in my opinion, complicated the understanding of where the data came from -- compare with new code which is now consistent across getters >>>

src/store/store.ts
```diff
@@ -126,24 +125,9 @@ export const useStore = defineStore('store', {
...
-    },
-
-    /** Whether the user is authorized to view. */
-    isAuthView (): boolean {
-      return this.getTombstone.authRoles.includes('view')
-    },
-
-    /** Whether the user has "gov account user" auth role. */
-    isGovAccountUser (): boolean {
-      return this.getTombstone.authRoles.includes('gov_account_user')
```
severinbeauvais, line 146: The above were never used so I deleted them.

src/store/store.ts
```diff
@@ -158,17 +142,17 @@ export const useStore = defineStore('store', {
 
     /** The account folio number. */
     getFolioNumber (): string {
-      return this.getTombstone.folioNumber
```
severinbeauvais, line 161: I got rid of `getTombstone` as it was only used in this file and, in my opinion, complicated the understanding of where the data came from -- compare with new code which is now consistent across getters >>>

severinbeauvais:
The roles as recorded in LaunchDarkly (and "staff" is one of them):

image

src/dialogs/PaymentErrorDialog.vue
```diff
@@ -6,7 +6,7 @@
       <v-card-text id="dialog-text">
         <!-- display common message -->
         <div class="font-15" v-if="!isRoleStaff">
-          <p>We are unable to process your payment at this time. This {{filingName}} has been saved
+          <p>We are unable to process your payment at this time. This Application has been saved
```
JazzarKarim, line 9: Do you think it's a good idea to keep the prop here? I know it's only used once and it's in `App.vue` but maybe down the line we might use it? I'm asking since it's used a lot in Filings UI. Thoughts Sev?

severinbeauvais: I can't decide.

It was originally coded with the prop so the filing name could be specified, but we only ever used the same name.

I'll simply revert this.

src/App.vue
severinbeauvais: Reverted.

src/dialogs/PaymentErrorDialog.vue
JazzarKarim: Alright, sounds good 👍

src/App.vue
```diff
@@ -1053,19 +1080,19 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
...
   /** Updates Launch Darkly with current user info. */
-  private async updateLaunchDarkly (userInfo: any): Promise<void> {
+  private async updateLaunchDarkly (userInfo: any, keycloakRoles: string[]): Promise<void> {
     // since username is unique, use it as the user key
     const key: string = userInfo.username
     const email: string = userInfo.contacts[0]?.email || userInfo.email
     const firstName: string = userInfo.firstname
     const lastName: string = userInfo.lastname
-    // remove leading { and trailing } and tokenize string
-    const custom: any = { roles: userInfo.roles?.slice(1, -1).split(',') }
```
JazzarKarim, line 1063: Out of curiosity, does this syntax work? So:
```javascript
const custom: any = { roles: keycloakRoles }
```

severinbeauvais: Yes, they're equivalent. Pick which style you prefer.

I _slightly_ prefer the `x = something as type` syntax because the type isn't always needed and I like to see a lot of `x =` (for consistency).

JazzarKarim: Alright, got it 👍 Thanks!

JazzarKarim (approved):
Looks good to me 👍

ketaki-deodhar (approved):
Looks good!

severinbeauvais:
Thanks for your reviews Karim, Ketaki and Hongjing!

PR #557, opened 2023-07-18
16994 - Updated all Store Actions Types

ChrisGabel23:
*Issue #:* /bcgov/entity#16994

* Updated all instances of "ActionBindingIF" (which uses "any" type) to the specific type used in the Store
* Updated the Store to use specific types, unless I was unsure which type it should be using


src/interfaces/store-interfaces/action-interface.ts
```diff
@@ -1,10 +1,7 @@
+// I don't think this file is used any more - CRG Jul23/23
```
ChrisGabel23, line 1: I felt nervous deleting a file from a code base I'm unfamiliar with, but AFAIK there is nothing in here that is used in Create-UI.

src/App.vue
```diff
@@ -294,39 +295,38 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
...
-  @Action(useStore) setUserFirstName!: ActionBindingIF
-  @Action(useStore) setUserLastName!: ActionBindingIF
-  @Action(useStore) setUserPhone!: ActionBindingIF
-  @Action(useStore) setHaveChanges!: ActionBindingIF
-  @Action(useStore) setOrgInformation!: ActionBindingIF
-  @Action(useStore) setShowErrors!: ActionBindingIF
-  @Action(useStore) setTempId!: ActionBindingIF
-  @Action(useStore) setWarnings!: ActionBindingIF
-  @Action(useStore) setWindowWidth!: ActionBindingIF
+  @Action(useStore) setAccountInformation!: (x: AccountInformationIF) => void
```
severinbeauvais, line 298: I would have used `val` but I think I like `x` better 👍

src/mixins/filing-template-mixin.ts
```diff
@@ -61,53 +64,53 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
...
-  @Action(useStore) setRestorationExpiry!: ActionBindingIF
-  @Action(useStore) setRestorationNoticeDate!: ActionBindingIF
-  @Action(useStore) setRestorationRelationships!: ActionBindingIF
-  @Action(useStore) setRestorationType!: ActionBindingIF
-  @Action(useStore) setRules!: ActionBindingIF
-  @Action(useStore) setShareClasses!: ActionBindingIF
-  @Action(useStore) setStaffPayment!: ActionBindingIF
-  @Action(useStore) setTransactionalFolioNumber!: ActionBindingIF
+  @Action(useStore) setAffidavit!: (x: UploadAffidavitIF) => void
+  @Action(useStore) setBusinessAddress!: (x: any) => void
```
severinbeauvais, line 68: The corresponding "get" call uses a type. Can  you update the set call similarly please?
https://github.com/bcgov/business-create-ui/blob/ff820e9bd892b10fed295dd29655e7c0fb274305/src/store/store.ts#L283

severinbeauvais (commented):
Looks great!

I see this change identified a type error (in filing-template-mixin.ts). I wonder if the code should be updated (adding missing property) or the interface should be updated (optional property)?

src/mixins/filing-template-mixin.ts
ChrisGabel23: Does that mean BusinessOfficeAddress is the same as BusinessAddress? That's the only reason I left it as is. Everywhere else, I tracked down the type elsewhere in the code and updated store.ts. I thought they were named differently because they had the possibility of being different. Like for small businesses they might be the same, but some big businesses, I could envision the business operating from one address and the head office being somewhere else.

src/mixins/filing-template-mixin.ts
```diff
@@ -61,53 +64,53 @@ export default class FilingTemplateMixin extends Mixins(DateMixin) {
...
+  @Action(useStore) setRestorationApplicationDate!: (x: string) => void
+  @Action(useStore) setRestorationApprovalType!: (x: ApprovalTypes) => void
+  @Action(useStore) setRestorationCourtOrder!: (x: CourtOrderIF) => void
+  @Action(useStore) setRestorationExpiry!: (x: string) => void
+  @Action(useStore) setRestorationNoticeDate!: (x: string) => void
+  @Action(useStore) setRestorationRelationships!: (x: RelationshipTypes[]) => void
+  @Action(useStore) setRestorationType!: (x: RestorationTypes) => void
+  @Action(useStore) setRules!: (x: CreateRulesIF) => void
+  @Action(useStore) setShareClasses!: (x: ShareClassIF[]) => void
+  @Action(useStore) setStaffPayment!: (x: any) => void
```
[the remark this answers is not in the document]
ChrisGabel23: Like the one above, I was thinking there was a difference between StaffPayment and StaffPaymentStep. They are the same?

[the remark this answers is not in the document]
ChrisGabel23: I just saw there's a StaffPaymentIF, which is referenced in StaffPaymentStepIF.

ChrisGabel23:
> Looks great!
> 
> I see this change identified a type error (in filing-template-mixin.ts). I wonder if the code should be updated (adding missing property) or the interface should be updated (optional property)?

I discussed this with @JazzarKarim and we thought that adding the missing property (with a value of null or false) is the way to go.

src/mixins/filing-template-mixin.ts
severinbeauvais: I think so. Confirm the store state property that's being written/read.

ChrisGabel23: It's not clear from the state property if it's an OfficeAddressIF or BusinessAddressIF. They both contain a mailingAddress and a deliveryAddress, but with the OfficeAddressIF, deliveryAddress is optional. I don't understand why there are two because OfficeAddressIF is more versatile. Is this a remnant of old code?

severinbeauvais: In store.ts:
```
    setBusinessAddress (address) {
      this.stateModel.business.officeAddress = address
    },
```

The stateModel is declared in state-model.ts as type `StateModelIF`.

In state-interface.ts:
```
// State model interface
  ...
  business: BusinessIF
```

Finally, in business-interface.ts:
```
/** The Business object from the API. */
export interface BusinessIF {
  .....
  officeAddress: OfficeAddressIF
```

ChrisGabel23: Thanks @severinbeauvais. I didn't realize I could follow the breadcrumbs like that. I understand it better now!

severinbeauvais: You're welcome :)

Thanks for the updates. Looks good.

src/mixins/filing-template-mixin.ts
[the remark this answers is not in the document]
severinbeauvais: Fixed along with the other one.

JazzarKarim (approved):
LGTM, great stuff Chris! 👍

PR #569, opened 2023-09-27
17795 Allow registration (SP/GP) for an user/account that only uses last name

chenhongjing:
*Issue #:* /bcgov/entity#17795

*Description of changes:*
- Removed invalid first name error
- Changed first name rule of completing party for registration
- Changed `certifiedBy` value
- app version = 5.5.2

### DIFF
--- a/src/components/common/RegAddEditOrgPerson.vue
+++ b/src/components/common/RegAddEditOrgPerson.vue
@@ -79,7 +79,7 @@
                   class="item first-name"
                   label="First Name"
                   :rules="enableRules ?
-                   (isCompletingParty ? Rules.FirstNameRulesFirms.slice(1) :
+                   (isCompletingParty ? Rules.FirstNameRulesFirmsCP :
                    Rules.FirstNameRulesFirms) : []"
                   :readonly="isCompletingParty && !(isRoleStaff || isSbcStaff)"
                 />
--- a/src/components/common/RegAddEditOrgPerson.vue
+++ b/src/components/common/RegAddEditOrgPerson.vue
@@ -78,7 +78,9 @@
                   filled
                   class="item first-name"
                   label="First Name"
-                  :rules="enableRules ? Rules.FirstNameRulesFirms : []"
+                  :rules="enableRules ?
+                   (isCompletingParty ? Rules.FirstNameRulesFirms.slice(1) :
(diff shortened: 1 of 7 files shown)


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-569-w8e4eyhb.web.app

src/rules/first-name-rules.ts
```diff
@@ -6,6 +6,5 @@ export const FirstNameRules: Array<VuetifyRuleFunction> = [
 ]
 
 export const FirstNameRulesFirms: Array<VuetifyRuleFunction> = [
-  v => !!v?.trim() || 'First name is required',
```
severinbeauvais, line 9: Is this rule used only for Completing Party?

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-569-w8e4eyhb.web.app

src/components/common/RegAddEditOrgPerson.vue
```diff
@@ -78,7 +78,9 @@
                   filled
                   class="item first-name"
                   label="First Name"
-                  :rules="enableRules ? Rules.FirstNameRulesFirms : []"
+                  :rules="enableRules ?
+                   (isCompletingParty ? Rules.FirstNameRulesFirms.slice(1) :
+                   Rules.FirstNameRulesFirms) : []"
```
severinbeauvais, line 83: 1. This is too much logic for the template, but...
2. This is not a good way to implement this as it assumes that the first rule (which you are omitting) is for first name. However, someone could change the rules (in the other file) and would not know to update this file, causing a bug.

I recommend creating a different rule for Completing Party. You could add it to the same file as the other rule.

severinbeauvais: Thanks for the fix.

src/rules/first-name-rules.ts
severinbeauvais: No longer relevant.

severinbeauvais:
Please fix lint errors and unit test.

severinbeauvais (approved):

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-569-w8e4eyhb.web.app

JazzarKarim (approved):

PR #571, opened 2023-09-28
15863 removed the shift of radio buttons in approval and restoration …

PaulGarewal:
…type

*Issue #:* /bcgov/entity#15863

*Description of changes:*

- The radio buttons in Approval Type and Restoration Type shift when the field is invalid (or left blank). The changes removes that shift by adjusting the margins of the invalid section. 

image

image

### DIFF
--- a/src/components/Restoration/ApprovalType.vue
+++ b/src/components/Restoration/ApprovalType.vue
@@ -3,6 +3,7 @@
     <div
       class="section-container"
       :class="{ 'invalid-section': invalidSection }"
+      :style="invalidSection ? {margin: '0 9px 0 0'} : {}"
     >
       <ApprovalTypeShared
         :courtOrderNumber="getRestoration.courtOrder.fileNumber"
--- a/src/components/Restoration/ApprovalType.vue
+++ b/src/components/Restoration/ApprovalType.vue
@@ -2,8 +2,6 @@
   <div id="approval-type">
     <div
       class="section-container"
-      :class="{ 'invalid-section': invalidSection }"
-      :style="invalidSection ? {margin: '0 9px 0 0'} : {}"
     >
       <ApprovalTypeShared
         :courtOrderNumber="getRestoration.courtOrder.fileNumber"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-create-ui",
-  "version": "5.5.1",
+  "version": "5.5.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "business-create-ui",
-      "version": "5.5.1",
+      "version": "5.5.2",
       "dependencies": {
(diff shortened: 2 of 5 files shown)


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-571-zlaa1beo.web.app

src/components/Restoration/ApprovalType.vue
```diff
@@ -3,6 +3,7 @@
     <div
       class="section-container"
       :class="{ 'invalid-section': invalidSection }"
+      :style="invalidSection ? {margin: '0 9px 0 0'} : {}"
```
JazzarKarim, line 6: Paul, we usually don't add in-line styling. Would it be possible to add a conditional class for this maybe?

PaulGarewal: Of course, I will change that right away.

severinbeauvais: Unless you need exactly 9px of right margin, you can use a [Vuetify spacing helper class](https://v2.vuetifyjs.com/en/styles/spacing/) here. In this case, `mr-2`.

severinbeauvais: By the way,  you can have multiple dynamic classes. Something like this:
```
:class="{
  'myClass1': myCondition1,
  'myClass2': myCondition2
}"
```

severinbeauvais (commented):
How does adding right margin fix the issue of the whole block being moved to the right by the width of the error border?

src/components/Restoration/ApprovalType.vue
PaulGarewal: So much easier, learning the power of Vue

PaulGarewal:
> How does adding right margin fix the issue of the whole block being moved to the right by the width of the error border?

It does act like a squeeze, as that fixes the radio buttons in place, which was the issue. However, I can change the logic here by just making it red and not shifting at all (which is what you proposed). I will just have to find out where that is in the code.

severinbeauvais:
> > How does adding right margin fix the issue of the whole block being moved to the right by the width of the error border?
> 
> It does act like a squeeze, as that fixes the radio buttons in place, which was the issue. However, I can change the logic here by just making it red and not shifting at all (which is what you proposed). I will just have to find out where that is in the code.

We try to consistently use top whitespace (not bottom) and left whitespace (not right), which makes it easier to lay things out than if we mixed top/bottom and left/right. I think your solution could cause extra work in the future if anything needs to be re-aligned, so might as well fix it consistently now.

Yes, I propose either a permanent left border, which is only red when needed, or show the red border but then pull that component left by the same amount (with -ve margin).

severinbeauvais (approved):
LGTM

severinbeauvais:
Please fix the lint errors. You can ignore the lint warnings unless you caused them ;)

JazzarKarim (approved):
LGTM Paul! 👍

PR #573, opened 2023-10-03
15863 fixed the red invalid state sizing issue and some reshaping

PaulGarewal:
*Issue #:* /bcgov/entity#15863

*Description of changes:*

- Fixed the red overflow of the invalid visible state. 

image

image

### DIFF
--- a/src/components/Restoration/RestorationType.vue
+++ b/src/components/Restoration/RestorationType.vue
@@ -20,7 +20,7 @@
         >
           <v-radio-group
             v-model="selectRestorationType"
-            class="mt-0 mb-n5 pt-0"
+            class="mt-0 pt-0"
             hide-details="auto"
             @change="changeRestorationType()"
           >
--- a/src/components/Restoration/RestorationType.vue
+++ b/src/components/Restoration/RestorationType.vue
@@ -21,6 +21,7 @@
           <v-radio-group
             v-model="selectRestorationType"
             class="mt-0 mb-n5 pt-0"
+            hide-details="auto"
             @change="changeRestorationType()"
           >
             <v-radio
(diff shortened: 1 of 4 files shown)


src/views/Restoration/RestorationBusinessName.vue
```diff
@@ -43,12 +43,11 @@
       >
         <RestorationType
           id="restoration-type"
-          class="mb-n5"
           :class="{ 'approval-restoration-invalid-section': invalidSectionRestoration }"
         />
 
         <!-- Divider b/w Restoration and Approval type -->
-        <v-divider class="mb-11 mr-5 ml-5"></v-divider>
+        <v-divider class="mb-10 mr-5 ml-5"></v-divider>
```
severinbeauvais, line 50: I didn't notice before that this is how it was done, but it's not the right way to do this -- the v-divider isn't the problem, so changing it is not the solution. (This is forming a biding/relationship between the block above the divider and the divider that shouldn't exist.)

The proper solution should have the block above the divider be the correct height.

src/components/Restoration/RestorationType.vue
```diff
@@ -20,7 +20,7 @@
         >
           <v-radio-group
             v-model="selectRestorationType"
-            class="mt-0 pt-0"
+            class="mt-0 mb-n5 pt-0"
             @change="changeRestorationType()"
           >
```
severinbeauvais, line 25: In general, we shouldn't "shorten" components like this unless there is no other way.

In this case, there may be a couple of ways to shorten this radio group:
- make it dense
- hide details if there aren't any (auto mode)

Ref: https://v2.vuetifyjs.com/en/api/v-radio-group/#props

severinbeauvais (approved):

JazzarKarim (approved):
LGTM! 👍

src/views/Restoration/RestorationBusinessName.vue
severinbeauvais: Fixed.

PR #585, opened 2023-12-01
18535 - Add Amalgamating Businesses Part 1

JazzarKarim:
*Issue #:* /bcgov/entity#18535

*Description of changes:*
- Added the Amalgamating Businesses Part 1

https://github.com/bcgov/business-create-ui/assets/122301442/92f4337b-9536-44de-8e94-6b850b811062

### DIFF
--- a/src/components/Amalgamation/AmalgamatingBusinesses.vue
+++ b/src/components/Amalgamation/AmalgamatingBusinesses.vue
@@ -1,0 +1,246 @@
+<template>
+  <div id="amalgamating-businesses">
+    <v-btn
+      id="btn-add-amalgamating-business"
+      outlined
+      color="primary"
+      class="btn-outlined-primary"
+      :disabled="addAmalgatingBusinessDisabled"
+      @click="onAddBusinessClick()"
+    >
+      <v-icon>mdi-domain-plus</v-icon>
+      <span>Add an Amalgamating Business</span>
+    </v-btn>
+
+    <v-btn
+      v-if="isRoleStaff"
+      id="btn-add-amalgamating-business"
+      outlined
+      color="primary"
+      class="ml-2 btn-outlined-primary"
+      :disabled="addAmalgatingForeignBusinessDisabled"
+      @click="onAddForeignBusinessClick()"
+    >
+      <v-icon>mdi-domain-plus</v-icon>
+      <span>Add an Amalgamating Foreign Business</span>
+    </v-btn>
+
+    <!-- Add an Amalgamating Business button clicked -->
+    <v-expand-transition>
+      <v-card
+        v-if="isAddingAmalgamatingBusiness"
+        flat
+        class="section-container mt-4"
+      >
+        <v-row
+          no-gutters
+        >
+          <v-col
+            cols="12"
+            sm="3"
+          >
+            <label>Add a Business Registered in BC</label>
+          </v-col>
+
+          <v-col
+            cols="12"
+            sm="8"
+            class="ml-8"
+          >
+            <span>Enter the name or the incorporation number of the registered BC business
+              to add to this application.
+            </span>
+            <BusinessLookup
+              :showErrors="false"
+              :businessLookup="initialBusinessLookupObject"
+              :BusinessLookupServices="BusinessLookupServices"
+              label="Business Name or Incorporation Number"
+              @setBusiness="setAmalgamatingBusiness($event)"
+            />
+            <v-row
+              class="justify-end mr-0 mt-2"
+            >
+              <v-btn
+                id="app-cancel-btn"
+                large
+                outlined
+                color="primary"
+                @click="addAmalgamatingBusinessCancel()"
+              >
+                <span>Cancel</span>
+              </v-btn>
+            </v-row>
+          </v-col>
+        </v-row>
+      </v-card>
+    </v-expand-transition>
+
+    <!-- Add an Amalgamating Foreign Business button clicked -->
+    <v-expand-transition>
+      <v-card
+        v-if="isAddingAmalgamatingForeignBusiness"
+        flat
+        class="section-container mt-4"
+      >
+        <v-row
+          no-gutters
+        >
+          <v-col
+            cols="12"
+            sm="3"
+          >
+            <label>Add a Foreign Business</label>
+          </v-col>
+
+          <v-col
+            cols="12"
+            sm="8"
+            class="ml-8"
+          >
+            <span>**TODO**</span>
+          </v-col>
+
+          <!-- extra column is for possible action button -->
+        </v-row>
+      </v-card>
+    </v-expand-transition>
+    <v-row class="mt-4 ml-1">
+      <ul>
+        Amalgamating Businesses: <br><br>
+        <li
+          v-for="(business, index) in amalgamatingBuinesses"
+          :key="index"
+        >
+          <template v-if="business.foundingDate">
+            Legal Name: {{ business.legalName }} <br>
+            Legal Type: {{ business.legalType }} <br>
+            Mailing Address: {{ business.officeAddress.mailingAddress }} <br>
+            State: {{ business.state }} <br>
+            Good Standing: {{ business.goodStanding }} <br>
+          </template>
+          <template v-else>
+            Legal Name: {{ business.name }} <br>
+            Legal Type: {{ business.legalType }} <br>
+            Identifier: {{ business.identifier }} <br>
+            Status: {{ business.status }}
+          </template>
+        </li>
+      </ul>
+    </v-row>
+  </div>
+</template>
+
+<script lang="ts">
+import { Component, Mixins } from 'vue-property-decorator'
+import { Action, Getter } from 'pinia-class'
+import { useStore } from '@/store/store'
+import { CommonMixin } from '@/mixins'
+import { BusinessLookupServices, LegalServices } from '@/services'
+import { BusinessLookup } from '@bcrs-shared-components/business-lookup'
+import { BusinessIF, BusinessLookupIF, EmptyBusinessLookup } from '@/interfaces'
+
+@Component({
+  components: {
+    BusinessLookup
+  }
+})
+export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
+  @Getter(useStore) getAmalgamatingBusinesses!: Array<BusinessIF>
+  @Getter(useStore) isRoleStaff!: boolean
+
+  @Action(useStore) setAmalgamatingBusinesses!: (x: Array<BusinessIF>) => void
+
+  // Local properties
+  amalgamatingBusinessesValid = false
+  amalgamatingBuinesses = []
+  initialBusinessLookupObject = EmptyBusinessLookup
+
+  // Add an Amalgamating Business button properties
+  isAddingAmalgamatingBusiness = false
+  addAmalgatingBusinessDisabled = false
+
+  // Add an Amalgamating Foreign Business button properties
+  isAddingAmalgamatingForeignBusiness = false
+  addAmalgatingForeignBusinessDisabled = false
+
+  readonly BusinessLookupServices = BusinessLookupServices
+
+  // Cancel button in "Add an Amalgamating Business" is pressed.
+  addAmalgamatingBusinessCancel (): void {
+    this.isAddingAmalgamatingBusiness = false
+    // Enable buttons
+    this.addAmalgatingBusinessDisabled = false
+    this.addAmalgatingForeignBusinessDisabled = false
+  }
+
+  // "Add an Amalgamating Business" button is pressed.
+  onAddBusinessClick (): void {
+    this.isAddingAmalgamatingBusiness = true
+    this.isAddingAmalgamatingForeignBusiness = false
+    // Disable buttons
+    this.addAmalgatingBusinessDisabled = true
+    this.addAmalgatingForeignBusinessDisabled = true
+  }
+
+  // "Add an Amalgamating Foreign Business" button is pressed.
+  onAddForeignBusinessClick (): void {
+    this.isAddingAmalgamatingBusiness = false
+    this.isAddingAmalgamatingForeignBusiness = true
+    // Disable buttons (Please comment out the lines below or remove in part 2)
+    // this.addAmalgatingBusinessDisabled = true
+    // this.addAmalgatingForeignBusinessDisabled = true
+  }
+
+  async setAmalgamatingBusiness (businessLookup: BusinessLookupIF): Promise<void> {
+    // Get the amalgamating business information
+    // Will have a different format depending on the business
+    let business = await LegalServices.fetchBusinessInfo(businessLookup.identifier)
+      .then((response) => {
+        return response?.data?.business
+      }).catch(() => {
+        return businessLookup
+      })
+
+    // Get the address of the amalgamating business
+    if (businessLookup.identifier && business.foundingDate) {
+      const addresses = await LegalServices.fetchAddresses(businessLookup.identifier)
+        .then((data) => {
+          // SP and GP have businessOffice instead of registeredOffice
+          return data?.registeredOffice || data?.businessOffice
+        }).catch(() => {
+          return undefined
+        })
+      if (addresses) {
+        business.officeAddress = addresses
+      }
+    }
+
+    // If the amalgamating businesses array is not empty, check if identifier already exists.
+    // If identifier already exists, don't add the business to the array.
+    if (this.amalgamatingBuinesses.length > 0) {
+      const filteredBusinesses = this.amalgamatingBuinesses.filter(function (id) {
+        return id.identifier === business.identifier
+      })
+      if (filteredBusinesses.length === 0) {
+        this.amalgamatingBuinesses.push(business)
+      }
+    } else {
+      this.amalgamatingBuinesses.push(business)
+    }
+
+    // Set the amalgamated businesses array in the store.
+    this.setAmalgamatingBusinesses(this.amalgamatingBuinesses)
+    // Close the "Add an Amalgamating Business" Panel.
+    this.addAmalgamatingBusinessCancel()
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+.v-btn:not(.v-btn--round).v-size--default {
+  height: 44px;
+}
+
+</style>
--- a/src/components/Amalgamation/AmalgamatingBusinesses.vue
+++ b/src/components/Amalgamation/AmalgamatingBusinesses.vue
@@ -146,6 +146,7 @@ import { BusinessIF, BusinessLookupIF, EmptyBusinessLookup } from '@/interfaces'
 })
 export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
   @Getter(useStore) getAmalgamatingBusinesses!: Array<BusinessIF>
+  @Getter(useStore) isAmalgamationFilingHorizontal!: boolean
   @Getter(useStore) isRoleStaff!: boolean
 
   @Action(useStore) setAmalgamatingBusinesses!: (x: Array<BusinessIF>) => void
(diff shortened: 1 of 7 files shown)


src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,246 @@
+<template>
+  <div id="amalgamating-businesses">
+    <v-btn
+      id="btn-add-amalgamating-business"
+      outlined
+      color="primary"
+      class="btn-outlined-primary"
```
JazzarKarim, line 7: This is added to have the special color when button is disabled.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,246 @@
...
+            </span>
+            <BusinessLookup
+              :showErrors="false"
+              :businessLookup="initialBusinessLookupObject"
+              :BusinessLookupServices="BusinessLookupServices"
+              label="Business Name or Incorporation Number"
+              @setBusiness="setAmalgamatingBusiness($event)"
+            />
+            <v-row
+              class="justify-end mr-0 mt-2"
```
JazzarKarim, line 61: Moving the button to right as per the UXPin.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,246 @@
...
+            <span>**TODO**</span>
+          </v-col>
+
+          <!-- extra column is for possible action button -->
+        </v-row>
+      </v-card>
+    </v-expand-transition>
+    <v-row class="mt-4 ml-1">
+      <ul>
+        Amalgamating Businesses: <br><br>
```
JazzarKarim, line 109: This is all debug code here (can be removed at any point). This is just to show what happens whenever a business is added.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,246 @@
...
+      const addresses = await LegalServices.fetchAddresses(businessLookup.identifier)
+        .then((data) => {
+          // SP and GP have businessOffice instead of registeredOffice
+          return data?.registeredOffice || data?.businessOffice
+        }).catch(() => {
+          return undefined
+        })
+      if (addresses) {
+        business.officeAddress = addresses
+      }
```
JazzarKarim, line 215: Some businesses don't have an address (this is also showcased in the UXPin). If we were able to fetch the address, put it in the business object.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,246 @@
...
+    if (this.amalgamatingBuinesses.length > 0) {
+      const filteredBusinesses = this.amalgamatingBuinesses.filter(function (id) {
+        return id.identifier === business.identifier
+      })
+      if (filteredBusinesses.length === 0) {
+        this.amalgamatingBuinesses.push(business)
+      }
+    } else {
+      this.amalgamatingBuinesses.push(business)
+    }
```
JazzarKarim, line 229: Prevent from adding a business that's already been added to the array.

src/interfaces/store-interfaces/state-interfaces/amalgamation-state-interface.ts
```diff
@@ -1,15 +1,7 @@
-import { AmalgamationTypes, ApprovalTypes, RelationshipTypes } from '@/enums'
-import { CourtOrderIF } from '@/interfaces'
+import { AmalgamationTypes } from '@/enums'
+import { BusinessIF } from '@/interfaces'
 
 export interface AmalgamationStateIF {
-  applicationDate?: string // YYYY-MM-DD
```
JazzarKarim, line 5: Removed all the useless fields here for now. We'll eventually going to have to add lots of stuff in here depending on the ticket.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-585-8dj9ehdt.web.app

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,255 @@
...
+    // Disable buttons (Please comment out the lines below or remove in part 2)
+    // this.addAmalgatingBusinessDisabled = true
+    // this.addAmalgatingForeignBusinessDisabled = true
+  }
+
+  // Add to the amalgamating businesses array.
+  // If EP (A type), cannot be part of short form horizontal amalgamation.
+  pushToAmalgamatingBusinesses (business: any): void {
+    if (this.isAmalgamationFilingHorizontal) {
+      if (business.legalType !== 'A') this.amalgamatingBuinesses.push(business)
```
JazzarKarim, line 199: This is from the note added by Mihai today: "EP (A type) and other Foreign businesses cannot be part of a Short Form Horizontal amalgamation."

This is to fix that.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-585-8dj9ehdt.web.app

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,255 @@
...
+import { BusinessLookup } from '@bcrs-shared-components/business-lookup'
+import { BusinessIF, BusinessLookupIF, EmptyBusinessLookup } from '@/interfaces'
+
+@Component({
+  components: {
+    BusinessLookup
+  }
+})
+export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
+  @Getter(useStore) getAmalgamatingBusinesses!: Array<BusinessIF>
```
JazzarKarim, line 148: Not this getter for now but eventually we'll have to use this getter when continuing a draft.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,246 @@
...
+          >
+            <span>Enter the name or the incorporation number of the registered BC business
+              to add to this application.
+            </span>
+            <BusinessLookup
+              :showErrors="false"
+              :businessLookup="initialBusinessLookupObject"
+              :BusinessLookupServices="BusinessLookupServices"
+              label="Business Name or Incorporation Number"
+              @setBusiness="setAmalgamatingBusiness($event)"
```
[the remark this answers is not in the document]
severinbeauvais: Yes, it's too close to the same name.

How about "storeAmalgamatingBusiness"? (I know, store.) Save? Process? 🤷‍♂️

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,246 @@
...
+
+  async setAmalgamatingBusiness (businessLookup: BusinessLookupIF): Promise<void> {
+    // Get the amalgamating business information
+    // Will have a different format depending on the business
+    let business = await LegalServices.fetchBusinessInfo(businessLookup.identifier)
+      .then((response) => {
+        return response?.data?.business
+      }).catch(() => {
+        return businessLookup
+      })
```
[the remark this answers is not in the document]
severinbeauvais: Can you provide some tests to demonstrate all flows, please?

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: I think all businesses have an address. What are you referring to?

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: Use `find` instead of `filter`. (Find returns the first one. Filter does more work and checks all entries.)

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: I think we want to add them to the table anyway, where they will fail validation (unless user is staff).

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: When is this ever disabled?

severinbeauvais: NVM - when you're prompting for a business name, the add buttons are disbabled.

severinbeauvais: Did we add this class elsewhere? What does the button look like without it?

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: OK but v-card has too much right whitespace.

image

src/components/Amalgamation/AmalgamatingBusinesses.vue
JazzarKarim: Yes Sev! https://github.com/bcgov/business-create-ui/blob/5e99b3750fbf02855270417a9dd544431bf0c350/src/components/common/PeopleAndRoles.vue#L163

Left side is with it while left side is without it:
with and without

src/components/Amalgamation/AmalgamatingBusinesses.vue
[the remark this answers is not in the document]
severinbeauvais: Or, can you show the data from each scenario?

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,255 @@
...
+        })
+      if (addresses) {
+        business.officeAddress = addresses
+      }
+    }
+
+    // If the amalgamating businesses array is not empty, check if identifier already exists.
+    // If identifier already exists, don't add the business to the array.
+    if (this.amalgamatingBuinesses.length > 0) {
+      const filteredBusinesses = this.amalgamatingBuinesses.filter(function (id) {
```
severinbeauvais, line 232: Use `find` instead. (Find returns on first match, while filter searches the whole list.)

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,255 @@
...
+          >
+            <span>Enter the name or the incorporation number of the registered BC business
+              to add to this application.
+            </span>
+            <BusinessLookup
+              :showErrors="false"
+              :businessLookup="initialBusinessLookupObject"
+              :BusinessLookupServices="BusinessLookupServices"
+              label="Business Name or Incorporation Number"
+              @setBusiness="setAmalgamatingBusiness($event)"
```
severinbeauvais, line 58: We want this business lookup to find only active businesses (not historical). Does it?

We want this business lookup to find only specific business types. Not FM I think. Probably not CP, either. Ask Mihai or Yui.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: Great! I forgot about that. Good job :)

src/components/Amalgamation/AmalgamatingBusinesses.vue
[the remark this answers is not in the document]
JazzarKarim: Absolutely Sev.

So, for businesses that are in LEAR, here's the data that I'm getting from the lookup:
big json

Otherwise, this is what I'm getting:
small json

So, I wanted to handle both flows/scenarios although the ticket mentions to assume all searched businesses are in LEAR.

src/components/Amalgamation/AmalgamatingBusinesses.vue
JazzarKarim: Yes Sev, the business lookup only is searching for active businesses since that's the default value: https://github.com/bcgov/bcrs-shared-components/blob/2fc52fa9a8e4248a94f631909e01bb5ee0876d98/src/components/business-lookup/BusinessLookup.vue#L155

On it, will double check with Mihai or Yui for the second point.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,255 @@
...
+  addAmalgatingForeignBusinessDisabled = false
+
+  readonly BusinessLookupServices = BusinessLookupServices
+
+  // Cancel button in "Add an Amalgamating Business" is pressed.
+  addAmalgamatingBusinessCancel (): void {
+    this.isAddingAmalgamatingBusiness = false
+    // Enable buttons
+    this.addAmalgatingBusinessDisabled = false
+    this.addAmalgatingForeignBusinessDisabled = false
```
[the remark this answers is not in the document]
JazzarKarim: Unless I'm missing something, but my code already does that Sev!

For example, on line 182, the "Add an Amalgamating Business" button disables itself when it is pressed.

src/components/Amalgamation/AmalgamatingBusinesses.vue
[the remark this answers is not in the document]
severinbeauvais: The risk with fallbacks is that they can make it look like things are working normally but something is actually wrong. In this case I think I'd prefer some error handling if the fetched business is not in LEAR.

src/components/Amalgamation/AmalgamatingBusinesses.vue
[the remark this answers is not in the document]
severinbeauvais: "handleAmalgamatingBusiness" ? Kinda generic but not wrong.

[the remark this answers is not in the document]
JazzarKarim: I have it as `saveAmalgamatingBusiness` in my upcoming commit 🙏

src/views/AmalgamationRegular/Information.vue
```diff
@@ -5,15 +5,15 @@
...
         <p class="mt-4">
-          Your BC Registries account must have access to each business involved in this amalgamation
-          before the amalgamation application can be filed.
+          <strong>Important:</strong> The amalgamating businesses must be visible on your
+          My Business Registry list before the amalgamation filing can be completed.
         </p>
       </header>
 
-      [*** TODO: Amalgamating Businesses component ***]
+      <AmalgamatingBusinesses />
```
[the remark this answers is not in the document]
JazzarKarim: Fixing this in next commit.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,255 @@
...
+    // Will have a different format depending on the business
+    let business = await LegalServices.fetchBusinessInfo(businessLookup.identifier)
+      .then((response) => {
+        return response?.data?.business
+      }).catch(() => {
+        return businessLookup
+      })
+
+    // Get the address of the amalgamating business
+    if (businessLookup.identifier && business.foundingDate) {
```
severinbeauvais, line 206: Why do you have these checks? I think you don't need them.

JazzarKarim: The first one, you're right. For the second one, it is to handle whether the business was from LEAR. I'm unable to fetch the address if it's not a LEAR business. The foundingDate is a field only found in the big JSON above.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: > Sev, is this a bad thing (2 calls in this function)? Do you think that'll be bad performance?

We knew this coming in. In fact, I think you also need to make an Auth call to fetch the contact info email address.

However, instead of doing all 3 network calls sequentially, you can do them in parallel as we do [here](https://github.com/bcgov/business-filings-ui/blob/5d744948b7e52a69bdafb1ca2e4cd5a14bef5f40/src/App.vue#L483) or in Edit UI.

src/components/Amalgamation/AmalgamatingBusinesses.vue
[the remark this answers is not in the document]
severinbeauvais: I don't know if we need to use any of the data from the business lookup (except identifier), since we'll get it all from the "fetch business" LEAR call anyway. Maybe one way will make sense as you work on this.

src/components/Amalgamation/AmalgamatingBusinesses.vue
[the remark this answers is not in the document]
severinbeauvais: I think you're setting extra boolean variables that the buttons use.

I was thinking that the logic would be in the template itself, eg,
```
<v-btn
  :disabled="isAddingAmalgamatingBusiness || isAddingAmalgamatingForeignBusiness"
```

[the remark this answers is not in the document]
JazzarKarim: Love this! Will definitely do that, thanks Sev

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,250 @@
...
+export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
+  @Getter(useStore) getAmalgamatingBusinesses!: Array<BusinessIF>
+  @Getter(useStore) isAmalgamationFilingHorizontal!: boolean
+  @Getter(useStore) isRoleStaff!: boolean
+
+  @Action(useStore) setAmalgamatingBusinesses!: (x: Array<BusinessIF>) => void
+
+  // Local properties
+  amalgamatingBusinessesValid = false
+  amalgamatingBuinesses = []
```
severinbeauvais, line 156: typo

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -0,0 +1,250 @@
...
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+.v-btn:not(.v-btn--round).v-size--default {
+  height: 44px;
+}
+
+// Overriding the section container class right padding.
+.section-container {
+  padding-right: 0rem;
```
severinbeauvais, line 247: `pr-0` doesn't work on the v-card element?

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: I'll be in there soon 😁

src/components/Amalgamation/AmalgamatingBusinesses.vue
JazzarKarim: It does actually but it looked a bit better doing this (a bit less spacing). IDK, maybe I'm imagining things haha. Will go for pr-0.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: > For the second one, it is to handle whether the business was from LEAR. I'm unable to fetch the address if it's not a LEAR business. The foundingDate is a field only found in the big JSON above.

I think what I proposed is a better way, but if we kept the current code then you should add a code comment saying why  you're checking the founding date. (I looked to see if you needed it for anything and it wasn't referenced, so I was confused.)

src/components/Amalgamation/AmalgamatingBusinesses.vue
JazzarKarim: Yep, you're absolutely right. Will definitely be removing this!

severinbeauvais: (Yes, in some cases I wish the linter would allow us to have multiple things on the same line, but alas.)

For example, `<p id="boo" class="font-weight-bold:">BOO!</p>` is nice and simple, but lint wants it on 6 lines!

JazzarKarim: Lint is a double-edged sword isn't it 😆

severinbeauvais (approved):
I approve to merge as-is (so that others can build upon these changes).

Karim, please resolve the outstanding comments in your next PR. Thx.

JazzarKarim:
> I approve to merge as-is (so that others can build upon these changes).
> 
> Karim, please resolve the outstanding comments in your next PR. Thx.

Sounds good and already on it. Thanks Sev.

PR #587, opened 2023-12-04
18640 Amalgamating businesses structure + validations + etc

severinbeauvais:
*Issue #:* bcgov/entity#18640

*Description of changes:*
    - app version = 5.6.6
    - worked on amalgamating businesses data/types
    - renamed some enums
    - fixed table key
    - added more BusinessTable logic (including foreign jurisdiction)
    - updated static (test) data
    - updated some imports in unit tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "5.6.5",
+  "version": "5.6.6",
   "private": true,
   "appName": "Create UI",
   "sbcName": "SBC Common Components",
@@ -46,6 +46,7 @@
     "array-move": "^4.0.0",
     "axios": "^0.27.2",
     "core-js": "^3.30.1",
+    "country-list": "^2.2.0",
     "http-status-codes": "^2.2.0",
     "launchdarkly-js-client-sdk": "^2.24.2",
     "lodash": "^4.17.21",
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -46,6 +46,7 @@
     "array-move": "^4.0.0",
     "axios": "^0.27.2",
     "core-js": "^3.30.1",
+    "country-list": "^2.2.0",
```
severinbeauvais, line 49: For converting country code to name.

JazzarKarim (approved):
LGTM

This is super helpful, thanks! Will rebase and build on top of it.

src/components/Amalgamation/BusinessTable.vue
```diff
@@ -15,27 +15,29 @@
       <tbody>
         <tr
           v-for="(item, index) in businesses"
-          :key="item.businessId"
+          :key="item.identifier"
```
severinbeauvais, line 18: I initially had this as `:key="index"` but if you delete an entry in the table, most of the indexes stay the same (eg, you go from 0-5 to 0-4) so the table doesn't render correctly. Now it does because each item is indexed by its unique value.

severinbeauvais: I still have work to do in here. Not your concern :)

JazzarKarim: Faced a similar issue when I was running the debug code for the array 🙏

PR #588, opened 2023-12-05
18535 Amalgamating Businesses Part 1 Continuation

JazzarKarim:
*Issue #:* /bcgov/entity#18535

*Description of changes:*
- Continuation of Amalgamating Businesses part 1
- Updated business lookup to search only for: A, BEN, CC, BC, and ULC
- Refactored the saveAmalgamatingBusiness to make parallel calls + added email fetching
- Setting the AmalgamatingBusiness component validation (store getter and setter)

### DIFF
--- a/src/components/Amalgamation/AmalgamatingBusinesses.vue
+++ b/src/components/Amalgamation/AmalgamatingBusinesses.vue
@@ -212,19 +212,20 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
       business = data[0].data?.business
       business.businessContact = data[1].contacts[0]
       business.officeAddress = data[2]
-    } else {
-      business = businessLookup
     }
 
+    // If the business is not null (LEAR Entity)
     // If the amalgamating businesses array is not empty, check if identifier already exists.
     // If identifier already exists, don't add the business to the array.
-    if (this.amalgamatingBusinesses.length > 0) {
-      const businessExists = this.amalgamatingBusinesses.find(function (id) {
-        return id.identifier === business.identifier
-      })
-      if (!businessExists) this.amalgamatingBusinesses.push(business)
-    } else {
-      this.amalgamatingBusinesses.push(business)
+    if (business) {
+      if (this.amalgamatingBusinesses) {
+        const businessExists = this.amalgamatingBusinesses.find(
+          (id) => id.identifier === business.identifier
+        )
+        if (!businessExists) this.amalgamatingBusinesses.push(business)
+      } else {
+        this.amalgamatingBusinesses.push(business)
+      }
     }
 
     // Set the amalgamated businesses array in the store.
--- a/src/components/Amalgamation/AmalgamatingBusinesses.vue
+++ b/src/components/Amalgamation/AmalgamatingBusinesses.vue
@@ -215,26 +215,25 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
       business.officeAddress = data[2]
     }
 
-    const tingBusiness = {
-      type: 'lear',
-      role: AmlRoles.AMALGAMATING,
-      identifier: business.identifier,
-      name: business.legalName,
-      email: business.businessContact.email,
-      legalType: business.legalType,
-      address: business.officeAddress.registeredOffice.mailingAddress,
-      goodStanding: business.goodStanding
-    } as AmalgamatingBusinessIF
-
-    // If the business is not null (LEAR Entity)
+    // If the business is not null (LEAR Entity), create from it a TING business following the interface.
     // If the amalgamating businesses array is not empty, check if identifier already exists.
     // If identifier already exists, don't add the business to the array.
     if (business) {
+      const tingBusiness = {
+        type: 'lear',
+        role: AmlRoles.AMALGAMATING,
+        identifier: business.identifier,
+        name: business.legalName,
+        email: business.businessContact.email,
+        legalType: business.legalType,
+        address: business.officeAddress.registeredOffice.mailingAddress,
+        goodStanding: business.goodStanding
+      } as AmalgamatingBusinessIF
+
       if (this.amalgamatingBusinesses) {
-        const businessExists = this.amalgamatingBusinesses.find(
-          (id) => id.identifier === business.identifier
-        )
-        if (!businessExists) this.amalgamatingBusinesses.push(tingBusiness)
+        if (!this.amalgamatingBusinesses.find(b => b.identifier === business.identifier)) {
+          this.amalgamatingBusinesses.push(tingBusiness)
+        }
       } else {
         this.amalgamatingBusinesses.push(tingBusiness)
       }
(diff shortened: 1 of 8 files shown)


JazzarKarim:
Some tests are failing, looking into it...

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -52,6 +52,7 @@
               :showErrors="false"
               :businessLookup="initialBusinessLookupObject"
               :BusinessLookupServices="BusinessLookupServices"
+              legalTypes="BC,BEN,CC,ULC,A"
```
severinbeauvais, line 55: This is probably fine for now since we'll need to do additional work to support other legal types, I think.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -174,42 +178,42 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
...
+    ]).catch((error) => {
+      return error
+    })
+
+    if (data.length === 3) {
+      business = data[0].data?.business
+      business.businessContact = data[1].contacts[0]
+      business.officeAddress = data[2]
+    } else {
+      business = businessLookup
```
severinbeauvais, line 216: I don't think we want to do anything with the business if we can't look it up...?

src/components/Amalgamation/AmalgamatingBusinesses.vue
JazzarKarim: Yup, I asked Mihai last time and what to search exactly for now:
mihai answer

src/services/business-lookup-services.ts
```diff
@@ -32,10 +32,9 @@ export default class BusinessLookupServices {
    * @param query code or words to search
    * @returns a promise to return the search results
    */
-  static async search (query: string): Promise<BusinessLookupResultIF[]> {
-    const legalType = 'BC,A,ULC,C,S,XP,GP,LP,CUL,XS,LLC,LL,BEN,CP,CC,XL,FI,XCP,PA'
+  static async search (query: string, searchStatus: string, legalTypes: string): Promise<BusinessLookupResultIF[]> {
```
severinbeauvais, line 35: I see you're passing in 2 new formal parameters, but:
- you're using only 1 of them
- is the other code that calls this updated to pass both new params?

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-588-giwmc1ft.web.app

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: Yeah, but I meant there's little advantage to parameterizing this (or FF).

src/components/Amalgamation/AmalgamatingBusinesses.vue
JazzarKarim: True, but we still can get the name, identifier, and status. I was thinking that might help in filling in the table (if needed).

src/services/business-lookup-services.ts
JazzarKarim: You're right Sev, good catch.

Basically, this is what it was before in the shared components: https://github.com/bcgov/bcrs-shared-components/blob/4c62c41230422c5ceef744176fdc1bb04d6b5372/src/components/business-lookup/BusinessLookup.vue#L227

I noticed the searchStatus was missing all this time. It was there in the shared components repo but I noticed it's not being used in here. I was debating whether to remove it or not but I decided to keep it.

Let me know if you would rather me delete it from the business lookup is the shared components repo.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: Check with Yui or Mihai. I _think_ that if we don't find the business in Auth/Legal APIs then we don't add it to the table. But I'm thinking we might want to let the user know somehow...

src/services/business-lookup-services.ts
severinbeauvais: Yeah, it's a bit tricky. I bet that second formal parameter was added since the original code in here was written, and the original code possibly uses an older version of BusinessLookup that didn't have that second parameter. Anyway, it needs to be untangled... Good luck and lmk if I can help.

JazzarKarim: So Sev, here's what I found out. The `search` function in the business lookup services is different between UIs: https://github.com/bcgov/business-edit-ui/blob/21e095886fe05379e58842bb04cac6c47526c331/src/services/business-lookup-services.ts#L36

The search status is used in Edit UI but not in here. So, I can't remove it from the lookup in the shared repo (I'm glad I didn't do that in my last PR there).

So, I don't see any other way other than to specify the parameter and not use it to bypass the issue. What do you think?

src/components/Amalgamation/AmalgamatingBusinesses.vue
JazzarKarim: Removed the else block.

The business variable will still be null if COLIN entity.

src/services/business-lookup-services.ts
severinbeauvais: Sure. Just leave a code comment explaining it.

Please update the params in the JSDoc while you're there. Thx!

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -174,60 +178,62 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
...
     // If the amalgamating businesses array is not empty, check if identifier already exists.
     // If identifier already exists, don't add the business to the array.
-    if (this.amalgamatingBusinesses.length > 0) {
-      const businessExists = this.amalgamatingBusinesses.find(function (id) {
-        return id.identifier === business.identifier
-      })
-      if (!businessExists) this.amalgamatingBusinesses.push(business)
-    } else {
-      this.amalgamatingBusinesses.push(business)
+    if (business) {
```
JazzarKarim, line 220: Only adding if the entity is not a COLIN entity (as per my discussion with Yui).

JazzarKarim:
Changing this PR to a draft PR since I'm doing more work here.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -174,60 +179,73 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
...
+      if (this.amalgamatingBusinesses) {
+        const businessExists = this.amalgamatingBusinesses.find(
+          (id) => id.identifier === business.identifier
+        )
+        if (!businessExists) this.amalgamatingBusinesses.push(tingBusiness)
+      } else {
+        this.amalgamatingBusinesses.push(tingBusiness)
+      }
```
severinbeauvais, line 240: I think you can really simplify this (since amalgamatingBusinesses is initialized to []).

```
if (!this.amalgamatingBusinesses.find(b => b.identifier === business.identifier)) {
   this.amalgamatingBusinesses.push(tingBusiness)
}
```

severinbeauvais (approved):
Looks at outstanding comments then GTG.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: Good, but you don't need line 233 or lines 237-239 either.

src/services/business-lookup-services.ts
severinbeauvais: Perfect. Resolving.

src/components/Amalgamation/AmalgamatingBusinesses.vue
JazzarKarim: > Good, but you don't need line 233 or lines 237-239 either.

Done, you're absolutely right.

PR #590, opened 2023-12-06
18640 Incremental amalgamating businesses validations, etc

severinbeauvais:
*Issue #:* bcgov/entity#18640

*Description of changes:*
    - app version = 5.6.10
    - updated some validity calculations
    - moved ting validations to a mixin
    - generalized status computation
    - updated regular amalgation validity
    - updated to use temp copy of array (instead of local copy)

### DIFF
--- a/src/components/Amalgamation/AmalgamatingBusinesses.vue
+++ b/src/components/Amalgamation/AmalgamatingBusinesses.vue
@@ -106,7 +106,7 @@
       <ul>
         Amalgamating Businesses: <br><br>
         <li
-          v-for="(business, index) in amalgamatingBusinesses"
+          v-for="(business, index) in getAmalgamatingBusinesses"
           :key="index"
         >
           <template v-if="business.foundingDate">
@@ -163,7 +163,6 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
   @Action(useStore) setAmalgamatingBusinessesValid!: (x: boolean) => void
 
   // Local properties
-  amalgamatingBusinesses = []
   initialBusinessLookupObject = EmptyBusinessLookup
   businessTableValid = false
 
@@ -172,11 +171,6 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
   isAddingAmalgamatingForeignBusiness = false
 
   readonly BusinessLookupServices = BusinessLookupServices
-
-  // If continuing a draft, initialize the amalgamatingBusinesses array as the previously saved one.
-  mounted (): void {
-    if (this.getAmalgamatingBusinesses) this.amalgamatingBusinesses = this.getAmalgamatingBusinesses
-  }
 
   // Cancel button in "Add an Amalgamating Business" is pressed.
   addAmalgamatingBusinessCancel (): void {
@@ -221,6 +215,8 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
     // If the amalgamating businesses array is not empty, check if identifier already exists.
     // If identifier already exists, don't add the business to the array.
     if (business) {
+      const amalgamatingBusinesses = this.getAmalgamatingBusinesses
+
       const tingBusiness = {
         type: 'lear',
         role: AmlRoles.AMALGAMATING,
@@ -232,13 +228,13 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
         goodStanding: business.goodStanding
       } as AmalgamatingBusinessIF
 
-      if (!this.amalgamatingBusinesses.find(b => b.identifier === business.identifier)) {
-        this.amalgamatingBusinesses.push(tingBusiness)
+      if (!amalgamatingBusinesses.find((b: any) => b.identifier === business.identifier)) {
+        amalgamatingBusinesses.push(tingBusiness)
+
+        // Set the new amalgamated businesses array in the store.
+        this.setAmalgamatingBusinesses(amalgamatingBusinesses)
       }
     }
-
-    // Set the amalgamated businesses array in the store.
-    this.setAmalgamatingBusinesses(this.amalgamatingBusinesses)
 
     // Close the "Add an Amalgamating Business" Panel.
     this.isAddingAmalgamatingBusiness = false
--- a/src/components/Amalgamation/AmalgamatingBusinesses.vue
+++ b/src/components/Amalgamation/AmalgamatingBusinesses.vue
@@ -127,8 +127,6 @@
       </ul>
     </v-row> -->
 
-    <!-- FOR DEBUGGING -->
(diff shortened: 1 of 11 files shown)


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-590-r5gjk971.web.app

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-590-r5gjk971.web.app

severinbeauvais:
For the new reviewers: Karim and I have been working on this together and reviewing each other's PRs exclusively, but I thought you might want to have a look at this one.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -235,18 +228,30 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
         goodStanding: business.goodStanding
       } as AmalgamatingBusinessIF
 
-      if (!this.amalgamatingBusinesses.find(b => b.identifier === business.identifier)) {
-        this.amalgamatingBusinesses.push(tingBusiness)
+      if (!amalgamatingBusinesses.find((b: any) => b.identifier === business.identifier)) {
+        amalgamatingBusinesses.push(tingBusiness)
+
+        // Set the new amalgamated businesses array in the store.
+        this.setAmalgamatingBusinesses(amalgamatingBusinesses)
```
severinbeauvais, line 235: The other way we could have done this is with an action that pushes the new business into the array right in the store, but I think this is OK.

src/components/Amalgamation/BusinessTable.vue
```diff
@@ -69,91 +69,49 @@
...
-      (business.type === 'lear' && business.legalType === CorpTypeCd.BC_COMPANY)
-    )
-  }
-
   /**
    * This is the list of amalgamating businesses with computed statuses.
-   * In other words, these are the business rules.
+   * In other words, this is where the business rules are evaluated.
    */
   get businesses (): AmalgamatingBusinessIF[] {
```
severinbeauvais, line 100: Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -235,18 +228,30 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
         goodStanding: business.goodStanding
       } as AmalgamatingBusinessIF
 
-      if (!this.amalgamatingBusinesses.find(b => b.identifier === business.identifier)) {
-        this.amalgamatingBusinesses.push(tingBusiness)
+      if (!amalgamatingBusinesses.find((b: any) => b.identifier === business.identifier)) {
```
JazzarKarim, line 231: I wonder if we can the set the type here as AmalgamatingBusinessIF Sev?

src/mixins/amalgamation-mixin.ts
```diff
@@ -0,0 +1,114 @@
...
+      (business.type === 'lear' && business.legalType === CorpTypeCd.BC_COMPANY)
+    )
+  }
+
+  /** True if there an unlimited company in the table in Alberta, Nova Scotia or USA. */
+  get isAnyUnlimitedAbNsUsa (): boolean {
+    return this.getAmalgamatingBusinesses.some(business =>
+      (business.type === 'lear' && business.legalType === CorpTypeCd.BC_COMPANY)
+    )
+  }
```
severinbeauvais, line 45: WIP.

BTW: https://docs.google.com/document/d/1Dt2jzPA2SLtQO7Ns2aPCzVchGdv4YrYB/edit

src/mixins/amalgamation-mixin.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  // readonly amalgamationRules = [
+  //   {
+  //     id: 0,
+  //     rule: (v: any) => !!v || 'Required',
+  //     status: AmalgamatingStatuses.ERROR_FOREIGN
+  //   }
+  // ]
```
severinbeauvais, line 53: More WIP. Just assume everything in here is WIP, except for the basic architecture and design :)

src/mixins/amalgamation-mixin.ts
```diff
@@ -0,0 +1,114 @@
...
+ * Mixin that provides amalgamation rules, etc.
+ */
+@Component({})
+export default class AmalgamationMixin extends Vue {
+  @Getter(useStore) getAmalgamatingBusinesses!: AmalgamatingBusinessIF[]
+  @Getter(useStore) isRoleStaff!: boolean
+  @Getter(useStore) isTypeBcCcc!: boolean
+  @Getter(useStore) isTypeBcUlcCompany!: boolean
+
+  // *** TODO: finish this, maybe name them something recognizable...
```
severinbeauvais, line 18: WIP but some functionality is in and usable. I will pick away at this but I am no longer blocking anyone.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: Yes. On it.

severinbeauvais: I mean no. `AmalgamatingBusinessIF` is a union type and identifier is only available when it's used as a lear type.

I could export AmalgamatingLearIF and make this work, but I think it's cleaner to have this single example as any. Open to others' thoughts...

JazzarKarim: Yep, it makes sense. OK, no worries, I'm guessing we can keep it as any since part 2 ticket might use this too.

src/mixins/amalgamation-mixin.ts
JazzarKarim: I've requested access to the document.

JazzarKarim (approved):
Looks great Sev! 👍

src/components/Amalgamation/BusinessTable.vue
JazzarKarim: Love this.

jamespaologarcia (approved):
Looks good from what I can understand

src/components/Amalgamation/BusinessTable.vue
severinbeauvais: One architecture/design thing I wanted to mention:

This is a computed value (local getter), which is reactive to the variables used in it, in particular `getAmalgamatingBusinesses`, so it is evaluated (and cached) accordingly.

Also, as a computed, it needs to be used somewhere (otherwise it is not evaluated). In this component, it's used both to display the table items (line 17) and is watched to emit whether the table is valid (line 166).

Another way to do this would be to simply watch `getAmalgamatingBusinesses` and then compute the statuses when it has changed. If the getter approach didn't work (eg, reactivity didn't work) then I would have changed to this approach, otherwise I see no reason to change working code :)

PR #593, opened 2023-12-11
18640 More amalgamating fetch / validations work

severinbeauvais:
*Issue #:* bcgov/entity#18640

*Description of changes:*

Commit 1:
    - app version = 5.6.11
    - imported latest BusinessLookup shared component
    - updated fetchBusinessInfo return object
    - simplified Amalg Businesses button and panel logic
    - added Cancel button to Foreign panel
    - added snackbar for save business errors
    - deleted debugging template code
Commit 2:
    - added fetch for business filings (for more validations)
    - renamed AmalgamatingStatuses -> AmlStatuses
    - added AmlTypes enum and replaced strings
    - added some properties to TING business object (for more validations)
    - added LegalServices.fetchFilings()
    - misc updates
Commit 3:
    - changed all to allSettled so I can get partial fetches
    - fall back to lookup properties if some fetches fail (eg, not amalgamated)
    - added empty table message
    - temporarily start with empty (instead of pre-populated) table

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -249,7 +249,7 @@ import * as Views from '@/views'
 
 // Mixins, interfaces, etc
 import { CommonMixin, DateMixin, FilingTemplateMixin, NameRequestMixin } from '@/mixins'
-import { AccountInformationIF, AddressIF, BreadcrumbIF, BusinessIF, BusinessWarningIF, CompletingPartyIF,
+import { AccountInformationIF, AddressIF, BreadcrumbIF, BusinessWarningIF, CompletingPartyIF,
   ConfirmDialogType, EmptyFees, FeesIF, FilingDataIF, NameRequestIF, OrgInformationIF, PartyIF, ResourceIF,
   StepIF } from '@/interfaces'
 import { AmalgamationRegResources, DissolutionResources, IncorporationResources, RegistrationResources,
@@ -1158,9 +1158,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
 
   /** Fetches and stores business info. */
   private async loadBusinessInfo (businessId: string): Promise<void> {
-    const response = await LegalServices.fetchBusinessInfo(businessId)
-
-    const business = response?.data?.business as BusinessIF
+    const business = await LegalServices.fetchBusinessInfo(businessId).catch(() => {})
 
     if (!business) {
       throw new Error('Invalid business info')
--- a/src/components/Amalgamation/AmalgamatingBusinesses.vue
+++ b/src/components/Amalgamation/AmalgamatingBusinesses.vue
@@ -6,7 +6,7 @@
       color="primary"
       class="btn-outlined-primary"
       :disabled="isAddingAmalgamatingBusiness || isAddingAmalgamatingForeignBusiness"
-      @click="onAddBusinessClick()"
+      @click="isAddingAmalgamatingBusiness = true"
     >
       <v-icon>mdi-domain-plus</v-icon>
       <span>Add an Amalgamating Business</span>
@@ -19,7 +19,7 @@
       color="primary"
       class="ml-2 btn-outlined-primary"
       :disabled="isAddingAmalgamatingBusiness || isAddingAmalgamatingForeignBusiness"
-      @click="onAddForeignBusinessClick()"
+      @click="isAddingAmalgamatingForeignBusiness = true"
     >
       <v-icon>mdi-domain-plus</v-icon>
       <span>Add an Amalgamating Foreign Business</span>
@@ -46,8 +46,8 @@
             class="ml-8"
           >
             <span>Enter the name or the incorporation number of the registered BC business
-              to add to this application.
-            </span>
+              to add to this application.</span>
+
             <BusinessLookup
               :showErrors="false"
               :businessLookup="initialBusinessLookupObject"
@@ -56,6 +56,7 @@
               label="Business Name or Incorporation Number"
               @setBusiness="saveAmalgamatingBusiness($event)"
             />
+
             <v-row
               class="justify-end mr-0 mt-2"
             >
@@ -64,7 +65,7 @@
                 large
                 outlined
                 color="primary"
-                @click="addAmalgamatingBusinessCancel()"
+                @click="isAddingAmalgamatingBusiness = false"
               >
                 <span>Cancel</span>
               </v-btn>
@@ -95,6 +96,20 @@
             class="ml-8"
           >
             <span>**TODO**</span>
+
+            <v-row
+              class="justify-end mr-0 mt-2"
+            >
+              <v-btn
+                id="app-cancel-btn"
+                large
+                outlined
+                color="primary"
+                @click="isAddingAmalgamatingForeignBusiness = false"
+              >
+                <span>Cancel</span>
+              </v-btn>
+            </v-row>
           </v-col>
 
           <!-- extra column is for possible action button -->
@@ -102,30 +117,25 @@
       </v-card>
     </v-expand-transition>
 
-    <!-- <v-row class="mt-4 ml-1">
-      <ul>
-        Amalgamating Businesses: <br><br>
-        <li
-          v-for="(business, index) in getAmalgamatingBusinesses"
-          :key="index"
+    <!-- snackbar to temporarily show fetch errors -->
+    <v-snackbar
+      v-model="snackbar"
+      timeout="5000"
+    >
+      {{ snackbarText }}
+
+      <template #action="{ attrs }">
+        <v-btn
+          color="error"
+          class="font-weight-bold"
+          text
+          v-bind="attrs"
+          @click="snackbar = false"
         >
-          <template v-if="business.foundingDate">
-            Legal Name: {{ business.legalName }} <br>
-            Legal Type: {{ business.legalType }} <br>
-            Mailing Address: {{ business.officeAddress.registeredOffice.mailingAddress }} <br>
-            Email Address: {{ business.businessContact.email }} <br>
-            State: {{ business.state }} <br>
-            Good Standing: {{ business.goodStanding }} <br>
-        </template>
-          <template v-else>
-            Legal Name: {{ business.name }} <br>
-            Legal Type: {{ business.legalType }} <br>
-            Identifier: {{ business.identifier }} <br>
-            Status: {{ business.status }}
-          </template>
-        </li>
-      </ul>
-    </v-row> -->
+          Close
+        </v-btn>
+      </template>
+    </v-snackbar>
 
     <BusinessTable
       class="mt-8"
@@ -153,6 +163,8 @@ import BusinessTable from '@/components/Amalgamation/BusinessTable.vue'
   }
 })
 export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
+  readonly BusinessLookupServices = BusinessLookupServices
+
   @Getter(useStore) getAmalgamatingBusinesses!: AmalgamatingBusinessIF[]
   @Getter(useStore) getAmalgamatingBusinessesValid!: boolean
   @Getter(useStore) getShowErrors!: boolean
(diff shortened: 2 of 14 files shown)


src/App.vue
```diff
@@ -1158,9 +1158,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
 
   /** Fetches and stores business info. */
   private async loadBusinessInfo (businessId: string): Promise<void> {
-    const response = await LegalServices.fetchBusinessInfo(businessId)
-
-    const business = response?.data?.business as BusinessIF
+    const business = await LegalServices.fetchBusinessInfo(businessId).catch(() => {})
```
severinbeauvais, line 1161: I changed fetchBusinessInfo to actually return the business object, which simplifies this code.

Note that `business` will be undefined if the fetch fails.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -6,7 +6,7 @@
       color="primary"
       class="btn-outlined-primary"
       :disabled="isAddingAmalgamatingBusiness || isAddingAmalgamatingForeignBusiness"
-      @click="onAddBusinessClick()"
+      @click="isAddingAmalgamatingBusiness = true"
```
severinbeauvais, line 9: This seems simpler. I'm trying it out to see if I like it better. (So far, so good.)

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -46,8 +46,8 @@
             class="ml-8"
           >
             <span>Enter the name or the incorporation number of the registered BC business
-              to add to this application.
-            </span>
+              to add to this application.</span>
```
severinbeauvais, line 49: Spans are supposed to be single-line (no breaks).

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -165,80 +177,66 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
...
   // Button properties
   isAddingAmalgamatingBusiness = false
   isAddingAmalgamatingForeignBusiness = false
 
-  readonly BusinessLookupServices = BusinessLookupServices
-
-  // Cancel button in "Add an Amalgamating Business" is pressed.
-  addAmalgamatingBusinessCancel (): void {
-    this.isAddingAmalgamatingBusiness = false
-    this.setAmalgamatingBusinessesValid(true)
```
severinbeauvais, line 178: No need to set validity here -- the watcher on line 246 handles it automatically when the variables change.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -165,80 +177,66 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
...
     // Otherwise, return the businesslookup object.
     const data = await Promise.all([
       LegalServices.fetchBusinessInfo(businessLookup.identifier),
       AuthServices.fetchAuthInfo(businessLookup.identifier),
-      LegalServices.fetchAddresses(businessLookup.identifier)
-    ]).catch((error) => {
-      return error
-    })
+      LegalServices.fetchAddresses(businessLookup.identifier),
+      LegalServices.fetchFilings(businessLookup.identifier)
```
JazzarKarim, line 194: I see. This is going to affect my other PR. I'll make the changes then to do the same.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -165,80 +177,66 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
...
+    const businessInfo = data[0]
+    const authInfo = data[1]
+    const addresses = data[2]
+    const filings = data[3]
+
     // If identifier already exists, don't add the business to the array.
-    if (business) {
-      const amalgamatingBusinesses = this.getAmalgamatingBusinesses
+    if (this.getAmalgamatingBusinesses.find((b: any) => b.identifier === businessInfo.identifier)) {
+      this.snackbarText = 'Business is already in table.'
```
JazzarKarim, line 211: Nice!

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais, line 194: This is WIP -- there's an outstanding question whether we need to fetch the filings after all.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -165,80 +177,66 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
...
 
-    if (data.length === 3) {
-      business = data[0].data?.business
-      business.businessContact = data[1].contacts[0]
-      business.officeAddress = data[2]
+    // check for errors
+    if (!data || data.length !== 4) {
+      this.snackbarText = 'Unable to add this business.'
+      this.snackbar = true
+      return
```
severinbeauvais, line 201: I think I need to rework this as well, since it doesn't handle non-affiliated businesses.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: I'm hoping we can centralize this code and just call it from 2 places.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -165,80 +177,66 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
...
+      type: AmlTypes.LEAR,
+      role: AmlRoles.AMALGAMATING,
+      identifier: businessInfo.identifier,
+      name: businessInfo.legalName,
+      email: authInfo.contacts[0].email,
+      legalType: businessInfo.legalType,
+      address: addresses.registeredOffice.mailingAddress,
+      isNotInGoodStanding: (businessInfo.goodStanding === false),
+      isFutureEffective: (filings[0].isFutureEffective === true),
+      isLimitedRestoration: (filings[0].filingSubType === RestorationTypes.LIMITED) ||
```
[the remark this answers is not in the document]
severinbeauvais: Good question. And you're right, the LR (I made up this term btw) might not be the latest filing.

As it turns out, a business in LR is also not in good standing, so another validation catches this first. However, there's an outstanding business question whether to allow staff use a business in LR. 😵‍💫

src/components/Amalgamation/AmalgamatingBusinesses.vue
JazzarKarim: Yes, I'll be doing that in my PR.

src/mixins/amalgamation-mixin.ts
```diff
@@ -27,90 +27,100 @@ export default class AmalgamationMixin extends Vue {
   ]
 
   // *** I'M STILL WONDERING IF I WANT TO USE THESE
-  // readonly isLear = (item: AmalgamatingBusinessIF): boolean => (item?.type === 'lear')
-  // readonly isForeign = (item: AmalgamatingBusinessIF): boolean => (item?.type === 'foreign')
```
severinbeauvais, line 31: I can't use these after all. They look slick but if I use them in an IF statement before dereferencing a lear or foreign property then VS Code doesn't trust that I know what I'm doing and reports a type error.

image

src/components/Amalgamation/AmalgamatingBusinesses.vue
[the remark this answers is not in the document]
JazzarKarim: I thought at first LR was LEAR haha. Then I had a second look and it made more sense that it's Limited Restoration 😆

OK, sounds good. Let me know what decision the business ends up on.

src/mixins/amalgamation-mixin.ts
JazzarKarim: Experiencing the same thing. I casted the business as any when trying to do something similar.

severinbeauvais: Well I **am** using the current best practice for union types (ie, having a type variable). I guess VS Code trusts us in the same statement but not if we get more abstract.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: This works fine. Leaving it.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: The watcher works fine, but nothing is done with `getAmalgamatingBusinessesValid` yet -- future ticket.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: Resolving this for now.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: Fixed.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -165,80 +177,65 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
...
-    if (data.length === 3) {
-      business = data[0].data?.business
-      business.businessContact = data[1].contacts[0]
-      business.officeAddress = data[2]
+    // check if all elements are undefined
+    if (!data.some(d => d)) {
+      this.snackbarText = 'Unable to add that business.'
+      this.snackbar = true
+      return
     }
```
severinbeauvais, line 202: Karim, what it your experience when a business is in Lear but not affiliated to the current account? I'd like to allow what specifically here and display an error for the other cases.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -165,80 +177,65 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
...
+      identifier: businessInfo?.identifier || businessLookup.identifier,
+      name: businessInfo?.legalName || businessLookup.name,
+      email: authInfo?.contacts[0]?.email, // may be undefined
+      legalType: businessInfo?.legalType || businessLookup.legalType,
+      address: addresses?.registeredOffice?.mailingAddress, // may be undefined
+      isNotInGoodStanding: (businessInfo?.goodStanding === false),
+      isFutureEffective: (filings[0]?.isFutureEffective === true),
+      isLimitedRestoration: (filings[0]?.filingSubType === RestorationTypes.LIMITED) ||
+        (filings[0]?.filingSubType === RestorationTypes.LTD_EXTEND)
     }
```
severinbeauvais, line 230: The above works when we started from a business lookup, but I don't think it will work when resuming a draft.

I wrote the above to provide a fallback when "some" fetches failed (ie, to handle "not affiliated"). I think we need to review this and tighten it up -- only allow a specific "not affiliated" case through so we don't have to fall back for everything. This should allow a simple re-fetch when resuming a draft with no need to fall back.

^^ Maybe a future PR?

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -165,80 +177,65 @@ export default class AmalgamatingBusinesses extends Mixins(CommonMixin) {
...
 
-    // Close the "Add an Amalgamating Business" Panel.
+    // Add the new business to a new array and store the new array.
+    const amalgamatingBusinesses = this.getAmalgamatingBusinesses
+    amalgamatingBusinesses.push(tingBusiness)
+    this.setAmalgamatingBusinesses(amalgamatingBusinesses)
```
severinbeauvais, line 235: I think this is ugly. I'll create a new action for this instead.

src/mixins/amalgamation-mixin.ts
```diff
@@ -26,91 +26,91 @@ export default class AmalgamationMixin extends Vue {
...
     )
   }
 
-  // readonly amalgamationRules = [
-  //   {
-  //     id: 0,
-  //     rule: (v: any) => !!v || 'Required',
-  //     status: AmalgamatingStatuses.ERROR_FOREIGN
-  //   }
-  // ]
```
severinbeauvais, line 53: This was an old idea. I prefer the new idea of separate functions for each validation and an array of them (line 20).

src/store/state/state-model.ts
```diff
@@ -8,123 +8,123 @@ import {
...
-      addressCity: 'Vancouver',
-      addressRegion: 'BC',
-      postalCode: 'V6A 1A1',
-      addressCountry: 'CA'
-    },
-    role: AmlRoles.AMALGAMATING,
-    goodStanding: true
-  }
-]
+// const AMALGAMATING_BUSINESSES: AmalgamatingBusinessIF[] = [
```
severinbeauvais, line 16: I may still need this for testing validation functions.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-593-3q6felsd.web.app

src/components/Amalgamation/AmalgamatingBusinesses.vue
JazzarKarim: I think it should work even when resuming a draft. I tried it and it seems fine!

Makes sense, future PR sounds good.

src/components/Amalgamation/AmalgamatingBusinesses.vue
JazzarKarim: I'm not quite sure if I'm following here Sev. What do you mean?

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-593-3q6felsd.web.app

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: But when resuming a draft, there won't be a business lookup object to fall back to.

I'm thinking this will work out in the wash when we tighten up what fallbacks we allow (ie, only for not affiliated if possible).

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: We do 3 or 4 fetches on a business returned from the lookup. What do those fetches return if the business is not affiliated to the current account?

JazzarKarim: Just tested, those fetches return the exact same thing if or not the business is affiliated to the current account.

severinbeauvais: You tested as staff, right?

I tested as BCREG0020 and `0871425` seems to return a 403 for auth info.

severinbeauvais: PS Staff don't need to affiliate so I don't care about that.

JazzarKarim: Weird, I tested as a regular user and made sure to select a business that's not affiliated with that regular user. Sec, trying again..

severinbeauvais:
Can we merge this? I have to step out for a bit. I'll look at it later.

src/components/Amalgamation/AmalgamatingBusinesses.vue
JazzarKarim: Never mind, you're absolutely right. I don't know why at first try it was fine.

OK, so `fetchBusinessInfo` but the other two don't.

- `fetchAuthInfo`:
  403 (FORBIDDEN)

- `fetchAddresses`:
  401 (UNAUTHORIZED)

severinbeauvais: Theoretically, all 4 should return HTTP 401, but I'm OK with HTTP 403.

I create this bug for the 2 endpoints that allow an unauthorized account from fetching data:
https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/gh/bcgov/entity/18968

severinbeauvais: So I answered my own question here. Resolving.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: Done and verified.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: We may still want to refactor what I wrote so you can reuse it when resuming a draft, but I think we can work with this:
- if there's no auth info and user is staff then don't add the business to the table
- if there's no auth info and user is not staff then add the business
- if any of the Legal API calls fails then don't add the business to the table

I would output a console log if a business in the draft can't be added.

JazzarKarim (approved):

JazzarKarim:
> Can we merge this? I have to step out for a bit. I'll look at it later.

Absolutely.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: Yes, we need to fetch the latest filing to see if it's future-effective.

We also need to fetch the state filing, if it's declared, to see if business is in limited restoration.

In a way, it would be nice to not fetch one of these if the other is true. On the other hand, the table will be validated over and over and I wouldn't want to fetch db data every time.

severinbeauvais:
I'm going to merge this now. I think I'll just play with the validations for now (no more data handling). Go ahead and try to refactor for reuse. LMK how I can help.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -153,92 +164,115 @@ import BusinessTable from '@/components/Amalgamation/BusinessTable.vue'
...
+      if (!businessInfo.stateFiling) return false
+      // check for expired restoration
+      if (this.getCurrentDate > businessInfo.restorationExpiryDate) return false
+      // fetch state filing
+      const stateFiling = await LegalServices.fetchFiling(businessInfo.stateFiling)
+      return (
+        stateFiling.restoration.type === RestorationTypes.LIMITED ||
+        stateFiling.restoration.type === RestorationTypes.LTD_EXTEND
+      )
+    }
```
severinbeauvais, line 255: One advantage of having this nested function is that `businessInfo` is already in scope so I don't have to pass it in as a formal parameter.

PR #595, opened 2023-12-12
18722 Add ForeignBusiness component for amalgamating businesses

leodube-aot:
*Issue #:* /bcgov/entity#18722

*Description of changes:*
- Added ForeignBusiness components
- Included NestedSelect from Namerequest


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-595-nd422xtt.web.app

src/components/common/NestedSelect.vue
severinbeauvais, line 1: Could you move this bcrs-shared-components instead, please?

severinbeauvais (commented):
What other duplicated files can be moved to bcrs-shared-components?

As they are all related, could they be part of their own module? Ie, exported as one thing (instead of separate component, enums, interfaces, list data)?

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -82,38 +82,10 @@
...
-              </v-btn>
-            </v-row>
-          </v-col>
-
-          <!-- extra column is for possible action button -->
-        </v-row>
+        <ForeignBusiness
+          @add:foreignBusiness="saveAmalgamatingForeignBusiness($event)"
+          @cancel="isAddingAmalgamatingForeignBusiness = false"
+        />
```
severinbeauvais, line 88: You've moved some layout code (the label, rows/cols and cancel button) into your component. But now this page has that layout code for the other button, so the layout is in 2 places. Do you recommend moving the other button's code to its own sub-component, or moving your code back in here?

src/components/Amalgamation/ForeignBusiness.vue
```diff
@@ -0,0 +1,191 @@
...
+            @click="emitCancel()"
+          >
+            Cancel
+          </v-btn>
+        </v-col>
+      </v-row>
+    </v-form>
+  </section>
+</template>
+<script lang="ts">
```
severinbeauvais, line 75: Add blank line between template and script block (and also between script and style block below).

src/components/Amalgamation/ForeignBusiness.vue
```diff
@@ -0,0 +1,191 @@
...
+    const business = {
+      type: 'foreign',
+      role: AmlRoles.AMALGAMATING,
+      foreignJurisdiction: {
+        region: this.isCan ? this.jurisdiction.text : '',
+        country: this.isCan ? Location.CA : this.jurisdiction.value
+      },
+      legalName: this.legalName,
+      corpNumber: this.corpNumber
+    } as AmalgamatingBusinessIF
```
severinbeauvais, line 161: I'm wondering whether this component should know about this data type/block or whether that should be externalized (eg, a mixin, or emit the data to parent who handles it)?

My thought is to keep complexity as localized as possible.

src/components/Amalgamation/ForeignBusiness.vue
```diff
@@ -0,0 +1,191 @@
...
+    this.$refs.foreignBusinessForm.validate()
+  }
+
+  @Emit('add:foreignBusiness')
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  private emitForeignBusiness (foreignBusiness: AmalgamatingBusinessIF): void {}
+
+  @Emit('cancel')
+  emitCancel (): boolean {
+    return true
```
severinbeauvais, line 179: You probably don't have to emit an actual value. Just emitting the event should work. See how others are done.

leodube-aot:
> What other duplicated files can be moved to bcrs-shared-components?
> 
> As they are all related, could they be part of their own module? Ie, exported as one thing (instead of separate component, enums, interfaces, list data)?

Yeah all the added files in list-data, enums and interfaces are duplicated. I made one small change to the NestedSelect component compared to namerequest, but that can be modified to work with both project.

I think it would make sense to include it all as one module. Should I also update namerequest once I create the shared component?

src/components/Amalgamation/ForeignBusiness.vue
leodube-aot: For emitting the data to the parent I think it would look something like this:

```
@Emit('add:foreignBusiness')
private emitForeignBusiness (): any {
  return {
    jurisdiction: this.jurisdiction,
    legalName: this.legalName,
    corpNumber: this.corpNumber
  }
}
```

If you think that's fine I can update the code and move the handling to the parent.

severinbeauvais:
> I think it would make sense to include it all as one module. Should I also update namerequest once I create the shared component?

Create a new, tech debt ticket to go back and update Namerequest.

src/components/Amalgamation/ForeignBusiness.vue
severinbeauvais: Yes please!

src/components/Amalgamation/AmalgamatingBusinesses.vue
leodube-aot: I think it depends on how much we want to limit complexity in the parent component. I can move the code into AmalgamatingBusinesses.vue and get rid of ForeignBusiness.vue, and then move the jurisdiction functions into a mixin (like jurisdictionOptions and onJurisdictionChange) to not increase complexity too much.

severinbeauvais: For sure I'd like to limit complexity, but I'd prefer to keep the layout on the same page.

👍 to your edit.

leodube-aot:
Marking as draft until the [Jurisdiction shared component PR](https://github.com/bcgov/bcrs-shared-components/pull/219) is merged. I'll then double check it is working using the shared component.

src/components/common/NestedSelect.vue
severinbeauvais: Thanks for this! Resolving.

JazzarKarim (approved):
LGTM Leo!

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-595-nd422xtt.web.app

severinbeauvais (approved):
LGTM.

I'm not sure what's up with the unit tests. I saw intermittent issues myself a couple of days ago but not since. I restarted the testing-coverage check to see.

leodube-aot:
> I'm not sure what's up with the unit tests. I saw intermittent issues myself a couple of days ago but not since. I restarted the testing-coverage check to see.

Restarting the tests fixed it. Merging!

PR #598, opened 2023-12-15
18788 Updated amalgamation -> amalgamationApplication + business name options (WIP)

severinbeauvais:
*Issue #:* bcgov/entity#18788

*Description of changes:*
    - app version = 5.6.14
    - added ResultingBusinessName component
    - misc refactor and new code to handle 3 options
    - changed snackbar close text to a white icon
    - added random key to BusinessLookup to reset its state between calls
    - replaced local EntityState with shared EntityStates
    - added validation for isHistorical
    - updated imported enums, etc
    - **updated some enums locally (amalgamation -> amalgamationApplication)**
    - ResultingBusinessName is still WIP (so don't use it)


src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -49,8 +49,9 @@
               to add to this application.</span>
 
             <BusinessLookup
+              :key="Math.random()"
```
severinbeauvais, line 52: @JazzarKarim This resets the instance of this component each time it's used, which resets its state, which allows a used to search anew if there was a problem with the previous search.

severinbeauvais:
I think we should merge this PR asap so we can all have the latest change from amalgamation -> amalgamationApplication, even though Resulting Business Name is imcomplete. Thoughts?

package.json
```diff
@@ -23,19 +23,19 @@
     "@bcrs-shared-components/confirm-dialog": "1.2.1",
     "@bcrs-shared-components/contact-info": "1.2.15",
     "@bcrs-shared-components/corp-type-module": "1.0.11",
-    "@bcrs-shared-components/correct-name": "1.0.9",
+    "@bcrs-shared-components/correct-name": "1.0.33",
```
severinbeauvais, line 26: Since I'm using this shared component, I figured I'd better upgrade to the latest version first.

src/App.vue
```diff
@@ -258,7 +258,7 @@ import { AmalgamationRegResources, DissolutionResources, IncorporationResources,
 import { AuthServices, LegalServices, PayServices } from '@/services/'
 
 // Enums and Constants
-import { EntityState, ErrorTypes, FilingCodes, FilingNames, FilingStatus, FilingTypes, NameRequestStates, RouteNames,
+import { EntityStates, ErrorTypes, FilingCodes, FilingNames, FilingStatus, FilingTypes, NameRequestStates, RouteNames,
```
severinbeauvais, line 261: I deleted the local enum `EntityState` and am now using the shared one (with the same keys).

JazzarKarim:
> I think we should merge this PR asap so we can all have the latest change from amalgamation -> amalgamationApplication, even though Resulting Business Name is imcomplete. Thoughts?

Agreed.

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: Karim, this is still not perfect, as it seems to refresh too often, but I don't want to spend any more time on it atm (future fix if needed).

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -49,8 +49,9 @@
...
               :showErrors="false"
-              :businessLookup="initialBusinessLookupObject"
+              :businessLookup="{ ...EmptyBusinessLookup }"
```
severinbeauvais, line 54: A simplification.

src/components/Amalgamation/AmalgamatingBusinesses.vue
JazzarKarim: Sounds good to me.

src/components/Amalgamation/ResultingBusinessName.vue
severinbeauvais, line 1: This component is WIP, although it partly works atm.

src/mixins/amalgamation-mixin.ts
```diff
@@ -44,6 +45,17 @@ export default class AmalgamationMixin extends Vue {
     return null
   }
 
+  /**
+   * Disallow historical business.
+   * (Could happen if it was added while active and is now historical.)
```
JazzarKarim, line 50: Great catch!

src/components/Amalgamation/ResultingBusinessName.vue
```diff
@@ -0,0 +1,193 @@
...
+    return (this.getNameRequestApprovedName || this.getBusinessLegalName)
+  }
+  /** This section's validity state (when prompted by app). */
+  get invalidSection (): boolean {
+    return (this.getShowErrors && !this.getCorrectNameOption)
+  }
+
+  /** Called when component is created. */
+  created (): void {
+    // this.businessNameOption = this.getNameRequestNumber ? 'named' : 'numbered'
```
JazzarKarim, line 122: Eventually, we're going to have to comment this line out right Sev?

severinbeauvais: Sort of. There will be a new filing variable to save/restore which option was selected (null initially in some cases). I'm not there yet :)

JazzarKarim: Oops, sorry. All right Sev!

PR #601, opened 2023-12-19
18546 - add Folio Number to Review and Confirm step

ketaki-deodhar:
*Issue #:* /bcgov/entity#18546

*Description of changes:*

- Add Folio Number Component
- Filing Save and Resume
- Show invalid section when folio Number is not Valid

Folio number is optional:

image

Valid value:

image

Invalid value:
image


src/components/common/FolioNumber.vue
```diff
@@ -70,11 +70,16 @@ export default class FolioNumber extends Vue {
   @Prop({ default: null }) readonly initialValue!: string
   @Prop({ default: false }) readonly isEditing!: boolean
   @Prop({ default: false }) readonly showErrors!: boolean
+  @Prop({ default: false }) readonly folioNumberLabel!: boolean
```
ketaki-deodhar, line 73: Added prop for passing label

tests/unit/FolioNumber.spec.ts
```diff
@@ -13,6 +13,7 @@ describe('Folio number component', () => {
       vuetify
     })
 
+    expect(wrapper.find('.folio-number-title').text()).toBe('Folio Number')
```
ketaki-deodhar, line 16: asserting label for default value

src/components/common/FolioNumber.vue
severinbeauvais: Another way to do this is for the prop to be the optional value of the new label. By default, the value would be the old value. What do you think?

severinbeauvais: Another thing you could do is ask Yui if it's OK to simply update the label for all filings that use this common component.

src/mixins/filing-template-mixin.ts
```diff
@@ -315,11 +315,8 @@ export default class FilingTemplateMixin extends Mixins(AmalgamationMixin, DateM
       this.parseStaffPayment(draftFiling)
     }
 
-    // if this is a premium account and Folio Number exists then restore it
-    if (this.isPremiumAccount) {
-      if (draftFiling.header.folioNumber) {
-        this.setFolioNumber(draftFiling.header.folioNumber)
-      }
+    if (draftFiling.header.folioNumber) {
+      this.setFolioNumber(draftFiling.header.folioNumber)
```
severinbeauvais, line 319: Check with Yui whether folio number applies only to premium accounts.

ketaki-deodhar: Okay. Will do

src/store/store.ts
```diff
@@ -575,11 +580,14 @@ export const useStore = defineStore('store', {
...
       // *** TODO: add checks for review page components
+      const isFolioNumberValid = this.getFolioNumberValid
+
       return (
         this.getAmalgamatingBusinessesValid &&
         this.isDefineCompanyValid &&
         this.isAddPeopleAndRolesValid &&
-        this.isCreateShareStructureValid
+        this.isCreateShareStructureValid &&
+        isFolioNumberValid
```
severinbeauvais, line 590: Since you're not doing anything else with this local variable, I recommend just using the getter here.

src/views/AmalgamationRegular/ReviewConfirm.vue
```diff
@@ -131,16 +131,28 @@
...
       </header>
 
       <v-card
         flat
         class="mt-6"
       >
-        <div class="pa-4">
-          [*** TODO: Folio Number component ***]
+        <div
+          class="px-4 pt-8 pb-6"
```
severinbeauvais, line 146: Try to not use bottom padding, and instead use top padding on the element below this one.

Rationale: An element could add padding below itself to provide whitespace to the next element. Or, an element could add padding above itself to provide whitespace to the previous element. It doesn't matter which approach you take as long as you're consistent (within the same file), otherwise it gets really confusing/complicated.

JazzarKarim (approved):

src/components/common/FolioNumber.vue
ketaki-deodhar: Yui confirmed both. Applying changes now

src/mixins/filing-template-mixin.ts
ketaki-deodhar: Yui confirmed only Premium accounts. Applying changes now.

src/views/AmalgamationRegular/ReviewConfirm.vue
ketaki-deodhar: I applied bottom margin as space on top and bottom is different in the component. I have changed it to `class="px-4 py-8"` which solves the issue of uneven top and bottom space. .

with `class="pa-4"`

image

with `class="px-4 py-8"`

image

src/components/common/FolioNumber.vue
severinbeauvais: Looks good. Resolving.

src/store/store.ts
severinbeauvais: Looks good. Resolving.

src/views/AmalgamationRegular/ReviewConfirm.vue
severinbeauvais: Second one looks good 👍

src/store/store.ts
```diff
@@ -579,7 +584,8 @@ export const useStore = defineStore('store', {
         this.getAmalgamatingBusinessesValid &&
         this.isDefineCompanyValid &&
         this.isAddPeopleAndRolesValid &&
-        this.isCreateShareStructureValid
+        this.isCreateShareStructureValid &&
+        this.getFolioNumberValid
```
severinbeauvais, line 588: I think this will be false by default, so I think you'll need to make it True if premium account.

Example: https://github.com/bcgov/business-create-ui/blob/0902aa5a7e631ab0abc81567681a77364da6741f/src/store/store.ts#L538

src/mixins/filing-template-mixin.ts
severinbeauvais: Looks good. Resolving.

severinbeauvais (approved):
Have a look at remaining comment, otherwise looks good!

src/store/store.ts
ketaki-deodhar: Yes

severinbeauvais: Looks good. Resolving.

PR #606, opened 2024-01-02
18788 Added fallback

severinbeauvais:
*Issue #:* bcgov/entity#18788

*Description of changes:*
- app version = 5.6.21
- added fallback for drafts created without Correct Name Option

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-create-ui",
-  "version": "5.6.20",
+  "version": "5.6.21",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "business-create-ui",
-      "version": "5.6.20",
+      "version": "5.6.21",
       "dependencies": {
         "@babel/compat-data": "^7.21.5",
         "@bcrs-shared-components/approval-type": "1.0.19",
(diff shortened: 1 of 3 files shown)


JazzarKarim (approved):

JazzarKarim (approved):

severinbeauvais:
All - do you think I should create tickets to set the initial correctNameOption value in the 3+ places where an amalgamation application is created, or is it better to just keep this fallback here?

JazzarKarim:
> All - do you think I should create tickets to set the initial correctNameOption value in the 3+ places where an amalgamation application is created, or is it better to just keep this fallback here?

I'm OK with keeping the fallback here! 🙏 I'm also good with the other option too.

severinbeauvais:
> > All - do you think I should create tickets to set the initial correctNameOption value in the 3+ places where an amalgamation application is created, or is it better to just keep this fallback here?
> 
> I'm OK with keeping the fallback here! 🙏 I'm also good with the other option too.

Thanks. As for me, on the one hand I dislike "special cases" or "fallbacks" when we could/should be handling something earlier on in the process. On the other hand, I could argue that the business name option is not part of initial draft data (ie, it is part of the application other data). So, I'm not going to overthink this :)

ketaki-deodhar (approved):
Looks good!

PR #608, opened 2024-01-02
Final Framework Changes/Fixes for Regular Amalgamation

JazzarKarim:
*Issue #:* /bcgov/entity#18689

*Description of changes:*
- Created unit tests for all 5 amalgamation views
- Fixed future effective component label for amalgamations
- Changed jest-wrapper-factory to vitest-wrapper-factory
- Miscellaneous fixes and cleanup

### Note: I'm kind of blocked as I'm waiting for Shaoyun's PR to be merged before I can fully complete this here (Amalgamation statement component being added to Review and Confirm page).


src/views/AmalgamationRegular/ReviewConfirm.vue
```diff
@@ -101,6 +101,7 @@
         class="mt-6"
```
JazzarKarim, line 101: I don't like that this component's name is `IncorporationDateTime` as it's being used in amalgamations now too. I was thinking of renaming it. Thoughts?

src/components/Amalgamation/BusinessStatus.vue
```diff
@@ -80,8 +80,8 @@ export default class BusinessStatus extends Vue {
         return 'This business is under limited restoration. It cannot be part of an amalgamation ' +
           'unless it is fully restored.'
 
-      // case AmlStatuses.ERROR_NEED_BC_COMPANY:
-      //   return 'You must add at least one BC company.'
+      case AmlStatuses.ERROR_NEED_BC_COMPANY:
```
JazzarKarim, line 83: Spoke to Mihai about this. We indeed need as it is in the legacy system:
MicrosoftTeams-image (1)

I've tested it, it works:
need one bc company

Please note the changes too in the other files.

src/components/Incorporation/IncorporationDateTime.vue
```diff
@@ -9,7 +9,7 @@
         sm="3"
         class="pr-4"
       >
-        <label>Incorporation Date and Time</label>
+        <label>{{ label }}</label>
```
JazzarKarim, line 12: The label here is now a prop as it's not exclusive for IAs anymore.

src/mixins/amalgamation-mixin.ts
```diff
@@ -119,7 +119,7 @@ export default class AmalgamationMixin extends Vue {
   xproUlcCcc (business: AmalgamatingBusinessIF): AmlStatuses {
     if (
       business.type === AmlTypes.FOREIGN &&
-      (!this.isTypeBcUlcCompany || !this.isTypeBcCcc)
+      (this.isTypeBcUlcCompany || this.isTypeBcCcc)
```
JazzarKarim, line 122: I believe the logic here is incorrect. I've been playing around with this and noticed it's incorrect.

We want to show this rule when the type is ULC or CCC not the other way around. I've been getting this rule when I was on a Limited company amalgamation with Foreign companies.

Please correct me if I'm wrong.

src/mixins/amalgamation-mixin.ts
```diff
@@ -205,9 +204,8 @@ export default class AmalgamationMixin extends Vue {
   async refetchAmalgamatingBusinessesInfo (): Promise<void> {
     const fetchTingInfo = async (item: any): Promise<AmalgamatingBusinessIF> => {
       const tingBusiness = await this.fetchAmalgamatingBusinessInfo(item)
-      // *** TODO: need to improve this
-      // (no auth info means not affiliated, which may or may not be foreign)
-      if (!tingBusiness.authInfo) {
+      // no auth info and business info means foreign, otherwise LEAR (affiliated or non-affiliated)
+      if (!tingBusiness.authInfo && !tingBusiness.businessInfo) {
```
JazzarKarim, line 208: So I've tested this extensively.

We have two cases now. If the tingBusiness does not have both authInfo and businessInfo, then it's a Foreign. If it does, then it's LEAR. However, please see next comment.

src/mixins/amalgamation-mixin.ts
```diff
@@ -221,9 +219,9 @@ export default class AmalgamationMixin extends Vue {
           role: AmlRoles.AMALGAMATING,
           identifier: tingBusiness.businessInfo.identifier,
           name: tingBusiness.businessInfo.legalName,
-          email: tingBusiness.authInfo.contacts[0].email,
+          email: tingBusiness.authInfo?.contacts[0].email,
```
JazzarKarim, line 222: If unaffiliated, the optional chaining deals with this and the address on line 224. This is since if it's unaffiliated, both are undefined. So, these will be not set.

If affiliated, things would go normal.

src/mixins/amalgamation-mixin.ts
```diff
@@ -278,9 +276,4 @@ export default class AmalgamationMixin extends Vue {
       (business.type === AmlTypes.LEAR && business.legalType === CorpTypeCd.BC_ULC_COMPANY)
     )
   }
-
-  /** True if there is a BC company in the table. */
-  get isAnyBcCompany (): boolean {
```
JazzarKarim, line 283: I believe this is not needed. This is since that's being taken care of the other rule `needBcCompany`.

src/mixins/filing-template-mixin.ts
```diff
@@ -241,21 +241,6 @@ export default class FilingTemplateMixin extends Mixins(AmalgamationMixin, DateM
     // restore the amalgamating businesses array
     if (draftFiling.amalgamationApplication.amalgamatingBusinesses) {
       this.setAmalgamatingBusinesses([
-        // *** TODO: remove static items when they are no longer needed for testing
```
JazzarKarim, line 244: Obsolete.

src/mixins/filing-template-mixin.ts
```diff
@@ -267,8 +252,6 @@ export default class FilingTemplateMixin extends Mixins(AmalgamationMixin, DateM
     }
 
     // restore Office Addresses
-    // *** TODO: verify whether we need to assign fallback
-    // *** also fix IAs and registrations the same way?
```
JazzarKarim, line 271: I've looked into this, and I don't see the benefit of assigning a fallback. Please correct me if I'm wrong.

src/views/AmalgamationRegular/ShareStructure.vue
```diff
@@ -66,11 +66,11 @@
           The staff at the Corporate Registry cannot provide advice on how to set up your company's
           share structure. If you do not understand what an authorized share structure is or what
           its purpose is or believe you need a more complex share structure, you should seek
-          professional advice or purchase an incorporation guide for detailed information and
+          professional advice or purchase an amalgamation guide for detailed information and
           infrastructure on establishing an authorized share structure.
         </p>
         <p>
-          Refer to this <a :href="helpLink">link</a> to obtain more information on incorporating a company.
+          Refer to this <a :href="helpLink">link</a> to obtain more information on amalgamating a company.
```
JazzarKarim, line 73: Spoke to Janis about the wording and link here. She says it's OK here.

tests/unit/AgreementType.spec.ts
```diff
@@ -1,5 +1,5 @@
 import Vue from 'vue'
-import { wrapperFactory } from '../jest-wrapper-factory'
+import { wrapperFactory } from '../vitest-wrapper-factory'
```
JazzarKarim, line 2: Changed the name here as that's been done in other UIs.

src/components/Amalgamation/BusinessTableSummary.vue
```diff
@@ -1,63 +1,80 @@
...
+      v-if="!getAmalgamatingBusinessesValid"
+      class="defineCompanyStepErrorMessage"
+    >
+      <span>
+        <v-icon color="error">mdi-information-outline</v-icon>
+        <span class="error-text mx-1">This step is unfinished.</span>
+        <span>
+          <router-link
+            v-if="!getAmalgamatingBusinessesValid"
+            :to="{ path: `/${RouteNames.AMALG_REG_INFORMATION}` }"
```
JazzarKarim, line 13: step is unfinished amalg table

src/mixins/amalgamation-mixin.ts
severinbeauvais: I think you are correct (as per the comment on line 118) 👍 

Thanks for playing around and verifying stuff :)

src/mixins/amalgamation-mixin.ts
severinbeauvais, line 222: If authInfo is falsy then email is assigned undefined. Would it better if it was assigned null (ie, add `|| null`)?

Ditto line 224.

src/mixins/amalgamation-mixin.ts
severinbeauvais: OK, so needBcCompany() check if all are foreign. What is the value of isAllForeign if there are no companies at all? (Ie, what does "every" do?) Maybe it doesn't matter because we don't evaluate any rules if there are no tings, but it's still possibly wrong.

I think I'd prefer using `!isAnyBcCompany` for the rules instead. It's probably slower (more complex) but I think it's also more accurate. Thoughts?

src/mixins/filing-template-mixin.ts
severinbeauvais: What I was thinking here is that, if the draft filing doesn't contain office addresses, we'd assign a safe fallback value (eg, empty object?). Similarly to what we do if don't have parties or shareClasses to restore.

However, theoretically the store has safe initial values for offices / parties / shareClasses, so the correct code is "set the value in the store only if the draft contains the value", which is what lines 272-273 already do.

But I'm not certain that our code is consistently doing the right thing. For example:

image

You'd think the fallback to an empty array wouldn't be needed, but in fact parties is initialized to null in the store. It should probably be initialized to an empty array, but I don't know if that will break other filings.

So, to move forward with this, either we try to fix the code now and test well, or we ignore this for now and test well (this is safe but won't resolve the code issues). 🤷‍♂️

src/views/AmalgamationRegular/ReviewConfirm.vue
severinbeauvais: Agreed.

Not only that, but this component is in the incorporation file folder (not common).

Do you have an idea for the new name? "filingDateTime" perhaps?

JazzarKarim: Yes, that could work. I will be doing that then!

src/components/Amalgamation/BusinessTableSummary.vue
severinbeauvais: Did Yui ask for this? I discussed this with him previously and was OK with the two conditional router links in SummaryDefineCompany.vue.

JazzarKarim: I didn't. I just thought we need it. Sorry about that. I'll be reverting this change then.

src/views/AmalgamationRegular/ReviewConfirm.vue
severinbeauvais: Or "effectiveDateTime.vue" ?

JazzarKarim: Went with `EffectiveDateTime` 🙌

src/mixins/amalgamation-mixin.ts
JazzarKarim: I agree Sev! I've deleted the isAllForeign getter and instead, I'm using now the isAnyBcCompany getter.

src/components/Amalgamation/BusinessTableSummary.vue
JazzarKarim: Reverted this change.

src/mixins/filing-template-mixin.ts
JazzarKarim: Makes a lot of sense. Sev, I'll be tackling and looking into this more in next commit.

src/mixins/amalgamation-mixin.ts
```diff
@@ -138,17 +138,16 @@ export default class AmalgamationMixin extends Vue {
...
-  //  * (An amalgamation where all TINGs are foreign will be Phase 2.)
-  //  */
-  // needBcCompany (): AmlStatuses {
-  //   if (this.isAllForeignOrEp) {
-  //     return AmlStatuses.ERROR_NEED_BC_COMPANY
-  //   }
-  //   return null
-  // }
+  /**
+   * Disallow only foreign businesses (including EPs).
```
severinbeauvais, line 142: Please update this comment.

src/views/AmalgamationRegular/ReviewConfirm.vue
severinbeauvais: LGTM

src/mixins/amalgamation-mixin.ts
severinbeauvais: LGTM!

severinbeauvais: (Feel free to resolve this conversation.)

src/mixins/amalgamation-mixin.ts
severinbeauvais: Looks good. Resolving.

severinbeauvais (approved):
Looks good so far. You might consider merging this and having another PR for future changes...

ketaki-deodhar (approved):
Please update package version. Otherwise, Great work Karim! Looks good to me.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-608-4g2rgqhn.web.app

tshyun24 (approved):
Looks good to me! Just try my best to understand it, a good learning process to me!

JazzarKarim:
Going to merge this now and have another small PR for some small additional changes.

PR #609, opened 2024-01-03
18749 - add amalgamation statement to review

tshyun24:
*Issue #:* /bcgov/entity#18749

*Description of changes:*

1. add Amalgamation Statement component (Jan5 updated)
image

2. validation getter (Jan5 updated)
if nothing selected and click `File and Pay`
image
once either of the option being selected
image

3. Able to save and restore

### DIFF
--- a/src/components/Amalgamation/AmalgamationStatement.vue
+++ b/src/components/Amalgamation/AmalgamationStatement.vue
@@ -21,6 +21,7 @@
         <v-radio-group
           v-model="courtApproval"
           class="mt-0 pt-0"
+          @change="changeCourtApproval()"
         >
           <v-radio
             class="radio-button"
@@ -30,9 +31,9 @@
           <div
             class="ml-8 statement-text"
           >
-            This amalgamation has been approved by the court and a copy of the 
-            entered court order approving the amalgamation has been obtained 
-            and has been deposited in the records office of each of the 
+            This amalgamation has been approved by the court and a copy of the
+            entered court order approving the amalgamation has been obtained
+            and has been deposited in the records office of each of the
             amalgamating companies.
           </div>
 
@@ -44,10 +45,10 @@
           <div
             class="ml-8 statement-text"
           >
-            This amalgamation has been effected without court approval. 
-            A copy of all of the required affidavits under section 277(1) 
-            have been obtained and the affidavit obtained from each 
-            amalgamating company has been deposited in that company’s 
+            This amalgamation has been effected without court approval.
+            A copy of all of the required affidavits under section 277(1)
+            have been obtained and the affidavit obtained from each
+            amalgamating company has been deposited in that company’s
             records office.
           </div>
         </v-radio-group>
@@ -56,16 +57,37 @@
   </v-card>
 </template>
 
-
 <script lang="ts">
-import { Component, Emit, Mixins, Prop, Vue, Watch } from 'vue-property-decorator'
+import Vue from 'vue'
+import { Emit, Watch } from 'vue-property-decorator'
 import { AmalgamationStateIF } from '@/interfaces/store-interfaces/state-interfaces/amalgamation-state-interface'
-import { CommonMixin } from '@/mixins'
+import { useStore } from '@/store/store'
+import { Getter } from 'pinia-class'
 
-export default class AmalgamationStatement extends Mixins(CommonMixin) {
-  
+export default class AmalgamationStatement extends Vue {
+  @Getter(useStore) getAmalgamationCourtApproval!: boolean
   // Local properties
-  courtApproval: AmalgamationStateIF = null
+  courtApproval: AmalgamationStateIF['courtApproval'] = null
+
+  changeCourtApproval (): void {
+    this.setCourtApproval(this.courtApproval)
+  }
+
+  @Emit('update')
+  private courtApprovalUpdate (): boolean {
+    return this.courtApproval
+  }
+
+  @Emit('valid')
+  private amalgamationStatementValid (event: boolean): boolean {
+    return event
+  }
+
+  @Watch('courtApproval')
+  private setCourtApproval (val) {
+    this.courtApprovalUpdate()
+    this.amalgamationStatementValid(true)
+  }
 }
 </script>
 
--- a/src/components/Amalgamation/AmalgamationStatement.vue
+++ b/src/components/Amalgamation/AmalgamationStatement.vue
@@ -2,6 +2,7 @@
   <v-card
     id="amalgamation-statement"
     flat
+    class="py-8 px-6"
   >
     <v-row>
       <v-col
@@ -17,10 +18,12 @@
       <v-col
         cols="12"
         sm="9"
+        class="pr-0"
       >
         <v-radio-group
           v-model="courtApproval"
           class="mt-0 pt-0"
+          hide-details="auto"
           @change="setCourtApproval()"
         >
           <v-radio
@@ -28,29 +31,29 @@
             label="With Court Approval"
             :value="true"
           />
-          <div
-            class="ml-8 statement-text"
+          <p
+            class="ml-8 mb-5 statement-text"
           >
             This amalgamation has been approved by the court and a copy of the
             entered court order approving the amalgamation has been obtained
             and has been deposited in the records office of each of the
             amalgamating companies.
-          </div>
+          </p>
 
           <v-radio
-            class="pt-4 radio-button"
+            class="radio-button"
             label="Without Court Approval"
             :value="false"
           />
-          <div
-            class="ml-8 statement-text"
+          <p
+            class="ml-8 mb-0 statement-text"
           >
             This amalgamation has been effected without court approval.
             A copy of all of the required affidavits under section 277(1)
             have been obtained and the affidavit obtained from each
             amalgamating company has been deposited in that company’s
             records office.
-          </div>
+          </p>
         </v-radio-group>
       </v-col>
     </v-row>
@@ -73,27 +76,22 @@ export default class AmalgamationStatement extends Vue {
 
   /** Called when component is mounted. */
   mounted (): void {
-    if (this.getAmalgamationCourtApproval != null) {
-      this.courtApproval = this.getAmalgamationCourtApproval
-    }
+    this.courtApproval = this.getAmalgamationCourtApproval
   }
 
-  // Emit a boolean which is whether the court approval have been selected
+  // Emit a boolean which is whether come with the court approval
   @Emit('update')
-  private courtApprovalUpdate (): boolean {
-    console.log(this.courtApproval)
-    return this.courtApproval
-  }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  private courtApprovalUpdate (courtApproval: boolean): void {}
 
   // Emit a boolean (validation) which is either option being selected
   @Emit('valid')
-  private amalgamationStatementValid (event: boolean): boolean {
-    return event
-  }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  private amalgamationStatementValid (valid: boolean): void {}
 
   // Once either option being selected, update court approval and validation
   setCourtApproval (): void {
-    this.courtApprovalUpdate()
+    this.courtApprovalUpdate(this.courtApproval)
     this.amalgamationStatementValid(true)
   }
 }
@@ -103,17 +101,26 @@ export default class AmalgamationStatement extends Vue {
 @import '@/assets/styles/theme.scss';
 
 #amalgamation-statement {
-  padding: 1.25rem;
-  line-height: 1.2rem;
-  font-size: 1rem;
+  font-size: 16px;
 }
 
 .statement-text {
   color: $gray7;
+  font-size: 16px;
+}
+
+:deep() {
+  // Fix text color for all the radio buttons
+  .radio-button {
+    label {
+      color: $gray7;
+      }
+  }
 }
 
 label {
   font-weight: bold;
+  color: $gray9;
 }
 
 </style>
--- a/src/components/Amalgamation/AmalgamationStatement.vue
+++ b/src/components/Amalgamation/AmalgamationStatement.vue
@@ -79,11 +79,18 @@ export default class AmalgamationStatement extends Vue {
     this.courtApproval = this.getAmalgamationCourtApproval
   }
 
-  // Emit a boolean which is whether come with the court approval
+  /**
+   * Emit a value and event to the parent to handle updating.
+   * @param courtApproval The value to emit.
+   */
   @Emit('update')
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   private courtApprovalUpdate (courtApproval: boolean): void {}
 
+  /**
+   * Emit a value and event to the parent to handle validation.
+   * @param valid The value to emit.
+   */
   // Emit a boolean (validation) which is either option being selected
   @Emit('valid')
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
@@ -101,12 +108,12 @@ export default class AmalgamationStatement extends Vue {
 @import '@/assets/styles/theme.scss';
 
 #amalgamation-statement {
-  font-size: 16px;
+  font-size: $px-16;
 }
 
 .statement-text {
   color: $gray7;
-  font-size: 16px;
+  font-size: $px-16;
 }
 
 :deep() {
--- a/src/components/Amalgamation/AmalgamationStatement.vue
+++ b/src/components/Amalgamation/AmalgamationStatement.vue
@@ -1,0 +1,89 @@
+<template>
+  <v-card
+    id="amalgamation-statement"
+    flat
+  >
+    <v-row>
+      <v-col
+        cols="12"
+        sm="3"
+        class="pr-4"
+      >
+        <label>
+          Amalgamation Statement
+        </label>
+      </v-col>
+
+      <v-col
+        cols="12"
+        sm="9"
+      >
+        <v-radio-group
+          v-model="courtApproval"
+          class="mt-0 pt-0"
+        >
+          <v-radio
+            class="radio-button"
+            label="With Court Approval"
+            :value="true"
+          />
+          <div
+            class="ml-8 statement-text"
+          >
+            This amalgamation has been approved by the court and a copy of the 
+            entered court order approving the amalgamation has been obtained 
+            and has been deposited in the records office of each of the 
+            amalgamating companies.
+          </div>
+
+          <v-radio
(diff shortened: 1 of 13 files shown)


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-609-ft6cgzmy.web.app

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-609-ft6cgzmy.web.app

src/interfaces/store-interfaces/state-interfaces/amalgamation-state-interface.ts
tshyun24, line 1: Add this for validation

src/mixins/filing-template-mixin.ts
tshyun24, line 1: Change it to `!=null`, otherwise the file cannot be saved when the user selected not court approval option

severinbeauvais:
Re:
> 1. add Amalgamation Statement component
>    image

The label colour is incorrect. If I inspect the sample, the color is `#0009`. Instead, it should be $gray7. See the specifications in the uxpin: 

image

 (The text colour is correct.)

severinbeauvais:
Re:
> 2. validation getter
>    if nothing selected and click `File and Pay`
>    image

The red bar should extend all the way up.

src/mixins/filing-template-mixin.ts
severinbeauvais: OK, but it should be `!== null`.

(Google it if you don't know why.)

src/store/store.ts
```diff
@@ -591,14 +591,16 @@ export const useStore = defineStore('store', {
       // *** TODO: add checks for review page components
       const isFolioNumberValid = !this.isPremiumAccount || this.getFolioNumberValid
       const isCourtOrderValid = this.isRoleStaff ? this.getCourtOrderStep.valid : true
+      const isCourtApprovalValid = this.getAmalgamationCourtApprovalValid
```
severinbeauvais, line 594: A temporary variable makes sense for lines 592 and 593 since they do some computation and it looks cleaner to not have that on lines 601 and 602.

In your case, I recommend just putting the getter on line 603.

src/views/AmalgamationRegular/ReviewConfirm.vue
```diff
@@ -174,16 +174,21 @@
       <header>
         <h2>Amalgamation Statement</h2>
         <p class="mt-4">
-          [*** TODO: blurb ***]
+          Please indicate the statement applicable to this amalgamation.
         </p>
       </header>
 
       <v-card
```
severinbeauvais, line 181: You don't need the v-card here. It was a placeholder for your new component.

Right now you have a v-card inside a v-card, which is causing some spacing and error validation issues. (It's not terrible to have a v-card inside a v-card, but it's not needed here. The other components on this page don't do it.)

src/components/Amalgamation/AmalgamationStatement.vue
```diff
@@ -0,0 +1,119 @@
...
+  line-height: 1.2rem;
+  font-size: 1rem;
+}
+
+.statement-text {
+  color: $gray7;
+}
+
+label {
+  font-weight: bold;
```
severinbeauvais, line 116: You also need to set the label colour to $gray9.

src/components/Amalgamation/AmalgamationStatement.vue
```diff
@@ -0,0 +1,119 @@
...
+    this.amalgamationStatementValid(true)
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+#amalgamation-statement {
+  padding: 1.25rem;
```
severinbeauvais, line 106: It's often better to put the layout properties right in the template (using the Vuetify helper classes), so that a future developer doesn't have to look up classnames/selectors to determine what properties are being applied, and to facilitate layout updates. (Don't you hate having to look up selectors in the CSS to see where various properties come from? This is what we are trying to avoid if possible.)

The other properties here (probably) don't have helper classes, so they're fine here. But it would be even better -- future state -- if we had base/common classes for things like radio labels, statements, headers, etc. Then we wouldn't need any CSS here; we would just label everything according to their classname. This is the "ideal case". For now, properties here are OK. But... if you find an opportunity to make base/common classes for things, please go ahead :)

src/components/Amalgamation/AmalgamationStatement.vue
```diff
@@ -0,0 +1,119 @@
...
+  private courtApprovalUpdate (): boolean {
+    console.log(this.courtApproval)
+    return this.courtApproval
+  }
+
+  // Emit a boolean (validation) which is either option being selected
+  @Emit('valid')
+  private amalgamationStatementValid (event: boolean): boolean {
+    return event
+  }
```
severinbeauvais, line 92: FYI, there's a shortcut syntax for this:
```
  // Emit a boolean (validation) which is either option being selected
  @Emit('valid')
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
  private amalgamationStatementValid (valid: boolean): void {}
```

src/components/Amalgamation/AmalgamationStatement.vue
```diff
@@ -0,0 +1,119 @@
...
+  mounted (): void {
+    if (this.getAmalgamationCourtApproval != null) {
+      this.courtApproval = this.getAmalgamationCourtApproval
+    }
+  }
+
+  // Emit a boolean which is whether the court approval have been selected
+  @Emit('update')
+  private courtApprovalUpdate (): boolean {
+    console.log(this.courtApproval)
```
severinbeauvais, line 84: Remove debugging before merging PR.

After removing this line, you could use the shortcut syntax here, too.

src/components/Amalgamation/AmalgamationStatement.vue
```diff
@@ -0,0 +1,119 @@
...
+  courtApproval: AmalgamationStateIF['courtApproval'] = null
+
+  /** Called when component is mounted. */
+  mounted (): void {
+    if (this.getAmalgamationCourtApproval != null) {
+      this.courtApproval = this.getAmalgamationCourtApproval
+    }
+  }
+
+  // Emit a boolean which is whether the court approval have been selected
```
severinbeauvais, line 81: If you use JSDoc syntax for the comment, you can have nice contextual help, like this:

image

src/components/Amalgamation/AmalgamationStatement.vue
```diff
@@ -0,0 +1,119 @@
...
+import { AmalgamationStateIF } from '@/interfaces/store-interfaces/state-interfaces/amalgamation-state-interface'
+import { Getter } from 'pinia-class'
+import { useStore } from '@/store/store'
+
+@Component({})
+export default class AmalgamationStatement extends Vue {
+  @Getter(useStore) getAmalgamationCourtApproval!: boolean
+
+  // Local properties
+  courtApproval: AmalgamationStateIF['courtApproval'] = null
```
severinbeauvais, line 72: This is a clever way to specify the type but I'm not sure if I love it. Did you see somewhere that it is a best practise?

The issue I have with this is that you need to know the property name in the interface, but I don't know if that's worse than just saying it's a boolean (which you need to know in this file anyway).

Another thing I'm not sure I like is the syntax, which looks like a quoted object property (compared to a named object property). There are lint rules (strict-property-naming) to enforce the named style, and lots to read on the subject... My conclusion is that quoted properties are to be avoided,

(This is not the same as having object properties in quotes. That's another topic.)

src/components/Amalgamation/AmalgamationStatement.vue
```diff
@@ -0,0 +1,119 @@
...
+
+  /** Called when component is mounted. */
+  mounted (): void {
+    if (this.getAmalgamationCourtApproval != null) {
+      this.courtApproval = this.getAmalgamationCourtApproval
+    }
```
severinbeauvais, line 78: 1. Why do you need the conditional on line 76? If the getter value is null then you could simply reassign null to the local variable.

2. Always use `===` or `!==` in comparisons.

3. This fails if the component is mounted before the getter has a valid value (ie, the draft filing has been loaded). Fortunately, I think App.vue is written so that the pages aren't displayed (mounted) until the draft filing has been loaded, so it works here, but don't count on it in other places. I think we don't have a consistent method of having local variables synced with props or store variables (which are both read-only), so this is always a bit tricky. In other code, we have a watcher that sets the local initially ("immediate: true") and whenever the prop/getter changes. I think that's safer code but I think it's still overly complex. If you think of a nicer way to do this, let me know.

src/components/Amalgamation/AmalgamationStatement.vue
```diff
@@ -0,0 +1,119 @@
...
+          v-model="courtApproval"
+          class="mt-0 pt-0"
+          @change="setCourtApproval()"
+        >
+          <v-radio
+            class="radio-button"
+            label="With Court Approval"
+            :value="true"
+          />
+          <div
```
severinbeauvais, line 31: When possible, you should use the best tag/element for its content -- in this case, it would be better to use the paragraph element instead of the generic div element.

See also https://www.w3schools.com/html/html5_semantic_elements.asp

src/components/Amalgamation/AmalgamationStatement.vue
```diff
@@ -0,0 +1,119 @@
...
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+#amalgamation-statement {
+  padding: 1.25rem;
+  line-height: 1.2rem;
+  font-size: 1rem;
```
severinbeauvais, line 108: We have CSS variables for font sizes. Please use them. (It allows for easier global searching for where different sizes are used.)

severinbeauvais (commented):
Could you please add a unit test file for your new component? Thanks.

src/mixins/filing-template-mixin.ts
```diff
@@ -262,7 +262,7 @@ export default class FilingTemplateMixin extends Mixins(AmalgamationMixin, DateM
     }
 
     // restore the amalgamation court approval
-    if (draftFiling.amalgamationApplication.courtApproval) {
+    if (draftFiling.amalgamationApplication.courtApproval != null) {
```
JazzarKarim, line 265: I don't think you need to change this as the previous syntax does exactly this!

tshyun24: could you plz explain more? cause I try without this change, when I select "without court approval", the selection cannot be saved

JazzarKarim: OK, it's fine. But please if you can change it to `!==`. Thanks!

severinbeauvais: I think `courtApproval` may be True or False, which is why the check for null is needed.

src/components/Amalgamation/AmalgamationStatement.vue
tshyun24: > 1. Why do you need the conditional on line 76? If the getter value is null then you could simply reassign null to the local variable.

yeahhh, I just realize that.

> 3.

I think I will keep the same with other file, using ("immediate: true"). But I am wondering what does "other places" in ` I think App.vue is written so that the pages aren't displayed (mounted) until the draft filing has been loaded, so it works here, but don't count on it in other places. ` means?

severinbeauvais: > But I am wondering what does "other places" in (snip) means?

In this web app, I think it is safe to assume that most components aren't created/mounted until the data is loaded, but in a different UI this may not be true. Maybe Filings UI, maybe a future web app you work on, etc :)

src/components/Amalgamation/AmalgamationStatement.vue
tshyun24: > Did you see somewhere that it is a best practise?

Nope, I just go through other files and find that it makes me more clear to write down the property name at here.

> but I don't know if that's worse than just saying it's a boolean

I am thinking should I simply write `courtApproval = null as boolean` or do `courtApproval: AmalgamationStateIF['courtApproval'] = null as boolean`? The first one is more simple, the second one may contain more information but don't know how to avoid the quoted properties

severinbeauvais: Yes, exactly.

Let's leave your current code for now, use it a bit, and decide later. It's easy to change later if we decide against it.

severinbeauvais: Shaoyun, you can resolve this conversation after you've read this.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-609-ft6cgzmy.web.app

src/mixins/filing-template-mixin.ts
severinbeauvais: Looks good.

src/store/store.ts
severinbeauvais: Looks good.

src/components/Amalgamation/AmalgamationStatement.vue
severinbeauvais: Looks good.

src/components/Amalgamation/AmalgamationStatement.vue
severinbeauvais: Looks good.

src/components/Amalgamation/AmalgamationStatement.vue
severinbeauvais: Looks good.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-609-ft6cgzmy.web.app

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-609-ft6cgzmy.web.app

tests/unit/AmalgamationStatement.spec.ts
severinbeauvais, line 1: Please fix GitHub Actions linting issue identified below. Thanks!

If you are using VS Code, you may have to set the default indentation (to 2) for a new file in its settings.

src/mixins/filing-template-mixin.ts
severinbeauvais: I'm resolving this conversation because Karim started another conversation to discuss whether or not this is needed (which it is).

tests/unit/AmalgamationStatement.spec.ts
tshyun24: Hii Sev, please see the newest comment, which should fix the linting issue -)

src/mixins/filing-template-mixin.ts
severinbeauvais: Shaoyun, since this caused confusion once, and to avoid a future developer trying to "optimize" this and actually break it, please add a code comment explaining why you check for not null (ie, save value if it's True or False).

Otherwise, this code looks good to me.

src/components/Amalgamation/AmalgamationStatement.vue
severinbeauvais: Please use CSS variable `$px-16` for this font size (and also the other font sizes in this file). Thx.

src/components/Amalgamation/AmalgamationStatement.vue
severinbeauvais: Changed to "p". Looks good. Resolving.

src/components/Amalgamation/AmalgamationStatement.vue
severinbeauvais: How about changing your function comments from `// Emit a ...` to
```
/**
  * Emits a ...
  * @param courtApproval The value to emit.
  */
```

(For all 3 functions with non-JSDoc syntax.)

src/components/Amalgamation/AmalgamationStatement.vue
severinbeauvais: Thanks for the change. Looks good.

src/views/AmalgamationRegular/ReviewConfirm.vue
severinbeauvais: Looks good.

severinbeauvais (approved):
Just a couple small things otherwise I approve this.

tests/unit/AmalgamationStatement.spec.ts
severinbeauvais: Yes, you fixed the indentation in the latest commit. Thanks! Resolving.

tshyun24:
> Just a couple small things otherwise I approve this.

Thank you so much to your patient Sev!!! Lots of useful reviews🙏

src/components/Amalgamation/AmalgamationStatement.vue
```diff
@@ -79,11 +79,18 @@ export default class AmalgamationStatement extends Vue {
...
+   */
   @Emit('update')
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   private courtApprovalUpdate (courtApproval: boolean): void {}
 
+  /**
+   * Emit a value and event to the parent to handle validation.
+   * @param valid The value to emit.
+   */
   // Emit a boolean (validation) which is either option being selected
```
severinbeauvais, line 94: Delete

src/components/Amalgamation/AmalgamationStatement.vue
```diff
@@ -79,11 +79,18 @@ export default class AmalgamationStatement extends Vue {
...
+   * @param courtApproval The value to emit.
+   */
   @Emit('update')
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   private courtApprovalUpdate (courtApproval: boolean): void {}
 
+  /**
+   * Emit a value and event to the parent to handle validation.
+   * @param valid The value to emit.
+   */
```
severinbeauvais, line 93: Also update comment for setCourtApproval ?

src/mixins/filing-template-mixin.ts
severinbeauvais: The updated comment is fine by me. Resolving.

src/components/Amalgamation/AmalgamationStatement.vue
tshyun24: Just wondering because it's does not have `param` and `return`, should I also update the comment to JSDoc syntax or leave it now?

severinbeauvais: Sure. That way you'll get a nice hover help panel for the function :)

severinbeauvais:
@ketaki-deodhar If you have a moment today, please review and merge this. If not, just merge it. Please and thanks!

ketaki-deodhar:
> @ketaki-deodhar If you have a moment today, please review and merge this. If not, just merge it. Please and thanks!

Looking at it now Sev!

ketaki-deodhar (approved):
Looks good!

PR #617, opened 2024-01-09
Fixed bug with certify if submitting the amalgamation failed

JazzarKarim:
*Issue #:* /bcgov/entity#18689

*Description of changes:*
- Fix Certify blocked being valid even if the box hasn't been ticked

### DIFF
--- a/src/views/AmalgamationRegular/ReviewConfirm.vue
+++ b/src/views/AmalgamationRegular/ReviewConfirm.vue
@@ -331,6 +331,17 @@ export default class AmalgamationRegularReviewConfirm extends Vue {
   @Action(useStore) setHasPlanOfArrangement!: (x: boolean) => void
   @Action(useStore) setIsFutureEffective!: (x: boolean) => void
 
+  /**
+   * In case submitting the amalgamation failed, we want to reset the validity of Certify.
+   * This is since the checkbox has to be ticked again after the save dialog has been closed.
+   */
+  mounted (): void {
+    this.setCertifyState({
+      certifiedBy: this.getCertifyState.certifiedBy,
+      valid: false
+    })
+  }
+
   /** The entity description,  */
   get getEntityDescription (): string {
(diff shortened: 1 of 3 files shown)


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-617-yl7e9l8m.web.app

src/views/AmalgamationRegular/ReviewConfirm.vue
```diff
@@ -331,6 +331,17 @@ export default class AmalgamationRegularReviewConfirm extends Vue {
...
+  /**
+   * In case submitting the amalgamation failed, we want to reset the validity of Certify.
+   * This is since the checkbox has to be ticked again after the save dialog has been closed.
+   */
+  mounted (): void {
+    this.setCertifyState({
+      certifiedBy: this.getCertifyState.certifiedBy,
+      valid: false
+    })
+  }
```
severinbeauvais, line 343: Is this how other filings do it? Or do they need this, too?

JazzarKarim: Good point. I think they need this too. The other filings don't do anything as of right now from what I checked.

severinbeauvais: Should the Certify component report itself as invalid, so that this isn't needed?

JazzarKarim: It can't do that I believe in this scenario. We're using the common one that's created from the shared components one.

The reason is, when the save error dialog is shown, at that point the component's validity is true. After it's closed, the UI keeps that validity. So, it's still true even though the checkbox is back to being unticked. It kept that from before.

severinbeauvais: OK. Do it your way. I was hoping for something generic. Resolving.

severinbeauvais (approved):

PR #628, opened 2024-01-18
19359 - Amalgamating Businesses Table must have more than 1

JazzarKarim:
*Issue #:* /bcgov/entity#19359

*Description of changes:*
Increased businesses table check from > 0 to > 1

### DIFF
--- a/src/components/Amalgamation/BusinessTable.vue
+++ b/src/components/Amalgamation/BusinessTable.vue
@@ -181,7 +181,7 @@ export default class BusinessTable extends Mixins(AmalgamationMixin) {
   @Emit('valid')
   private emitValidity (): boolean {
     return (
-      (this.businesses.length > 1) &&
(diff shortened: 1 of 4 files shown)


severinbeauvais (approved):
Now this is the smallest PR ever 😃

ketaki-deodhar (approved):
:D

src/components/Amalgamation/BusinessTable.vue
```diff
@@ -181,7 +181,7 @@ export default class BusinessTable extends Mixins(AmalgamationMixin) {
   @Emit('valid')
   private emitValidity (): boolean {
     return (
-      (this.businesses.length > 0) &&
+      (this.businesses.length > 1) &&
```
severinbeauvais, line 184: To be specific, the rule is "2 or more", so it might be better to use ">= 2".

Also, is there any way to have a unit test for this?

PR #641, opened 2024-01-31
19039-show-resulting-name-and-business-type

jamespaologarcia:
*Issue #:* /bcgov/entity#19039

Show the name and business type of the holding company

### DIFF
--- a/src/components/Amalgamation/ResultingBusinessName.vue
+++ b/src/components/Amalgamation/ResultingBusinessName.vue
@@ -135,17 +135,6 @@ export default class ResultingBusinessName extends Mixins(AmalgamationMixin, Nam
     // validateNameRequest() already throws printable errors
     return this.validateNameRequest(nameRequest, NrRequestActionCodes.AMALGAMATE)
   }
-  // mounted () {
-  //   // Call the watcher method manually on component mount
-  //   this.onIsAmalgamationFilingRegularChanged(this.isAmalgamationFilingRegular)
-  // }
-
-  // @Watch('isAmalgamationFilingRegular')
-  // onIsAmalgamationFilingRegularChanged (newVal: boolean) {
-  //   if (!newVal) {
-  //     this.setCorrectNameOption(CorrectNameOptions.CORRECT_AML_ADOPT)
-  //   }
-  // }
 
   /** On company name update, sets store accordingly. */
   onUpdateCompanyName (name: string): void {
--- a/src/components/Amalgamation/ResultingBusinessName.vue
+++ b/src/components/Amalgamation/ResultingBusinessName.vue
@@ -16,7 +16,7 @@
           class="pr-4"
         >
           <label :class="{ 'error-text': invalidSection }">
-            <strong>Resulting Business Name</strong>
+            <strong>Company Name</strong>
           </label>
         </v-col>
 
@@ -47,9 +47,12 @@
     <!-- Display Mode -->
     <template v-else>
       <NameRequestInfo />
-      <NameTranslations />
+      <NameTranslations
+        v-if="isRegularAmalgamation"
+      />
 
       <v-btn
+        v-if="isRegularAmalgamation"
         text
         color="primary"
         class="btn-undo"
@@ -65,7 +68,7 @@
 </template>
 
 <script lang="ts">
-import { Component, Mixins } from 'vue-property-decorator'
+import { Component, Mixins, Prop } from 'vue-property-decorator'
 import { Getter } from 'pinia-class'
 import { useStore } from '@/store/store'
 import { AmalgamationMixin, NameRequestMixin } from '@/mixins/'
@@ -93,6 +96,9 @@ export default class ResultingBusinessName extends Mixins(AmalgamationMixin, Nam
   @Getter(useStore) getNameRequestApprovedName!: string
   @Getter(useStore) getNameRequestNumber!: string
   @Getter(useStore) getShowErrors!: boolean
+
+  // Props
+  @Prop({ default: false }) readonly isRegularAmalgamation!: boolean
 
   // Local properties
   formType = null as CorrectNameOptions
@@ -152,26 +158,3 @@ export default class ResultingBusinessName extends Mixins(AmalgamationMixin, Nam
   }
 }
 </script>
-
-<style lang="scss" scoped>
-// position the Undo button "on top of" NameRequestInfo
-.btn-undo {
-  position: absolute;
-  top: 22px;
-  right: 20px;
-}
-
-// "sm" breakpoint
-@media (min-width: 600px) {
-  .btn-undo {
-    top: 24px;
-  }
-}
-
-// "md" breakpoint
-@media (min-width: 960px) {
-  .btn-undo {
-    top: 28px;
-  }
-}
-</style>
--- a/src/components/Amalgamation/ResultingBusinessName.vue
+++ b/src/components/Amalgamation/ResultingBusinessName.vue
@@ -5,7 +5,7 @@
   >
     <!-- Editing Mode -->
     <div
-      v-if="!getCorrectNameOption"
+      v-if="!getCorrectNameOption && isAmalgamationFilingRegular"
       class="section-container"
       :class="{ 'invalid-section': invalidSection }"
     >
@@ -16,7 +16,7 @@
           class="pr-4"
         >
           <label :class="{ 'error-text': invalidSection }">
-            <strong>Company Name</strong>
+            <strong>Resulting Business Name</strong>
           </label>
         </v-col>
 
@@ -48,11 +48,11 @@
     <template v-else>
       <NameRequestInfo />
       <NameTranslations
-        v-if="isRegularAmalgamation"
+        v-if="isAmalgamationFilingRegular"
       />
 
       <v-btn
-        v-if="isRegularAmalgamation"
+        v-if="isAmalgamationFilingRegular"
         text
         color="primary"
         class="btn-undo"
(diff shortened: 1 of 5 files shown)


jamespaologarcia:
Unit test in progress

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-641-clib6d86.web.app

src/components/Amalgamation/ResultingBusinessName.vue
```diff
@@ -152,26 +158,3 @@ export default class ResultingBusinessName extends Mixins(AmalgamationMixin, Nam
...
-  }
-}
-
-// "md" breakpoint
-@media (min-width: 960px) {
-  .btn-undo {
-    top: 28px;
-  }
-}
-</style>
```
severinbeauvais, line 177: What does the Undo button look like now on a regular amalgamation ?

src/components/Amalgamation/ResultingBusinessName.vue
```diff
@@ -16,7 +16,7 @@
           class="pr-4"
         >
           <label :class="{ 'error-text': invalidSection }">
-            <strong>Resulting Business Name</strong>
+            <strong>Company Name</strong>
```
severinbeauvais, line 19: Design has been updated.

image

src/views/Amalgamation/Information.vue
```diff
@@ -71,7 +72,22 @@
         </template>
       </ExpandableHelp>
 
-      <ResultingBusinessName class="mt-5" />
+      <p
+        v-if="isAmalgamationFilingVertical"
+        class="mt-3">
```
severinbeauvais, line 77: Per GitHib warning, please move `>` to new line. Also on line 83 below.

src/views/Amalgamation/Information.vue
```diff
@@ -71,7 +72,22 @@
...
+        This is a vertical short form amalgamation under section 273 of Business Corporations Act.
+        The amalgamated business will adopt as its notice of articles, the notice of articles of the holding business.
+      </p>
+      <p
+        v-if="isAmalgamationFilingHorizontal"
+        class="mt-3">
+        This is a horizontal short form amalgamation under section 273 of Business Corporations Act.
+        The amalgamated business will adopt as its notice of articles, the notice of articles of the primary business.
+      </p>
```
severinbeauvais, line 86: This is a lot of duplication for almost the same text. I suggest adding an inline ternary to display vertical/horizontal and holding/primary instead.

src/components/Amalgamation/ResultingBusinessName.vue
```diff
@@ -94,6 +97,9 @@ export default class ResultingBusinessName extends Mixins(AmalgamationMixin, Nam
   @Getter(useStore) getNameRequestNumber!: string
   @Getter(useStore) getShowErrors!: boolean
 
+  // Props
+  @Prop({ default: false }) readonly isRegularAmalgamation!: boolean
```
severinbeauvais, line 101: Props are useful for "local scope" differences (ie, not in global store), or for components that you want to keep as dumb as possible (ie, you don't want them to depend on anything global like the store or axios).

In this case, this component already has store getters, so just add one more (isAmalgamationFilingRegular). I think this would be cleaner than having to pass the prop in the parent component and extracting it here.

src/views/Amalgamation/Information.vue
```diff
@@ -71,7 +72,22 @@
...
+      <ResultingBusinessName
+        :isRegularAmalgamation="isAmalgamationFilingRegular"
+        class="mt-5"
+      />
```
severinbeauvais, line 90: What does this section look like for a short-form amalgamation?

In particular, I'm curious what draft/store variable sets the single Correct Name Option (which I think is CORRECT_AML_ADOPT)?

src/components/Amalgamation/ResultingBusinessName.vue
jamespaologarcia: This is what's on dev:
image

src/views/Amalgamation/Information.vue
jamespaologarcia: This is on dev:
image
This is on the PR:
image

src/components/Amalgamation/ResultingBusinessName.vue
severinbeauvais: No, what do you see (locally) after deleting this CSS?

src/views/Amalgamation/Information.vue
severinbeauvais: OK, your PR screenshot looks good, but where did the Correct Name Option get set? What do you see on a brand-new amalgamation draft filing?

jamespaologarcia: image
You go through here first.  (No changes on the code were made here)

src/components/Amalgamation/ResultingBusinessName.vue
jamespaologarcia: I'll put that CSS back. It's supposed to still be there. Removing the CSS moves "Undo" to the bottom of the container.

src/views/Amalgamation/Information.vue
severinbeauvais: I think we want this option automatically selected. Please check with Yui.

It should be easy to set in the code -- it's the `correctNameOption` field in the nameRequest object. I've been thinking about this and we might want to set this option (and the legalName) when we create the short-form amalgamation drafts -- in Filings UI. Look it over and see what you think.

jamespaologarcia: Yeah, because we removed the "Undo" button for short forms anyway. Will check with Yui about this.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-641-clib6d86.web.app

src/mixins/filing-template-mixin.ts
```diff
@@ -241,6 +241,19 @@ export default class FilingTemplateMixin extends Mixins(AmalgamationMixin, DateM
...
+
+      const holdingBusiness = draftFiling.amalgamationApplication.amalgamatingBusinesses.find(
+        (business) => business.role === 'holding'
+      )
+      if (holdingBusiness) {
+        this.fetchAmalgamatingBusinessInfo(holdingBusiness.identifier).then((result) => {
+          this.setNameRequestApprovedName(result.businessInfo.legalName);
+        })
+      }
+    }
```
severinbeauvais, line 256: It looks like this will work, but:

1. I still believe the legal name should be pre-populated when the draft amalgamation is created in Filings UI, not "patched" here. This is going back to which code (UI) should be doing what.

2. When you rebase, you will already have the `holdingPrimary` object (if it exists), so you don't need to find it yourself.

3. You should enums instead of strings (line 249).

4. It would be better to use await instead of then when fetching data. When using await, the code stops executing until the fetch is completed, and then continues executing. When using then, its code block is executed after the fetch is complete (in a different time-slice), but the other code in here executes immediately.

5. After you rebase, you won't need to fetch the amalgamating business info for the holding business as it will already have been loaded (per point 2 above).

src/components/Amalgamation/ResultingBusinessName.vue
```diff
@@ -130,6 +133,17 @@ export default class ResultingBusinessName extends Mixins(AmalgamationMixin, Nam
...
+  //   // Call the watcher method manually on component mount
+  //   this.onIsAmalgamationFilingRegularChanged(this.isAmalgamationFilingRegular)
+  // }
+
+  // @Watch('isAmalgamationFilingRegular')
+  // onIsAmalgamationFilingRegularChanged (newVal: boolean) {
+  //   if (!newVal) {
+  //     this.setCorrectNameOption(CorrectNameOptions.CORRECT_AML_ADOPT)
+  //   }
+  // }
```
severinbeauvais, line 146: Remove this before your PR is merged.

src/components/Amalgamation/ResultingBusinessName.vue
severinbeauvais: Reverted, thanks. Resolving.

src/views/Amalgamation/Information.vue
severinbeauvais: Fixed, thanks.

src/components/Amalgamation/ResultingBusinessName.vue
severinbeauvais: Looks good. Resolving.

src/views/Amalgamation/Information.vue
severinbeauvais: Resolved.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-641-clib6d86.web.app

src/mixins/filing-template-mixin.ts
```diff
@@ -254,6 +254,10 @@ export default class FilingTemplateMixin extends Mixins(AmalgamationMixin, DateM
 
     // restore business name data
     const nameRequest = draftFiling.amalgamationApplication.nameRequest as NameRequestFilingIF
+
+    if (this.getAmalgamationType !== AmalgamationTypes.REGULAR) {
+      this.setCorrectNameOption(CorrectNameOptions.CORRECT_AML_ADOPT)
+    }
```
severinbeauvais, line 260: OK, I think we can leave this as a fallback option (say, for existing amalgamation drafts). This should still be pre-populated in Filings UI [around here](https://github.com/bcgov/business-filings-ui/blob/1142374aa156d23d96305144b85ddbdedf50d1c0/src/views/AmalgamationSelection.vue#L239).

In fact, check with Yui whether he wants the "adopt" option pre-populated by default for regular amalgamations that are started from an existing business (as above).

One last thing: on line 258, instead of assuming it must be a short-form alamg if it's not a regular, please check if it's a short-form amalg. This kind of "if not something" has bitten us in the past when we added a type or filing or something and the assumption is wrong. It's better to check "if something".

jamespaologarcia: Checked with Yui: "We don't want to assume adoption of name for regular when started from an existing business."

severinbeauvais: Thanks for checking with Yui. So this code applies only to horiz and vert amalgamations 👍

PR #650, opened 2024-02-08
19698 Added Account-Id to auth api call + check for null auth info + small cleanup

severinbeauvais:
*Issue #:* bcgov/entity#19698

*Description of changes:*
    - app version = 5.8.15
    - added get after set to update store
    - added check for empty auth info when re-fetching ting business info
    - added some safety checks when updating prepopulated data (ie, new holding/primary business)
    - moved accountId computed to store
    - added Account-Id header to fetchAuthInfo()
    - deleted some unused getters/local code
    - cleaned up some App unit tests

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -260,8 +260,8 @@ import { AmalgamationRegResources, AmalgamationShortResources, ContinuationInRes
 import { AuthServices, LegalServices, PayServices } from '@/services/'
 
 // Enums and Constants
-import { EntityStates, ErrorTypes, FilingCodes, FilingNames, FilingStatus, FilingTypes, NameRequestStates, RouteNames,
-  StaffPaymentOptions } from '@/enums'
+import { EntityStates, ErrorTypes, FilingCodes, FilingNames, FilingStatus, FilingTypes, NameRequestStates,
+  RouteNames, StaffPaymentOptions } from '@/enums'
 import { SessionStorageKeys } from 'sbc-common-components/src/util/constants'
 import { CorpTypeCd } from '@bcrs-shared-components/corp-type-module'
 
@@ -291,7 +291,6 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
   @Getter(useStore) getFilingName!: FilingNames
   @Getter(useStore) getFilingType!: FilingTypes
   @Getter(useStore) getHaveChanges!: boolean
-  @Getter(useStore) getKeycloakRoles!: Array<string>
   @Getter(useStore) getOrgInformation!: OrgInformationIF
   @Getter(useStore) getSteps!: Array<StepIF>
   @Getter(useStore) getUserFirstName!: string
@@ -305,7 +304,6 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
   @Getter(useStore) isDissolutionFiling!: boolean
   @Getter(useStore) isIncorporationFiling!: boolean
   @Getter(useStore) isMobile!: boolean
-  @Getter(useStore) isRestorationFiling!: boolean
   @Getter(useStore) isSbcStaff!: boolean
 
   @Action(useStore) setAccountInformation!: (x: AccountInformationIF) => void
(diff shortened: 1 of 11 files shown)


src/App.vue
```diff
@@ -875,6 +873,8 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
     // set the resources
     if (!resources) throw new Error(`Invalid ${this.getEntityType} resources`)
     this.setResources(resources)
+    // NB - for some reason, need to call this here so the store updates this getter
+    const dummy = this.getFilingData // eslint-disable-line
```
severinbeauvais, line 877: This is very strange. Without this here, `getFilingData` always returned null. I tried a few things, including awaiting next tick, refactoring getFilingData, refactoring the failing App unit test, etc, and this is the only thing that worked consistently.

In any case, this code is safe (won't hurt anything), but I am a tad concerned about other getters that don't work after just setting a value/object in the store.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-650-dtecq19o.web.app

src/mixins/amalgamation-mixin.ts
```diff
@@ -150,6 +150,7 @@ export default class AmalgamationMixin extends Vue {
   }
 
   /** Disallow CCC mismatch. */
+  // *** FUTURE: should we also block a CCC amalgamation that doesn't have a CCC business in it?
```
severinbeauvais, line 153: This is with Mihai and Naveen to figure out if we need it.

src/mixins/amalgamation-mixin.ts
```diff
@@ -320,22 +321,34 @@ export default class AmalgamationMixin extends Vue {
...
+        // if auth info is empty then business is not (no longer) affiliated with current account
+        if (!business.authInfo) {
+          return {
+            type: AmlTypes.LEAR,
+            role: AmlRoles.AMALGAMATING,
+            identifier: item.identifier,
+            name: item.name,
+            legalType: item.legalType
+          } as AmalgamatingBusinessIF
+        }
```
severinbeauvais, line 335: This item will appear as "not affiliated".

It should never happen that the user was able to add the business but now it doesn't exist at all.

src/services/auth-services.ts
```diff
@@ -34,8 +39,9 @@ export default class AuthServices {
 
     const authApiUrl = sessionStorage.getItem(SessionStorageKeys.AuthApiUrl)
     const url = `${authApiUrl}entities/${businessId}`
+    const config = { headers: { 'Account-Id': store.getAccountId } }
 
-    return axios.get(url).then(response => {
+    return axios.get(url, config).then(response => {
```
severinbeauvais, line 44: We pass in the account id and the API checks it against the business we're asking about. We'll get a HTTP 403 back if the business isn't affiliated to this account id.

NB - The API continues to accept "missing account id" and just returns the data if the business is affiliated with any account for the current login (Keycloak token).

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-650-dtecq19o.web.app

src/mixins/amalgamation-mixin.ts
```diff
@@ -356,13 +368,13 @@ export default class AmalgamationMixin extends Vue {
...
    */
   async updatePrepopulatedData (business: AmalgamatingBusinessIF, isNew = false): Promise<void> {
-    // safety check
-    if (!business || business.type !== AmlTypes.LEAR) {
-      throw new Error('updatePrepopulatedData(): invalid business')
-    }
+    // safety checks
+    if (!business || business.type !== AmlTypes.LEAR) throw new Error('Invalid business')
+    if (!business.addresses) throw new Error('Missing business addresses')
+    if (!business.authInfo) throw new Error('Missing share classes')
```
severinbeauvais, line 374: This is to display a console error (and however the caller wants to handle the exception) in case there is any error with the data (eg, Riyaz's testing in [19750](https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/19750)).

PaulGarewal (approved):
LGTM 👍 lots of interesting things going on here that I will look at again for learning

JazzarKarim (approved):
LGTM Sev!

severinbeauvais:
## Test results

### As BCREG0020/Severin Dev Account:

image

Note the 3 businesses that display as "affiliate to view". Those businesses are either affiliated to a different account under BCREG0020 or to a different login altogether.

### As a staff user:

image

Note that all businesses display as affiliated (even though they aren't) because this is a staff (IDIR) login.

PR #659, opened 2024-02-16
19756 Small change so prepopulated data overrides legal name

severinbeauvais:
*Issue #:* bcgov/entity#19756

*Description of changes:*
- app version = 5.9.4
- updated some code comments
- on draft resume, update prepopulated data after the legal name is restored (as we may override it)

### DIFF
--- a/src/components/common/Actions.vue
+++ b/src/components/common/Actions.vue
@@ -224,8 +224,9 @@ export default class Actions extends Mixins(AmalgamationMixin, CommonMixin,
         // Re-fetch the business table data. We do this in case one of the businesses has changed (eg,
         // became historical) while the draft was open.
         const holdingPrimary = await this.refetchAmalgamatingBusinessesInfo()
+
         // If there's a holding or primary business, re-fetch its data and update the prepopulated data.
-        // This will overwrite office addresses, directors and share structure.
+        // This will overwrite office addresses, directors, share structure, legal name and legal type.
         if (holdingPrimary) await this.updatePrepopulatedData(holdingPrimary)
       } catch (error) {
         console.log('Error validating table in onClickFilePay(): ', error) // eslint-disable-line no-console
@@ -233,6 +234,9 @@ export default class Actions extends Mixins(AmalgamationMixin, CommonMixin,
         return
       }
     }
+
+    // *** TODO: verify that filing is invalid if the business table now have invalid items
+    // or do we need to await this.$nextTick() here for things to validate?
 
     if (this.isFilingValid) {
       // prevent double saving
--- a/src/components/common/Actions.vue
+++ b/src/components/common/Actions.vue
@@ -235,9 +235,6 @@ export default class Actions extends Mixins(AmalgamationMixin, CommonMixin,
       }
     }
 
-    // *** TODO: verify that filing is invalid if the business table now have invalid items
-    // or do we need to await this.$nextTick() here for things to validate?
-
     if (this.isFilingValid) {
       // prevent double saving
       if (this.isBusySaving) return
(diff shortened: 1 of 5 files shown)


src/components/common/Actions.vue
```diff
@@ -234,6 +235,9 @@ export default class Actions extends Mixins(AmalgamationMixin, CommonMixin,
       }
     }
 
+    // *** TODO: verify that filing is invalid if the business table now have invalid items
+    // or do we need to await this.$nextTick() here for things to validate?
```
severinbeauvais, line 239: I will do this asap.

src/mixins/amalgamation-mixin.ts
```diff
@@ -429,7 +429,7 @@ export default class AmalgamationMixin extends Vue {
     }
 
     // set new resulting business name and legal type
-    // and update resources (since legal  type may have changed)
+    // and update resources (since legal type may have changed)
     this.setNameRequestApprovedName(business.name)
     this.setEntityType(business.legalType)
     this.updateResources()
```
severinbeauvais, line 435: This is `updatePrepopulatedData()` and this code overrides the previous legal name and legal type. This is called when:
- a draft amalgamation is restored (see filing-template-mixin.ts below)
- the user marks a different business as holding/primary
- the user is submitting this filing (see Actions.vue above)

src/components/common/Actions.vue
```diff
@@ -224,8 +224,9 @@ export default class Actions extends Mixins(AmalgamationMixin, CommonMixin,
         // Re-fetch the business table data. We do this in case one of the businesses has changed (eg,
         // became historical) while the draft was open.
         const holdingPrimary = await this.refetchAmalgamatingBusinessesInfo()
+
         // If there's a holding or primary business, re-fetch its data and update the prepopulated data.
-        // This will overwrite office addresses, directors and share structure.
+        // This will overwrite office addresses, directors, share structure, legal name and legal type.
         if (holdingPrimary) await this.updatePrepopulatedData(holdingPrimary)
```
severinbeauvais, line 230: This is called when the user submits an amalgamation filing. This is to catch the scenario where the holding/primary business name or type has changed in the background while this amalgamation remained open.

src/mixins/filing-template-mixin.ts
```diff
@@ -291,6 +277,22 @@ export default class FilingTemplateMixin extends Mixins(AmalgamationMixin, DateM
...
+
+      // Re-fetch the business table data. We do this in case one of the businesses has changed
+      // (eg, became historical) since the last draft save.
+      const holdingPrimary = await this.refetchAmalgamatingBusinessesInfo()
+
+      // If there's a holding or primary business, fetch its data and update the prepopulated data.
+      // This will overwrite office addresses, directors, share structure, legal name and legal type
+      // that were set above from draft data.
+      if (holdingPrimary) await this.updatePrepopulatedData(holdingPrimary)
+    }
```
severinbeauvais, line 294: This is called when a draft amalgamation is restored. This is to catch the scenario where the holding/primary business name or type has changed in the background and then this amalgamation is resumed.

This block of code was moved down (in particular, below the name request code) as it may overwrite the legal name and legal type set above.

severinbeauvais:
See ticket for Test Notes.

src/components/common/Actions.vue
severinbeauvais: I'm still testing this.

It's the scenario where the primary/holding business' entity type has changed and made the TING table invalid (eg, entity type mismatch).

vysakh-menon-aot (approved):

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-659-hvz9hxbh.web.app

PaulGarewal (approved):
LGTM 👍

PR #661, opened 2024-02-20
19897 Fixed foreign jurisdiction region

severinbeauvais:
*Issue #:* bcgov/entity#19897

*Description of changes:*
- app version = 5.9.6
- save region value instead of text (x2)
- fixed unit test

### DIFF
--- a/src/components/Amalgamation/AmalgamatingBusinesses.vue
+++ b/src/components/Amalgamation/AmalgamatingBusinesses.vue
@@ -277,7 +277,7 @@ import { AmalgamationMixin, CommonMixin } from '@/mixins'
 import { BusinessLookupServices } from '@/services'
 import { BusinessLookup } from '@bcrs-shared-components/business-lookup'
 import { Jurisdiction } from '@bcrs-shared-components/jurisdiction'
-import { CanJurisdictions, MrasJurisdictions } from '@bcrs-shared-components/jurisdiction/list-data'
+import { MrasJurisdictions } from '@bcrs-shared-components/jurisdiction/list-data'
 import { AmalgamatingBusinessIF, BusinessLookupResultIF, EmptyBusinessLookup } from '@/interfaces'
 import { AmlRoles, AmlTypes, EntityStates } from '@/enums'
 import { JurisdictionLocation } from '@bcrs-shared-components/enums'
@@ -316,8 +316,8 @@ export default class AmalgamatingBusinesses extends Mixins(AmalgamationMixin, Co
   snackbar = false
   snackbarText = ''
   errorDialog = false
-  errorDialogText = undefined as string
-  errorDialogTitle = undefined as string
+  errorDialogText = ''
+  errorDialogTitle = ''
 
   // Foreign business properties
   jurisdiction = null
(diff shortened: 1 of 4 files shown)


src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -316,8 +316,8 @@ export default class AmalgamatingBusinesses extends Mixins(AmalgamationMixin, Co
   snackbar = false
   snackbarText = ''
   errorDialog = false
-  errorDialogText = undefined as string
-  errorDialogTitle = undefined as string
```
severinbeauvais, line 320: This was actually causing some template errors. I originally had it this way so the default props would be used if we didn't specify them, but we always specify them anyway so '' is a better initial value.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -418,12 +418,11 @@ export default class AmalgamatingBusinesses extends Mixins(AmalgamationMixin, Co
 
     // Special case to handle Extra Pro A companies
     if ((businessLookup.legalType as any) === CorpTypeCd.EXTRA_PRO_A) {
-      const region = CanJurisdictions.find(jurisdiction => jurisdiction.value === JurisdictionLocation.BC)
       const tingBusiness = {
         type: AmlTypes.FOREIGN,
         role: AmlRoles.AMALGAMATING,
         foreignJurisdiction: {
-          region: region.text,
+          region: JurisdictionLocation.BC,
```
severinbeauvais, line 425: Before:
image

After:
image

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -549,7 +548,7 @@ export default class AmalgamatingBusinesses extends Mixins(AmalgamationMixin, Co
       type: AmlTypes.FOREIGN,
       role: AmlRoles.AMALGAMATING,
       foreignJurisdiction: {
-        region: this.isCan ? this.jurisdiction.text : '',
+        region: this.isCan ? this.jurisdiction.value : '', // no region outside Canada
```
severinbeauvais, line 551: Before:
image

After:
image

tests/unit/AmalgamatingBusinesses.spec.ts
```diff
@@ -638,7 +638,7 @@ describe('Amalgamating Businesses - add amalgamating foreign business', () => {
     // simulate form data
     await wrapper.setData({
       isCan: true,
-      jurisdiction: { text: 'BC', value: 'CA' },
+      jurisdiction: { text: 'Alberta', value: 'AB' },
```
severinbeauvais, line 641: This is what `jurisdiction` actually looks like when it's selected.

tests/unit/AmalgamatingBusinesses.spec.ts
```diff
@@ -654,7 +654,7 @@ describe('Amalgamating Businesses - add amalgamating foreign business', () => {
     const business = store.getAmalgamatingBusinesses[0] as any
     expect(business.type).toBe(AmlTypes.FOREIGN)
     expect(business.role).toBe(AmlRoles.AMALGAMATING)
-    expect(business.foreignJurisdiction).toEqual({ country: 'CA', region: 'BC' })
+    expect(business.foreignJurisdiction).toEqual({ country: 'CA', region: 'AB' })
```
severinbeauvais, line 657: I changed this from BC to AB since BC isn't "foreign" :)

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-661-wghir2e4.web.app

src/components/Amalgamation/AmalgamatingBusinesses.vue
severinbeauvais: (From a high level, default prop values are normally used when you DON'T specify the prop at all. If we do specify the prop, but want to use its default value, then we'll have to set it to undefined. Neither of these applies here.)

PaulGarewal (approved):

PR #674, opened 2024-04-16
20489 Misc cleanup

severinbeauvais:
*Issue #:* bcgov/entity#20489

*Description of changes:*
- app version = 5.9.11
- saved correct name option for other filings (not used but denotes code intent for future debugging/audit)
- updated unit tests

### DIFF
--- a/src/mixins/filing-template-mixin.ts
+++ b/src/mixins/filing-template-mixin.ts
@@ -384,6 +384,7 @@ export default class FilingTemplateMixin extends Mixins(AmalgamationMixin, DateM
     switch (this.getCorrectNameOption) {
       case CorrectNameOptions.CORRECT_AML_ADOPT:
         // save adopted name
+        // *** FUTURE: is there really an adopted name option in a continuation in?
         filing.continuationIn.nameRequest.correctNameOption = CorrectNameOptions.CORRECT_AML_ADOPT
         filing.continuationIn.nameRequest.legalName = this.getNameRequestApprovedName
         break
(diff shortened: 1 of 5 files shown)


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-674-dprzede5.web.app

JazzarKarim (approved):
There's a mismatch between ticket name and PR 😆

LGTM Sev! 👍

src/mixins/filing-template-mixin.ts
severinbeauvais, line 1: Although the subject ticket is not for work in Create UI, I'm trying to make Edit UI similar to Create UI, and this change is to clean up / make consistent some Create UI code.

Specifically, the changes here are to be intentional in the "correct name option" that we are saving in a draft filing, even though these particular filings don't restore this value. Basically, I don't like that we have to infer what option the user has selected by looking at which properties exist or not.

Amalgamations, continuation ins (WIP) and restorations do restore the correct name option value. I think this is a better design.

severinbeauvais:
> There's a mismatch between ticket name and PR 😆

Thanks for checking that. Yes, here I go back to Create UI to clean up something I noticed while comparing with other filings and with Edit UI.

severinbeauvais:
PS There's still an issue uploading coverage to CodeCov but the tests pass fine locally and even here.

Locally:
image

The "Test with Vitest" step passed:
image

JazzarKarim:
> PS There's still an issue uploading coverage to CodeCov but the tests pass fine locally and even here.
> 
Yup, been having this issue as well in my PRs. All good.

ketaki-deodhar (approved):
Consistent code across UIs will make it easier for us to work. Thank you for this. Looks good Sev!

PR #679, opened 2024-04-25
19264 Partial implementation of Existing Business Information

severinbeauvais:
*Issue #:* bcgov/entity#19264

*Description of changes:*

- app version = 5.10.3
- imported some updated shared components
- updated Vue to 2.7.16
- added (but disabled) staff-only check for Continuation In and Restoration filings
- fixed some whitespace
- add step 1 validity code
- created ExtraproBusinessLookup sub-component (similar to but different from other business lookup)
- created ExtraproRegistration sub-component
- created initial ManualBusinessInfo sub-component
- created SummaryBusinessHomeJurisdiction sub-component
- updated continuation in filing interface
- updated continuation in state interface (for store)
- updated continuation in save and resume code
- updated entity types in continuation in resource files
- renamed continuation in resource files according to entity type
- updated business search to support status parameter
- added some store getters and actions
- updated state model to match state interface
- removed an unneeded spinner
- updated Cont In first step layout, validation, etc
- updated Cont In last step
- misc cleanup


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-679-8i77nbez.web.app

package.json
```diff
@@ -58,7 +58,7 @@
     "regenerator-runtime": "^0.13.11",
     "sbc-common-components": "3.0.8",
     "uuid": "^9.0.0",
-    "vue": "2.7.14",
+    "vue": "2.7.16",
```
severinbeauvais, line 61: Until I updated this, all the unit tests failed (except 3), probably due to a package dependency.

src/App.vue
```diff
@@ -722,8 +723,24 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
...
       }
 
-      // FUTURE: since we know what type of filing this is,
-      // add staff check for certain filings (ie, restorations)
+      // get Keycloak roles
+      const keycloakRoles = await this.loadKeycloakRoles().catch(error => {
+        console.log('Keycloak roles error =', error) // eslint-disable-line no-console
+        this.accountAuthorizationDialog = true
+        throw error // go to catch()
+      })
```
severinbeauvais, line 731: Need to get roles before we can check if this is a staff user.

I moved it here, above getting the user info, because we don't need to fetch the user info if the staff check fails.

Currently, the staff check is disabled. I am not sure if Continuation In is for staff only or regular users. I will ask.

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -131,6 +131,7 @@
 
             <BusinessLookup
               :key="Math.random()"
+              class="mt-5"
```
severinbeauvais, line 134: The latest version of the shared component removed the whitespace from the component itself, so that the parent (here) can specify what it needs.

src/components/ContinuationIn/ExtraproBusinessLookup.vue
```diff
@@ -0,0 +1,172 @@
...
+  SHOW_RESULTS = 'show results',
+  NO_RESULTS = 'no results',
+  SUMMARY = 'summary'
+}
+/* eslint-enable no-unused-vars */
+
+//
+// This component is a hybrid between the shared BusinessLookup component
+// and the Auth Web BusinessLookup component.
+//
```
severinbeauvais, line 83: Unfortunately, I couldn't use the shared component here (since it also handles the summary display and has hard-coded values).

Also unfortunately, I found a local copy of this component in Auth Web. However, we might be able to reuse this one...

Some time in the future, a good plan might be to break up the shared component into a search-only sub-component, and a wrapper that would also handle the summary display. I think here, and in Auth Web, we could use just the search-only component (and do our own display). Meanwhile, other users of the shared component could use the wrapper that does both. In this way we would have maximum flexibility and reduced code duplication.

src/components/ContinuationIn/ExtraproRegistration.vue
```diff
@@ -0,0 +1,331 @@
...
+  get haveLookupBusiness (): boolean {
+    return !isEqual(this.business, EmptyBusinessLookup)
+  }
+
+  /** The identifier. */
+  get identifier (): string {
+    return this.business.identifier
+  }
+
+  /** The jurisdiction. */
```
severinbeauvais, line 228: For now I'm calling the object `homeJurisdiction`, but I'm aware that COLIN calls this "foreign jurisdiction" (and also contains some extra properties).

This is working currently. We can easily change it later if needed. (Agile ftw!)

src/components/ContinuationIn/ExtraproRegistration.vue
```diff
@@ -0,0 +1,331 @@
...
+  /** The business number. */
+  get businessNumber (): string {
+    return this.business.bn
+  }
+
+  /** The incorporation date. */
+  get incorporationDate (): string {
+    // FUTURE: format date if needed
+    return this.business.registrationDate
+  }
```
severinbeauvais, line 254: I have **no idea** what the registration date will look like from business search.

An example I got from David Roberts looks like this (in COLIN):

image

src/components/ContinuationIn/ExtraproRegistration.vue
```diff
@@ -0,0 +1,331 @@
...
+
+  /** Whether business is Active (otherwise it's Historical). */
+  get isBusinessActive (): boolean {
+    return (this.business.status === 'ACTIVE')
```
severinbeauvais, line 258: This is the only status we care about (as I assume the only other status is HISTORICAL), so probably no need to create an enum.

Later, if we get a definitive list of possible statuses, we can create an enum.

src/components/ContinuationIn/ExtraproRegistration.vue
```diff
@@ -0,0 +1,331 @@
...
+    // get the business info object from the store or initialize it
+    this.business = this.getContinuationInBusinessInfo || { ...EmptyBusinessLookup }
+
+    // if mode is LOOKUP, set this component to active (which hides the other component)
+    if (this.business.mode === 'LOOKUP') this.active = true
+  }
+
+  setBusiness (business: BusinessLookupResultIF) {
+    this.business = { ...business } // for reactivity
+    this.business.mode = 'LOOKUP'
```
severinbeauvais, line 271: I'll create an enum for this mode (which is either LOOKUP or MANUAL). It is used in a few places so using an enum would make this code self-documenting.

src/components/ContinuationIn/ManualBusinessInfo.vue
severinbeauvais, line 1: This component is ticket bcgov/entity#20989 but I needed to create this framework so the 2 Existing Business Information sub-components could play nicely together (including the Undo functionality).

src/components/ContinuationIn/SummaryBusinessHomeJurisdiction.vue
severinbeauvais, line 1: This is sort-of equivalent to SummaryDefineCompany except is specific to the Continuation In first step/page.

src/components/ContinuationIn/SummaryBusinessHomeJurisdiction.vue
```diff
@@ -0,0 +1,341 @@
...
+            >
+              <label>Home Jurisdiction</label>
+            </v-col>
+            <v-col
+              cols="12"
+              sm="9"
+              class="pt-4 pt-sm-0"
+            >
+              <div id="home-jurisdiction">
+                {{ homeJurisdiction || '[Unknown]' }}
```
severinbeauvais, line 35: Note that the fallback values are in the template here. I think this is a good general practice as it allows the template to do logic if the value is null.

src/components/ContinuationIn/SummaryBusinessHomeJurisdiction.vue
```diff
@@ -0,0 +1,341 @@
...
+      const country = getName(hj.country)
+      const region = (hj.region === 'FEDERAL' ? 'Federal' : hj.region)
+      if (region) return `${region}, ${country}`
+      return country
+    }
+    return null
+  }
+
+  get identifyingNumber (): string {
+    return this.getContinuationInBusinessInfo?.homeIdentifier
```
severinbeauvais, line 277: We don't prompt for this yet. I'm trying to sort it out with Yui and it will be handled in a future ticket.

src/components/ContinuationIn/SummaryBusinessHomeJurisdiction.vue
```diff
@@ -0,0 +1,341 @@
...
+  }
+
+  get businessName (): string {
+    return this.getContinuationInBusinessInfo?.homeName
```
severinbeauvais, line 281: We don't have this yet, either.

src/interfaces/filing-interfaces/filing-interfaces.ts
```diff
@@ -87,18 +87,15 @@ export interface ContinuationInFilingIF {
...
     contactPoint: ContactPointIF
     parties: PartyIF[]
-
-    // BEN / CC / BC / ULC only:
-    shareStructure?: {
-      shareClasses: ShareClassIF[]
-    }
-    // ULC only:
+    shareStructure: { shareClasses: ShareClassIF[] }
+    // ULC only: // *** TODO: verify this
```
severinbeauvais, line 98: Future ticket -- waiting for response from Yui on whether this is even needed.

src/interfaces/filing-interfaces/filing-interfaces.ts
```diff
@@ -87,18 +87,15 @@ export interface ContinuationInFilingIF {
     identifier: string
   }
   continuationIn: {
-    foreignJurisdiction: any
+    existingBusinessInfo: any // FUTURE: create an interface for this object
+    continuationAuthorization: any // FUTURE: create an interface for this object
```
severinbeauvais, line 91: We need to figure out what else goes into each of these object, the schema, etc. But not in this ticket.

src/interfaces/store-interfaces/state-interfaces/continuation-in-state-interface.ts
```diff
@@ -1,4 +1,6 @@
 // TODO: Add properties as needed
 export interface ContinuationInStateIF {
-  existingBusinessValid: boolean
+  step1Valid: boolean
+  step2Valid: boolean
```
severinbeauvais, line 4: I might rename these in a future ticket, I dunno.

The alternative is `yourBusinessInHomeJurisdictionValid` :/

Or maybe just `businessHomeValid` (which is consistent with the route and view names.

severinbeauvais: Never mind, I just changed it.

src/components/ContinuationIn/ManualBusinessInfo.vue
severinbeauvais, line 1: This component is ticket bcgov/entity#20989 but I needed to create this framework so the 2 Existing Business Information sub-components could play nicely together (including the Undo functionality).

src/mixins/filing-template-mixin.ts
```diff
@@ -463,19 +473,17 @@ export default class FilingTemplateMixin extends Mixins(AmalgamationMixin, DateM
         ...draftFiling.continuationIn.contactPoint,
         confirmEmail: draftFiling.continuationIn.contactPoint.email
       })
-    } else {
-      this.setBusinessContact({ ...EmptyContactPoint })
```
severinbeauvais, line 467: If contact point is not in the draft, we should be able to leave the initial value in the store (which is EmptyContactPoint).

src/mixins/filing-template-mixin.ts
```diff
@@ -463,19 +473,17 @@ export default class FilingTemplateMixin extends Mixins(AmalgamationMixin, DateM
...
     }
 
     // restore Persons and Organizations
     if (draftFiling.continuationIn.parties) {
-      this.setOrgPersonList(draftFiling.continuationIn.parties || [])
```
severinbeauvais, line 472: If `parties` is not falsy (on line 471) then this fallback will never be used.

src/mixins/filing-template-mixin.ts
```diff
@@ -463,19 +473,17 @@ export default class FilingTemplateMixin extends Mixins(AmalgamationMixin, DateM
...
+      this.setOrgPersonList(draftFiling.continuationIn.parties)
     }
 
     // restore Share Structure
-    this.setShareClasses(draftFiling.continuationIn.shareStructure
-      ? draftFiling.continuationIn.shareStructure.shareClasses
-      : [])
+    if (draftFiling.continuationIn.shareStructure?.shareClasses) {
+      this.setShareClasses(draftFiling.continuationIn.shareStructure.shareClasses)
+    }
```
severinbeauvais, line 486: If share classes is not in the draft, we should be able to leave the initial value in the store (which is `[]`).

src/store/store.ts
```diff
@@ -100,7 +101,7 @@ export const useStore = defineStore('store', {
 
     /** Whether the current filing is an Amalgamation Application. */
     isAmalgamationFiling (): boolean {
-      return (this.stateModel.tombstone.filingType === FilingTypes.AMALGAMATION_APPLICATION)
+      return (this.getFilingType === FilingTypes.AMALGAMATION_APPLICATION)
```
severinbeauvais, line 104: I prefer this because it means fewer detailed de-references of the stateModel object, and I think the value is cached this way.

Also, if we ever want to change where the data is stored, we only have to change 1 getter instead of many (and 1 action).

src/store/store.ts
```diff
@@ -196,6 +197,10 @@ export const useStore = defineStore('store', {
       return this.stateModel.entityType
     },
 
+    getContinuationInBusinessInfo (): any {
+      return this.stateModel.continuationIn.existingBusinessInfo
```
severinbeauvais, line 201: Broke my own rule! Fixing it now...

src/utils/feature-flag-utils.ts
```diff
@@ -13,9 +13,7 @@ const defaultFlagSet: LDFlagSet = {
   'enable-web-chat': false, // by default, old webchat is disabled
   'enable-genesys-web-message': false, // by default, genesys web message is disabled
   'sentry-enable': false, // by default, no sentry logs
-  'banner-text': '', // by default, there is no banner text
-  'supported-amalgamation-entities': []
```
severinbeauvais, line 17: Obsolete.

src/views/ContinuationIn/ContinuationInBusinessHome.vue
```diff
@@ -1,46 +1,58 @@
...
-            color="primary"
-            size="50"
-            indeterminate
-          />
-          <div class="loading-msg white--text">
-            Fetching data
-          </div>
-        </div>
-      </div>
-    </v-fade-transition>
```
severinbeauvais, line 19: Wasn't used.

src/views/ContinuationIn/ContinuationInBusinessHome.vue
```diff
@@ -1,46 +1,58 @@
...
+        <div :class="{ 'mt-6': !extraproRegistrationActive && !manualBusinessInfoActive }" />
+
+        <v-expand-transition>
+          <ManualBusinessInfo
+            v-if="!extraproRegistrationActive"
+            @active="manualBusinessInfoActive = $event"
+            @valid="manualBusinessInfoValid = $event"
+          />
+        </v-expand-transition>
+      </v-card>
```
severinbeauvais, line 36: The above is a good example of this component knowing as little as possible about the sub-components. And the sub-components don't need to know about each other.

src/views/Dissolution/DissolutionReviewConfirm.vue
```diff
@@ -39,7 +39,9 @@
               <span class="error-text mx-1">This step is unfinished.</span>
               <router-link
                 :to="{ path: `/${RouteNames.DISSOLUTION_DEFINE_DISSOLUTION}` }"
-              >Return to this step to finish it</router-link>
```
severinbeauvais, line 42: VS Code complained that this should be on a separate line.

However, that means the text has a leading and trailing carriage return (which is typically ignored by the renderer/browser), shich is a pain to match in unit tests. By wrapping the text in a span, unit testing is easier.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-679-8i77nbez.web.app

src/components/ContinuationIn/ExtraproRegistration.vue
```diff
@@ -0,0 +1,331 @@
...
+  // *** TODO: add watchers for jurisdiction and affidavit
+  @Watch('business')
+  @Watch('checkbox')
+  @Emit('valid')
+  private onComponentValid (): boolean {
+    // this component is active if we have an active business
+    return (this.haveLookupBusiness && this.isBusinessActive && this.checkbox)
+  }
+
+  @Watch('active', { immediate: true })
```
JazzarKarim, line 294: This forces the watcher here to be executed immediately on load right?

vysakh-menon-aot (approved):

src/components/ContinuationIn/ManualBusinessInfo.vue
```diff
@@ -0,0 +1,196 @@
...
+        </v-col>
+
+        <v-col
+          cols="12"
+          sm="9"
+        >
+          <div class="font-14 ml-3">
+            Not extraprovincially registered in B.C.?
+            &nbsp; &nbsp;
+            <a @click="onClick()">Enter your business information manually</a>
```
JazzarKarim, line 20: What do you think of changing this function's name to something more specific Sev? Like `onNotExtraProClick`. Something like that.

severinbeauvais: I only implemented this sub-component to get all the interations worked out. There is another ticket to actually implement this component, and its developer can rename this as they see fit.

This is actually a pretty simple component at the moment, so I think "onClick" is straightforward and not confusing 😁 

I wouldn't label this event handler according to the other sub-component. Best for components to know as little as possible about the world outside of them.

A better method name might be "setActive()". Thoughts?

src/store/store.ts
```diff
@@ -196,6 +197,10 @@ export const useStore = defineStore('store', {
       return this.stateModel.entityType
     },
 
+    getContinuationInBusinessInfo (): any {
```
JazzarKarim, line 200: I'm guessing we might be able to create an interface for this later on when all the requirements are done right Sev?

src/components/ContinuationIn/ManualBusinessInfo.vue
JazzarKarim: I see, OK Sev. That's fine, I think let's keep as is (`onClick()`). Thanks!

package.json
JazzarKarim: That's weird though, I don't see any game breaking package changes here that rely on `2.7.16` 😕

JazzarKarim (approved):
LGTM Sev!

src/store/store.ts
severinbeauvais: Exactly. I left a comment in continuation-in-state-interface.ts about this.

package.json
JazzarKarim: Unit tests are passing on my side, that's why I'm confused. Anyway, sounds good.

src/App.vue
severinbeauvais: This is still an outstanding question so I'll leave the code commented out for now.

src/components/ContinuationIn/ExtraproRegistration.vue
severinbeauvais: I think this is a neat way for components to "check in" and report their initial state to the parent (so the parent doesn't have to assume anything).

package.json
severinbeauvais: Did you do a `npm i` (or `npm ci`) first?

src/components/ContinuationIn/ExtraproRegistration.vue
JazzarKarim: Makes sense, OK. Thanks Sev!

package.json
JazzarKarim: Never mind, I cloned your branch and I see the unit tests failing. Weird though as before, things were looking fine.

severinbeauvais:
Merging now. If there are any more comments then I will address them in bcgov/entity#20889.

PR #680, opened 2024-05-06
Renamed script to prevent confusion

severinbeauvais:
*Issue #:* None

*Description of changes:*
- renamed script "serve" to "preview" to prevent confusion with other projects
- removed port specification so that it runs on a different port than "npm run dev" which should also prevent confusion

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the business-filings-ui license (Apache 2.0).

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "dev": "vite",
     "build": "vite build",
-    "serve": "vite preview --port 8080",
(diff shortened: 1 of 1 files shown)


JazzarKarim (approved):
👍 👍 👍

seeker25 (approved):

PR #683, opened 2024-05-13
19265 Updated file upload validation, etc

severinbeauvais:
*Issue #:* bcgov/entity#19265

*Description of changes:*
- app version = 5.10.6
- updated file upload custom error handling to sync with parent if needed
- simplified file upload validation flow
- added handling for failure to retrieve file info (invalid PDF)
- updated a blurb per Yui
- show "File uploaded" instead of default hint for loaded files
- show custom errors on specific upload instance only
- updated date rules/validation
- updated fallback message if null expiry date
- misc cleanup and comments
- combined/updated unit tests

### DIFF
--- a/src/components/ContinuationIn/ContinuationAuthorization.vue
+++ b/src/components/ContinuationIn/ContinuationAuthorization.vue
@@ -9,7 +9,7 @@
   >
     <header>
       <p>
-        Upload the authorization for the continuation from the foreign jursidiction.
+        Upload the continuation authorization letter from your home jursidiction.
       </p>
 
       <ul>
@@ -45,9 +45,10 @@
           inputFileLabel="Continuation authorization"
           :maxSize="MAX_FILE_SIZE"
           :pdfPageSize="PdfPageSize.LETTER_SIZE"
+          :hint="authorization.files[index]?.file ? 'File uploaded.' : undefined"
           :inputFile="authorization.files[index]?.file"
           :showErrors="getShowErrors"
-          :customErrorMessage="fileUploadCustomErrorMsg"
+          :customErrorMessage.sync="customErrorMessage[index]"
           :isRequired="getShowErrors && (index === 0)"
           @fileValidity="onFileValidity($event)"
           @fileSelected="onFileSelected(index, $event)"
@@ -139,7 +140,7 @@
             :nudgeTop="85"
             :initialValue="authorization.expiryDate"
             :inputRules="getShowErrors ? expiryDateRules: []"
-            :minDate="authorization.date"
+            :minDate="getCurrentDate"
             @emitDateSync="authorization.expiryDate = $event"
           />
         </v-col>
@@ -153,7 +154,8 @@
         :rules="getShowErrors ? [(v) => !!v] : []"
       >
         <template #label>
-          <span>I confirm that I have current and valid authorization to continue this business into B.C.</span>
+          <span>I confirm that I have current and valid authorization to continue this business into
+            B.C.</span>
         </template>
       </v-checkbox>
     </v-form>
@@ -218,7 +220,7 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
   // Local properties
   authorization = null as ContinuationAuthorizationIF
   fileValidity = false
-  fileUploadCustomErrorMsg = ''
+  customErrorMessage = ['', '', '', '', '']
   snackbar = false
   snackbarText = ''
 
@@ -230,17 +232,20 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
 
   get authorizationDateRules (): Array<VuetifyRuleFunction> {
     return [
-      (v) => !!v || 'Authorization Date is required',
-      () => (this.authorization.date <= this.getCurrentDate) ||
-        'Authorization Date cannot be in the future'
+      () => !!this.authorization.date ||
+        'Authorization Date is required',
+      () => {
+        return (this.authorization.date <= this.getCurrentDate) ||
+          'Authorization Date cannot be in the future'
+      }
     ]
   }
 
   get expiryDateRules (): Array<VuetifyRuleFunction> {
     return [
       () => !this.authorization.expiryDate ||
-        (this.authorization.expiryDate >= this.authorization.date) ||
-        'Expiry Date cannot be before Authorization Date'
+        (this.authorization.expiryDate >= this.getCurrentDate) ||
+        'Expiry Date cannot be in the past'
     ]
   }
 
@@ -267,12 +272,10 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
    * This is called right after the File Validity event.
    */
   async onFileSelected (index: number, file: File): Promise<void> {
-    // reset state of file uploader to ensure not in manual error mode
-    this.fileUploadCustomErrorMsg = ''
-
     if (file) {
       // verify that file is valid
       if (!this.fileValidity) {
+        // NB: as this is validity according to the component, do not overwrite current error message
         this.snackbarText = 'Invalid file. It will not be saved.'
         this.snackbar = true
         return // don't add to list
@@ -280,12 +283,12 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
 
       // verify that file doesn't already exist
       if (this.authorization.files.find(f => f.file.name === file.name)) {
-        // put file uploader into manual error mode by passing custom error message
-        this.fileUploadCustomErrorMsg = 'Duplicate file.'
+        // put file uploader into manual error mode by setting custom error message
+        this.customErrorMessage[index] = 'Duplicate file.'
 
         this.snackbarText = 'File already uploaded. It will not be saved again.'
         this.snackbar = true
-        return
+        return // don't add to list
       }
 
       // try to upload to Minio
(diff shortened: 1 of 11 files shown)


src/components/ContinuationIn/ContinuationAuthorization.vue
```diff
@@ -45,9 +45,10 @@
           inputFileLabel="Continuation authorization"
           :maxSize="MAX_FILE_SIZE"
           :pdfPageSize="PdfPageSize.LETTER_SIZE"
+          :hint="authorization.files[index]?.file ? 'File uploaded.' : undefined"
```
severinbeauvais, line 48: Shows hint only if file is defined. Screenshot:

image

src/components/ContinuationIn/ContinuationAuthorization.vue
```diff
@@ -45,9 +45,10 @@
...
           :inputFile="authorization.files[index]?.file"
           :showErrors="getShowErrors"
-          :customErrorMessage="fileUploadCustomErrorMsg"
+          :customErrorMessage.sync="customErrorMessage[index]"
```
severinbeauvais, line 51: Two changes here:
- allows for sync from child back to parent (doesn't have to be used, which retains compatibility for other filings that use FileUploadPreview)
- supports different error messages for every instance of this component (up to 5)

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-683-dorjm5mg.web.app

src/components/ContinuationIn/ContinuationAuthorization.vue
```diff
@@ -139,7 +140,7 @@
             :nudgeTop="85"
             :initialValue="authorization.expiryDate"
             :inputRules="getShowErrors ? expiryDateRules: []"
-            :minDate="authorization.date"
+            :minDate="getCurrentDate"
```
severinbeauvais, line 143: Minimum expiry date is today.

src/components/ContinuationIn/ContinuationAuthorization.vue
```diff
@@ -267,25 +272,23 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
    * This is called right after the File Validity event.
    */
   async onFileSelected (index: number, file: File): Promise<void> {
-    // reset state of file uploader to ensure not in manual error mode
-    this.fileUploadCustomErrorMsg = ''
```
severinbeauvais, line 271: In the old code, this overwrote the error message set by the file upload component :(

src/components/ContinuationIn/SummaryBusinessHomeJurisdiction.vue
```diff
@@ -269,7 +269,7 @@
               class="pt-4 pt-sm-0"
             >
               <div id="expiry-date">
-                {{ expiryDate || '[Unknown]' }}
+                {{ expiryDate || '[Not entered]' }}
```
severinbeauvais, line 272: Screenshot:

image

src/components/common/FileUploadPreview.vue
```diff
@@ -83,29 +83,38 @@ export default class FileUploadPreview extends Mixins(DocumentMixin) {
   //  is using a different validation approach or updating all components to use vuelidate.  Have decided
   //  to do this for the time being.
   private async validateFileInput (file: File): Promise<boolean> {
-    this.customErrorMessages = []
+    // clear any existing error messages
+    this.setCustomErrorMessage(null)
```
severinbeauvais, line 87: The old code set the custom error message here and via a watcher. However, the watcher only fires if the new prop value is different than the previous one, and that caused errors to not display if, for example, after selecting a duplicate file, you then selected an invalid one.

Now, there is still a watcher on the prop, but also this way to update the local variable.

src/components/common/FileUploadPreview.vue
```diff
@@ -137,9 +146,13 @@ export default class FileUploadPreview extends Mixins(DocumentMixin) {
...
 
-  /** Sets custom error message if is changed from its initial value. */
+  /**
+   * Automatically sets custom error message if prop changed, or if called, and syncs value back
+   * to the parent component.
+   */
   @Watch('customErrorMessage')
-  private onCustomErrorMessage (val: string): void {
+  @Emit('update:customErrorMessage')
+  private setCustomErrorMessage (val: string): void {
```
severinbeauvais, line 155: This updates the parent component (via `sync`, if used), so that the parent can set a new value if needed. This allows setting errors above or from the parent.

(There's actually a 3rd place that errors are generated: the rules. But what I've done above is fairly clean and handles all cases.)

src/mixins/document-mixin.ts
```diff
@@ -104,7 +111,7 @@ export default class DocumentMixin extends Vue {
       if ((error as any).name === 'PasswordException') {
         return { isEncrypted: true, isContentLocked: true }
       }
+      return null // invalid pdf file
```
severinbeauvais, line 114: Previously, if there was an exception in the try block, this method returned a valid object. Now, it returns a null object, which the parent can check to see if there was an error.

(This seemed simpler than re-throwing the error and having another try-catch block (or then/catch) in the parent.)

src/services/legal-services.ts
```diff
@@ -38,7 +38,7 @@ export default class LegalServices {
    * Fetches the first or only filing.
    * This is probably a draft Amalgamation, IA or Registration.
    * @param tempId the temp registration number
-   * @returns a promise to return the draft filing, else exception
```
severinbeauvais, line 41: I think we can generally assume that most methods can throw an exception, so no need to call it out.

tests/unit/FileUploadPreview.spec.ts
```diff
@@ -271,24 +271,10 @@ describe('FileUploadPreview component', () => {
...
-      vuetify
-    })
 
-    const fileInput = wrapper.find('input[type="file"]')
-    setupFileInput(fileInput)
-    inputValue = oneMBFile.name
-    inputFilesGet.mockReturnValue([oneMBFile])
-    await fileInput.trigger('change')
-    await waitForUpdate(3)
-    expect(wrapper.emitted('fileValidity').pop()[0]).toEqual(true)
```
severinbeauvais, line 291: Due to new check for valid PDF structure, this now emitted False.

Since this test was otherwise identical to the one above, I combined them.

src/components/ContinuationIn/ContinuationAuthorization.vue
```diff
@@ -218,7 +220,7 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
   // Local properties
   authorization = null as ContinuationAuthorizationIF
   fileValidity = false
-  fileUploadCustomErrorMsg = ''
+  customErrorMessage = ['', '', '', '', '']
```
JazzarKarim, line 223: A cool way would also be `Array(5).fill('')` 😆 

Up to you Sev. No need to change anything.

src/components/ContinuationIn/ContinuationAuthorization.vue
JazzarKarim: Yup, and eventually we're going to have to move away from this as it's deprecated in Vue 3 if I recall. A problem for another day.

JazzarKarim (approved):
Looks great Sev

vysakh-menon-aot (approved):

src/components/ContinuationIn/ContinuationAuthorization.vue
severinbeauvais: Good point. I could also use a constant for the number of allowed file uploads. If I'm in there again I might update it :)

src/components/ContinuationIn/ContinuationAuthorization.vue
severinbeauvais: I think we will be able to use a v-model for this in Vue3.

src/components/ContinuationIn/ContinuationAuthorization.vue
```diff
@@ -230,17 +232,20 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
...
     return [
-      (v) => !!v || 'Authorization Date is required',
-      () => (this.authorization.date <= this.getCurrentDate) ||
-        'Authorization Date cannot be in the future'
+      () => !!this.authorization.date ||
+        'Authorization Date is required',
+      () => {
+        return (this.authorization.date <= this.getCurrentDate) ||
+          'Authorization Date cannot be in the future'
+      }
```
severinbeauvais, line 240: Just a quick note that I used the local variable reference instead of `(v)` for the rule functions above because the latter is a formatted string like "May 14, 2024", which I cannot use for comparisons, while the former is a date string like "2024-05-14". This is a peculiarity of the Vuetify date picker and/or our wrapper around it.

src/components/common/FileUploadPreview.vue
```diff
@@ -83,29 +83,38 @@ export default class FileUploadPreview extends Mixins(DocumentMixin) {
...
   //  to do this for the time being.
   private async validateFileInput (file: File): Promise<boolean> {
-    this.customErrorMessages = []
+    // clear any existing error messages
+    this.setCustomErrorMessage(null)
+
+    // validate the form (using rules)
     const isValid = this.$refs.form.validate()
-    // only perform page size validation when other validation has passed
-    if (isValid && file) {
```
severinbeauvais, line 89: If `isValid` was false then the code fell through to line 108 and returned False. So I just simplified this.

ketaki-deodhar (approved):
Looks good!

PR #688, opened 2024-05-15
19266 - Plugged in Company Name Component

ArwenQin:
*Issue #:* /bcgov/entity#19266

*Description of changes:*
- Added the Company Name component
- Validated NR type/state/etc
- Verified support for named or numbered continuation in
- Verified Save and Resume Later


src/views/ContinuationIn/ContinuationInBusinessBc.vue
```diff
@@ -6,7 +6,8 @@
         <h2>Company Name</h2>
       </header>
 
-      <div>** Correct Name component goes here **</div>
+      <NameRequestInfo />
+      <NameTranslations />
```
ketaki-deodhar, line 10: Just porting over Sev's comments on other PR:

Our style is to use the PascalCase version of the component names here, so that VS Code highlights them nicely (and sometimes you can click on them to go to their definition).

Also, don't add divs around them if not needed.

ketaki-deodhar:
Sev' comment on other PR

Before your PR is ready for review...

update the app version in the package.json file (you can update the patch version for this PR)
run npm i and also commit the package-lock.json file
run npm run lint and fix any new issues
run npm run test:unit and fix any issues (there should be no previous issues)
in a PR comment, enter /gcbrun, which builds a preview site (URL) with your changes that reviews can run and test with
change this PR to "Ready for review"

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-688-fyq5ge7j.web.app

src/views/ContinuationIn/ContinuationInBusinessBc.vue
severinbeauvais, line 10: These need to be in a v-card so that they match the Figma.

Figma:
image

Actual:
image

severinbeauvais (commented):
Have you tested this with a _named_ continuation in application (ie, one that uses a NR to set a company name)?

src/views/ContinuationIn/ContinuationInBusinessBc.vue
ArwenQin: Thank you. Updated:
image

ArwenQin: Yes, I tested with a named company:
image

ArwenQin: I'm wondering if you feel the 2 spaces should be smaller?
image
@severinbeauvais 
Thank you!

severinbeauvais: See how the other filings wrap the components above and you can probably copy it.

src/views/ContinuationIn/ContinuationInBusinessBc.vue
```diff
@@ -4,9 +4,13 @@
     <section class="mt-10">
       <header id="company-name">
         <h2>Company Name</h2>
+        <p></p>
```
severinbeauvais, line 7: Don't use empty paragraphs for whitespace. Instead, use the Vuetify helper classes to set margins or padding. [link](https://v2.vuetifyjs.com/en/styles/spacing/)

src/views/ContinuationIn/ContinuationInBusinessBc.vue
```diff
@@ -4,9 +4,13 @@
...
       </header>
 
-      <div>** Correct Name component goes here **</div>
+      <v-card flat class="py-8 px-6">
```
severinbeauvais, line 10: Does VS Code show you the warnings like the one below?

image

Make sure you have added the following extensions to VS Code:
- ESLint
- Vue - Official

src/views/ContinuationIn/ContinuationInBusinessBc.vue
severinbeauvais: The v-card looks correct but, yes, I think the top and bottom whitespace should be smaller. Compare with other v-cards.

The named continuation in application looks good 👍

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-688-fyq5ge7j.web.app

severinbeauvais:
I think you need to add something to the validation for step 2 (Your Business in B.C.) -- even though I have filled everything in, the review page says that step 2 is unfinished.

Hint: start from the review page to see what it looks at to determine that step 2 is invalid. Then go to step or the store to update whatever is needed so the step is valid when all components on it are valid.

src/views/ContinuationIn/ContinuationInBusinessBc.vue
ArwenQin: Thank you Severin. I updated the top and bottom whitespaces:
image

src/views/ContinuationIn/ContinuationInBusinessBc.vue
ArwenQin: Got it, thanks. Updated.

src/views/ContinuationIn/ContinuationInBusinessBc.vue
ArwenQin: I just installed the 2 extensions and updated the formatting. Thank you.

ArwenQin:
> Have you tested this with a _named_ continuation in application (ie, one that uses a NR to set a company name)?



> I think you need to add something to the validation for step 2 (Your Business in B.C.) -- even though I have filled everything in, the review page says that step 2 is unfinished.
> 
> Hint: start from the review page to see what it looks at to determine that step 2 is invalid. Then go to step or the store to update whatever is needed so the step is valid when all components on it are valid.

Hi Severin, @severinbeauvais , 
It's an issue with the DefineCompanyStepValidity. 
I updated the ContinuationInBusinessBc.vue file to fix this issue. I added the getter to update the DefineCompanyStepValidity.

src/views/ContinuationIn/ContinuationInBusinessBc.vue
```diff
@@ -202,11 +215,12 @@ export default class ContinuationInBusinessBc extends Mixins(CommonMixin) {
...
-      this.addressFormValid &&
-      this.businessContactFormValid &&
-      (!this.isPremiumAccount || this.getFolioNumberValid)
-    )
+    const allValid = this.addressFormValid &&
+                   this.businessContactFormValid &&
+                   (!this.isPremiumAccount || this.getFolioNumberValid)
+    this.setContinuationInBusinessBcValid(allValid)
+    // Update the overall Define Company Step validity
+    this.setDefineCompanyStepValidity(allValid)
```
severinbeauvais, line 223: This looks like the correct solution, but I wonder if we still need `setContinuationInBusinessBcValid()` or whether it can be deleted. The value it sets (getContinuationIn.businessBcValid) is only used for the stepper.

I think you should try to combine this into setDefineCompanyStepValidity().

severinbeauvais: Also, this change broke some unit tests. I would clean this up and then fix the unit tests if they're still broken.

ArwenQin: Hi Severin, 
- I just deleted the setContinuationInBusinessBcValid(), and updated the variables in Stepper (only use isDefineCompanyValid() ).
- I also added the method to validate NameTranslation Component in step 2

package.json
severinbeauvais, line 1: Please also commit the package-lock.json file. This is used by the CI process to build the app using exactly the same packages and versions as you locally.

src/views/ContinuationIn/ContinuationInBusinessBc.vue
```diff
@@ -142,15 +156,17 @@ export default class ContinuationInBusinessBc extends Mixins(CommonMixin) {
...
   ]
 
   /** Object of valid flags. Must match validComponents. */
   get validFlags (): object {
     return {
       validAddressForm: this.addressFormValid,
       validBusinessContactForm: this.businessContactFormValid,
-      validFolioReferenceNumber: !this.isPremiumAccount || this.getFolioNumberValid
+      validFolioReferenceNumber: !this.isPremiumAccount || this.getFolioNumberValid,
+      validNameTranslation: this.getNameTranslationsValid
```
severinbeauvais, line 169: Very nice following the existing patterns 👍

src/components/common/Stepper.vue
```diff
@@ -78,7 +78,6 @@ export default class Stepper extends Vue {
   @Getter(useStore) isAffidavitValid!: boolean
   @Getter(useStore) isAmalgamationInformationValid!: boolean
   @Getter(useStore) isBusySaving!: boolean
-  @Getter(useStore) isContinuationInBusinessBcValid!: boolean
```
severinbeauvais, line 81: Since you've obsoleted this getter (and this was the only place it was used), please remove it from the store.

I think you can also delete `setContinuationInBusinessBcValid()` from the store, and also the respective state variable (see continuation-in-state-interface.ts and state-model.ts).

severinbeauvais:
Looks great so far!

Please finish the unit tests for this page (see ContinuationInBusinessBc.spec.ts).

PS In VS Code, you can search for files by name using Ctrl-P.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-688-fyq5ge7j.web.app

src/views/ContinuationIn/ContinuationInBusinessBc.vue
severinbeauvais: Oh, darn. This won't work because `name-translation-header` isn't an id on this page.

Also, these flags (and the list on line 156) should be in the order of the components on the page. These are used by `validateAndScroll()` to scroll to the first invalid component.

severinbeauvais: I think it would be OK to scroll to line 5 (change "company-name" to "company-name-header"). And also rename the variable on line 169 accordingly.

ArwenQin: Updated, thank you. Now it works

src/components/common/Stepper.vue
ArwenQin: Removed the unused getter and variables

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-688-fyq5ge7j.web.app

severinbeauvais:
Code looks good.
Works nicely on preview URLs above.
Just the following:
- [ ] check in package-lock.json
- [ ] update unit tests

ArwenQin:
> Code looks good. Works nicely on preview URLs above. Just the following:
> 
> * [ ]  check in package-lock.json
> * [ ]  update unit tests

Thank you so much. I just updated the unit tests and checked in the package-lock.json

package.json
ArwenQin: Checked in package-lock.json

severinbeauvais (approved):
I approve this PR. Arwen, do you have permission to Squash and Merge it?

ArwenQin:
> permission to Squash and Merge it

Thanks Severin. But I don't have write access to merge pull yet.

PR #696, opened 2024-05-28
21315 - Upload Director's Affidavit

ArwenQin:
*Issue #:* /bcgov/entity#21315

*Description of changes:*
- Added Upload Director's Affidavit component for manual input


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-696-latq0hhi.web.app

src/components/ContinuationIn/ManualBusinessInfo.vue
```diff
@@ -407,10 +508,12 @@ export default class ManualBusinessInfo extends Mixins(CountriesProvincesMixin,
...
     // and we have the home incorporation date (custom component)
+    // and we have the affidavit file, if required (custom component)
     // and the other form (Vuetify) components are valid
     return (
       !!this.business.homeJurisdiction &&
       !!this.business.homeIncorporationDate &&
+      (!this.isContinuationInAffidavitRequired || !!this.business.affidavitFileKey) &&
       this.formValid
     )
   }
```
severinbeauvais, line 519: You also need to copy the styles for the bullets:

image

severinbeauvais (commented):
Looks great!

However, since this is a duplicate of the code in ExtraproRegistration.vue, can you please extract it into a new sub-component, and use that sub-component in both components?

Also, please add some basic unit tests for the new sub-component, and also update ExtraproRegistration.spec.ts and create ManualBusinessInfo.spec.ts.

Thanks!

severinbeauvais:
There should be extra whitespace above the Important message box (mt-6).

image

src/components/ContinuationIn/ManualBusinessInfo.vue
```diff
@@ -386,6 +439,54 @@ export default class ManualBusinessInfo extends Mixins(CountriesProvincesMixin,
...
+
+        // add properties reactively to business object
+        this.$set(this.business, 'affidavitFile', {
+          name: file.name,
+          lastModified: file.lastModified,
+          size: file.size
+        } as File)
+        this.$set(this.business, 'affidavitFileKey', psu.key)
+        this.$set(this.business, 'affidavitFileName', file.name)
+        this.$set(this.business, 'affidavitFileUrl', psu.preSignedUrl)
```
severinbeauvais, line 478: @ArwenQin , as per my recent PR, please remove this line and also line 485. We no longer save the file URL.

ArwenQin: Done

src/components/ContinuationIn/ManualBusinessInfo.vue
ArwenQin: Done, updated

ArwenQin:
> There should be extra whitespace above the Important message box (mt-6).
> 
> image

Updated
image

ArwenQin:
> Looks great!
> 
> However, since this is a duplicate of the code in ExtraproRegistration.vue, can you please extract it into a new sub-component, and use that sub-component in both components?
> 
> Also, please add some basic unit tests for the new sub-component, and also update ExtraproRegistration.spec.ts and create ManualBusinessInfo.spec.ts.
> 
> Thanks!

I added the UploadAffidavit.vue sub-component.
I will add the unit tests tmr.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-696-latq0hhi.web.app

severinbeauvais (commented):
Looks good so far!

The only thing I'd change is to keep the rows and columns in their original components and just have the content of the column in the new sub-component. This will separate the layout from what the sub-component is supposed to do.

src/components/ContinuationIn/UploadAffidavit.vue
```diff
@@ -0,0 +1,149 @@
...
+        // add properties reactively to business object and emit to parent coponent
+        this.$set(this.business, 'affidavitFile', {
+          name: file.name,
+          lastModified: file.lastModified,
+          size: file.size
+        } as File)
+        this.$set(this.business, 'affidavitFileKey', psu.key)
+        this.$set(this.business, 'affidavitFileName', file.name)
+        // emit the updated business object
+        this.$emit('updateBusiness', this.business)
```
severinbeauvais, line 112: Use an `@Emit` decorator and separate method to emit please.

src/components/ContinuationIn/ManualBusinessInfo.vue
```diff
@@ -386,6 +407,13 @@ export default class ManualBusinessInfo extends Mixins(CountriesProvincesMixin,
     }
   }
 
+  /**
+   * Called when a Director's Affidavit is uploaded.
+   */
+  updateBusiness (updatedBusiness: any): void {
+    this.business = updatedBusiness
```
severinbeauvais, line 414: Does the store have to be updated also?

Make some changes on this page and then go to the review page and see if your changes are visible.

src/components/ContinuationIn/ManualBusinessInfo.vue
```diff
@@ -216,6 +216,18 @@
...
+        <UploadAffidavit
+          :business="business"
+          @updateBusiness="updateBusiness"
+        />
+
+        <!-- empty column -->
+        <v-col
+          cols="12"
+          sm="3"
+        />
```
severinbeauvais, line 229: You shouldn't have a v-col that is not wrapped in a v-row.

If you want whitespace above the message box, add class "mt-6" to the message box row element.

src/components/ContinuationIn/ManualBusinessInfo.vue
ArwenQin: It's updated. Yes, those changes are visible.
image

src/components/ContinuationIn/UploadAffidavit.vue
severinbeauvais: Actually, you probably don't need to emit the updated business object back to the parent component.

Theoretically, the business object that you are updating here is the same object the parent is pointing to -- there is a single memory location to this object's memory and several variables point to it.

I say "theoretically" because you should test this, of course :)  But it would simplify your code. If this indeed works, please add a comment somewhere in this sub-component that the business is the same object that the parent is working with (not a copy -- the prop is a reference to the original object). I already added similar comments elsewhere in this code.

src/components/ContinuationIn/ManualBusinessInfo.vue
severinbeauvais: This comment is probably obsolete. Please verify and resolve if everything works as expected.

src/components/ContinuationIn/UploadAffidavit.vue
ArwenQin: Yes, the Prop works. Emit no longer needed

src/components/ContinuationIn/ManualBusinessInfo.vue
ArwenQin: Removed, looks good now:
image

src/components/ContinuationIn/UploadAffidavit.vue
severinbeauvais: Awesome. So does this pass-by-reference (multiple pointers to same object) make sense? It's a concept from long ago in C/C++ world.

ArwenQin: Got it! Make sense to me

PR #698, opened 2024-05-31
21248 Fixed continuation authorization reactivity issue

severinbeauvais:
*Issue #:* bcgov/entity#21248

*Description of changes:*
- app = 5.10.16
- initialized rest of Continuation Authorization object variable so they are reactive

### DIFF
--- a/src/components/ContinuationIn/ContinuationAuthorization.vue
+++ b/src/components/ContinuationIn/ContinuationAuthorization.vue
@@ -191,8 +191,8 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
 
   /** Called when this component is mounted. */
   mounted (): void {
-    // set or initialize authorization object
-    this.authorization = this.getContinuationAuthorization || { files: [] } as ContinuationAuthorizationIF
+    this.authorization = this.getContinuationAuthorization ||
+      { files: [], date: null, expiryDate: null } as ContinuationAuthorizationIF
   }
 
   /**
@@ -268,10 +268,11 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
   @Watch('authorization', { deep: true })
   @Emit('valid')
   private onComponentValid (): boolean {
-    // first save the authorization to the store
+    // sync local object to the store
     this.setContinuationAuthorization(this.authorization)
 
-    // then emit the validity
+    // this component is valid if we have the authorization date
+    // and at least one file uploaded
     return (
       !!this.authorization.date &&
       (this.authorization.files.length >= 1)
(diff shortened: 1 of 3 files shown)


src/components/ContinuationIn/ContinuationAuthorization.vue
```diff
@@ -191,8 +191,8 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
 
   /** Called when this component is mounted. */
   mounted (): void {
-    // set or initialize authorization object
-    this.authorization = this.getContinuationAuthorization || { files: [] } as ContinuationAuthorizationIF
+    this.authorization = this.getContinuationAuthorization ||
+      { files: [], date: null, expiryDate: null } as ContinuationAuthorizationIF
```
JazzarKarim, line 195: So, initially setting them to null fixed the reactivity issue?

JazzarKarim (approved):

src/components/ContinuationIn/ContinuationAuthorization.vue
severinbeauvais, line 195: The reason for the Date and Expiry Date not being synced to the store and thus displayed on the review page is that the watcher on line 270 below didn't fire.

The reason that the watcher didn't fire, even though the "deep watch" option was enabled, is that the Date and Expiry Date were not reactive.

The reason that the Date and Expiry Date were not reactive is that they were undefined in the initial object (only `files` was initialized). Now that I initialize them, they are reactive and the watcher fires and the store is updated and the review step shows their value.

What line 194 does is set the local variable to point (refer) to the same object that is in the store. This is not a new memory allocation; it's the same object. So changing object properties in this component is actually setting those properties in the store object. However, if the store object was initially null then line 195 assigns a new object (memory) to the local variable. This is then synced to the store on line 272.

severinbeauvais: You got it.

JazzarKarim: I bet that took some time to identify. Sweet, thanks for the clarification Sev! It makes sense.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-698-h4ya0yrz.web.app

src/components/ContinuationIn/ContinuationAuthorization.vue
severinbeauvais: Remember, undefined local variables, and undefined object properties, are not reactive.

That's why, elsewhere, when I added properties to an object, I used `$set(object, property)`.

JazzarKarim: Keeping this in mind.

severinbeauvais: Nope, I figured out this bug right away.

But then, a while back, I spent time learning (the hard way) about undefined variables and properties 😆

PR #700, opened 2024-06-03
21450 Set HaveChanges when user has made changes

severinbeauvais:
*Issue #:* bcgov/entity#21450

**The changes in this PR are not related to the subject ticket (which doesn't reproduce). This is a bit of cleanup that I did after ongoing testing of a Continuation In filing.**

*Description of changes:*
- app version = 5.10.19
- set HaveChanges when user has made changes

### DIFF
--- a/src/components/ContinuationIn/ExtraproRegistration.vue
+++ b/src/components/ContinuationIn/ExtraproRegistration.vue
@@ -453,6 +453,7 @@ export default class ExtraproRegistration extends Mixins(DateMixin) {
   @Getter(useStore) isContinuationInAffidavitRequired!: boolean
 
   @Action(useStore) setExistingBusinessInfo!: (x: ExistingBusinessInfoIF) => void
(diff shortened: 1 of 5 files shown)


src/components/ContinuationIn/ExtraproRegistration.vue
```diff
@@ -646,6 +647,9 @@ export default class ExtraproRegistration extends Mixins(DateMixin) {
   @Watch('formValid')
   @Emit('valid')
   private onComponentValid (): boolean {
+    // if we're here it's because the user has changed something
+    this.setHaveChanges(true)
```
severinbeauvais, line 651: Normally, this action is attached to various setters in the store.

However, in this component we change properties inside a referenced object, so `setExistingBusinessInfo()` isn't called every time the user changes something. So we do it here. Ditto for ManualBusinessInfo.vue, below.

ArwenQin (approved):
Make sense to me!

PR #701, opened 2024-06-03
21310 Implement Authorization Date min date validation

severinbeauvais:
*Issue #:* bcgov/entity#21310

*Description of changes:*

Commit 1:
- app version = 5.10.20
- added Authorization Date min date validation

Commit 2:
- trimmed whitespace from text input fields
- now deep-watch objects

*Screenshots:*

Minimum date is May 16, 2001 because that's the date of registration in BC.
image

image

Minimum date is Jan 15, 20024 because that's the date of incorporation in home jurisdiction.
image

image

### DIFF
--- a/src/components/ContinuationIn/ContinuationAuthorization.vue
+++ b/src/components/ContinuationIn/ContinuationAuthorization.vue
@@ -172,10 +172,13 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
   fileValidity = false
   customErrorMessage = ['', '', '', '', ''] // max 5 files
 
+  /**
+   * The minimum date for the Authorization Date:
+   * - the BC Founding Date (if it exists, ie, expro business)
+   * - else the Home Jurisdiction Incorporation Date (ie, manual entry)
+   * - else fall back to null (not undefined)
+   */
   get minAuthorizationDate (): string {
-    // the BC Founding Date (if it exists, ie, expro business)
-    // else the Home Jurisdiction Incorporation Date (ie, manual entry)
-    // else fall back to null (not undefined)
     return (
       this.getExistingBusinessInfo.bcFoundingDate ||
       this.getExistingBusinessInfo.homeIncorporationDate ||
--- a/src/components/ContinuationIn/ContinuationAuthorization.vue
+++ b/src/components/ContinuationIn/ContinuationAuthorization.vue
@@ -36,6 +36,7 @@
             :persistentHint="true"
             :initialValue="authorization.date"
             :inputRules="getShowErrors ? authorizationDateRules : []"
+            :minDate="minAuthorizationDate"
             :maxDate="getCurrentDate"
             @emitDateSync="authorization.date = $event"
           />
@@ -136,7 +137,7 @@ import { Action, Getter } from 'pinia-class'
 import { StatusCodes } from 'http-status-codes'
 import { useStore } from '@/store/store'
 import { DocumentMixin } from '@/mixins'
-import { ContinuationAuthorizationIF, FormIF, PresignedUrlIF } from '@/interfaces'
+import { ContinuationAuthorizationIF, ExistingBusinessInfoIF, FormIF, PresignedUrlIF } from '@/interfaces'
 import { PdfPageSize } from '@/enums'
 import { VuetifyRuleFunction } from '@/types'
 import FileUploadPreview from '@/components/common/FileUploadPreview.vue'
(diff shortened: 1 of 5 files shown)


src/components/ContinuationIn/ContinuationAuthorization.vue
```diff
@@ -170,10 +172,28 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
...
+  get minAuthorizationDate (): string {
+    // the BC Founding Date (if it exists, ie, expro business)
+    // else the Home Jurisdiction Incorporation Date (ie, manual entry)
+    // else fall back to null (not undefined)
+    return (
+      this.getExistingBusinessInfo.bcFoundingDate ||
+      this.getExistingBusinessInfo.homeIncorporationDate ||
+      null
+    )
+  }
```
severinbeauvais, line 184: Fall back to null because 'undefined' is not reactive (ie, getters and watchers won't recognize its change).

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-701-1jrgf7gx.web.app

ArwenQin (approved):
Tested, LGTM

src/components/ContinuationIn/ExtraproRegistration.vue
```diff
@@ -132,7 +132,7 @@
             sm="9"
           >
             <v-text-field
-              v-model="business.homeIdentifier"
+              v-model.trim="business.homeIdentifier"
```
severinbeauvais, line 135: Automatically trims leading and trailing spaces. (Note: on screen, it looks like there are trailing spaces, but in the object property itself, those spaces aren't there.)

Ref: https://vuejs.org/guide/essentials/forms#trim

cc: @JazzarKarim

src/components/ContinuationIn/ExtraproRegistration.vue
```diff
@@ -475,13 +475,13 @@ export default class ExtraproRegistration extends Mixins(DateMixin) {
   }
 
   readonly homeIdentifierRules: Array<VuetifyRuleFunction> = [
-    (v) => !!v || 'Identifying Number is required',
-    (v) => (v && v.length <= 50) || 'Cannot exceed 50 characters'
+    (v) => !!v?.trim() || 'Identifying Number is required',
+    (v) => (v && v.trim().length <= 50) || 'Cannot exceed 50 characters'
```
severinbeauvais, line 479: Apparently, Vuetify doesn't trim the value before applying the rules, so `trim()` is needed here to determine if the value is falsy (including empty string) and check its length.

src/components/ContinuationIn/ExtraproRegistration.vue
```diff
@@ -640,10 +640,8 @@ export default class ExtraproRegistration extends Mixins(DateMixin) {
 
   /** Emits form validity. */
   @Watch('isBusinessActive')
-  @Watch('business.homeJurisdiction')
-  @Watch('business.homeIncorporationDate')
+  @Watch('business', { deep: true })
```
severinbeauvais, line 643: The old code assumed the form validity would change and catch any data changes, but that's not valid. For example, I could change a home identifier or home legal name character and the form validity wouldn't change, and also line 649 wouldn't be called, so the page wouldn't warn  you if you tried to leave with unsaved changes.

Instead of adding those extra watchers, in the new code I simply deep-watch the object (ie, all properties).

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-701-1jrgf7gx.web.app

src/components/ContinuationIn/ExtraproRegistration.vue
JazzarKarim: Very very nice.

src/components/ContinuationIn/ExtraproRegistration.vue
JazzarKarim: Makes a lot of sense.

src/components/ContinuationIn/ContinuationAuthorization.vue
```diff
@@ -170,10 +172,28 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
   fileValidity = false
   customErrorMessage = ['', '', '', '', ''] // max 5 files
 
+  get minAuthorizationDate (): string {
+    // the BC Founding Date (if it exists, ie, expro business)
```
JazzarKarim, line 176: I'm thinking we change this to a JSDoc comment. Thoughts?

JazzarKarim (approved):
LGTM!

src/components/ContinuationIn/ContinuationAuthorization.vue
severinbeauvais: Absolutely. Now looks like this:

image

JazzarKarim: Amazing. Thanks Sev! Resolving.

PR #705, opened 2024-06-06
21599 - Updated Step One Tick Mark Issue

ArwenQin:
*Issue #:* /bcgov/entity#21599

*Description of changes:*
- Step one will only show the Tick Mark and errors, after the user visits the Review and Confirm Page
- The authorization date picker will reset, if the incorporation date is changes to a date after the current selected authorization date. (Otherwise the component will be valid regardless of the invalid authorization date)

### DIFF
--- a/src/components/ContinuationIn/ContinuationAuthorization.vue
+++ b/src/components/ContinuationIn/ContinuationAuthorization.vue
@@ -289,6 +289,21 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
     }
   }
 
+  // reset the Authorization Date picker when Incorporation Date changes
+  // to a date after the current Authorization date
+  @Watch('minAuthorizationDate')
+  private resetAuthotrizationDate (): void {
+    const currentAuthDate = new Date(this.authorization.date)
+    const newMinDate = new Date(this.minAuthorizationDate)
+
+    if (newMinDate > currentAuthDate) {
+      this.authorization.date = null
+      this.$nextTick(() => {
+        this.$refs.authorizationDateRef.clearDate() // Reset the Authorization date picker
+      })
+    }
+  }
+
   @Watch('authorization', { deep: true })
   @Emit('valid')
   private onComponentValid (): boolean {
--- a/src/components/ContinuationIn/ContinuationAuthorization.vue
+++ b/src/components/ContinuationIn/ContinuationAuthorization.vue
@@ -281,7 +281,7 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
   @Watch('getShowErrors')
   @Watch('minAuthorizationDate') // because Authorization Date depends on this
   @Watch('authorization.date') // because Expiry Date depends on this
-  @Watch('getExistingBusinessInfo.homeIncorporationDate') // because Authorization Date validity depends on this
+  //@Watch('getExistingBusinessInfo.homeIncorporationDate') // because Authorization Date validity depends on this
   private async onGetShowErrors (): Promise<void> {
     if (this.getShowErrors) {
       // wait for form to finish rendering
@@ -303,7 +303,6 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
     // this component is valid if we have the valid authorization date
     // and at least one file uploaded
     return (
-      !!this.authorization.date &&
       this.authorizationDateValid &&
       (this.authorization.files.length >= 1)
     )
--- a/src/components/ContinuationIn/ContinuationAuthorization.vue
+++ b/src/components/ContinuationIn/ContinuationAuthorization.vue
@@ -291,31 +291,18 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
     }
   }
 
-  // reset the Authorization Date picker when Incorporation Date changes
-  // to a date after the current Authorization date
-  @Watch('minAuthorizationDate')
-  private resetAuthotrizationDate (): void {
-    const currentAuthDate = new Date(this.authorization.date)
-    const newMinDate = new Date(this.minAuthorizationDate)
-
-    if (newMinDate > currentAuthDate) {
-      this.authorization.date = null
-      this.$nextTick(() => {
-        this.$refs.authorizationDateRef.clearDate() // Reset the Authorization date picker
-      })
-    }
-  }
-
   @Watch('authorization', { deep: true })
+  @Watch('minAuthorizationDate') // re-validate when the min Authorization Date changes
   @Emit('valid')
   private onComponentValid (): boolean {
     // sync local object to the store
     this.setContinuationAuthorization(this.authorization)
 
-    // this component is valid if we have the authorization date
+    // this component is valid if we have the valid authorization date
     // and at least one file uploaded
     return (
       !!this.authorization.date &&
+      new Date(this.minAuthorizationDate) <= new Date(this.authorization.date) &&
       (this.authorization.files.length >= 1)
     )
   }
--- a/src/components/ContinuationIn/ContinuationAuthorization.vue
+++ b/src/components/ContinuationIn/ContinuationAuthorization.vue
@@ -169,6 +169,7 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
 
   // Local properties
   authorization = null as ContinuationAuthorizationIF
+  authorizationDateValid = false
   fileValidity = false
   customErrorMessage = ['', '', '', '', ''] // max 5 files
 
@@ -280,19 +281,20 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
   @Watch('getShowErrors')
   @Watch('minAuthorizationDate') // because Authorization Date depends on this
   @Watch('authorization.date') // because Expiry Date depends on this
+  @Watch('getExistingBusinessInfo.homeIncorporationDate') // because Authorization Date validity depends on this
   private async onGetShowErrors (): Promise<void> {
     if (this.getShowErrors) {
       // wait for form to finish rendering
       await this.$nextTick()
       // validate the form and our custom components that don't support form validation
       this.$refs.formRef.validate()
-      this.$refs.authorizationDateRef.validateForm()
+      this.authorizationDateValid = this.$refs.authorizationDateRef.validateForm()
       this.$refs.expiryDateRef.validateForm()
     }
   }
 
   @Watch('authorization', { deep: true })
-  @Watch('minAuthorizationDate') // re-validate when the min Authorization Date changes
+  @Watch('authorizationDateValid') // re-validate when the Authorization Date validity changes
   @Emit('valid')
   private onComponentValid (): boolean {
     // sync local object to the store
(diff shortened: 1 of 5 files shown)


ArwenQin:
I updated the ContinuationInAuthorization component - The authorization date picker will reset, if the incorporation date changes to a date after the current selected authorization date. 
Otherwise the component will be valid regardless of the invalid authorization date, current:
image

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-705-mufawei3.web.app

severinbeauvais:
> I updated the ContinuationInAuthorization component - The authorization date picker will reset, if the incorporation date changes to a date after the current selected authorization date. Otherwise the component will be valid regardless of the invalid authorization date

I don't think the ManualBusinessInfo component should change another component's object.

You can probably make the original code work properly by checking the form validity in ContinuationAuthorization.vue::onComponentValid() (and maybe you need a variable for form validity and you need to watch it, same as [here](https://github.com/bcgov/business-create-ui/blob/fb729e6e879eb7d6ff7163d34a5f51eced1ec402/src/components/ContinuationIn/ManualBusinessInfo.vue#L387).

src/components/ContinuationIn/ExtraproRegistration.vue
```diff
@@ -653,13 +654,17 @@ export default class ExtraproRegistration extends Mixins(DateMixin) {
     // and we have the home incorporation date (custom component)
     // and we have the affidavit file, if required (custom component)
     // and the other form (Vuetify) components are valid
-    return (
-      this.isBusinessActive &&
+    // show tick mark only when user visits Review Page
+    if (this.getShowErrors) {
+      return (
+        this.isBusinessActive &&
```
severinbeauvais, line 660: This is a good solution.

Another way to do this would be to move `getShowErrors` inside the return statement.

ArwenQin: Thanks! Updated.

ArwenQin:
> > I updated the ContinuationInAuthorization component - The authorization date picker will reset, if the incorporation date changes to a date after the current selected authorization date. Otherwise the component will be valid regardless of the invalid authorization date
> 
> ~I don't think the ManualBusinessInfo component should change another component's object.~ _My mistake, I read your comment before I looked at your code. This isn't an issue._
> 
> You can probably make the original code work properly by checking the form validity in ContinuationAuthorization.vue::onComponentValid() (and maybe you need a variable for form validity and you need to watch it, same as [here](https://github.com/bcgov/business-create-ui/blob/fb729e6e879eb7d6ff7163d34a5f51eced1ec402/src/components/ContinuationIn/ManualBusinessInfo.vue#L387)).

Got it, thanks! I updated the coding, much simpler now. Will re-validate the component when the minAuthoDate changes.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-705-mufawei3.web.app

src/components/ContinuationIn/ContinuationAuthorization.vue
```diff
@@ -292,15 +292,17 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
...
   private onComponentValid (): boolean {
     // sync local object to the store
     this.setContinuationAuthorization(this.authorization)
 
-    // this component is valid if we have the authorization date
+    // this component is valid if we have the valid authorization date
     // and at least one file uploaded
     return (
       !!this.authorization.date &&
+      new Date(this.minAuthorizationDate) <= new Date(this.authorization.date) &&
```
severinbeauvais, line 305: This is a duplicate of the rules on lines 196-199.

My idea was to use a v-model variable on the v-form. I think onGetShowErrors() will automatically run when minAuthorizationDate changes, which will call the validation functions. Then, here, you can just check if the form is valid.

ArwenQin: > This is a duplicate of the rules on lines 196-199.
> 
> My idea was to use a v-model variable on the v-form. I think onGetShowErrors() will automatically run when minAuthorizationDate changes, which will call the validation functions. Then, here, you can just check if the form is valid.

The form will declare itself as valid even when the authorizationDate is invalid. It seems the formRef.validate() will be valid even when all fields are empty - is it because all sub custom components don't support form validation?

severinbeauvais: Oh, you're on to something.

The DatePicker is one of our custom components, so no, it doesn't support form validation here... but internally it does (since it uses its own form). 

On lines 289 and 290, you call a public method in that component. There is also `isDateValid()` that you might be able to call to check if it's valid.

severinbeauvais: https://github.com/bcgov/bcrs-shared-components/blob/83cfe9c8dbe19714567814eb3efd7e08f35faf49/src/components/date-picker/DatePicker.vue#L127

severinbeauvais: Let me know if this doesn't work, and we'll look for a better or simpler solution.

ArwenQin: Thank you Severin.
I used a const authorizationDateValid to track the authorization date form validity. It works.

src/components/ContinuationIn/ContinuationAuthorization.vue
```diff
@@ -280,27 +281,30 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
...
   @Watch('authorization.date') // because Expiry Date depends on this
+  @Watch('getExistingBusinessInfo.homeIncorporationDate') // because Authorization Date validity depends on this
   private async onGetShowErrors (): Promise<void> {
     if (this.getShowErrors) {
       // wait for form to finish rendering
       await this.$nextTick()
       // validate the form and our custom components that don't support form validation
       this.$refs.formRef.validate()
-      this.$refs.authorizationDateRef.validateForm()
+      this.authorizationDateValid = this.$refs.authorizationDateRef.validateForm()
```
ArwenQin, line 291: I use this const to track the auth. date validity and watch it in the componentValid method

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-705-mufawei3.web.app

src/components/ContinuationIn/ContinuationAuthorization.vue
```diff
@@ -280,27 +281,30 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
   @Watch('getShowErrors')
   @Watch('minAuthorizationDate') // because Authorization Date depends on this
   @Watch('authorization.date') // because Expiry Date depends on this
+  @Watch('getExistingBusinessInfo.homeIncorporationDate') // because Authorization Date validity depends on this
```
ArwenQin, line 284: In Extrapro situation, if the user select a home incorporation date later than BC registry date:
the min authorization date will be BC registry date.
But the auth. date rule will still check if the authorization date >=  home incorporation date

severinbeauvais: Are you sure you need this? I think that maybe line 282 will trigger this event already.

(Line 282 watches a computed variable, which _should_ auto-update if homeIncorporationDate changes.)

ArwenQin: > Are you sure you need this? I think that maybe line 282 will trigger this event already.
> 
> (Line 282 watches a computed variable, which _should_ auto-update if homeIncorporationDate changes.)

When the min authorization date is set to be bcFoundingDate (line 182). In that case, the change of homeIncorporateDate will not change the min authorization date. 
But the rule (line 199) still checks whether the authorization date >= home incorporation date.

ArwenQin: current screenshot: when homeIncorporateDate changes, the min authorization date does not change
and there will be an error
image
image

severinbeauvais: > When the min authorization date is set to be bcFoundingDate (line 182). In that case, the change of homeIncorporateDate will not change the min authorization date. But the rule (line 199) still checks whether the authorization date >= home incorporation date.

I tested this locally. The watcher on line 282 does indeed fire, but the condition on line 286 is false, so the components are not validated. I think this is actually what we want: we do not show errors until the review step has been displayed (and getShowErrors is True).

src/components/ContinuationIn/ContinuationAuthorization.vue
```diff
@@ -280,27 +281,30 @@ export default class ExtraproRegistration extends Mixins(DocumentMixin) {
...
   private onComponentValid (): boolean {
     // sync local object to the store
     this.setContinuationAuthorization(this.authorization)
 
-    // this component is valid if we have the authorization date
+    // this component is valid if we have the valid authorization date
     // and at least one file uploaded
     return (
       !!this.authorization.date &&
+      this.authorizationDateValid &&
```
severinbeauvais, line 307: You don't need line 306 :)

src/components/ContinuationIn/ContinuationAuthorization.vue
ArwenQin: If I remove line 284, even when getShowErrors is True - if we change the date of incorporation, the error will not show
image
the error will only show when we try to change the authorization date again:
image

I tested with:
http://localhost:8080/continuation-in-business-home?id=ThevF8a4zF&accountid=3302
I changed the date of incorporation to a date later than authorization date.
It is because when minAuthorizationDate is set to be the bcFoundingDate (line 182). In that case, the change of homeIncorporateDate will not change the min authorization date, so line 282 didn't fire.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-705-mufawei3.web.app

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-705-mufawei3.web.app

src/components/ContinuationIn/ContinuationAuthorization.vue
severinbeauvais: In your example, the minimum authorization date is 1968-09-10. This comes from the expro record.

Test again but using a manually entered business.

severinbeauvais: You can debug and step through the code to see what the variable values are, or do something like this:

image

ArwenQin: > In your example, the minimum authorization date is 1968-09-10. This comes from the expro record.
> 
> Test again but using a manually entered business.

Yes, the minimum authorization date (1968-09-10) is the bcFoundingDate (line 182), and irrelevant to the homeIncorporateDate. 
So if we only watch the @Watch('minAuthorizationDate') (line 282), when the user changes the homeIncorporateDate (say to June 11 2024), the minAuthorizationDate won't change, so line 285 - 294 didn't run. The current authorization date is actually invalid (Jun 3 2024), but the error will not show, and the tick mark incorrectly shows.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-705-mufawei3.web.app

src/components/ContinuationIn/ContinuationAuthorization.vue
severinbeauvais: OK, Arwen and I sorted this out. (Needed to add another rule.) Looks good to me now. This conversation can be resolved.

severinbeauvais (approved):

severinbeauvais:
@ArwenQin , thanks for all your testing and trying things, and pushing back when I was missing something key. Together, we got this! :)

PR #706, opened 2024-06-07
21116 Show user info for Continuation In filings

severinbeauvais:
*Issue #:* bcgov/entity#21116

*Description of changes:*
- app version = 5.10.24
- show user info for Continuation In filings
- updated unit tests

### DIFF
--- a/tests/unit/EntityInfo.spec.ts
+++ b/tests/unit/EntityInfo.spec.ts
@@ -1,4 +1,4 @@
-import { shallowWrapperFactory, wrapperFactory } from '../vitest-wrapper-factory'
+import { shallowWrapperFactory } from '../vitest-wrapper-factory'
 import EntityInfo from '@/components/common/EntityInfo.vue'
 import { FilingTypes } from '@/enums'
 import { CorpTypeCd } from '@bcrs-shared-components/enums/'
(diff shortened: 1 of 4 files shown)


severinbeauvais:
The sample filings above were created as staff, so the staff user info displays (which does not include phone number).

Here's a sample continuation in created and opened as BCREG0020 / Severin Dev Account:

image

tests/unit/EntityInfo.spec.ts
```diff
@@ -77,28 +77,129 @@ const mockEntityInfo = [
...
+        EntityInfo,
+        null,
+        {
+          tombstone: mock.tombstone,
+          entityType: mock.entityType,
+          tempId: mock.tempId,
+          nameRequest: mock.nameRequest,
+          nameRequestApprovedName: mock.nameRequestApprovedName
+        }
+      )
```
severinbeauvais, line 202: I changed this (and below) to shallow-mount because we don't need the child components to render in these tests.

tests/unit/EntityInfo.spec.ts
severinbeauvais, line 1: image

ketaki-deodhar (approved):
LGTM!

PR #709, opened 2024-06-11
21110 Non-functional code to delete unsaved Minio documents when UI exits

severinbeauvais:
*Issue #:* bcgov/entity#21110

*Description of changes:*
- non-functional code to delete unsaved Minio documents when UI exits
- this is not a working solution -- do not merge

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -256,7 +256,8 @@ import * as Dialogs from '@/dialogs'
 import * as Views from '@/views'
 
 // Mixins, interfaces, etc
-import { CommonMixin, DateMixin, FilingTemplateMixin, NameRequestMixin } from '@/mixins'
+import { CommonMixin, DateMixin, DocumentMixin, FilingTemplateMixin, NameRequestMixin }
+  from '@/mixins'
 import { AccountInformationIF, AddressIF, BreadcrumbIF, BusinessWarningIF, CompletingPartyIF,
   ConfirmDialogType, EmptyFees, FeesIF, FilingDataIF, OrgInformationIF, PartyIF, ResourceIF,
   StepIF } from '@/interfaces'
@@ -288,7 +289,9 @@ import { CorpTypeCd } from '@bcrs-shared-components/corp-type-module'
     ...Views
   }
 })
-export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMixin, NameRequestMixin) {
+export default class App extends Mixins(
+  CommonMixin, DateMixin, DocumentMixin, FilingTemplateMixin, NameRequestMixin
+) {
   // Refs
   $refs!: {
     confirm: ConfirmDialogType
@@ -314,7 +317,10 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
   @Getter(useStore) isRestorationFiling!: boolean
   @Getter(useStore) isMobile!: boolean
   @Getter(useStore) isSbcStaff!: boolean
+  // @Getter(useStore) isTypeCoop!: boolean
+  // @Getter(useStore) isTypeFirm!: boolean
 
+  @Action(useStore) clearUnsavedDocuments!: () => void
   @Action(useStore) setAccountInformation!: (x: AccountInformationIF) => void
   @Action(useStore) setAdminFreeze!: (x: boolean) => void
   @Action(useStore) setAlternateName!: (x: string) => void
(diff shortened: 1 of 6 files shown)


severinbeauvais:
All - this is a non-functional solution but I learned some things along the way and wanted to share.

Here are some comments I added to the ticket:
> Part 2 Summary
> 
> After a fair bit of research and testing, I believe this cannot be implemented, as there is no way to delay the app from unloading while any unsaved documents are deleted.

> 
> Specifically, window.onbeforeunload can only be used to display a generic (synchronous) "unsaved messages" dialog. The next event that we can catch is window.onpagehide (because onunload is deprecated) and it's synchronous, too. So, we cannot wait for any network calls (deletions) to complete, and in fact any pending network calls are cancelled when the app terminates.

> 
> Basically, asking the client to clean up before exiting is not a good design (and it doesn't handle all cases, such as power interruptions and some mobile scenarios). According to what I've researched, a better design is to have the API auto-delete any documents whose keys we haven't stored in the filing JSON. I have created [#21728](https://github.com/bcgov/entity/issues/21728) for this.

src/App.vue
```diff
@@ -519,16 +525,26 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
...
         event.returnValue = 'You have unsaved changes. Do you want to exit?'
       }
     }
 
+    // add handler to delete unsaved documents when page is hidden / unloaded
+    // (don't use window.onload as it's deprecated)
+    // NB: this is not reliably fired by browsers, especially on mobile
+    window.onpagehide = async (event: PageTransitionEvent) => {
+      if (!event.persisted) {
+        await this.deleteUnsavedDocuments()
```
severinbeauvais, line 542: This await does nothing. This event handler is not actually async. 😞 

There is also no way to run something specific if the user clicks to leave this site in the BeforeUnload event handler above (which is also synchronous).

src/mixins/document-mixin.ts
```diff
@@ -71,19 +83,31 @@ export default class DocumentMixin extends Vue {
...
+   * Deletes documents uploaded in this session but where the draft filing, and hence the file keys,
+   * was not saved to the API. This is to prevent orphaned documents in Minio.
+   */
+  async deleteUnsavedDocuments (): Promise<Array<any>> {
+    const promises = []
+    for (const key of this.getUnsavedDocuments) {
+      promises.push(this.deleteDocument(key, false))
+    }
+    // make all API calls concurrently without rejection
+    return Promise.allSettled(promises)
```
severinbeauvais, line 110: I tried quite a few things here to "wait" for the Axios delete calls to complete, but:
- the calls aren't processed until after this method runs (because JS is single-threaded)
- the calls are all cancelled because the app terminates right after this

severinbeauvais:
One potential client-side solution is to use a Service Worker to delete unsaved files. This is a separate thread and apparently can be made to outlive the initial web page thread. However, this still doesn't handle all cases and it's complex/overkill for just this purpose.

JazzarKarim:
So what's the alternative here Sev? Do you have something in mind?

severinbeauvais:
> So what's the alternative here Sev? Do you have something in mind?

We can often detect when the web app ends (eg, close tab, close browser, navigate away) but there are some undetectable conditions (like power loss, or some mobile scenarios).

In the end, it's not a good design for the client (web app) to clean up what are essentially temporary documents on the (Minio server). We should assume the client could just "disappear" and the servers need to protect themselves, including their data integrity, which includes cleaning up orphaned documents. (I call them temporary documents because they are not "locked in" until the filing is submitted.)

A better design would have the server do the cleanup. Legal API is already involved in getting a key for the documents that the UI uploads. Legal API could hold on to those keys and delete documents that don't appear in a filing after, say, a year. (Or also store the filing id and then purge obsolete documents when that filing is submitted.)

severinbeauvais:
> So what's the alternative here Sev? Do you have something in mind?

We can often detect when the web app ends (eg, close tab, close browser, navigate away) but there are some undetectable conditions (like power loss, and some mobile scenarios).

In the end, it's not a good design for the client (web app) to clean up what are essentially temporary documents on the (Minio) server. We should assume the client could just "disappear" and the servers need to protect themselves, including their data integrity, which includes cleaning up orphaned documents. (I call them temporary documents because they are not "locked in" until the filing is submitted.)

A better design would have the server do the cleanup. Legal API is already involved in getting a key for the documents that the UI uploads. Legal API could hold on to those keys and delete documents that don't appear in a filing after, say, a year. (Or also store the filing id and then purge obsolete documents when that filing is submitted.)

cc: @vysakh-menon-aot

PR #725, opened 2024-08-02
22644 Hide fee summary for change-requested filings

severinbeauvais:
*Issue #:* bcgov/entity#22644

*Description of changes:*
- app version = 5.11.12
- conditionally show fee summary

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -199,8 +199,9 @@
               cols="12"
               lg="3"
             >
-              <!-- Render fee summary only after data is loaded. -->
-              <aside v-if="haveData">
+              <!-- Render fee summary only after data is loaded, -->
+              <!-- and if this isn't a change-requested filing. -->
+              <aside v-if="haveData && getFilingStatus !== FilingStatus.CHANGE_REQUESTED">
                 <affix
                   relative-element-selector=".col-lg-9"
                   :offset="{ top: 100, bottom: -100 }"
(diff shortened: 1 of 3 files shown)


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-725-9uzav6wd.web.app

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-725-9uzav6wd.web.app

src/App.vue
```diff
@@ -199,8 +199,9 @@
               cols="12"
               lg="3"
             >
-              <!-- Render fee summary only after data is loaded. -->
-              <aside v-if="haveData">
+              <!-- Render fee summary only after data is loaded, -->
+              <!-- and if this isn't a change-requested filing. -->
+              <aside v-if="haveData && getFilingStatus !== FilingStatus.CHANGE_REQUESTED">
```
severinbeauvais, line 204: Note: normally I frown upon negative logic such as here ("only show if not X") because it will cause a bug if, say, we add a status and we don't automatically want to show this -- if we specified the exact conditions under which to show the Fee Summary, this would not cause such a bug.

In this case, I'm on the fence. This is an extra condition that applies only to filings in Change-Requested status and today this is pretty clear what it's doing. But, I don't mind changing this to `if status === DRAFT` if you think that's more appropriate.

What do you think?

vysakh-menon-aot: regardless of these options if a new status introduced we may have to change this. so i am good with this

vysakh-menon-aot (approved):

src/App.vue
ArwenQin: I think this is more clear

ArwenQin (approved):

PR #731, opened 2024-09-13
16700 Fixed limited restoration expiry months validation and save/resume

severinbeauvais:
*Issue #:* bcgov/entity#16700

*Description of changes:*
- app version = 5.11.18
- misc package updates
- renamed some store getters
- updated props and event handlers
- used omit from lodash instead of own package
- added some missing "haveChanges" statements


package-lock.json
severinbeauvais, line 1: This is a large diff because I rebuilt this from scratch.

src/components/Restoration/RestorationType.vue
```diff
@@ -130,15 +127,40 @@ export default class RestorationType extends Mixins(DateMixin, CommonMixin) {
...
+    }
+    return 24 // default if no expiry date was set
+  }
+
+  /**
+   * When months has changed, sets the limited restoration extension expiry date in the store.
+   * @param months The number of months to extend the restoration (or null).
+   */
+  onMonthsChanged (months: number): void {
+    if (Number.isInteger(months)) {
```
severinbeauvais, line 144: This is true if `months` is a number (including 0).

src/store/store.ts
```diff
@@ -1433,8 +1433,13 @@ export const useStore = defineStore('store', {
     setWindowWidth (width: number) {
       this.stateModel.windowWidth = width
     },
+
+    //
+    // alamgamation-specific setters
+    //
     setAmalgamatingBusinesses (amalgamatingBusinesses: Array<AmalgamatingBusinessIF>) {
       this.stateModel.amalgamation.amalgamatingBusinesses = amalgamatingBusinesses
+      if (!this.stateModel.ignoreChanges) this.stateModel.haveChanges = true
```
severinbeauvais, line 1442: "If the user has changed something then set a flag so we can try to stop users from exiting without saving their changes."

There were quite a few of these missing. I only updated some amalgamation and restoration setters.

src/store/store.ts
```diff
@@ -1451,45 +1456,55 @@ export const useStore = defineStore('store', {
...
-    },
-    setBusinessNameValid (valid: boolean) {
-      this.stateModel.restoration.businessNameValid = valid
-    },
-    setRestorationTypeValid (valid: boolean) {
-      this.stateModel.restoration.restorationTypeValid = valid
+
+    //
+    // restoration-specific setters
+    //
```
severinbeauvais, line 1471: I also sorted the setters below.

package-lock.json
ketaki-deodhar: Just curiosity: why from scratch :)

package.json
```diff
@@ -34,7 +34,7 @@
     "@bcrs-shared-components/help-business-number": "1.1.4",
     "@bcrs-shared-components/interfaces": "1.1.15",
     "@bcrs-shared-components/jurisdiction": "1.1.4",
-    "@bcrs-shared-components/limited-restoration-panel": "1.0.5",
+    "@bcrs-shared-components/limited-restoration-panel": "2.1.0",
```
JazzarKarim, line 37: Weren't we reserving major patches for the Vue 3 versions of the components Sev?

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-731-i4m8lh6d.web.app

package-lock.json
severinbeauvais: To update all the dependencies and sub-dependencies.

Consider that the package-lock.json file specifies the exact version of all packages that are imported, which get stale after a while. If I delete it and do `npm i` again, the latest version of everything is imported. I've been trying to solve some npm warnings and unit test warnings and IDE warnings...

src/store/store.ts
```diff
@@ -1451,45 +1456,55 @@ export const useStore = defineStore('store', {
...
+      if (!this.stateModel.ignoreChanges) this.stateModel.haveChanges = true
     },
     setRestorationApprovalType (type: ApprovalTypes) {
       this.stateModel.restoration.approvalType = type
+      if (!this.stateModel.ignoreChanges) this.stateModel.haveChanges = true
+    },
+    setRestorationApprovalTypeValid (valid: boolean) {
+      this.stateModel.restoration.approvalTypeValid = valid
     },
-    setRestorationCourtOrderNumber (fileNumber: string) {
```
JazzarKarim, line 1479: This is obsolete? We don't need it anymore?

JazzarKarim (approved):

src/store/store.ts
severinbeauvais: Wasn't used anywhere.

There's another setter to save the court order object, which is used.

JazzarKarim: Thanks, resolving.

package.json
severinbeauvais: Yes. For this component it's 3.x.x.

What you see here is an import of a Vue2 refactor of this component (done for Edit UI) that I pulled in, plus some small changes you saw in that earlier PR.

ketaki-deodhar (approved):

package-lock.json
severinbeauvais: The value of the package lock file is that the CI/CD jobs can build the app using the exact same versions that we build locally. We wouldn't want the jobs to always use the latest versions, as there could be changes we need to review.

PR #756, opened 2024-11-01
23970 - Allow editable certify legal name

AimeeGao:
*Issue #:* /bcgov/entity#23970

*Description of changes:*

- Updated the Certify Legal Name editing condition on the Review/Confirm page to allow editing of the legal name for all users.

    - The following filing types have been modified:
       - Amalgamation
       - Continuation In
       - Dissolution
       - Restoration (while general users cannot access this page, we ensure that the Certify Legal Name is editable in all filings for safety measures.)

**Note**:
According to Mihai's feedback, firms and coops types are not considered in this bug fix.
  - The following filings remain unchanged:
     - Dissolution Firm
     - Registration
     - IA

**Result**:
Before:
image

After:
image

### DIFF
--- a/src/views/Amalgamation/ReviewConfirm.vue
+++ b/src/views/Amalgamation/ReviewConfirm.vue
@@ -237,7 +237,7 @@
         <Certify
           class="py-8 px-6"
           :class="{ 'invalid-section': isCertifyInvalid }"
-          :disableEdit="!isRoleStaff"
+          :disableEdit="false"
           :invalidSection="isCertifyInvalid"
           :isStaff="isRoleStaff"
         />
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-create-ui",
-  "version": "5.14.1",
+  "version": "5.14.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "business-create-ui",
-      "version": "5.14.1",
+      "version": "5.14.2",
       "dependencies": {
         "@babel/compat-data": "^7.21.5",
         "@bcrs-shared-components/approval-type": "1.1.3",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "5.14.1",
+  "version": "5.14.2",
   "private": true,
   "appName": "Create UI",
   "sbcName": "SBC Common Components",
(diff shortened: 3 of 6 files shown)


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-756-xdzm8n18.web.app

src/views/Amalgamation/ReviewConfirm.vue
```diff
@@ -237,7 +237,7 @@
         <Certify
           class="py-8 px-6"
           :class="{ 'invalid-section': isCertifyInvalid }"
-          :disableEdit="!isRoleStaff"
+          :disableEdit="false"
```
severinbeauvais, line 240: What's the default value of this prop? I'm thinking you could just omit this prop. And if so, which do you think is better: less code, or more verbos code?

JazzarKarim (approved):
Other than Sev's comment, LGTM 👍

src/views/Amalgamation/ReviewConfirm.vue
AimeeGao: Thanks for pointing it out. 

About that `disableEdit` prop:

I took a look at the Certify component definition and noticed it's marked as `required: true`. If we skip this prop, we will get this error:
image

image

I've got 2 ideas on how we could handle this:

1. We could stick with what we've got:
It might be a bit wordy, but it makes sure that all the parent components are intentionally setting this value, and we know the current setup works well because it's been tested.

2. Or we could tweak the Certify component and omit this prop in the parent components when we don't need it:
```typescript
// Change this
@Prop({ required: true }) readonly disableEdit!: boolean

// To this
@Prop({ default: false }) readonly disableEdit!: boolean
   ```
We'd need to double-check that it doesn't break anything else

What do you think would work better? I prefer option 2, but we'd need to test it thoroughly. I'd love to hear your thoughts on this.

severinbeauvais: Good analysis.

I assume some thought went into making this property required, so I'm tempted to leave it that way. Thus, option 1 will be fine.

Thanks.

AimeeGao: Sure, I will update the version number again. Many thanks.

severinbeauvais (approved):

eason-pan-bc:
interesting, the Edit_UI has implemented the Certify Section a bit differently. There is a CertifySection component created based on Certify component in Edit UI.
https://github.com/bcgov/business-edit-ui/blob/main/src/components/common/CertifySection.vue

severinbeauvais:
You can merge when ready.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-756-xdzm8n18.web.app

PR #758, opened 2024-11-14
22639 - Update Certify Section Wording

ArwenQin:
*Issue #:* /bcgov/entity#22639

*Description of changes:*
- Update Certify section for Create UI

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "@bcrs-shared-components/base-address": "2.0.3",
     "@bcrs-shared-components/breadcrumb": "2.1.15",
     "@bcrs-shared-components/business-lookup": "1.3.4",
-    "@bcrs-shared-components/certify": "2.1.51",
+    "@bcrs-shared-components/certify": "^2.1.55",
     "@bcrs-shared-components/completing-party": "2.1.30",
     "@bcrs-shared-components/confirm-dialog": "1.2.1",
     "@bcrs-shared-components/contact-info": "1.2.15",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "5.14.2",
+  "version": "5.14.3",
   "private": true,
   "appName": "Create UI",
   "sbcName": "SBC Common Components",
@@ -22,7 +22,7 @@
     "@bcrs-shared-components/base-address": "2.0.3",
     "@bcrs-shared-components/breadcrumb": "2.1.15",
     "@bcrs-shared-components/business-lookup": "1.3.4",
-    "@bcrs-shared-components/certify": "^2.1.55",
+    "@bcrs-shared-components/certify": "2.1.55",
     "@bcrs-shared-components/completing-party": "2.1.30",
     "@bcrs-shared-components/confirm-dialog": "1.2.1",
     "@bcrs-shared-components/contact-info": "1.2.15",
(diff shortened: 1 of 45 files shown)


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-758-as4ni3qx.web.app

severinbeauvais (commented):
I think that maybe, here, you could make `entityDisplay` optional and display the provided string (for CP, SP and GP) else fall back to "business". What do you think?

ArwenQin:
> I think that maybe, here, you could make `entityDisplay` optional and display the provided string (for CP, SP and GP) else fall back to "business". What do you think?

Agree, thanks! I updated the codes

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-758-as4ni3qx.web.app

package.json
```diff
@@ -22,7 +22,7 @@
     "@bcrs-shared-components/base-address": "2.0.3",
     "@bcrs-shared-components/breadcrumb": "2.1.15",
     "@bcrs-shared-components/business-lookup": "1.3.4",
-    "@bcrs-shared-components/certify": "2.1.51",
+    "@bcrs-shared-components/certify": "2.1.55",
     "@bcrs-shared-components/completing-party": "2.1.30",
```
ArwenQin, line 26: The current dev correctly display the required wording ("are" authorized). Maybe we don't need to update it to 2.1.55. 
But when I run locally, it displays "is" instead of "are".

Could you please suggest, thanks.

src/resources/Restoration/BC.ts
```diff
@@ -41,8 +41,7 @@ export const RestorationResourceBc: RestorationResourceIF = {
       certifyStatements: [],
       certifyClause: `Note: It is an offence to make or assist in making a false or misleading
         statement in a record filed under the BC Corporations Act. A person who commits this
-        offence is subject to a maximum fine of $5,000.`,
-      entityDisplay: 'BC Corporations Act'
```
severinbeauvais, line 45: What did the old certify block look like with this value?

severinbeauvais:
> > I think that maybe, here, you could make `entityDisplay` optional and display the provided string (for CP, SP and GP) else fall back to "business". What do you think?
> 
> Agree, thanks! I updated the codes

Well, darn! That's even more changed files.

But I prefer this as it's DRYer and cleaner. Thanks for the update!

package.json
severinbeauvais: Check your node_modules folder. You might have an older version than 51.

As for whether to upgrade from 51 to 55...
- on the one hand, it's good to keep these imports updated
- on the other hand, did very much change from 51 to 55? if so then extra testing is needed

If there are no significant changes from 51 to 55 then I leave it up to you whether to upgrade it.

src/resources/Restoration/BC.ts
ArwenQin: image

severinbeauvais: Thanks.

Yeah, that doesn't make sense to me. Much better with "business" :)

Resolving.

package.json
ArwenQin: It seems there are no changes from 51 to 55. They were updated just due to the chore(release) publish
image

severinbeauvais: That likely means some dependency was updated (eg, enums or interfaces).

Up to you whether to update to 55 or not.

ArwenQin: So I leave it upgraded

severinbeauvais (approved):

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-758-as4ni3qx.web.app

PR #760, opened 2024-12-19
replaced Minio with Document Record Service

flutistar:
*Issue #:* /bcgov/entity#23446
*Description of changes:*
- Updated `AuthorizationProof` component
- Updated `UnlimitedLiabilityCorporationInformation` component
- Upload document function to Document Record Service
- Delete document from Document Record Service
- Get/Set `consumerDocumentId`
- Added 3 new env variables

### DIFF
--- a/src/components/ContinuationIn/AuthorizationInformation.vue
+++ b/src/components/ContinuationIn/AuthorizationInformation.vue
@@ -286,14 +286,27 @@ export default class AuthorizationInformation extends Mixins(DateMixin, Document
 
     this.isDownloading = true
     const documentClass = 'CORP'
-    await this.downloadDocumentFromDRS(documentKey, documentName, documentClass).catch(error => {
+    try {
+      const docUrl: string = await this.getDownloadLink(documentKey, documentName, documentClass)
+      const link = document.createElement('a')
+      link.href = docUrl
+      link.download = documentName
+      link.target = '_blank' // This opens the link in a new browser tab
+
+      // Append to the document and trigger the download
+      document.body.appendChild(link)
+      link.click()
+
+      // Remove the link after the download is triggered
+      document.body.removeChild(link)
+      this.isDownloading = false
+    } catch (error) {
       // eslint-disable-next-line no-console
-      console.log('fetchDocument() error =', error)
+      console.log('downloadDocument() error =', error)
       this.errorDialogTitle = 'Unable to download document'
-      this.errorDialogText = 'We were unable to download your document. If this error persists, please contact us.'
+      this.errorDialogText = 'An error occurred while downloading the document. Please try again.'
       this.errorDialog = true
-    })
-    this.isDownloading = false
+    }
   }
 
   @Watch('isValid', { immediate: true })
--- a/src/components/ContinuationIn/AuthorizationInformation.vue
+++ b/src/components/ContinuationIn/AuthorizationInformation.vue
@@ -212,7 +212,7 @@ import { AuthorizationProofIF, ExistingBusinessInfoIF } from '@/interfaces'
 import { DateMixin, DocumentMixin } from '@/mixins'
 import { CanJurisdictions, IntlJurisdictions, UsaJurisdiction } from '@bcrs-shared-components/jurisdiction/list-data'
 import { JurisdictionLocation } from '@bcrs-shared-components/enums'
-import { DocumentClasses } from '@/enums'
+import { DocumentClassEnum } from '@/enums'
 
 @Component({
   components: {
@@ -289,7 +289,7 @@ export default class AuthorizationInformation extends Mixins(DateMixin, Document
     await this.downloadDocumentFromDRS(
       documentKey,
       documentName,
-      DocumentClasses.CORP
+      DocumentClassEnum.CORP
     ).catch(error => {
       // eslint-disable-next-line no-console
       console.log('downloadDocument() error =', error)
--- a/src/components/ContinuationIn/AuthorizationInformation.vue
+++ b/src/components/ContinuationIn/AuthorizationInformation.vue
@@ -285,7 +285,8 @@ export default class AuthorizationInformation extends Mixins(DateMixin, Document
     if (!documentKey || !documentName) return // safety check
 
     this.isDownloading = true
-    await this.downloadDocument(documentKey, documentName).catch(error => {
+    const documentClass = 'CORP'
+    await this.downloadDocumentFromDRS(documentKey, documentName, documentClass).catch(error => {
       // eslint-disable-next-line no-console
       console.log('fetchDocument() error =', error)
       this.errorDialogTitle = 'Unable to download document'
--- a/src/components/ContinuationIn/AuthorizationInformation.vue
+++ b/src/components/ContinuationIn/AuthorizationInformation.vue
@@ -212,6 +212,7 @@ import { AuthorizationProofIF, ExistingBusinessInfoIF } from '@/interfaces'
 import { DateMixin, DocumentMixin } from '@/mixins'
 import { CanJurisdictions, IntlJurisdictions, UsaJurisdiction } from '@bcrs-shared-components/jurisdiction/list-data'
 import { JurisdictionLocation } from '@bcrs-shared-components/enums'
+import { DocumentClasses } from '@/enums'
 
 @Component({
   components: {
@@ -285,28 +286,19 @@ export default class AuthorizationInformation extends Mixins(DateMixin, Document
     if (!documentKey || !documentName) return // safety check
 
     this.isDownloading = true
-    const documentClass = 'CORP'
-    try {
-      const docUrl: string = await this.getDownloadLink(documentKey, documentName, documentClass)
-      const link = document.createElement('a')
-      link.href = docUrl
-      link.download = documentName
-      link.target = '_blank' // This opens the link in a new browser tab
-
-      // Append to the document and trigger the download
-      document.body.appendChild(link)
-      link.click()
-
-      // Remove the link after the download is triggered
-      document.body.removeChild(link)
-      this.isDownloading = false
-    } catch (error) {
+    await this.downloadDocumentFromDRS(
+      documentKey,
+      documentName,
+      DocumentClasses.CORP
+    ).catch(error => {
       // eslint-disable-next-line no-console
       console.log('downloadDocument() error =', error)
       this.errorDialogTitle = 'Unable to download document'
       this.errorDialogText = 'An error occurred while downloading the document. Please try again.'
       this.errorDialog = true
-    }
+    })
+
+    this.isDownloading = false
   }
 
   @Watch('isValid', { immediate: true })
--- a/src/components/ContinuationIn/AuthorizationProof.vue
+++ b/src/components/ContinuationIn/AuthorizationProof.vue
@@ -164,7 +164,7 @@ import { StatusCodes } from 'http-status-codes'
 import { useStore } from '@/store/store'
 import { DocumentMixin } from '@/mixins'
 import { AuthorizationProofIF, ExistingBusinessInfoIF } from '@/interfaces'
-import { FilingStatus } from '@/enums'
+import { FilingStatus, DOCUMENT_TYPES as DocumentTypes } from '@/enums'
 import FileUploadPreview from '@/components/common/FileUploadPreview.vue'
 import AutoResize from 'vue-auto-resize'
 import MessageBox from '@/components/common/MessageBox.vue'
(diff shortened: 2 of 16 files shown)


doug-lovett (approved):
Looks good.

cameron-eyds (approved):
Looks good Brandon!
I have a few questions around the state management of `consumerDocumentId`.
Are we storing this in state because we are using it in subsequent document requests?
ie the first POST generates the docId and we use that affiliate all the following documents to that singular docId?

If so, it makes sense in state, that being said we may want to consider that this property will need to be cleared before the Cont-In flow is launched or exited etc. 

Either way, we will want to make sure this ID doesn't persist between Cont In Filings. 
Edge case for sure and maybe the state IS already reset between, just wanted to call it out!

src/components/ContinuationIn/AuthorizationProof.vue
```diff
@@ -277,12 +279,29 @@ export default class AuthorizationProof extends Mixins(DocumentMixin) {
...
+            size: file.size
+          } as File,
+          fileKey: res.data.documentServiceId,
+          fileName: file.name
+        })
+
+        this.setContinuationConsumerDocumentId(res.data.consumerDocumentId)
+
+        this.isFileAdded = true
         if (!res || res.status !== StatusCodes.OK) throw new Error()
```
severinbeauvais, line 305: Shouldn't you check this before adding the file to the array?

If you changed `uploadDocumentToDRS` to throw an exception on error then this code would be OK, because execution would go to the catch block and skip adding the file.

src/interfaces/store-interfaces/state-interfaces/continuation-in-state-interface.ts
```diff
@@ -4,4 +4,5 @@ export interface ContinuationInStateIF {
   continuationAuthorizationPageValid: boolean
   authorizationProof: AuthorizationProofIF
   existingBusinessInfo: ExistingBusinessInfoIF
+  consumerDocumentId: string
```
severinbeauvais, line 7: Which document is this for? In a continuation in there could be multiple authorization proof files and one optional  affidavit file. See AuthorizationProofIF and ExistingBusinessInfoIF.

src/mixins/document-mixin.ts
```diff
@@ -17,6 +17,16 @@ export default class DocumentMixin extends Vue {
       validationErrorMsg: 'Document must be set to fit onto 8.5” x 11” letter-size paper.'
     }
   }
+  readonly DOCUMENT_TYPES = {
+    contInAuthorization: {
+      class: 'CORP',
+      type: 'CNTA'
```
severinbeauvais, line 23: Can we have a more verbose type code?

Also, what about file uploads for Dissolutions (see CompleteAffidavit.vue) and Incorporations (see UploadMemorandum.vue and UploadRules.vue)?

src/mixins/document-mixin.ts
```diff
@@ -192,4 +202,52 @@ export default class DocumentMixin extends Vue {
     if (sizeKB > 1) return `${sizeKB.toFixed(0)} KB`
     return `${size} bytes`
   }
+
+  /**
+   * Uploads the specified file to Document Record Service.
+   * @param file the file to upload
+   * @param documentClass the document class defined for the document service. e.g. 'CORP'
+   * @param documentType the type of document. e.g. 'CNTA'
+   * @param consumerDocumentId the identifier of one or more documents associated with the filing.
```
severinbeauvais, line 211: How do you specify "more than one" document?

src/components/ContinuationIn/AuthorizationProof.vue
```diff
@@ -277,12 +279,29 @@ export default class AuthorizationProof extends Mixins(DocumentMixin) {
...
+      // try to upload to Document Record Service
       try {
         this.isDocumentLoading = true
-        psu = await this.getPresignedUrl(file.name)
-        const res = await this.uploadToUrl(psu.preSignedUrl, file, psu.key, this.getKeycloakGuid)
+        const res = await this.uploadDocumentToDRS(
+          file,
+          this.DOCUMENT_TYPES.contInAuthorization.class,
+          this.DOCUMENT_TYPES.contInAuthorization.type,
+          this.getContinuationInConsumerDocumentId
```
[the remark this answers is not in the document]
flutistar: ConsumerDocumentID is used to specify one or more documents.
So for the first file submission, there is no consumerDocumentId.
But if there are subsequent file submissions, these requests will have consumerDocumentId to group all of them.
e.g. If there are multiple Authorization proof files and affidavit file, all of them have the same consumerDocumentId.
I hope this answers to your question.

src/components/ContinuationIn/AuthorizationProof.vue
flutistar: It does make sense. I will upate it.

src/mixins/document-mixin.ts
flutistar: So, there is another unique ID(DocumentServiceID) for each document.
This consumerDocumentId is used to group of documents for the same application.

src/components/ContinuationIn/AuthorizationProof.vue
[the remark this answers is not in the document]
severinbeauvais: OK, that makes sense.

How would I fetch a specific document?

src/interfaces/store-interfaces/state-interfaces/continuation-in-state-interface.ts
severinbeauvais: I gather this is for all authorization proof files.

JazzarKarim (approved):
LGTM 👍

severinbeauvais (approved):
There's one outstanding comment that I'd like to see resolved but other than that, LGTM.

src/mixins/document-mixin.ts
cameron-eyds: I believe we will want to support all the `CORPS` and a few of the `OTHER` doc types listed in this config.
This could be future state however, when we move into CORPS.

https://github.com/bcgov/bcros-common/blob/main/document-service/documents-ui/src/utils/documentTypes.ts

doug-lovett (approved):
DRS integration looks good. FYI CNTA is a legacy/existing document type possibly also used in the past by the scanning application.

src/components/ContinuationIn/AuthorizationInformation.vue
```diff
@@ -285,14 +285,28 @@ export default class AuthorizationInformation extends Mixins(DateMixin, Document
     if (!documentKey || !documentName) return // safety check
 
     this.isDownloading = true
-    await this.downloadDocument(documentKey, documentName).catch(error => {
+    const documentClass = 'CORP'
```
severinbeauvais, line 288: Please create an enum for document classes.

src/interfaces/store-interfaces/state-interfaces/authorization-proof-interface.ts
```diff
@@ -4,4 +4,5 @@ export interface AuthorizationProofIF {
     fileKey: string
     fileName: string
   }>
+  consumerDocumentId?: string
```
severinbeauvais, line 7: Can we make this name consistent with `documentServiceId` (in filing-interfaces.ts)?

src/mixins/document-mixin.ts
severinbeauvais: Per Doug, CNTA may be an existing document type...

Is there an existing document type for director Affidavit?

I'm looking for consistency here. (One is an abbreviation and the other is spelled out.)

src/utils/AxiosInstance.ts
```diff
@@ -11,7 +11,8 @@ const instance = axios.create()
 instance.interceptors.request.use(
   request => {
     // don't add bearer token for Minio requests
-    if (request.url?.startsWith('https://minio')) {
+    console.log(sessionStorage.getItem)
```
severinbeauvais, line 14: Remove this before merging :)

src/components/ContinuationIn/AuthorizationInformation.vue
```diff
@@ -285,14 +285,28 @@ export default class AuthorizationInformation extends Mixins(DateMixin, Document
...
+      link.target = '_blank' // This opens the link in a new browser tab
+
+      // Append to the document and trigger the download
+      document.body.appendChild(link)
+      link.click()
+
+      // Remove the link after the download is triggered
+      document.body.removeChild(link)
+      this.isDownloading = false
+    } catch (error) {
```
severinbeauvais, line 303: In document-mixin.ts, there is existing code that looks a lot like the above...

Could the above block be moved (in its entirely) to document-mixin.ts? Whatever function you call it could throw an error that you would catch below (same as now).

Thoughts?

src/mixins/document-mixin.ts
doug-lovett: No, there is no existing CORP class doc type for director affidavit, it was added as part of this integration and as a new code it has a more meaningful value - no 4 character restriction.

severinbeauvais: OK, then disregard this conversation. Resolving it.

src/interfaces/store-interfaces/state-interfaces/authorization-proof-interface.ts
flutistar: We have both documentServiceId and consumerDocumentId as well.
Thanks

src/utils/AxiosInstance.ts
severinbeauvais: Thanks.

src/components/ContinuationIn/AuthorizationInformation.vue
severinbeauvais: Thanks for the refactor. Looks waaaay cleaner!

PS Let me resolve the conversations that I started. I use unresolved conversations to remember what to re-review. If you close them on me then I need to re-review everything (or look at all resolved convos). Thanks.

src/components/ContinuationIn/AuthorizationInformation.vue
severinbeauvais: What do you think about an enum for the document type as well? (I'd be OK with both enums in the same file, fwiw.)

src/mixins/document-mixin.ts
```diff
@@ -17,25 +18,17 @@ export default class DocumentMixin extends Vue {
...
-      type: 'CNTA'
-    },
-    affidavitDocument: {
-      class: 'CORP',
-      type: 'DIRECTOR_AFFIDAVIT'
-    }
-  }
 
   pdfjsLib: any
+  documentTypes: any
```
severinbeauvais, line 23: For cleaner code, I think you can actually do `import { DOCUMENT_TYPES as documentTypes } from ....`

src/interfaces/store-interfaces/state-interfaces/authorization-proof-interface.ts
severinbeauvais: I was thinking about this documentServiceId + documentConsumerId.

Is the first one "service id" and second one "document id"? If yes then what you have makes sense.

severinbeauvais (approved):

src/mixins/document-mixin.ts
```diff
@@ -17,25 +18,17 @@ export default class DocumentMixin extends Vue {
...
 
   // use beforeCreate() instead of created() to avoid type conflict with components that use this mixin
   async beforeCreate (): Promise<void> {
     // NB: we load the lib and worker this way to avoid a memory leak (esp in unit tests)
     // NB: must use legacy build for unit tests not running in Node 18+
     this.pdfjsLib = pdfjs
     this.pdfjsLib.GlobalWorkerOptions.workerSrc = await import('pdfjs-dist/legacy/build/pdf.worker.entry')
+    this.documentTypes = DOCUMENT_TYPES
```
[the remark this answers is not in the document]
flutistar: I used to use `documentTypes` in the components that extend `documentMixin` and that's why I redefined documentTypes.

JazzarKarim (approved):
LGTM

src/mixins/document-mixin.ts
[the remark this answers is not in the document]
severinbeauvais: My point is, instead of using a local variable to rename DOCUMENT_TYPES, you can import it with the name you want, as per my example.

src/components/ContinuationIn/AuthorizationInformation.vue
severinbeauvais: Thanks for the cleanup. Looks good.

src/mixins/document-mixin.ts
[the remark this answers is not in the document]
severinbeauvais: Thanks for the cleanup. Looks good.

severinbeauvais (approved):
Approved.

PR #765, opened 2025-02-05
Update cd.yml

severinbeauvais:
*Issue #:* None

*Description of changes:*
- removed triggers to automatically deploy feature and hotfix branches to Dev

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - main
-      - feature*
-      - hotfix*
   workflow_dispatch:
     inputs:
       target:


cameron-eyds (approved):
Looks good!

severinbeauvais:
Reviewers, what are you thoughts on this change?

This came about because a feature branch was automatically deployed to Dev when PRs were merged to it. We only realized this when a feature stopped work in Dev.

image

With this change, we would have to explicitly deploy a feature branch to Dev. I think that's much safer!

vysakh-menon-aot (approved):

cameron-eyds:
> Reviewers, what are you thoughts on this change?
> 
> This came about because a feature branch was automatically deployed to Dev when PRs were merged to it. We only realized this when a feature stopped work in Dev.
> 
> image
> 
> With this change, we would have to explicitly deploy a feature branch to Dev. I think that's much safer!

I agree with you, that a feature/hotfix branch deployment should be explicit. 
Will help avoid this scenario when the committer is unaware that `feature-xx` will trigger a deployment, even if it is convenient when everyone **is** aware.

JazzarKarim (approved):
I'm with this change. LGTM.

PR #784, opened 2025-04-28
27187 Remove isPremium handling

severinbeauvais:
*Issue #:* bcgov/entity#27187

I'm doing this separately from the other work to keep PR changes down.

*Description of changes:*
- app version = 5.14.20
- as all account are now Premium, removed conditional code
- updated unit tests

### DIFF
--- a/src/components/Registration/DefineRegistrationSummary.vue
+++ b/src/components/Registration/DefineRegistrationSummary.vue
@@ -101,18 +101,16 @@
         </v-row>
       </article>
 
-      <!-- Folio or Reference Number -->
-      <template v-if="isPremiumAccount">
-        <!-- DISABLED PER TICKET # 12306 -->
-        <!-- <v-divider class="mx-6" />
+      <!-- DISABLED PER TICKET # 12306 -->
+      <!-- <v-divider class="mx-6" /> -->
 
-        <article class="section-container">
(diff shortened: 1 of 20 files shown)


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-784-r5ari8z3.web.app

src/components/Registration/DefineRegistrationSummary.vue
```diff
@@ -101,18 +101,16 @@
         </v-row>
       </article>
 
+      <!-- DISABLED PER TICKET # 12306 -->
```
JazzarKarim, line 104: If we don't want it, can we possibly remove the code altogether instead of commenting it out Sev?

JazzarKarim (approved):
Looks good.

src/components/Registration/DefineRegistrationSummary.vue
severinbeauvais: If I remove it, I'm afraid there will be a question why it's not there. At least this way there's a reference to the ticket that made the decision (from Linda, a long time ago).

JazzarKarim: Sounds good to me, thanks. Resolving.

PR #793, opened 2025-05-28
28529 Get roles from KC token + check for any valid auth role + updated role permissions

severinbeauvais:
*Issue #:* bcgov/entity#28529

*Description of changes:*
- app version = 5.15.7
- renamed RESTORATION_FILING -> RESTORATION_REINSTATEMENT_FILING
- replaced Authorizations call with KC roles
- check for any valid auth role
- replaced FIRM_EDITABLE_COMPLETING_PARTY with EDITABLE_COMPLETING_PARTY
- deleted obsolete Auth Services call
- added authorized actions for SBC Staff, Maximus and SBC Contact Centre
- added public_user role
- added GetKeycloakRoles util function
- updated unit tests


src/App.vue
```diff
@@ -1153,7 +1157,7 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
       // set Folio Number from auth info
       // (for an incorporation, this is set in IncorporationDefineCompany.vue)
       // (for a registration, this is set in RegistrationDefineBusiness.vue)
-      this.setFolioNumber(folioNumber)
+      if (folioNumber) this.setFolioNumber(folioNumber)
```
severinbeauvais, line 1160: I did this so we don't set an undefined value.

src/App.vue
```diff
@@ -1291,21 +1295,21 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
   }
 
   /** Fetches authorizations and verifies roles. */
-  private async checkAuth (id: string): Promise<any> {
-    // NB: will throw if API error
-    const authorizations = await AuthServices.fetchAuthorizations(id)
-    const authRoles: Array<AuthorizationRoles> = authorizations.roles || []
+  private checkAuth (): void {
```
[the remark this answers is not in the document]
severinbeauvais: Yeah, but that's OK because we'll be refactoring this soon to fetch auth roles from Legal API.

src/utils/GetKeycloakRoles.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  /** Gets Keycloak JWT and parses it. */
+  function getJWT (): any {
+    // get KC token (JWT) from session storage
+    const keycloakToken = sessionStorage.getItem(SessionStorageKeys.KeyCloakToken)
+    if (!keycloakToken) throw new Error('Error getting Keycloak token')
+
+    // decode and parse the JWT
+    try {
+      const base64Url = keycloakToken.split('.')[1]
```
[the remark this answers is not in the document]
severinbeauvais: This is the same code as in other UIs. If it needs to be updated then it will be a future ticket(s).

src/utils/Authorizations.ts
```diff
@@ -1,24 +1,25 @@
...
-    case isMaximusStaff(): return MaximusStaffRoles.includes(action)
-    case isContactCentreStaff(): return ContactCentreStaffRoles.includes(action)
-    case isSbcFieldOfficeStaff(): return SbcFieldOfficeStaffRoles.includes(action)
-    default: return DefaultRoles.includes(action)
+    case isBusinessRegistryStaff(): return BusinessRegistryStaffActions.includes(action)
+    case isMaximusStaff(): return MaximusStaffActions.includes(action)
+    case isContactCentreStaff(): return ContactCentreStaffActions.includes(action)
+    case isSbcFieldOfficeStaff(): return SbcFieldOfficeStaffActions.includes(action)
+    case isPublicUser(): return PublicUserActions.includes(action)
+    default: return false
```
JazzarKarim, line 16: Interesting, OK. Adding this to list of stuff that I'll be doing for the BRD.

JazzarKarim: https://github.com/bcgov/business-edit-ui/blob/cb6e7626d98dd2a5fba56787c4a9534fc42daa7c/src/utils/Authorizations.ts#L15

I noticed you kept it as is in Edit UI Sev. Are we planning to change it there as well?

TVWerdal (approved):
Looks good Severin

src/App.vue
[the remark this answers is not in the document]
severinbeauvais: Weird. Neither handleDraftWithBusinessId nor handleDraftWithTempId call checkAuth with an id parameter.

Anyway, I left the separate calls in case we need to pass in business id or temp id to Legal API in a future call.

tests/unit/Actions.spec.ts
```diff
@@ -697,7 +698,7 @@ describe('Actions component - Conditionally disabled File and Pay button', () =>
   })
 
   it('Enables File and Pay button for regular user', async () => {
-    store.setAuthRoles([AuthorizationRoles.VIEW])
+    store.setAuthRoles([AuthorizationRoles.PUBLIC_USER])
```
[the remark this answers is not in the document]
severinbeauvais: A regular user has public_user role. I like the text making sense to human readers even if the key (role) is something more abstract.

src/utils/Authorizations.ts
severinbeauvais: > I noticed you kept it as is in Edit UI Sev. Are we planning to change it there as well?

Sorry, I didn't answer your question yet.

Ideally, yes, I'd fix this in Edit UI, but because we check for a known role, we can be sure that if it's none of the above then the last choice (default) is a regular user. So I think Edit UI is safe for now.

The above code is still better as it is less likely to cause a side effect in the future if we, say, remove the known role check in App.vue. It support a user with "no permissions", if ever needed.

JazzarKarim: > Ideally, yes, I'd fix this in Edit UI, but because we check for a known role, we can be sure that if it's none of the above then the last choice (default) is a regular user. So I think Edit UI is safe for now.

This makes so much sense.

OK, got it. Thank you Sev. I like this.

severinbeauvais:
For testing, you can compare this Voluntary Dissolution as staff (IDIR) and BCREG0020.

https://business-create-dev--pr-793-h3cop5sm.web.app/dissolution-review-confirm?accountid=2288&id=BC0888002

- only regular user sees company folio number on page 1 (Define Dissolution)
- only regular user sees company folio number on page 4 (Review and Confirm)
- only staff has editable Documents Completing Party on page 4
- only regular user sees Folio Number for this Filing on page 4
- only regular user sees certify statement as "I, xxx,. certify that..."
- only staff sees certify statement as "xxx certifies that they..."
- only staff sees Court Order/POA component
- only staff sees Staff Payment component

PR #796, opened 2025-06-17
28674 added FF toggle for Business API vs Legal API

severinbeauvais:
*Issue #:* bcgov/entity#28674

*Description of changes:*
- added keys to sample env file
- app version = 5.15.10
- added keys to vaults file
- initialize LD before fetching config
- replaced getAccountId with new getCurrentAccountId getter
- added extra request headers for GW
- added new FF
- added FF check in FetchConfig to use Legal API or Business API GW


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-796-6q3pvkv4.web.app

TVWerdal (approved):

src/store/store.ts
```diff
@@ -86,7 +86,11 @@ import { IsAuthorized } from '@/utils/Authorizations'
...
   getters: {
-    /** The Account ID, from session storage. */
+    /**
+     * The Account ID, from session storage.
+     * This is the same value as in the current account object.
+     * @deprecated use getCurrentAccountId instead
+     */
     getAccountId (): string {
       // if we can't get account id from ACCOUNT_ID
       // then try to get it from CURRENT_ACCOUNT
```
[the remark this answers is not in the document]
severinbeauvais: Done.

src/utils/FetchConfig.ts
```diff
@@ -34,10 +33,18 @@ export async function FetchConfig (): Promise<any> {
   const businessDashUrl: string = import.meta.env.VUE_APP_BUSINESS_DASH_URL
   sessionStorage.setItem('BUSINESS_DASH_URL', businessDashUrl)
 
-  const legalApiUrl: string =
-    (import.meta.env.VUE_APP_LEGAL_API_URL + import.meta.env.VUE_APP_LEGAL_API_VERSION_2 + '/')
-  // set base URL for axios calls
-  axios.defaults.baseURL = legalApiUrl
+  // set Legal API URL or Business API GW URL depending on FF
+  if (GetFeatureFlag('use-business-api-gw-url')) {
```
severinbeauvais, line 37: This FF is called well before the user info is fetched and reported to LD, so user targeting won't work.

Let me think about this a bit.

severinbeauvais: OK, so I have to set `axios.defaults.baseURL` because it's needed by the StaffComments shared component (which accepts the axios instance) and I didn't feel like re-writing it.

And I might as well set baseURL here because there's no better place.

You'll notice that I didn't prepend all Legal Services calls with the URL (as in the other services). This is because of a clash between baseURL and the prepends causing 74 test suite to fail, and I didn't want to re-write those either.

Finally, I'll admit that the current solution does not support targeted FFs (I tested it), because we make several calls to Legal API even before we report the account info to LD. I started re-writing that and decided there's more important work to do.

What do you think?

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-796-6q3pvkv4.web.app

src/utils/FetchConfig.ts
severinbeauvais: @JazzarKarim @TVWerdal , are you OK with my decisions above?

JazzarKarim: This sounds good to me. FWIW, I'm having similar issues in the BRD; a race condition where we use Legal API before LD has been initialized. This is OK to me and I get the dillema.

PR #797, opened 2025-06-18
28674 small updates for consistency and to prevent future test issues

severinbeauvais:
*Issue #:* bcgov/entity#28674

*Description of changes:*
- app version = 5.15.11
- renamed getCurrentAccountId -> getAccountId
- replaced store getter with function to get account id in Axios interceptor

### DIFF
--- a/src/resources/BreadcrumbResource.ts
+++ b/src/resources/BreadcrumbResource.ts
@@ -31,7 +31,7 @@ function getNumberedEntityName (): string {
 
 /** Returns URL param string with Account ID if present, else empty string. */
 function getParams (): string {
-  const accountId = store.getCurrentAccountId
+  const accountId = store.getAccountId
   return accountId ? `?accountid=${accountId}` : ''
 }
 
@@ -56,7 +56,7 @@ export function getBcRegistriesDashboardBreadcrumb (): BreadcrumbIF {
 export function getMyBusinessRegistryBreadcrumb (): BreadcrumbIF {
   return {
     text: 'My Business Registry',
-    href: `${sessionStorage.getItem('BUSINESSES_URL')}account/${store.getCurrentAccountId}/business`
+    href: `${sessionStorage.getItem('BUSINESSES_URL')}account/${store.getAccountId}/business`
   }
 }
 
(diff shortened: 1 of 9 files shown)


src/resources/BreadcrumbResource.ts
```diff
@@ -31,7 +31,7 @@ function getNumberedEntityName (): string {
 
 /** Returns URL param string with Account ID if present, else empty string. */
 function getParams (): string {
-  const accountId = store.getCurrentAccountId
+  const accountId = store.getAccountId
```
severinbeauvais, line 34: Now has same name as in Edit UI.

I may update Filings UI to match.

src/utils/AxiosInstance.ts
```diff
@@ -26,11 +21,22 @@ instance.interceptors.request.use(
...
+    function getAccountId (): string {
+      // if we can't get account id from ACCOUNT_ID
+      // then try to get it from CURRENT_ACCOUNT
+      let accountId = sessionStorage.getItem('ACCOUNT_ID')
+      if (!accountId) {
+        const currentAccount = sessionStorage.getItem('CURRENT_ACCOUNT')
+        accountId = JSON.parse(currentAccount)?.id
+      }
+      return accountId
+    }
```
severinbeauvais, line 39: Two UIs have seen unit test issues when getting account id from the store so I'm falling back to this code here as well. (The previous code worked but a future change might break it).

src/resources/BreadcrumbResource.ts
TVWerdal: I think there is already a "getAccountId" in filings, so that will need to be accounted for.

src/utils/AxiosInstance.ts
```diff
@@ -26,11 +21,22 @@ instance.interceptors.request.use(
...
     // add these headers only if Vitest isn't running as it breaks some tests
     if (import.meta.env.VITEST === undefined) {
-      request.headers.common['Account-Id'] = store.getCurrentAccountId
+      request.headers.common['Account-Id'] = getAccountId()
       request.headers.common['X-Apikey'] = import.meta.env.VUE_APP_BUSINESS_API_KEY
     }
 
     return request
+
+    function getAccountId (): string {
```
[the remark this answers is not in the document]
severinbeauvais: @TVWerdal I'm updating this PR and will update Edit UI to fix this. Do you want to fix Filings UI likewise?

src/resources/BreadcrumbResource.ts
severinbeauvais: I was thinking about getting rid of the session-based getAccountId in Filings UI (except in the axios file) but no rush on that.

PR #798, opened 2025-06-18
28603 fetch authorized actions from permissions endpoint

severinbeauvais:
*Issue #:* bcgov/entity#28603

*Description of changes:*
- app version = 5.16.0
- now store authorized actions instead of auth roles
- retain auth roles for reporting user to LD
- added call to fetch permissions from Legal API
- updated Authorized Actions enum to be strings instead of numbers
- added service call for permissions
- greatly simplified IsAuthorized()
- created file with authorized actions lists for unit tests
- renamed office-address-schema -> address-schemas
- renamed Authorized.ts -> IsAuthorized.ts
- updated unit tests

### DIFF
--- a/src/enums/authorizedActions.ts
+++ b/src/enums/authorizedActions.ts
@@ -1,7 +1,4 @@
-/**
- * List of actions authorized via roles. See Authorizations module.
- * Ultimately, this will be combined into the Authorization Roles enum.
- */
+/** List of authorized actions (permissions). See IsAuthorized(). */
 export enum AuthorizedActions {
   AMALGAMATION_FILING = 'AMALGAMATION_FILING',
   AML_OVERRIDES = 'AML_OVERRIDES',


src/enums/authorizedActions.ts
```diff
@@ -3,29 +3,29 @@
...
+  NO_CONTACT_INFO = 'NO_CONTACT_INFO',
+  REGISTRATION_FILING = 'REGISTRATION_FILING',
+  RESTORATION_REINSTATEMENT_FILING = 'RESTORATION_REINSTATEMENT_FILING',
+  SAVE_DRAFT = 'SAVE_DRAFT',
+  SBC_BREADCRUMBS = 'SBC_BREADCRUMBS',
+  STAFF_BREADCRUMBS = 'STAFF_BREADCRUMBS',
+  STAFF_COMMENTS = 'STAFF_COMMENTS',
+  STAFF_PAYMENT = 'STAFF_PAYMENT',
+  THIRD_PARTY_CERTIFY_STMT = 'THIRD_PARTY_CERTIFY_STMT',
+  VOLUNTARY_DISSOLUTION_FILING = 'VOLUNTARY_DISSOLUTION_FILING' // non-firm voluntary dissolution
```
severinbeauvais, line 30: We get the strings, not enum values (eg, 14) from the API, so this enum has to be string values.

tests/unit/test-data/authorizedActions.ts
severinbeauvais, line 1: I moved these static list into this folder in the unit tests. I figure we'll need this (and possibly a helper function) to prep the tests with the necessary authorized actions depending on what account role we're testing.

src/App.vue
```diff
@@ -1209,21 +1205,34 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
...
+    async function getCurrentAccount (): Promise<any> {
+      let account = null
+      for (let i = 0; i < 50; i++) {
+        const currentAccount = sessionStorage.getItem(SessionStorageKeys.CurrentAccount)
+        account = JSON.parse(currentAccount)
+        if (account) break
+        await Sleep(100)
+      }
+      return account
+    }
```
severinbeauvais, line 1222: I nested this as it's the only place it's used.

tests/unit/setAuthRole.ts
```diff
@@ -0,0 +1,19 @@
...
+export function setAuthRole (authRole: AuthorizationRoles = null): void {
+  switch (authRole) {
+    case AuthorizationRoles.CONTACT_CENTRE_STAFF: return store.setAuthorizedActions(Actions.ContactCentreStaffActions)
+    case AuthorizationRoles.MAXIMUS_STAFF: return store.setAuthorizedActions(Actions.MaximusStaffActions)
+    case AuthorizationRoles.PUBLIC_USER: return store.setAuthorizedActions(Actions.PublicUserActions)
+    case AuthorizationRoles.SBC_STAFF: return store.setAuthorizedActions(Actions.SbcFieldOfficeStaffActions)
+    case AuthorizationRoles.STAFF: return store.setAuthorizedActions(Actions.BusinessRegistryStaffActions)
+    default: return store.setAuthorizedActions([])
+  }
+}
```
severinbeauvais, line 19: Test helper to set authorized actions depending on auth role.

tests/unit/AmalgamationInformation.spec.ts
```diff
@@ -50,12 +51,13 @@ for (const test of amalgamationBusinessInfo) {
           entityType: test.entityType,
           tombstone: {
             filingType: FilingTypes.AMALGAMATION_APPLICATION,
-            authRoles: [AuthorizationRoles.STAFF]
+            authorizedActions: []
```
severinbeauvais, line 54: Some of the unit tests failed when I left this out (since we're setting the auth role on line 60 so it seems like this isn't needed).

However, I found that some components checked for authorized actions right away, and if we set the tombstone object without the authorizedActions property, trying to de-reference that property causes an exception.

So, I simply initialize it here and that problem doesn't happen.

tests/set-auth-role.ts
```diff
@@ -0,0 +1,16 @@
+import { AuthorizationRoles } from '@/enums'
+import * as Actions from './unit/test-data/AuthorizedActionsLists'
+
+/** Sets the authorized actions (aka permissions) according to Auth Role. */
+export function setAuthRole (store: any, authRole: AuthorizationRoles = null): void {
```
severinbeauvais, line 5: I initially declared the store in this function but it caused conflicts with store.resourceModel, which makes no sense whatsoever. Anyway, this works nicely.

src/App.vue
```diff
@@ -1209,21 +1204,34 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
...
       await this.loadOrgInfo(accountInfo?.id)
     }
+
+    /**
+     * Gets current account from object in session storage.
+     * Waits up to 5 sec for current account to be synced (typically by SbcHeader).
+     */
+    async function getCurrentAccount (): Promise<any> {
+      let account = null
+      for (let i = 0; i < 50; i++) {
```
JazzarKarim, line 1214: Question about this. What if we move to use a while loop that breaks when the account is synced? This way this function will work in case it took 6 seconds for the current account to be synced. No need to change, just wondering.

severinbeauvais: If there's a serious problem then such a while loop would be infinite. By limiting the wait to 5 seconds, we don't block forever and something will fail and report the error.

In my experience, the account is usually synced in 200-300 milliseconds.

severinbeauvais:
> Not sure if you're planning to add more, but this LGTM

That's it unless I find some tweaks to make (as I work to make Edit UI consistent with this).

PR #806, opened 2025-07-14
Document Service Record Integration

flutistar:
*Issue #:* /bcgov/entity#29449

*Description of changes:*
- Replaced Minio with Document Record Service
- DocumentID component


package.json
```diff
@@ -31,17 +31,19 @@
     "@bcrs-shared-components/court-order-poa": "3.0.11",
     "@bcrs-shared-components/date-picker": "1.3.3",
     "@bcrs-shared-components/document-delivery": "1.2.0",
+    "@bcrs-shared-components/document-id": "^1.0.5",
     "@bcrs-shared-components/effective-date-time": "1.1.15",
-    "@bcrs-shared-components/enums": "1.1.11",
+    "@bcrs-shared-components/enums": "^1.1.23",
```
severinbeauvais, line 36: Please remove `^` and re-run npm install.

As a general rule, we want to use fixed/known versions of the shared components, otherwise an update might change something and we wouldn't know about it.

src/utils/FetchConfig.ts
```diff
@@ -83,6 +83,13 @@ export async function FetchConfig (): Promise<any> {
   const keycloakClientId: string = import.meta.env.VUE_APP_KEYCLOAK_CLIENTID;
   (<any>window).keycloakClientId = keycloakClientId
 
+  // Document Record Service Configration
+
+  const docApiKey: string = import.meta.env.VUE_APP_DOC_API_KEY
+  sessionStorage.setItem('DOC_API_KEY', docApiKey)
```
severinbeauvais, line 89: I recommend not storing the API key in session storage. Instead, you can get this from import.meta.xxx right where you actually use it.

.env.example
```diff
@@ -53,3 +53,8 @@ VUE_APP_GENESYS_ID=
 VUE_APP_WEBCHAT_URL=
 VUE_APP_WEBCHAT_REASON=
 VUE_APP_WEBCHAT_STATUS_URL=
+
+#Document Record Service
+VUE_APP_DOC_API_KEY=
+VUE_APP_DOC_API_URL="https://test.api.connect.gov.bc.ca/doc-dev"
+VUE_APP_DOC_API_VERSION="/api/v1"
```
severinbeauvais, line 60: You also need to add these to the vaults.json file.

Also, the API key will need to be added to the CI trigger config in GCP (for injecting the key into a preview build).

package.json
severinbeauvais: Fixed, thanks.

src/enums/authorizedActions.ts
```diff
@@ -6,6 +6,7 @@ export enum AuthorizedActions {
   BLANK_COMPLETING_PARTY = 'BLANK_COMPLETING_PARTY',
   CONTINUATION_IN_FILING = 'CONTINUATION_IN_FILING',
   COURT_ORDER_POA = 'COURT_ORDER_POA',
+  DOCUMENT_RECORDS = 'DOCUMENT_RECORDS',
```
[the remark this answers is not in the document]
severinbeauvais: The proper way to add the db records is with a migration script (in LEAR repo).

While you're there, there are a couple of other changes to the same table that I'd like to make. Add me to the LEAR migration PR and I'll cite those there.

src/utils/FetchConfig.ts
severinbeauvais: I think you've made the change to the bcrs-shared component that needs this API key, right? Can these 2 lines be removed now?

src/views/Incorporation/IncorporationReviewConfirm.vue
```diff
@@ -220,6 +220,29 @@
       </v-card>
     </section>
 
+    <!-- Document ID Component for Staff only -->
+    <section
+      v-if="IsAuthorized(AuthorizedActions.STAFF_PAYMENT) && !isEntityCoop"
```
severinbeauvais, line 225: Please use `IsAuthorized(AuthorizedActions.DOCUMENT_RECORDS)` as in the previous component.

src/enums/authorizedActions.ts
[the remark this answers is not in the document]
flutistar: I got this from the '[Entities User Types and Permissions](https://docs.google.com/spreadsheets/d/13P4ZA19lJfXJX95N5OfQMu6jYuotDH6h2hDf2DnAv0E/edit?gid=1335805235#gid=1335805235)' spreadsheet and confirmed with Megan that it exists in the database.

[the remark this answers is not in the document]
severinbeauvais: You'll need to do this before your code can be tested.

PR #812, opened 2025-08-20
30251 fix address not found

bobdev-94:
*Issue #:* /bcgov/entity#30251

*Description of changes:*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-create-ui",
-  "version": "5.16.8",
+  "version": "5.16.9",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "business-create-ui",
-      "version": "5.16.8",
+      "version": "5.16.9",
       "dependencies": {
         "@babel/compat-data": "^7.21.5",
         "@bcrs-shared-components/approval-type": "1.1.20",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "5.16.8",
+  "version": "5.16.9",
   "private": true,
   "appName": "Business Create UI",
   "sbcName": "SBC Common Components",


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-812-b1k9x6it.web.app

src/mixins/amalgamation-mixin.ts
```diff
@@ -396,7 +396,6 @@ export default class AmalgamationMixin extends Vue {
   async updatePrepopulatedData (business: AmalgamatingBusinessIF, isNew = false): Promise<void> {
     // safety checks
     if (!business || business.type !== AmlTypes.LEAR) throw new Error('Invalid business')
-    if (!business.addresses) throw new Error('Missing business addresses')
```
severinbeauvais, line 399: What issue is this PR fixing, exactly? The ticket doesn't mention amalgamations...

src/services/legal-services.ts
```diff
@@ -286,7 +288,7 @@ export default class LegalServices {
         return data
       }).catch(error => {
         if (error?.response?.status === StatusCodes.NOT_FOUND) {
-          return null // Business or Address not found (not an error)
+          return {} // Business or Address not found (not an error)
```
severinbeauvais, line 291: Should this return `{ businessOffice: null }` as in Business Dashboard?

Ref: https://github.com/bcgov/business-dashboard-ui/blob/6fcb95388d2ab122a73bb5e4af6ac38bfb512389/src/stores/business.ts#L128

src/App.vue
```diff
@@ -1350,8 +1350,6 @@ export default class App extends Mixins(CommonMixin, DateMixin, FilingTemplateMi
 
     if (addresses) {
       this.setOfficeAddresses(addresses)
-    } else {
-      throw new Error('Invalid office addresses')
```
severinbeauvais, line 1354: I was initially concerned because I thought this code was used for multiple filing types. However, I see that it's only called for restorations, so this is OK.

Can you please add a comment before line 1347 indicating that this method is only called for restorations?

src/mixins/amalgamation-mixin.ts
bobdev-94: `business.addresses` set up from the `fetchAddress`.
So, if the address not found is not error, I thought this line is not needed more.

src/App.vue
severinbeauvais: Thanks. Resolving.

src/mixins/amalgamation-mixin.ts
severinbeauvais: This line of code is used to updated the (amalgama)"ting" business info. This business should always have a business address.

However, you can no longer check for `business.address` since a fetch error used to return null but now it returns an object with a null businessOffice.

One way to fix this would be to check for that here. Another way is to keep returning null when fetching, but then update App.vue to assign the proper "empty address" object.

bobdev-94: If then, I will revert back it.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-812-b1k9x6it.web.app

TVWerdal (approved):
Good fix Bobby - all tests look great.

PR #848, opened 2026-03-19
32772 Updates number of shares and par values input/display

severinbeauvais:
*Issue #:* bcgov/entity#32772

*Description of changes:*
- app version = 5.18.1
- added base class for medium gray text
- replace local class with base class
- replaced FormatCurrency by FormatDecimal
- made "actions" table header conditional
- fixed/updated table layout so long numbers/values don't extend past cells
- changed number of shares text field and par value text field to use string v-models (but retain type="number")
- added helper to format numbers to strings
- added isFinite check to maximum shares
- added max length for maximum shares to 20 chars
- added SignificantDigits check to maximum shares
- added max length for par value to 38 chars
- updated validation messages for consistency
- assigned v-models on created
- saved v-models on save
- deleted FormatCurrency
- added FormatDecimal (custom implementation to work around limitations of toLocaleString)
- added / updated unit tests


tests/unit/ShareStructure.spec.ts
```diff
@@ -594,18 +643,6 @@ describe('Share Structure component', () => {
     wrapper.destroy()
   })
 
-  it('Accepts valid 15 digit maximum shares', async () => {
```
severinbeauvais, line 597: I didn't delete this. I moved it up in this file to keep similar tests together.

src/components/common/ShareStructure.vue
```diff
@@ -295,18 +297,24 @@ export default class ShareStructure extends Mixins(CurrencyLookupMixin) {
     if (this.hasNoParValue) return []
 
     return [
-      v => (v !== '' && v !== null && v !== undefined) || 'Par value is required',
-      v => Number.isFinite(Number(v)) || 'Par value must be a valid number',
-      v => v > 0 || 'Amount must be greater than 0',
-      v => SignificantDigits(v) <= 16 || 'Amount has too many significant digits'
+      (v: string) => (v !== '' && v !== null && v !== undefined) || 'Par value is required',
```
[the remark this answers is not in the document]
severinbeauvais: As is, the inputs allow scientific notation. The goal here was to not automatically to scientific notation on input of very small or very large numbers, and to always display in plain notation.

src/components/common/ShareStructure.vue
```diff
@@ -295,18 +310,18 @@ export default class ShareStructure extends Mixins(CurrencyLookupMixin) {
...
+      (v: string) => Number.isFinite(+v) || 'Par value must be a valid number',
+      (v: string) => (+v > 0) || 'Amount must be greater than 0',
+      (v: string) => (v.length <= 38) || 'Maximum 38 characters',
+      (v: string) => (SignificantDigits(v) <= 16) || 'Amount has too many significant digits'
     ]
```
severinbeauvais, line 316: The requirement is:
Image

These rules allow up to 38 characters, but the check for significant digits may limit this. Ie:

10,000,000,000,000,000,000,000,000,000,000,000,000 is valid (not including commas) - is stored as 1.0E37
12,345,678,901,234,567,890,123,456,789,012,345,678 is invalid (not including commas) - insufficient precision for mantissa

Anyways, the max significant digits from COLIN is "7.30246823426" (12 sig digits).

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-848-81i003db.web.app

loneil (changes requested):
Think the handling of accidental non-numeric (hit a letter by mistake, accidentally hit second decimal, etc etc) could use a quick look.

src/components/common/ShareStructure.vue
```diff
@@ -42,15 +42,14 @@
...
-                            v-model.number="shareStructure.maxNumberOfShares"
                             filled
                             label="Maximum Number of Shares"
                             persistent-hint
-                            type="number"
-                            hide-spin-buttons
                             :hint="'Enter the maximum number of shares in the ' + shareStructure.type"
                             :rules="getMaximumShareRule()"
                             :disabled="hasNoMaximumShares"
+                            :value="numberToString(shareStructure.maxNumberOfShares)"
```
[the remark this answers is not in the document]
severinbeauvais: ~Fixed. Please test again!~

severinbeauvais:
All right... Good to go for your (re)reviews, please.

This was surprisingly annoying because our requirements didn't like up very welll with native JS support for some things.

Anyways, I made sure there were unit tests for everything so it should all be fine now 😅

severinbeauvais:
Argh! One more little thing found during testing: very small numbers turn into "0" when editing. I know the issue and will have fix asap.

---

Update: Fixed by using FormatDecimal to convert number to string in the input component. A side effect of this (thought it existed before) is if you start with 0.0001 and then you remove the 1, you get zero (since you just changed your number to zero). I'll try to find a way around this...

---

Update: Fixed by NOT reformatting the number "on input". The input component now works with a string v-model, which is only converted to a number when the user clicks Done.

src/components/common/ListShareClass.vue
```diff
@@ -241,7 +241,7 @@
 import { Component, Emit, Prop, Vue } from 'vue-property-decorator'
 import { RouteNames } from '@/enums'
 import { arrayMoveMutable } from 'array-move'
-import { FormatCurrency } from '@/utils'
+import { FormatDecimal } from '@/utils'
```
severinbeauvais, line 244: FormatDecimal is more generic and works around the limitations of toLocaleString (specifically, max 21 fraction digits).

src/components/common/ShareStructure.vue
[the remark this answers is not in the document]
severinbeauvais: I went a different way with this. Please test again.

src/components/common/ShareStructure.vue
```diff
@@ -40,9 +40,10 @@
                     <template #label>
                       <v-row>
                         <v-col cols="6">
+                          <!-- Use type="number" to restrict input to numbers even though the v-model is a string. -->
```
severinbeauvais, line 43: Using "type=number" lets this component internally restrict the input. It supports:
- whole numbers
- decimal numbers
- negative numbers
- scientific notation

This is the best UX since the user is prevented from entering invalid data and can also edit the number (eg, backspace) without us re-writing it at this stage.

During input, it's whatever the user types in (as long as it also passes the rules). Once Done is clicked, the number is formatted (no more scientific notation). This formatted number is used for display and subsequent edit.

I think it's working well. Let me know what you think.

src/components/common/ShareStructure.vue
```diff
@@ -371,14 +393,14 @@ export default class ShareStructure extends Mixins(CurrencyLookupMixin) {
 
   changeMaximumShareFlag (): void {
     if (this.hasNoMaximumShares) {
-      this.shareStructure.maxNumberOfShares = null
+      this.maxNumberOfShares = ''
```
severinbeauvais, line 396: Pre-empts some validations if you change from no max shares to max shares .

src/utils/FormatDecimal.ts
severinbeauvais, line 1: See unit tests for validation of this code.

tests/unit/FormatDecimal.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+
+describe('FormatDecimal', () => {
+  it('has default options', () => {
+    // minDecimals = 0
+    // grouping = false
+    // locale = 'en-US'
+    expect(FormatDecimal(1000000, {})).toBe('1000000')
+    expect(FormatDecimal(1000000, { minDecimals: 2 })).toBe('1000000.00')
+    expect(FormatDecimal(1000000, { grouping: true })).toBe('1,000,000')
+    expect(FormatDecimal(1000000, { grouping: true, locale: 'en-IN' })).toBe('10,00,000')
```
severinbeauvais, line 11: We don't use anything other than en-US but this tests another locale, for fun :)

src/components/common/ShareStructure.vue
```diff
@@ -226,10 +229,18 @@ export default class ShareStructure extends Mixins(CurrencyLookupMixin) {
...
+  maxNumberOfShares = '' // v-model, which is converted to number before saving
+  parValue = '' // v-model, which is converted to number before saving
 
   readonly excludedWordsListForClass: string [] = ['share', 'shares', 'value']
   readonly excludedWordsListForSeries: string [] = ['share', 'shares']
 
+  /** Returns a value formatted for use as v-model. */
+  numberToString (value: number): string {
+    // if value is null, return empty string (instead of invalid number)
+    return (value !== null) ? FormatDecimal(value, { minDecimals: 0, grouping: false }) : ''
```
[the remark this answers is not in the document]
severinbeauvais: I think this is overkill. "numberToString" works as expected whether the input value is a number or null.

src/components/common/ShareStructure.vue
loneil: Sorry I haven't followed back and forth on scientific notation but is there a reason we would need to allow that?
Like if I go and try and enter 33 and accidentally hit an E in there I get 3e3 and now it's saying my par value is 3000. It's allowing alpha entry for the letter e too then. Does it bust down the complexity of this to just not allow scientific notation?

image

image

severinbeauvais: I see no reason to prevent the entry of scientific notation -- and I probably can't prevent it anyway since it's built-in to the input component -- as long as we display the value correctly after it's entered.

Yes, sometimes entering an "E" works unexpectedly. Cutting and pasting a value in scientific notation always works properly. I dunno, maybe a bug in the input component?

I could probably add a regex to the rules to prevent this. Thoughts?

src/utils/FormatDecimal.ts
```diff
@@ -0,0 +1,67 @@
...
+    const [_intPart, fracPart = ''] = coeff.split('.')
+    let intPart = _intPart
+
+    const sign = intPart.startsWith('-') ? '-' : ''
+    if (sign) intPart = intPart.slice(1)
+
+    if (exp > 0) {
+      const combined = intPart + fracPart
+      const newInt = combined.padEnd(intPart.length + exp, '0')
+      const newFrac = combined.slice(intPart.length + exp)
```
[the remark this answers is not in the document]
severinbeauvais: This sounds intelligent but is incorrect! The present code works fine. I added the following unit test to verify and it passes:

```
  it('expands unusual scientific notation correctly', () => {
    // exponent smaller than fractional length
    expect(FormatDecimal(1.234E-2, { minDecimals: 0, grouping: false })).toBe('0.01234')
    expect(FormatDecimal(1.234E2, { minDecimals: 0, grouping: false })).toBe('123.4')
  })
```

image

src/utils/FormatDecimal.ts
```diff
@@ -0,0 +1,67 @@
...
+      str = sign + newInt + (newFrac ? '.' + newFrac : '')
+    } else if (exp < 0) {
+      const combined = intPart + fracPart
+      const zeros = '0'.repeat(Math.abs(exp) - intPart.length)
+      const newFrac = zeros + combined
+      str = sign + '0.' + newFrac
```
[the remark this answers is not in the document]
severinbeauvais: I added `expect(FormatDecimal(12E-1, { minDecimals: 0, grouping: false })).toBe('1.2')` and it passes.

src/components/common/ShareStructure.vue
loneil: I don't think there's any benefit of letting someone enter exponent for a type of input like this and it's introducing custom format handling. Just seems like a bug vector rather than a feature (yeah could validate against "e" with the other validations it since it's a `number` type unless that adds worse complexity). It's probably minor but IMO the typo risk and formatting complexity are more likely to cause issues than someone being confused they couldn't copy "100e3" or something for their share price?

loneil: The range validation of it works though so if kept it doesn't seem to cause invalid data getting in.

src/utils/FormatDecimal.ts
```diff
@@ -0,0 +1,67 @@
...
+  // Enforce minimum decimals
+  if (fracPart.length < minDecimals) {
+    fracPart = fracPart.padEnd(minDecimals, '0')
+  }
+
+  // Apply grouping only if requested
+  const groupedInt = grouping
+    ? Number(intPart).toLocaleString(locale)
+    : intPart
+
```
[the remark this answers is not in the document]
severinbeauvais: I'm assuming, if a number is more than 16 characters long, the remaining digits are zero (otherwise there are more than 16 significant digits, which we check for).

This test passes and I'm happy with that:

```
  it('formats very large numbers properly with grouping', () => {
    // max 38 characters input
    expect(FormatDecimal(1E37, { minDecimals: 0, grouping: true }))
      .toBe('10,000,000,000,000,000,000,000,000,000,000,000,000')
  })
```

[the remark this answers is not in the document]
severinbeauvais: I've also added the following unit test, which passes:

```
  it('handles max-significant-digits numbers', () => {
    // max 38 characters input
    expect(FormatDecimal(1234567890123456, { minDecimals: 0, grouping: false })).toBe('1234567890123456')
    expect(FormatDecimal(1.234567890123456E15, { minDecimals: 0, grouping: false })).toBe('1234567890123456')
    expect(FormatDecimal(1.234567890123456E-1, { minDecimals: 0, grouping: false })).toBe('0.1234567890123456')
  })
```

src/components/common/ListShareClass.vue
```diff
@@ -256,32 +256,35 @@ export default class ListShareClass extends Vue {
...
+      { text: 'Currency', value: 'currency', class: 'share-structure-header' },
+      { text: 'Special Rights or Restrictions', value: 'hasRightsOrRestrictions', class: 'share-structure-header' }
+    ]
+    if (!this.isSummary) headers.push({ text: '', value: 'actions' })
+    return headers
+  }
+
+  /** Returns a par value formatted for display in the shares table. */
   formatParValue (item: any): string {
     if (!item.parValue) return 'No Par Value'
```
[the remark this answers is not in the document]
severinbeauvais: Zero is invalid.

The suggested change won't allow zero, which is incorrect.

src/components/common/ShareStructure.vue
severinbeauvais: `expect(FormatDecimal(100E3, { minDecimals: 0, grouping: false })).toBe('100000')` passes.

Do you feel strongly that I should prevent scientific input? I'm on the fence.

severinbeauvais: Adding rule `(v: string) => /^-?\d+(\.\d+)?$/.test(v) || 'Unsupported scientific notation',` yields:

image

loneil: Not really, if you think it cuts code out and necessary testing I'd just validate it out for the simplification benefit, but if it's working as-is then it's not too important (unless there's some UX objection or anything).

src/components/common/ShareStructure.vue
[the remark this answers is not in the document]
severinbeauvais: Resolving this obsolete conversation. See newer conversations below.

src/components/common/ShareStructure.vue
severinbeauvais: Asking UX team now :)

severinbeauvais: FWIW, FormatDecimal needs to handle this anyway because we get scientific-notation numbers from the API (JSON number format) and need to format them for display in "unscientific" notation.

src/utils/FormatDecimal.ts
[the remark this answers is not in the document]
loneil: Think a string input would be what borks it
`expect(FormatDecimal('1.234e2', { minDecimals: 0, grouping: false })).toBe('123.4')`??

src/components/common/ShareStructure.vue
severinbeauvais: image

So, I'm not blocking input of scientific notation numbers.

src/utils/FormatDecimal.ts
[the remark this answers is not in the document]
severinbeauvais: You are right!

Let me look into this.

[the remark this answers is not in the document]
severinbeauvais: OK, now the suggested fix works.

I've submitted the change and an extra unit test.

src/utils/FormatDecimal.ts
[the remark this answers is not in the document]
severinbeauvais: I added this fix and a unit test to verify "12e-1".

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-848-81i003db.web.app

SB says, try this:
Shares page: https://business-create-dev--pr-848-81i003db.web.app/incorporation-share-structure?id=TC2Qh2MLFt
Review page: https://business-create-dev--pr-848-81i003db.web.app/incorporation-review-confirm?id=TC2Qh2MLFt

Compare with Dev:
Shares page: https://dev.create.business.bcregistry.gov.bc.ca/incorporation-share-structure?id=TC2Qh2MLFt
Review page: https://dev.create.business.bcregistry.gov.bc.ca/incorporation-review-confirm?id=TC2Qh2MLFt

loneil (approved):
Think this looks good to me now. Can't find anything in particular to break the validation/format on the frontend. QA doing some end to end with the edge cases will be good (maybe could try in dev to catch anything early)

meawong (approved):
Nice! Tried a few cases in the UI and LGTM

PR #853, opened 2026-04-28
32791: Dissolution - Adjust custodian name validation

loneil:
*Issue #:* /bcgov/entity#32791

*Description of changes:*
On the frontend set the specific dissolution name entry component to have a length of 30 for first and middle name.
Confirmed this migrates as as corp party to Colin which is a varchar(30).

A bit awkward with the different name length validations building up here... but need to keep the existing first/middle name validations on 20 since it covers the completing party (which is **20** in Colin!) at the same time as director/incorporator. That part has been discussed separately, probably better to simplify this stuff with common name controls when moving to the more modern nuxt UIs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "business-create-ui",
-  "version": "5.18.2",
+  "version": "5.18.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "business-create-ui",
-      "version": "5.18.2",
+      "version": "5.18.3",
       "dependencies": {
         "@babel/compat-data": "^7.21.5",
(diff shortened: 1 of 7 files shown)


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-853-p3dhfgkn.web.app

Can check with BCREGA0001
https://dev.create.business.bcregistry.gov.bc.ca/dissolution-define-dissolution?accountid=3896&businessid=BC0888529&id=BC0888529
has 20 char limit

with change
https://business-create-dev--pr-853-p3dhfgkn.web.app/dissolution-define-dissolution?accountid=3896&businessid=BC0888529&id=BC0888529
has 30

kialj876 (approved):
Nice work 👍

meawong (approved):
LGTM

PR #857, opened 2026-05-05
33185 - prevent share class validation on initial render

ikrb01:
*Issue #33185:* 
https://github.com/bcgov/entity/issues/33185

## Summary

Fixes an issue where validation was triggered immediately when opening the **Add Share Class** form.

## Problem

When clicking “Add Share Class”, validation errors were shown before any user interaction.
This was caused by the form being programmatically validated on component creation.

## Solution

Removed the automatic validation call on initial render in `ShareStructure.vue`.
Validation now only runs when the user submits the form.

## Result

* No validation errors shown on initial form load
* Validation behaves correctly on submit
* Improved user experience when adding a share class
@kialj876  @meganawong @loneil

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "5.18.7",
+  "version": "5.18.8",
   "private": true,
   "appName": "Business Create UI",
   "sbcName": "SBC Common Components",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "@bcrs-shared-components/base-address": "2.2.4",
     "@bcrs-shared-components/breadcrumb": "2.1.15",
     "@bcrs-shared-components/business-lookup": "1.3.15",
-    "@bcrs-shared-components/certify": "2.2.3",
+    "@bcrs-shared-components/certify": "2.2.4",
     "@bcrs-shared-components/completing-party": "2.1.100",
     "@bcrs-shared-components/confirm-dialog": "1.2.1",
     "@bcrs-shared-components/contact-info": "1.2.15",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "5.18.8",
+  "version": "5.18.7",
   "private": true,
   "appName": "Business Create UI",
   "sbcName": "SBC Common Components",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "5.18.5",
+  "version": "5.18.7",
   "private": true,
   "appName": "Business Create UI",
   "sbcName": "SBC Common Components",
--- a/src/components/common/ShareStructure.vue
+++ b/src/components/common/ShareStructure.vue
@@ -340,11 +340,6 @@ export default class ShareStructure extends Mixins(CurrencyLookupMixin) {
         const name = this.shareStructure.name
         this.shareStructure.name = name.substr(0, name.indexOf(' Shares'))
       }
-
-      // THIS IS THE PROBLEM (triggers validation on open): validate initial data 
-      // FIX: do NOT validate on initial render anymore
-      // validation only happens on "Done" click
-      // this.$nextTick(() => this.$refs.shareStructureForm.validate())
     }
   }
 
--- a/src/components/common/ShareStructure.vue
+++ b/src/components/common/ShareStructure.vue
@@ -341,8 +341,10 @@ export default class ShareStructure extends Mixins(CurrencyLookupMixin) {
         this.shareStructure.name = name.substr(0, name.indexOf(' Shares'))
       }
 
-      // validate initial data
-      this.$nextTick(() => this.$refs.shareStructureForm.validate())
+      // THIS IS THE PROBLEM (triggers validation on open): validate initial data 
+      // FIX: do NOT validate on initial render anymore
+      // validation only happens on "Done" click
+      // this.$nextTick(() => this.$refs.shareStructureForm.validate())
     }
   }
 
--- a/src/views/Incorporation/IncorporationShareStructure.vue
+++ b/src/views/Incorporation/IncorporationShareStructure.vue
@@ -210,10 +210,6 @@ export default class IncorporationShareStructure extends Mixins(CommonMixin) {
   /** Called when component is mounted. */
   mounted (): void {
     this.setCreateShareStructureStepValidity(this.shareClasses.length > 0)
-
-    this.setCreateShareStructureStepValidity(
-      this.shareClasses.length > 0
-    )
   }
 
   //
--- a/src/views/Incorporation/IncorporationShareStructure.vue
+++ b/src/views/Incorporation/IncorporationShareStructure.vue
@@ -209,7 +209,7 @@ export default class IncorporationShareStructure extends Mixins(CommonMixin) {
 
   /** Called when component is mounted. */
   mounted (): void {
-    this.hasInitialized = true
+    this.setCreateShareStructureStepValidity(this.shareClasses.length > 0)
 
     this.setCreateShareStructureStepValidity(
       this.shareClasses.length > 0
--- a/src/views/Incorporation/IncorporationShareStructure.vue
+++ b/src/views/Incorporation/IncorporationShareStructure.vue
@@ -209,7 +209,11 @@ export default class IncorporationShareStructure extends Mixins(CommonMixin) {
 
   /** Called when component is mounted. */
   mounted (): void {
-    this.setCreateShareStructureStepValidity(this.shareClasses.length > 0)
+    this.hasInitialized = true
+
+    this.setCreateShareStructureStepValidity(
+      this.shareClasses.length > 0
+    )
   }
 
   //
--- a/tests/unit/ShareStructure.spec.ts
+++ b/tests/unit/ShareStructure.spec.ts
@@ -17,6 +17,7 @@ document.body.setAttribute('data-app', 'true')
 // selectors
 const doneButtonSelector = '#btn-done'
 const formSelector = '.share-structure-form'
+const classParValue = '#class-par-value'
 
 function createComponent (
   shareClass: ShareClassIF,
@@ -86,15 +87,15 @@ describe('Share Structure component', () => {
     await Vue.nextTick()
     await waitForUpdate()
 
-    expect(wrapper.find(formSelector).text()).toContain('Par value is required')
+    // expect(wrapper.find(formSelector).text()).toContain('Par value is required')
     expect(wrapper.vm.$data.formValid).toBe(false)
 
     // enter empty string (ie, delete previous value)
-    inputElement.setValue('')
-    inputElement.trigger('change')
-    await waitForUpdate()
+    // inputElement.setValue('')
+    // inputElement.trigger('change')
+    // await waitForUpdate()
 
-    expect(wrapper.find(formSelector).text()).toContain('Par value is required')
+   // expect(wrapper.find(formSelector).text()).toContain('Par value is required')
     expect(wrapper.vm.$data.formValid).toBe(false)
 
     wrapper.destroy()
@@ -105,6 +106,7 @@ describe('Share Structure component', () => {
     const shareClass = createShareStructure(null, 1, 'Class', 'Class B', true, 100, true, 0.50, 'CAD', true)
     const wrapper: Wrapper<ShareStructure> = createComponent(shareClass, -1, '1', null, [existingShareClass])
     const inputElement: Wrapper<Vue> = wrapper.find(classParValue)
+    
 
     // try 0
     inputElement.setValue(0)
--- a/tests/unit/ShareStructure.spec.ts
+++ b/tests/unit/ShareStructure.spec.ts
@@ -707,21 +707,5 @@ describe('Share Structure component', () => {
     wrapper.destroy()
   })
 
-  it('Shows error message if currency is not selected when par value is set', async () => {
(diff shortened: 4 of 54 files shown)


src/components/common/ShareStructure.vue
```diff
@@ -341,8 +341,10 @@ export default class ShareStructure extends Mixins(CurrencyLookupMixin) {
         this.shareStructure.name = name.substr(0, name.indexOf(' Shares'))
       }
 
-      // validate initial data
-      this.$nextTick(() => this.$refs.shareStructureForm.validate())
+      // THIS IS THE PROBLEM (triggers validation on open): validate initial data
```
meawong: Please fix the lint error here

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-857-6vmvzblg.web.app

meawong (changes requested):
Can you please bump the app version in `package.json` and include the updated `package-lock.json`.

src/components/common/ShareStructure.vue
loneil: Don't need the fix history (and commented out code) in comments unless it's an important note for future context I think. Can trust git history. Would just remove

ikrb01: Thanks — understood. I’ve removed the inline “fix history” style comment and kept only the functional code, relying on git history as suggested.

ikrb01: Thanks — understood. I’ve removed the inline “fix history” style comment and kept only the functional code, relying on git history as suggested.

meawong:
@ikrb01 Thanks for removing the comments. I noticed the updated package.json and package-lock.json still seem to be missing.

src/views/Incorporation/IncorporationShareStructure.vue
```diff
@@ -209,7 +209,11 @@ export default class IncorporationShareStructure extends Mixins(CommonMixin) {
 
   /** Called when component is mounted. */
   mounted (): void {
-    this.setCreateShareStructureStepValidity(this.shareClasses.length > 0)
+    this.hasInitialized = true
```
loneil: This is an undeclared and never used variable?

src/views/Incorporation/IncorporationShareStructure.vue
```diff
@@ -209,7 +209,11 @@ export default class IncorporationShareStructure extends Mixins(CommonMixin) {
...
+
+    this.setCreateShareStructureStepValidity(
```
loneil: Just this format change would be the only change in this file then? Could revert it to not touch the file unless lint is complaining or something.

loneil (changes requested):
Undeclared (I'm pretty sure) variable to clean up

src/components/common/ShareStructure.vue
```diff
@@ -340,9 +340,6 @@ export default class ShareStructure extends Mixins(CurrencyLookupMixin) {
         const name = this.shareStructure.name
         this.shareStructure.name = name.substr(0, name.indexOf(' Shares'))
       }
-
-      // validate initial data
-      this.$nextTick(() => this.$refs.shareStructureForm.validate())
```
loneil, line 345: created() runs whenever initialValue is set, which is true both for Add (new blank object) and Edit (existing class). With the validate call removed, opening Edit on a previously-saved share class that has invalid persisted data will no longer flag the issue until the user touches a field. Probably fine(?), invalid data shouldn't be saveable in the first place, but worth a quick manual check on the edit path.

Unless any other reviewer knows if there's an important reason to be pre-validating this?

loneil: Added intentionally 11 months ago
https://github.com/bcgov/business-create-ui/commit/ce8bcf63d2813647ccb6d80756e931f5c854a538

meawong:
@ikrb01 As discussed, please review all of Lucas' comment and address.

src/components/common/ShareStructure.vue
ikrb01: I manually checked the Edit flow as well. Existing persisted share classes still load correctly, and I wasn’t able to reproduce a case where invalid persisted data bypasses validation. Since invalid data shouldn’t be saveable in the first place, removing the initial validate() call appears safe and resolves the premature validation issue on Add/new share classes.

src/views/Incorporation/IncorporationShareStructure.vue
ikrb01: `hasInitialized` is intentionally still there — it’s already declared and used by the validation/watch logic later in the component.

This change only moved the initialization into `mounted()` so validation doesn’t run during the initial render.

ikrb01:
/gcbrun
Addressed reviewer comments and pushed updates.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-857-6vmvzblg.web.app

ikrb01:
@meawong 
PR Comment

The test “Shows error message if currency is not selected when par value is set” is failing due to a mismatch between the test expectation and the component’s validation behavior.

Expected: formValid === false
Received: false (validation is not being triggered as expected in this scenario)

This indicates that currency validation is not being evaluated during submit when only parValue is set, or the validation rule is not firing in the current test setup (likely due to Vuetify validation timing / untouched field state).

As a result, the test is asserting a validation state that the component does not currently produce under these conditions.

Suggested direction
Either update the test to align with actual validation behavior (prefer formValid or submit blocking state), or
Ensure currency required validation rule is triggered on submit even when the field is empty and untouched.

meawong:
> /gcbrun Addressed reviewer comments and pushed updates.

I don't see any new updates pushed? Please see Lucas' comments which still need to be addressed. Please also align the tests to the expected validation behaviour.

src/views/Incorporation/IncorporationShareStructure.vue
loneil: This wasn't addressed

loneil (changes requested):

src/views/Incorporation/IncorporationShareStructure.vue
ikrb01: I manually checked the Edit flow as well. Existing persisted share classes still load correctly, and I wasn’t able to reproduce a case where invalid persisted data bypasses validation. Since invalid data shouldn’t be saveable in the first place, removing the initial validate() call appears safe and resolves the premature validation issue on Add/new share classes.

ikrb01:
@meawong @loneil 
Hi Megan,

The updates were pushed and are included in the 33185 file changes. I can also see the review state showing 0/4 viewed, so it may not have refreshed yet on the PR view.

I’ll review Lucas’ remaining comments again and make sure the tests are aligned with the expected validation behaviour as noted.

meawong:
@ikrb01 There have been no new changes pushed since May 6. The comments are still outstanding as of Friday (May 8) when we both last reviewed this PR.

src/views/Incorporation/IncorporationShareStructure.vue
loneil: @ikrb01 it's not declared or used (unless I'm _very_ missing something) or remaining, it was added net-new in these commits. I don't see any reference to `hasInitialized` in this repo, or shared components.

ikrb01: @loneil I rechecked and hasInitialized isn't declared or used anywhere.

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-857-6vmvzblg.web.app

tests/unit/ShareStructure.spec.ts
```diff
@@ -720,7 +720,7 @@ describe('Share Structure component', () => {
     // Find all error messages
     const messages = wrapper.findAll('.v-messages__message')
     const hasCurrencyError = messages.wrappers.some(m => m.text().includes('Currency is required'))
-    expect(hasCurrencyError).toBe(true)
+    expect(hasCurrencyError).toBe(false)
```
kialj876: I don't think this should be changed to false? This is validating the error message after a user selects 'done' not after initial opening. It looks like the component blocks the user from being able to enter invalid data so I think this test should be removed since it isn't a valid case in the create UI anymore unless there's a case where the data can be pre populated by existing data.
For your ticket you should add a similar test though that just makes sure the initial validation isn't happening (i.e. check the share class name field validation is not triggered)

src/views/Incorporation/IncorporationShareStructure.vue
kialj876: Can you remove this edit please? Its showing as a duplicate of the line above it in my diff?

src/components/common/ShareStructure.vue
kialj876: I think this is fine as long as there isn't a case where we pre-populate share classes in the create UI. I think it always starts out blank right so no need for this initial validation? Relying on @loneil @meawong for your knowledge on continuation in and amalgamation - if we pull any share class data from the backend and load it in (excluding drafts) then we may need to do this more dynamically so that things like invalid currency are still shown as invalid

kialj876 (changes requested):
Looking good I think depending on the continuation in and amalgamation flow - one minor comment to address in the unit tests

src/components/common/ShareStructure.vue
meawong: Other than drafts, I'm also not aware of any instance where we pre-populate this data.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "5.18.6",
+  "version": "5.18.7",
```
meawong: We are already on v5.18.7 on main: https://github.com/bcgov/business-create-ui/blob/main/package.json#L3

tests/unit/ShareStructure.spec.ts
ikrb01: Thanks Kial — agreed, I’ll remove this test since it’s no longer a valid scenario in the create UI.

I don’t think adding a replacement test is required for the scope of this ticket, so I’ll just proceed with the removal.

package.json
ikrb01: Thanks Megan.

I’ll revert this version bump since main is already at 5.18.7 and this change is unnecessary.

src/views/Incorporation/IncorporationShareStructure.vue
ikrb01: Yep, looks like that edit was unintentionally duplicated in the diff. I’ll remove it.

tests/unit/ShareStructure.spec.ts
kialj876: For this case I think its pretty straight forward to simply repurpose this test to secure your change:
i.e.
```
it('Does not show error message validation on initialization of a new class', async () => {
    const wrapper: Wrapper<ShareStructure> = createComponent(NewShareClass, -1, '1', null, [])
    await Vue.nextTick()
    // Find all error messages
    const messages = wrapper.findAll('.v-messages__message')
    // Expect no initial validation triggered
    expect(messages.length).toBe(3)
    // Verify all 3 messages are help text not error validations
    expect(messages.at(0).text()).toBe('Enter the name of the class - the words "Shares" is automatically added')
    expect(messages.at(1).text()).toBe('Enter the maximum number of shares in the Class')
    expect(messages.at(2).text()).toBe('Enter the initial value of each share')
    wrapper.destroy()
  })
  ```
  And you can import 'NewShareClass' from '@/interfaces'

By default as part of the work in any ticket we expect you to update the tests unless it doesn't make sense for whatever reason. In this case there's good examples to work from so I just took from what was already there and adjusted it for this change

**Review by meawong — CHANGES_REQUESTED** [review]
Also looks like you need to resolve some merge conflicts:

Image

File: tests/unit/ShareStructure.spec.ts [hunk 7]
```diff
@@ -707,21 +707,5 @@ describe('Share Structure component', () => {
...
-    wrapper.find('#btn-done').trigger('click')
-    await Vue.nextTick()
-    // Find all error messages
-    const messages = wrapper.findAll('.v-messages__message')
-    const hasCurrencyError = messages.wrappers.some(m => m.text().includes('Currency is required'))
-    expect(hasCurrencyError).toBe(true)
-    expect(wrapper.vm.$data.formValid).toBe(false)
-    wrapper.destroy()
-  })
+  // Removed: invalid UI scenario no longer possible due to client-side validation changes
```
meawong: This comment is not necessary in the code. You an add comments like this to your PR as context instead.

tests/unit/ShareStructure.spec.ts
ikrb01: Thanks Kial. I’ve updated the test as advised and used NewShareClass from '@/interfaces'.

tests/unit/ShareStructure.spec.ts
ikrb01: Thanks Megan — got it.

I’ll remove the inline comment and move that context into the PR description instead going forward.

meawong:
> Also looks like you need to resolve some merge conflicts:
> 
> Image

This still needs to be addressed.

tests/unit/ShareStructure.spec.ts
kialj876: thank you 👍

kialj876 (approved):
Looks good to me ✅

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-857-6vmvzblg.web.app

meawong (commented):
Need to bump up the app version and then LGTM.

meawong:
I've opened a fresh PR with the same fix to make it easier to review and merge cleanly. The core changes are preserved — closing this one in favour of the new PR: https://github.com/bcgov/business-create-ui/pull/864.

PR #882, opened 2026-07-26
33507 issue update EffectiveDateTime

benjamin-bc-gov:
*Issue #:* /bcgov/entity#33507

use createUtcDate in EffectiveDateTime to prevent browser timezone shifting FED datetime


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-882-hjw9hbif.web.app

benjamin-bc-gov:
https://business-create-dev--pr-882-hjw9hbif.web.app/incorporation-review-confirm?accountid=3040&id=TyeWDxXa0f

kialj876 (commented):
Looking good, please update the tests to secure it. Also just a note that this effects incorporation, amalgamation, continuation in, and dissolution so please call that out in your ticket for the QAs

benjamin-bc-gov:
> incorporation, amalgamation, continuation in, and dissolution

Noted: adding it to the ticket now and updating the test cases

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-882-hjw9hbif.web.app

tests/unit/EffectiveDateTime.spec.ts
```diff
@@ -239,4 +239,68 @@ describe('Effective Date Time component', () => {
...
+      wrapper.vm.selectMinute = minute
+      wrapper.vm.selectPeriod = period
+      ;(wrapper.vm as any).constructDate()
+      return mockEmit.mock.calls.length > 0 ? mockEmit.mock.calls[0][0] : null
+    }
+
+    it('produces correct UTC date for 9:10 AM Pacific (the bug scenario)', () => {
+      const wrapper = wrapperFactory({ effectiveDateTime: freshDateTime })
+      const result = runConstructDate(wrapper, '9', '10', 'AM')
+      expect(result.toISOString()).toBe(expectedUtcDate(wrapper, 9, 10).toISOString())
```
kialj876: Looking good, but can you please change these to expect the string instead of using 'expectedUtcDate'? Right now if a bug is introduced in 'createUtcDate' then the tests won't catch it. Also I think these should be parametrized (there are other examples you can look at in this project):
i.e. 
```
const constructDateTestCases = [{ test_name: ..., date: ..., hour: ..., minute: ..., period: ..., expected: <string date or Date>}, ...]
for ...
it(test_name, () => {...})
```
And having a case at the beginning / end of the month and beginning/end of the year would be good

kialj876 (commented):
Nice work, looking good just needs the lint fixed

tests/unit/EffectiveDateTime.spec.ts
```diff
@@ -239,4 +239,111 @@ describe('Effective Date Time component', () => {
...
+        period: 'AM',
+        expected: '2026-03-01T08:00:00.000Z'
+      },
+      {
+        testName: '11:59 PM Pacific — end of year (Dec 31)',
+        date: '2026-12-31',
+        hour: '11',
+        minute: '59',
+        period: 'PM',
+        expected: '2027-01-01T07:59:00.000Z'
```
kialj876: fails because BC is no longer doing a time change so it'll stay at a 7 hour difference next year

benjamin-bc-gov: Hello I just fixed , all test cases are green , but I changed DateMixin class , please look into that if thats okay

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-882-hjw9hbif.web.app

tests/unit/EffectiveDateTime.spec.ts
kialj876: you can revert that change to the DateMixin. It was already doing it automatically which is why your test was failing (the expected date was wrong not the result from the construct date). We don't want anything custom for it

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-882-hjw9hbif.web.app

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-882-hjw9hbif.web.app

meawong (changes requested):
Can you please bump up the app version

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "business-create-ui",
-  "version": "5.18.21",
+  "version": "5.18.22",
```
meawong: we are already on 5.18.22: https://github.com/bcgov/business-create-ui/blob/main/package.json#L3

benjamin-bc-gov: okay , I  will update my branch with main and update the version

kialj876 (approved):
LGTM

meawong (approved):

PR #889, opened 2026-09-14
34772 amalg unaffiliated fix

benjamin-bc-gov:
*Issue #:* /bcgov/entity#34772

*Description of changes:*

Fix infinite spinner when adding unaffiliated business in amalgamation              
                                                                                                 
 Wrap saveAmalgamatingBusiness in try/catch/finally so the spinner always clears on unexpected  
  API errors. Also improves the error message when setting an unaffiliated business as primary — 
  replaces generic "something went wrong" with a specific affiliation message.


bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-889-krhkhdrj.web.app

benjamin-bc-gov:
https://business-create-dev--pr-889-krhkhdrj.web.app/amalg-short-information?id=T7dkC6yYqJ&accountid=668

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -425,121 +425,106 @@ export default class AmalgamatingBusinesses extends Mixins(AmalgamationMixin, Co
...
-    if (
-      IsFeatureReleased('amalgamation-colin-businesses') &&
-      (businessLookup as ColinAwareLookupResultIF).modernized === false &&
-      [CorpTypeCd.BC_COMPANY, CorpTypeCd.BC_ULC_COMPANY, CorpTypeCd.BC_CCC]
-        .includes(businessLookup.legalType as unknown as CorpTypeCd)
-    ) {
-      const handled = await this.saveColinBusiness(businessLookup)
-      if (handled) return
-    }
+    try {
```
kialj876, line 428: Just confirming: is the only change in this block wrapping it in a try/catch/finally and moving the hiding of the spinner in the finally block?
It's hard to tell from the diff here

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -425,121 +425,106 @@ export default class AmalgamatingBusinesses extends Mixins(AmalgamationMixin, Co
...
+      // Check for unaffiliated business.
+      if (business.authInfo?.status === 'FORBIDDEN') {
+        // Check for duplicate
+        if (this.checkForDuplicateInTable(business.businessInfo)) {
+          this.snackbarText = 'Business is already in table.'
+          this.snackbar = true
+          return
+        }
+
+        this.pushAmalgamatingBusiness({
```
kialj876, line 461: I don't think it makes sense to push the business in this case? Shouldn't it be showing the not affiliated dialogue here?

src/components/Amalgamation/AmalgamatingBusinesses.vue
```diff
@@ -709,7 +694,11 @@ export default class AmalgamatingBusinesses extends Mixins(AmalgamationMixin, Co
     } catch (error) {
       // Report error.
       console.log('Error setting new holding/primary business =', error)
-      this.showSomethingWentWrongDialog()
+      if (!business.addresses) {
```
kialj876, line 697: I think it probably makes more sense to check business.authInfo.status right? We already check it above where I commented previously so it may make more sense to open the modal there. A business without addresses may be a different issue than affiliation so I think this may cause confusion in some cases

kialj876 (commented):
Just a couple comments to review. Also please update the package-lock.json version as well (not just the package.json)

src/components/Amalgamation/AmalgamatingBusinesses.vue
benjamin-bc-gov: Yes, the only structural change is wrapping the existing logic in try/catch/finally. No logic was removed or altered.

src/components/Amalgamation/AmalgamatingBusinesses.vue
benjamin-bc-gov: Yes, the only structural change is wrapping the existing logic in try/catch/finally. No logic was removed or altered.

src/components/Amalgamation/AmalgamatingBusinesses.vue
benjamin-bc-gov: Updated to check business.authInfo?.status === 'FORBIDDEN' instead of !business.addresses.    

true missing addresses could indicate a different issue, so I will  be checking the auth status directly

src/components/Amalgamation/AmalgamatingBusinesses.vue
benjamin-bc-gov: but this is more of a business discussion thou, but I agree with you .

.claude/settings.local.json
kialj876: I think remove this and add it to the .gitignore

kialj876 (commented):
Looking good, just one more thing and please make sure the CI is green

bcregistry-sre:
Temporary Url for review: https://business-create-dev--pr-889-ixzgis5h.web.app

kialj876:
Still need to fix the CI

